PR #13, opened 2020-11-26
remove graphql-tools and @miragejs/graphql workaround

### LINKED ISSUE #16: Update @graphql-tools library
We currently have some workarounds in Orion integration code that were needed until some fixes were released in @graphql-tools library. Let's update the library and remove the workaround in Atlas

kdembler:
### ~Review #10  first~

Upgrades `graphql-tools` and `@miragejs/graphql` to the newest versions, allowing to remove current workarounds

Resolves #16


package.json
```diff
@@ -67,6 +67,7 @@
     "@types/react-transition-group": "^4.4.0",
     "@types/video.js": "^7.3.10",
     "apollo": "^2.30.2",
+    "babel-jest": "26.6.3",
```
kdembler, line 70: I have completely no clue why, but suddenly tests started crashing, saying that `babel-jest` was not found. This is super weird since running `yarn why babel-jest` outputs that we already have this package at this version because of dependency of `react-scripts`. Adding it here explicitly fixes tests though 🤔

tsconfig.json
```diff
@@ -2,7 +2,11 @@
   "extends": "./tsconfig.paths.json",
   "compilerOptions": {
     "target": "es2019",
-    "lib": ["dom", "dom.iterable", "es2019"],
```
fcescob, line 5: I see this formatting keeps happening when running the react-scripts, does adding the tsconfig to .prettierignore fixes that?

package.json
fcescob: Interesting, let's just keep this for now then

tsconfig.json
kdembler: Yeah, the problem is I think that `react-scripts` enforces specific formatting but Prettier tries to enforce our styleguide so we go back and forth, that's why I added it to `.prettierignore` and we're just gonna leave it be

fcescob (approved):
LGTM

PR #69, opened 2020-11-27
switch channel and video playback views to paginated videos query

### LINKED ISSUE #49: Make channel and video views use paginated queries
Currently, both video playback view (more videos section) and channel view use a regular video grid for displaying the channel's videos. Furthermore, they don't use paginated queries but rather just request all videos from a specified channel via the channel's `videos` field. This works for now, but once channels start having more videos, pagination will be a must.

As part of this task, support for passing channel ID to the InfiniteVideoGrid should be added and that component should be used for the aforementioned views (unless we don't want infinite grids there, then we only need to use paginated queries)

kdembler:
Resolves #49


fcescob (approved):
LGTM!

PR #89, opened 2020-12-01
Replace z-index With zIndex Theme values everywhere

### LINKED ISSUE #83: Clean up z-index across the app
Add z-index values as part of the theme and use them across the app. Most importantly, fix the video player "play overlay" overlapping with the navbar:

fcescob:
Fixes #83

### DIFF
--- a/src/shared/theme/sizes.ts
+++ b/src/shared/theme/sizes.ts
@@ -6,16 +6,10 @@ export function sizes(n: number, raw?: boolean) {
 }
 
 export const zIndex = {
-  1: 10,
-  2: 20,
-  3: 30,
-  4: 40,
-  5: 50,
-  '-1': -10,
-  '-2': -20,
-  '-3': -30,
-  '-4': -40,
-  '-5': -50,
-  minusInfinity: -9999,
-  infinity: 9999,
+  background: -10,
+  farBackground: -20,
+  header: 100,
+  overlay: 10,
+  nearOverlay: 20,
+  globalOverlay: 999,
 }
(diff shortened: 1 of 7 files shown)


src/shared/theme/sizes.ts
```diff
@@ -4,3 +4,18 @@ export function sizes<B extends boolean>(n: number, raw?: B): B extends false ?
 export function sizes(n: number, raw?: boolean) {
   return raw ? base * n : `${base * n}px`
 }
+
+export const zIndex = {
+  1: 10,
```
kdembler, line 9: I think whether maybe we should choose some meaningful names for those. I'm afraid this will become the same problem as with the `xl` sizes. Also, for example the `inifnity` is used for the header, but if at some point we want a full page overlay, we'll need that to have a higher zIndex. For now I think we could use something like `header`, `background`, `localOverlay`. If in the future we would need more, we could then pick other meaningful names. What do you think about that?

src/shared/theme/sizes.ts
```diff
@@ -4,3 +4,18 @@ export function sizes<B extends boolean>(n: number, raw?: B): B extends false ?
 export function sizes(n: number, raw?: boolean) {
   return raw ? base * n : `${base * n}px`
 }
+
+export const zIndex = {
```
kdembler, line 8: I think whether maybe we should choose some meaningful names for those. I'm afraid this will become the same problem as with the `xl` sizes. Also, for example the `inifnity` is used for the header, but if at some point we want a full page overlay, we'll need that to have a higher zIndex. For now I think we could use something like `header`, `background`, `localOverlay`. If in the future we would need more, we could then pick other meaningful names. What do you think about that?

fcescob: I can see this being hard to use if the scale has a lot of values, but given that will only ever use a handful of values for a `z-index`, I agree that it would be an improvement over the current version and will change it

kdembler (approved):

PR #108, opened 2020-12-04
fix the video loading jump

### LINKED ISSUE #46: Fix video playback loading layout jump
In the video playback view, once the video metadata (title, etc.) finished loading, the placeholder for the video player is hidden and the video player itself is displayed. For some reason, for a brief second, the player will have a very low height before going back up to its proper sizing. This causes the page layout to jump around a bit. We need to alleviate that some way. Introduced in Joystream/joystream#1643 

Screenshot 2020-11-09 at 12 35 33 PM

kdembler:
Resolves #46

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -14,3 +14,7 @@
   font-weight: bold;
   src: url('./assets/fonts/Inter-Bold.ttf') format('truetype');
 }
+
+.video-js {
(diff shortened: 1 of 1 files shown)


fcescob (approved):
Layout jump is gone, good job!

PR #110, opened 2020-12-04
keep line breaks for the cover video subtext fixed

### LINKED ISSUE #101: Adjust cover video subtext width
Remove the `max-width` from the container, add a max-width to the subtext itself, use `ch` units so the line-breaks are consistent. `40ch` seemed to work fine

kdembler:
Resolves #101


fcescob (approved):
LGTM

PR #112, opened 2020-12-05
Scroll to top

### LINKED ISSUE #81: Fix scroll behaviour
1. Scroll to top on every navigation
2. Scroll to top on category change in Browse view

fcescob:
fixes #81

### DIFF
--- a/src/components/LayoutWithRouting.tsx
+++ b/src/components/LayoutWithRouting.tsx
@@ -8,14 +8,6 @@ import { Navbar, ViewErrorFallback } from '@/components'
 import { HomeView, VideoView, SearchView, ChannelView, BrowseView } from '@/views'
 import routes from '@/config/routes'
 
-const ScrollToTop: React.FC = () => {
-  const { pathname } = useLocation()
-  useEffect(() => {
-    window.scrollTo(0, 0)
-  }, [pathname])
-  return null
-}
-
 type RouteProps = {
   Component: React.ComponentType
 } & RouteComponentProps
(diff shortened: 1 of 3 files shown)


src/components/LayoutWithRouting.tsx
```diff
@@ -19,6 +27,7 @@ const Route: React.FC<RouteProps> = ({ Component, ...pathProps }) => {
         navigate('/')
       }}
     >
+      <ScrollToTop />
```
kdembler: Do we need that in every `Route` component? Wouldn't just one instance in `LayoutWithRouting` (under `Router`) work?

src/views/BrowseView.tsx
```diff
@@ -18,12 +18,19 @@ const BrowseView: React.FC<RouteComponentProps> = () => {
...
-      onCompleted: (data) => handleCategoryChange(data.categories[0]),
+      onCompleted: (data) => {
+        handleCategoryChange(data.categories[0])
+        firstCategoryChange.current = false
+      },
     }
   )
-
+  const headerRef = useRef<HTMLHeadingElement>(null)
+  const firstCategoryChange = useRef(true)
```
kdembler: 1. Why do we need that?
2. Why can't it be a regular state?

fcescob: 1. We want to scroll up whenever the category changes, but _only if_ it changed because a user selected another category. The `onCompleted` callback was calling `handleCategoryChange` and so the page was scrolling up down every time the Browse view was opened, regardless of user actions.
2. It can be state, changed it.

src/components/LayoutWithRouting.tsx
fcescob: I don't think that would work, the `Router` component will only render one child, while we want to render this on every route. I agree it is redundant to have it on every route, so I'm happy to change it if we come up with a more elegant solution.

src/views/BrowseView.tsx
kdembler: Okay, so we need to skip the scroll on initial category load, got it. It's little confusing at the moment with the naming. Maybe we don't need state at all? We could add an optional param to the `handleCategoryChange` that would skip the scrolling

src/components/LayoutWithRouting.tsx
kdembler: Ah, right, I forgot we're using `@reach/router`, we could do that in `react-router` I think. I found this approach: https://github.com/reach/router/issues/262#issuecomment-484487915 I think it's better than having that in each route, and later we can get rid of that once we move to `react-router`

src/views/BrowseView.tsx
fcescob: Good point actually. I'll add the parameter there

src/components/LayoutWithRouting.tsx
```diff
@@ -24,21 +24,28 @@ const Route: React.FC<RouteProps> = ({ Component, ...pathProps }) => {
...
-        <Route path={routes.video()} Component={VideoView} />
-        <Route path={routes.search()} Component={SearchView} />
-        <Route Component={BrowseView} path={routes.browse()} />
-        <Route Component={ChannelView} path={routes.channel()} />
-      </Router>
-    </MainContainer>
-  </>
-)
+const LayoutWithRouting: React.FC = () => {
+  globalHistory.listen(({ action }) => {
```
kdembler: This need to go into `useEffect`

kdembler (changes requested):

src/components/LayoutWithRouting.tsx
kdembler: @Gamaranto this doesn't work. You're currently calling unsubscribe function right after subscribing

kdembler (approved):

PR #114, opened 2020-12-07
channel cover adjustment

### LINKED ISSUE #98: Adjust aspect ratio of the channel cover image
We need to aim for something lower than 16:9, so it doesn't take as much vertical space.

As part of this task:
- lower the visible aspect ratio
- make sure the layout work on mobile
- make sure the loading state pattern looks okay
- lower the content overlap (maybe make it fixed)
- adjust the gradient

### LINKED ISSUE #99: Add channel cover loading transition
So it doesn't just snap when the metadata loads

kdembler:
Resolves #98
Resolves #99

Adjusted the aspect ratio for the channel cover and added pattern to cover transition. Also used that occasion to do some groundwork to clean up transitions. We will need to clear that further up - probably worth using new variables in other places as well


src/components/Navbar/Navbar.style.tsx
```diff
@@ -11,7 +11,7 @@ type NavbarStyleProps = {
 }
 
 export const StyledSearchbar = styled(Searchbar)`
-  transition: max-width 0.6s cubic-bezier(0.165, 0.84, 0.44, 1);
+  transition: max-width 0.6s ${transitions.easing};
```
fcescob: let's use the timings here as well

src/shared/components/VideoPreview/VideoPreview.styles.tsx
```diff
@@ -38,7 +38,7 @@ export const CoverHoverOverlay = styled.div`
   left: 0;
   opacity: 0;
 
-  transition: opacity 0.4s cubic-bezier(0.165, 0.84, 0.44, 1);
+  transition: opacity 0.4s ${transitions.easing};
```
fcescob: same here, let's not hardcode the duration

src/shared/components/GlobalStyle/GlobalStyle.tsx
```diff
@@ -42,6 +42,17 @@ const globalStyles = css`
   }
 `
 
-const GlobalStyle: React.FC = () => <Global styles={globalStyles} />
+type GlobalStyleProps = {
+  additionalStyles?: SerializedStyles[] | SerializedStyles
+}
+
+const GlobalStyle: React.FC<GlobalStyleProps> = ({ additionalStyles }) => {
+  const additionalStylesArray = additionalStyles
```
fcescob, line 50: Could you elaborate on why this is necessary and we can't just include all styles into the same `css` fragment?
I feel like global styles should be pretty narrow in general and there should a be good reason to extend them

kdembler: Yeah, this is so we can add Atlas-specific global styles to the app as well. This component lives in `shared` so we shouldn't add them here. Otherwise we would need to add two separate global styles components in `LayoutWithRouting`. Does that make sense?

src/components/Navbar/Navbar.style.tsx
kdembler: I'm not sure we should use the global timing here. I wanted to make sure all the loading transitions share the same timing. This isn't related to loading - it happens to share the same value. I can make that use `timings` as well but then we should add a separate value - just for this component, so I don't think that's worth it

src/shared/components/VideoPreview/VideoPreview.styles.tsx
kdembler: The same case as above

src/components/Navbar/Navbar.style.tsx
fcescob: I would argue it would still be good to have duration for this in the theme, in case we want to chain animations in the future or we just want to keep them consistent through the app, I think something like `fast`, `slow`, `regular` would do

src/shared/components/GlobalStyle/GlobalStyle.tsx
fcescob: Ah yes good point, let's keep it this way then.

src/components/Navbar/Navbar.style.tsx
kdembler: I'd say we shouldn't pick those arbitrary values at this time as it doesn't bring us any value. If we happen to need to keep some timings consistent let's put them here at that time with a proper naming

fcescob: There is little value now because we have few unsophisticated animations, but I can see several benefits:
- Lots of time saved during refactoring
- Prevention of a lot of subtle inconsistencies which would be hard to catch, like one animation being slower of faster than another one 
This will get more annoying to do as the number of animations and hardcoded values grow so I'm for doing it now that the hassle is low

kdembler: Okay, added `regular`

fcescob (approved):
LGTM! Good job!

PR #117, opened 2020-12-08
add loading state to avatar, add ChannelLink component

### LINKED ISSUE #100: Add a loading state to the Avatar component
This way we don't need explicit placeholder handling whenever the `Avatar` is used. As part of this task, we should also add a transition between those two states so that it isn't abrupt

kdembler:
## ~Review after #114~

Resolves #100


fcescob (approved):
LGTM! Good job!

PR #139, opened 2020-12-10
change index title

### LINKED ISSUE #92: Title of Atlas should be Joystream
Recall that Atlas is not a public name, its just an internal product name, consumers should just think of Joystream.

ZK3DEV:
Resolves #92

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4,13 +4,13 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="description" content="Joystream Atlas" />
-    <link rel="apple-touch-icon" sizes="180x180" href="favicon/apple-touch-icon.png">
-<link rel="icon" type="image/png" sizes="32x32" href="favicon/favicon-32x32.png">
-<link rel="icon" type="image/png" sizes="16x16" href="favicon/favicon-16x16.png">
-<link rel="manifest" href="favicon/site.webmanifest">
-<link rel="mask-icon" href="favicon/safari-pinned-tab.svg" color="#5bbad5">
-<meta name="msapplication-TileColor" content="#da532c">
-<meta name="theme-color" content="#4038FF">
+    <link rel="apple-touch-icon" sizes="180x180" href="favicon/apple-touch-icon.png" />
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -4,13 +4,13 @@
...
     <meta name="description" content="Joystream Atlas" />
-    <link rel="apple-touch-icon" sizes="180x180" href="favicon/apple-touch-icon.png">
-<link rel="icon" type="image/png" sizes="32x32" href="favicon/favicon-32x32.png">
-<link rel="icon" type="image/png" sizes="16x16" href="favicon/favicon-16x16.png">
-<link rel="manifest" href="favicon/site.webmanifest">
-<link rel="mask-icon" href="favicon/safari-pinned-tab.svg" color="#5bbad5">
-<meta name="msapplication-TileColor" content="#da532c">
-<meta name="theme-color" content="#4038FF">
+    <link rel="apple-touch-icon" sizes="180x180" href="favicon/apple-touch-icon.png" />
+    <link rel="icon" type="image/png" sizes="32x32" href="favicon/favicon-32x32.png" />
```
fcescob, line 8: I think this might be a formatting error, but any reason to add `/>` ? Both meta and link tags don't need an end tag,afaik the browser will ignore it.

fcescob (dismissed):

public/index.html
ZK3DEV: as far as I know its good practise to do that as it gives more like a xhtml structure which helps the browsers to parse  faster and more secure thats why the formatting tools add it.

ZK3DEV: let me know if u want it removed, as it wasnt like i went to each line to add it, code formatting does by default :D

kdembler: All the other `meta` tags we had there before (like `<meta charset="utf-8" />`) actually are self-closing already, so for the sake of consistency let's add that to the favicon ones as well - Prettier will actually add them as well

kdembler (approved):
Looks good to me. @Gamaranto anything to add?

fcescob:
> Looks good to me. @Gamaranto anything to add?

Nope, all for me as well.

fcescob (approved):

PR #146, opened 2020-12-14
Maximized playback must respect content aspect ratio

### LINKED ISSUE #141: Maximized playback must respect content aspect ratio
Currently it does not, it just fits content to screen.

Look at YouTube maximized mode for reference, basically make the video as tall as possible, and just center the video on the screen.

ZK3DEV:
Resolves #141


ZK3DEV:
so this is breaking the aspect ratio, from the quick look didn't see it affecting some other area so if it does and its needed can do a different fix

kdembler (approved):
Wow, I wouldn't suspect this to affect the actual playback, nice find!

I think we can safely remove this, it was added some time ago to allow landing page cover video playback but it since has been reworked a bit and doesn't seem to need it anymore

Thanks!

PR #148, opened 2020-12-14
Add License To VideoView

### LINKED ISSUE #78: Introduce display of license information on video playback
We need to introduce this, currently we have nothing. Some simple approach can be selected for now, adding it to the bottom of the description or something, it will later be designed properly, it was forgotten.

fcescob:
This PR fixes #78 ,  @bedeho and @blrhc please review if how it looks right now is good enough for Babylon.


src/api/schemas/extendedQueryNode.graphql
```diff
@@ -63,28 +63,34 @@ enum VideoMediaEncoding {
...
 type UserDefinedLicense {
-  text: String!
+  content: String!
 }
 
-union License = UserDefinedLicense | CreativeCommonsLicense
+union License = UserDefinedLicense | KnownLicense
 
+type LicenseEntity {
+  id: ID!
```
kdembler, line 80: Not sure if we should put all the fields here that the query node exposes. In general so far I've tried only including fields that we actually use. I guess things like `createdById` can be removed once the query node implementation changes so we shouldn't rely on them.

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -94,8 +100,7 @@ type VideoMedia {
   pixelWidth: Int!
 
   # Resolution height
-  pixelHeight: Int!
-
+  pixelHeight: Int
```
kdembler: This type shouldn't be changed

src/mocking/data/raw/coverVideo.json
```diff
@@ -22,5 +22,13 @@
     "handle": "SCHISM",
     "avatarPhotoUrl": "https://eu-central-1.linodeobjects.com/joystream/schismavatar.png",
     "coverPhotoUrl": null
+  },
+  "license": {
+    "id": "1503",
+    "attribution": null,
+    "type": {
+      "__typename": "UserDefinedLicense",
+      "content": "Free to consume and share, no derivatives or adaptations, copyright Jamie King 2020\n"
```
kdembler, line 31: Should we include `\n` at the end?

src/views/VideoView/VideoView.tsx
```diff
@@ -99,6 +102,23 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
         </DescriptionContainer>
+        <LicenseContainer>
+          {data?.video ? (
+            <>
+              <p>
+                License:{' '}
+                {data.video.license.type.__typename === 'KnownLicense' ? (
+                  <a href={data.video.license.type.url || ''}>{`${data.video.license.type.code}`}</a>
+                ) : (
+                  `${data.video.license.type.content}`
```
kdembler: I don't think we need that interpolation

src/views/VideoView/VideoView.tsx
```diff
@@ -99,6 +102,23 @@ const VideoView: React.FC<RouteComponentProps> = () => {
             </>
           )}
         </DescriptionContainer>
+        <LicenseContainer>
+          {data?.video ? (
+            <>
+              <p>
+                License:{' '}
+                {data.video.license.type.__typename === 'KnownLicense' ? (
+                  <a href={data.video.license.type.url || ''}>{`${data.video.license.type.code}`}</a>
```
kdembler: The link should have `target="_blank"` and that should also imply `rel="noreferrer noopener"`

kdembler (changes requested):

src/api/schemas/extendedQueryNode.graphql
fcescob: Good point, I will leave only the fields that we actually use

src/api/schemas/extendedQueryNode.graphql
fcescob: No idea how that happened, will revert the change

src/mocking/data/raw/coverVideo.json
fcescob: I just took the response from the query node for the production version of the cover video, I say let's leave it in for consistency even if its not needed.

src/views/VideoView/VideoView.tsx
fcescob: Good point, changed that.

src/views/VideoView/VideoView.tsx
fcescob: good catch, removed it.

src/mocking/data/raw/coverVideo.json
kdembler: Ok

src/views/VideoView/VideoView.tsx
```diff
@@ -99,6 +100,27 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
+          {data?.video ? (
+            <>
+              <p>
+                License:{' '}
+                {data.video.license.type.__typename === 'KnownLicense' ? (
+                  <a
+                    href={data.video.license.type.url || ''}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >{`${data.video.license.type.code}`}</a>
```
kdembler: We don't need that interpolation as well, missed that before

kdembler (approved):

PR #179, opened 2020-12-17
Add AsyncReducer and LocalStorageClient

### LINKED ISSUE #135: Create local store for personalised data
Based on #134, this should provide an interface implementation that will allow storing the data in local storage (I guess we don't need anything more fancy than that?). Data we will surely need for the time being:
- followed channels
- videos that were started but not finished (with the current timestamp)

fcescob:
Fixes #135


src/shared/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -9,13 +9,10 @@ type VideoPlayerProps = {
...
 } & VideoJsConfig
 
-const VideoPlayer: React.FC<VideoPlayerProps> = ({
-  className,
-  autoplay,
-  isInBackground,
-  playing,
-  ...videoJsConfig
-}) => {
+const VideoPlayer: React.ForwardRefRenderFunction<HTMLVideoElement, VideoPlayerProps> = (
```
kdembler: Changes to this file are out of scope

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+function readFromLocalStorage<T>(key: string, { deserialize = JSON.parse } = {}) {
+  const valueInLocalStorage = window.localStorage.getItem(key)
+  if (valueInLocalStorage) {
+    try {
+      return deserialize(valueInLocalStorage) as T
+    } catch (error) {
+      console.log(
```
kdembler: Let's use `console.error` here

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+function writeToLocalStorage<T>(key: string, value: T, { serialize = JSON.stringify } = {}) {
+  window.localStorage.setItem(key, serialize(value))
+}
+
+const watchedVideos = promisify(() => readFromLocalStorage<WatchedVideo[]>('watchedVideos') ?? [])
+const interruptedVideos = () =>
+  watchedVideos().then(
```
kdembler: Maybe here and below we should use `async/await` for consistency?

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
...
+  watchedVideos().then((videos) => videos.filter((video) => video.__typename === COMPLETED_VIDEO) as CompletedVideo[])
+
+const watchedVideo = (id: string) => watchedVideos().then((videos) => videos.find((v) => v.id === id) ?? null)
+const setWatchedVideo = async (
+  __typename: typeof COMPLETED_VIDEO | typeof INTERRUPTED_VIDEO,
+  id: string,
+  timestamp: number
+) => {
+  const currentVideos = await watchedVideos()
+  const isIncluded = currentVideos.find((v) => v.id === id)
```
kdembler: This name suggests this will be a `boolean`. Maybe we just call it `currentVideo`?

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+    case 'INTERRUPTED': {
+      writeToLocalStorage(
+        'watchedVideos',
+        currentVideos.map((v) => (v.id ? { __typename, id, timestamp } : v))
+      )
+      break
+    }
+    default: {
+      const newVideo = __typename === 'COMPLETED' ? { __typename, id } : { __typename, id, timestamp }
```
kdembler: Let's handle this outside of switch with just `if (!currentVideo)`, should be more clear what is happening

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
...
+      writeToLocalStorage(
+        'watchedVideos',
+        currentVideos.map((v) => (v.id === id ? { __typename, id } : v))
+      )
+      break
+    }
+    case 'INTERRUPTED': {
+      writeToLocalStorage(
+        'watchedVideos',
+        currentVideos.map((v) => (v.id ? { __typename, id, timestamp } : v))
```
kdembler: You need `v.id === id` here, now we will just overwrite every item :P

src/hooks/index.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect, useRef } from 'react'
+
+export function useInterval(callback: () => void, delay = 1000) {
```
kdembler: I don't think that's needed?

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
+import {
```
kdembler: Not sure if `api/` is the right place for this to exist. We don't really do anything api-related here.

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
...
+  CompletedVideo,
+  COMPLETED_VIDEO,
+  FollowedChannel,
+  InterruptedVideo,
+  INTERRUPTED_VIDEO,
+  UserPersonalData,
+  WatchedVideo,
+} from './types'
+
+function promisify<T>(fn: () => T): () => Promise<T>
```
kdembler: Let's not mix up `function x` and `const x = () =>` syntax, unless we come up with clear rules when to use which

src/api/localStorage/localStorageClient.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+const followedChannels = promisify(() => readFromLocalStorage<FollowedChannel[]>('followedChannels') ?? [])
+const isFollowingChannel = (id: string) => followedChannels().then((channels) => channels.some((ch) => ch.id === id))
+const setChannelFollowing = async (id: string, follow = true) => {
+  const currentFollowedChannels = await followedChannels()
+  const isFollowing = currentFollowedChannels.some((channel) => channel.id === id)
+
+  const newFollowedChannels = follow
```
kdembler: Do you think we could simplify this somehow? It's a little hard to follow the conditions. Maybe we could just make `newFollowedChannels` mutable and go through couple `if-else`?

src/hooks/usePersonalData.tsx
```diff
@@ -0,0 +1,124 @@
+import React, { useReducer, useContext, Dispatch } from 'react'
```
kdembler: I don't really understand what's going on here so it'd be best to hop on a call and talk about this

kdembler (changes requested):

src/hooks/usePersonalData/usePersonalData.tsx
```diff
@@ -0,0 +1,97 @@
...
+      }
+    }
+    case 'UPDATE_FOLLOWED_CHANNELS': {
+      return {
+        ...state,
+        ...action.followedChannels,
+      }
+    }
+    default: {
+      throw new Error(`Unhandled action type: ${(action as any).type}`)
```
kdembler: Do we need `action as any`? All `Action` will have a `type` so that shouldn't be a problem?

src/hooks/usePersonalData/usePersonalData.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+}
+type ContextValue = {
+  state: PersonalData
+  dispatch: Dispatch<Action>
+}
+const PersonalDataContext = React.createContext<ContextValue | undefined>(undefined)
```
kdembler, line 47: Maybe it would be safer to initialize this with a default value? So we don't have to worry about `undefined` when accessing the `state`. That can be
```ts
const PersonalDataContext = React.createContext<ContextValue>({
  state: {
    followedChannels: [],
    watchedVideos: [],
  },
  // eslint-disable-next-line @typescript-eslint/no-empty-function
  dispatch: (_) => {},
})
```

src/hooks/usePersonalData/usePersonalData.tsx
kdembler: Oh I see what's happening here. Our switch exhausts all the cases for Action type so in `default` there is no possible value. Maybe we shouldn't throw but just `return state` there? TypeScript will make sure we don't make a mistake

kdembler (approved):

Issue #172, opened 2020-12-15
The Category Picker Completely covers the screen on mobile


fcescob:
On mobile the `CategoryPicker` on the production version of the app will end up covering the whole screen, completely obfuscating videos.

![image](https://user-images.githubusercontent.com/29527327/102233094-293f0080-3ef0-11eb-93c6-b10cf9fa12ce.png)

The design does not cover this case, since the screen showed uses a low number of categories.

bedeho:
The current UI design is actually not that functional either, a better solution is needed, horizontal scroll?

![Screenshot 2020-12-15 at 16 32 36](https://user-images.githubusercontent.com/437292/102235970-37425080-3ef3-11eb-9786-bf4f73501596.png)

KubaMikolajczyk:
This pattern is getting a complete rework in the redesign of discover page.
Discover page : [Discover page link](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=9775%3A4626)
also https://github.com/Joystream/atlas/issues/874
Should we close this? @kdembler

kdembler:
Yes sir

PR #186, opened 2020-12-29
Category picker enhancement

ghhrmnzdh:
The previous category picker was getting **too** much **height** as a sticky container, making it unusable on **smaller** screens!

Screenshot from [_Browse_](https://play.joystream.org/browse) endpoint:

image

I'm suggesting an enhancement to the category picker by making it horizontally **scrollable** shown as below:

ezgif com-gif-maker

### DIFF
--- a/src/shared/components/CategoryPicker/CategoryPicker.style.ts
+++ b/src/shared/components/CategoryPicker/CategoryPicker.style.ts
@@ -4,8 +4,20 @@ import ToggleButton from '../ToggleButton'
 import { sizes, colors } from '@/shared/theme'
 
 export const Container = styled.div`
-  display: flex;
-  flex-wrap: wrap;
+  display: -webkit-box;
+  white-space: nowrap;
+  overflow: scroll;
+  ::-webkit-scrollbar {
+    display: none;
+  }
+  > * {
+    &:first-child {
+      margin-left: var(--global-horizontal-padding);
+    }
+    &:last-child {
+      margin-right: var(--global-horizontal-padding);
+    }
+  }
 `
 
 export const StyledPlaceholder = styled(Placeholder)`
--- a/src/shared/components/CategoryPicker/CategoryPicker.style.ts
+++ b/src/shared/components/CategoryPicker/CategoryPicker.style.ts
@@ -4,7 +4,7 @@ import ToggleButton from '../ToggleButton'
 import { sizes, colors } from '@/shared/theme'
 
 export const Container = styled.div`
-  display: -webkit-box;
+  display: block;
   white-space: nowrap;
   overflow: scroll;
   > * {
@@ -18,6 +18,7 @@ export const Container = styled.div`
 `
 
 export const StyledPlaceholder = styled(Placeholder)`
+  display: inline-block;
   margin: 0 ${sizes(3)} ${sizes(3)} 0;
 `
 
--- a/src/shared/components/CategoryPicker/CategoryPicker.style.ts
+++ b/src/shared/components/CategoryPicker/CategoryPicker.style.ts
@@ -22,5 +22,5 @@ export const StyledPlaceholder = styled(Placeholder)`
 `
 
 export const StyledToggleButton = styled(ToggleButton)`
-  margin: 0 ${sizes(3)} ${sizes(3)} 0;
+  margin-right: ${sizes(3)};
 `
--- a/src/shared/components/CategoryPicker/CategoryPicker.style.ts
+++ b/src/shared/components/CategoryPicker/CategoryPicker.style.ts
@@ -7,9 +7,6 @@ export const Container = styled.div`
   display: -webkit-box;
   white-space: nowrap;
   overflow: scroll;
-  ::-webkit-scrollbar {
-    display: none;
-  }
   > * {
     &:first-child {
       margin-left: var(--global-horizontal-padding);
--- a/src/shared/components/Placeholder/Placeholder.tsx
+++ b/src/shared/components/Placeholder/Placeholder.tsx
@@ -22,6 +22,7 @@ const pulse = keyframes`
 `
 
 const Placeholder = styled.div<PlaceholderProps>`
+  display: 'inline-flex';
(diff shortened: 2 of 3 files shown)


src/shared/components/CategoryPicker/CategoryPicker.style.ts
```diff
@@ -4,8 +4,20 @@ import ToggleButton from '../ToggleButton'
 import { sizes, colors } from '@/shared/theme'
 
 export const Container = styled.div`
-  display: flex;
-  flex-wrap: wrap;
+  display: -webkit-box;
```
kdembler: Why `-webkit-box` instead of just `flex`? As far as I can tell we actually don't need any custom `display` here at all, the default `block` should work?

src/shared/components/CategoryPicker/CategoryPicker.style.ts
```diff
@@ -4,8 +4,20 @@ import ToggleButton from '../ToggleButton'
...
+  white-space: nowrap;
+  overflow: scroll;
+  ::-webkit-scrollbar {
```
kdembler: We should also add `scrollbar-width: none;` to handle that for Firefox as well

bedeho:
Awesome, this was actually already an issue here
https://github.com/Joystream/atlas/issues/172

Is the scrolling free, or does it snap to the label boundaries?

However

1. There seems to be some issue in the animation where a scroll bar emerges, and thus shifts the layout.
2. What is done for a non-mobile case when the viewport is still just too narrow? I assume there are no buttons for clicking left/right? how will the interaction work then?

src/shared/components/CategoryPicker/CategoryPicker.style.ts
```diff
@@ -4,8 +4,20 @@ import ToggleButton from '../ToggleButton'
...
+    display: none;
+  }
+  > * {
```
kdembler, line 10: I think those should be living outside of the component itself, in `StyledCategoryPicker` in `BrowseView.styles`. Without any additional styling the component should work as part of the layout, only the Browse View makes it span the full screen width

src/shared/components/CategoryPicker/CategoryPicker.style.ts
```diff
@@ -4,8 +4,20 @@ import ToggleButton from '../ToggleButton'
...
+    &:first-child {
+      margin-left: var(--global-horizontal-padding);
+    }
+    &:last-child {
+      margin-right: var(--global-horizontal-padding);
+    }
+  }
 `
 
 export const StyledPlaceholder = styled(Placeholder)`
```
kdembler, line 20: We can adjust those styles now. All the items used to have a bottom margin to handle row wrapping. Now that all the items will reside in one row, we can get rid of that margin for items. At the same time we will need to update the padding for `StyledCategoryPicker` which purposefully sets a lower bottom padding because it takes this margin into account.

src/shared/components/Placeholder/Placeholder.tsx
```diff
@@ -22,6 +22,7 @@ const pulse = keyframes`
 `
 
 const Placeholder = styled.div<PlaceholderProps>`
+  display: 'inline-flex';
```
kdembler: `'inline-flex'` is an invalid property value here so it's not taken into account at all. If you wanted to use that, you'd need to remove the apostrophes from the value, but I don't think this property is actually needed here?

kdembler (changes requested):
Hey @ghhrmnzdh, thanks for taking your time to tackle this issue!

There are some small code-related problems I've listed in the comments. Please let me know if something's unclear or you need some help :)

src/shared/components/CategoryPicker/CategoryPicker.style.ts
ghhrmnzdh: I'm assuming using `flex` with `overflow` of `hidden` or `scroll` may collapse the margin from the last element that is overflowing. `-webkit-box` handles this. Would like to know if there is a way to handle it with flex?

kdembler: You're right about `flex` collapsing the margin. However, we cannot really use `-webkit-box` as it's from an old specification of flexbox and shouldn't really be used in production. Do we need flexbox here at all though? Seems that the default `block` works just fine

ghhrmnzdh: Now I remember why I was trying to use `inline-flex`.  If you set the `display` of the parent container to `block`, skeletons look like this:

image

So I handled it with `-webkit-box`, but you're right about its usage in production. It just handles both placeholder and buttons well, without changing the display of any. 

Ok, let's move on. I changed the parent display to `block`.Now, I think I should change the `StyledPlaceholder` display in `CategoryPicker.style` to `inline` or `inline-block` in order to work properly. Right?

ghhrmnzdh:
@bedeho

> Is the scrolling free, or does it snap to the label boundaries?

The scrolling is free, and I think scroll with snapping blocks the user's access to new categories as it opens **only** one option at a time.

> 1. There seems to be some issue in the animation where a scroll bar emerges, and thus shifts the layout.

There is no issue with the scroll animation, it's just me scrolling vertically accidentally after showing the horizontal scroll.

> 2. What is done for a non-mobile case when the viewport is still just too narrow? I assume there are no buttons for clicking left/right? how will the interaction work then?

You're right. It works as you can typically scroll in that area, but there should be buttons in none-mobile cases. We should add something like a carousel to it. I will look through the `carousel` that exists in the `components`.

bedeho:
I will ask the designer to review this issue on Monday, he may have tweaks or adjustments to this which can improve it, so lets hold off on merging until he has reviewed.

kdembler:
@bedeho any status on that?

bedeho:
Unfortunately I don't think they have gotten to this yes, they will cover it as part of the overall revamp of how our Figma is organized, and how it specifies responsive variations. So it may actually be a way off.

kdembler:
Closing this for the time being since we will require designers input to fix that. Thanks for your time anyway @ghhrmnzdh!

Issue #183, opened 2020-12-22
Pause/play using spacebar


bedeho:
This is a very natural interaction that I use all the time.

kdembler:
Closed via #187

PR #187, opened 2021-01-02
Add play/pause functionality on spacebar

ghhrmnzdh:
#183

Added functionality to `VideoView`. Still won't work on `FeaturedVideoHeader` and it scrolls down on the spacebar.

### DIFF
--- a/src/views/VideoView/VideoView.tsx
+++ b/src/views/VideoView/VideoView.tsx
@@ -32,9 +32,8 @@ const VideoView: React.FC<RouteComponentProps> = () => {
   const videoID = data?.video?.id
 
   const [playing, setPlaying] = useState<boolean>(true)
-  const handleUserKeyPress = useCallback((event) => {
-    const { key, keyCode } = event
-
+  const handleUserKeyPress = useCallback((event: Event) => {
+    const { keyCode } = event as KeyboardEvent
     if (keyCode === 32) {
       event.preventDefault()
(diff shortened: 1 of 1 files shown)


src/views/VideoView/VideoView.tsx
```diff
@@ -31,6 +31,24 @@ const VideoView: React.FC<RouteComponentProps> = () => {
 
   const videoID = data?.video?.id
 
+  const [playing, setPlaying] = useState<boolean>(true)
+  const handleUserKeyPress = useCallback((event) => {
```
kdembler: We should type `event` here, ideally that would be `event: KeyboardEvent` but `window.addEventListener` won't accept that unfortunately so we will need to use `event: Event`

src/views/VideoView/VideoView.tsx
```diff
@@ -31,6 +31,24 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
+    const { key, keyCode } = event
```
kdembler: Once we have `event` typed we can do `event as KeyboardEvent` here to get access to the data we need

kdembler: I don't think `key` is used anywhere?

kdembler (changes requested):
Overall looks good, just some small tweaks. Thanks!

kdembler (approved):
All good now :)

Issue #181, opened 2020-12-21
Disabled buttons looks bad


kdembler:
<img width="498" alt="Screenshot 2020-12-21 at 16 09 49" src="https://user-images.githubusercontent.com/12646744/102791254-06589480-43a7-11eb-8ef0-a73b5875ee39.png">

I guess they may be too bright - they look unattractive

ghhrmnzdh:
Add `loading` state for buttons or it's too distracting? This design also matches the `placeholder` that is being used accross the project. Also the text and icons should darken a little bit, for example `colors.gray[200]` is a good choice.

**Preview**:

![loading-state-button](https://user-images.githubusercontent.com/35824204/103434554-55f55680-4c18-11eb-9dbe-63ef59f185ef.gif)

bedeho:
Lol @kdembler I already had an issue for it, but very encouraging that our judgements are converging ;)

https://github.com/Joystream/atlas/issues/185

bedeho:
> Add `loading` state for buttons or it's too distracting? This design also matches the `placeholder` that is being used accross the project. Also the text and icons should darken a little bit, for example `colors.gray[200]` is a good choice.
> 
> **Preview**:
> 
> ![loading-state-button](https://user-images.githubusercontent.com/35824204/103434554-55f55680-4c18-11eb-9dbe-63ef59f185ef.gif)

I think the issue here is that what type of loading animation is used is not going to work universally in all contexts, which is not actually that surprising. We will get the designer to look at this and create a new guideline.

kdembler:
I think the immediate hotfix will be just darkening the background and/or text a bit to find something that feels visually good

kdembler:
Fixed via #191 for now. There are no other places in the app that actually use disabled buttons, so I guess once designers decide to use them, they'll add designs for them as well. Closing for now

PR #191, opened 2021-01-04
Visually fix disabled button

ghhrmnzdh:
#181

I think this styling properly matches the `placeholder` colors that already existed, and it looks disabled.

Preview:

image


kdembler (approved):
Looks good, thanks!

PR #198, opened 2021-01-07
Fix avatar text

### LINKED ISSUE #180: ChannelPreview broken without an avatar
Screenshot 2020-12-21 at 09 46 22

This text should be bigger and centred. Styles should be there somewhere, my guess is some parts change and we fail to apply them

drillprop:
Fixes #180 
image
image


kdembler (approved):
Nice work!

PR #199, opened 2021-01-07
add ability to fetch cover video data

### LINKED ISSUE #106: Graceful landing page video playback
Right now, the playback of the cover video begins the moment the clip has been loaded, there are two problems with this.

1. The transition from loading to showing the cover photo appears instant, or at least very fast, can this be slowed down and have a fadein if it does not.
2. The user is unable to perceive the actual cover if they have fast connection, so it just jumps straight into the video directly while the entire scene is still loading, often when lots of preview videos have not yet been rendered. Overall it means that there is just a lot happening simultaneously, which slightly disorients the user on what is happening. We could instead let the cover sit for a moment after, even if the video is available, and then gracefully transition into playing the video at a later time so that this happens a little bit later than the entire scene getting rendered to begin with.

### LINKED ISSUE #122: Remove hardcoded featured cover video
Currently all the data for featured cover video is hardcoded inside Atlas. This works for now but can be troublesome in the long run so we should make plans to get rid of that and instead fetch that from the query node. What will need to be done:
- define data model for cover video - it will likely require some other data that's available in a regular video model - for example the subtext displayed below the title - description may be too long
- add that data to the content directory 
- add support to the CLI to manage that
- add support to the query node
- add support to Atlas - some loading handling there will be required

### LINKED ISSUE #185: Loading of home page controls update
Seems like the default loading state does not look that great, and actually greates a bad temporary high contrast which is disurptive

Screenshot 2020-12-28 at 20 16 27

It will probably work better to not show any button during loading, and instead fade in the button itself just as with the background itself. Not 100% sure that will work either, but worth trying. If that also looks funky then a designer must look at it, so just premptively tagging it with design label for now.

kdembler:
This PR prepares the home view to be able to fetch the cover video data from the query node. This won't be supported for the time being so the data is still hardcoded, but loading state was added and some general improvements were made.

Resolves #122
Resolves #185 
Resolves #106


src/shared/components/Placeholder/Placeholder.tsx
```diff
@@ -27,7 +27,7 @@ const Placeholder = styled.div<PlaceholderProps>`
   margin-bottom: ${({ bottomSpace = 0 }) => getPropValue(bottomSpace)};
   border-radius: ${({ rounded = false }) => (rounded ? '100%' : '0')};
   background-color: ${colors.gray['400']};
-  animation: ${pulse} 0.5s cubic-bezier(0.4, 0, 0.6, 1) infinite;
+  animation: ${pulse} 0.8s cubic-bezier(0.4, 0, 0.6, 1) infinite;
```
fcescob, line 30: Any reason for this change?

src/api/hooks/index.ts
```diff
@@ -0,0 +1,2 @@
+import useCoverVideo from './coverVideo'
```
fcescob, line 1: Let's be consistent and move this to `src/hooks`

fcescob (changes requested):

src/shared/components/Placeholder/Placeholder.tsx
kdembler: Just visual, it seemed blinking too fast to me, especially now that the full cover video will be loading as well. It just feels nicer a bit slower

src/api/hooks/index.ts
kdembler: I believe this is a right place for that. `src/hooks` will be used for utility/logic related hooks. This is strictly related to the data fetching. Right now, because of query node restrictions, we return mocked value from `useCoverVideo` anyway, but once that's alleviated, we will do Apollo query inside it, so it makes sense to live in `api`

PR #201, opened 2021-01-07
Fix screen pattern for Browse view

### LINKED ISSUE #164: Browse screen pattern imperfection
Initially fixed here, but the pattern looks like this

Screenshot 2020-12-14 at 13 15 55

here is the design from the discovery [screen](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=1006%3A21679)

Screenshot 2020-12-14 at 13 16 14

The problems appear to be

- vertical alignment (too low, should touch top horizontal divider line)
- interactions with the categories is off (should be translucent categories and gradient on pattern, not occlude)
- sizing (does not appear to be due to any sort of resizing policy, as its invariant to all viewport changes).

mikkio-j:
- fix vertical alignment, so it sticks to the line above
- pattern interacts better with categories buttons (pattern can be seen through the buttons)
- put sizing to the pattern for different widths
pattern

### DIFF
--- a/src/views/BrowseView/BrowseView.style.ts
+++ b/src/views/BrowseView/BrowseView.style.ts
@@ -20,7 +20,7 @@ export const StyledCategoryPicker = styled(CategoryPicker)<IsAtTop>`
   top: ${NAVBAR_HEIGHT}px;
   padding: ${sizes(5)} var(--global-horizontal-padding) ${sizes(2)};
   margin: 0 calc(-1 * var(--global-horizontal-padding));
-  background-color: ${colors.black};
+  background-color: ${(props) => (props.isAtTop ? colors.transparent : colors.black)};
   border-bottom: 1px solid ${(props) => (props.isAtTop ? colors.black : colors.gray[800])};
 `
 export const StyledInfiniteVideoGrid = styled(InfiniteVideoGrid)`
@@ -36,6 +36,8 @@ export const IntersectionTarget = styled.div`
 `
 export const StyledBackgroundPattern = styled(BackgroundPattern)`
   position: absolute;
-  right: 0;
+  top: 154px;
+  right: 80px;
   z-index: ${zIndex.background};
+  transform: scale(1.3);
 `
--- a/src/views/BrowseView/BrowseView.style.ts
+++ b/src/views/BrowseView/BrowseView.style.ts
@@ -35,23 +35,27 @@ export const StyledInfiniteVideoGrid = styled(InfiniteVideoGrid)`
 export const Container = styled.div`
   position: relative;
   padding-top: ${sizes(14)};
-  overflow-x: hidden;
 `
 export const IntersectionTarget = styled.div`
   min-height: 1px;
 `
+export const StyledBackgroundPatternContainer = styled.div`
+  position: absolute;
+  top: 0;
+  width: 100%;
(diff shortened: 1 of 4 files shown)


src/views/BrowseView/BrowseView.style.ts
```diff
@@ -20,7 +25,7 @@ export const StyledCategoryPicker = styled(CategoryPicker)<IsAtTop>`
   top: ${NAVBAR_HEIGHT}px;
   padding: ${sizes(5)} var(--global-horizontal-padding) ${sizes(2)};
   margin: 0 calc(-1 * var(--global-horizontal-padding));
-  background-color: ${colors.black};
+  background-color: ${(props) => (props.isAtTop ? colors.transparent : colors.black)};
```
kdembler, line 28: Since we're changing that, I think it could be nice to add a `transition` for background so the change is more seamless

src/views/BrowseView/BrowseView.style.ts
```diff
@@ -30,12 +35,26 @@ export const StyledInfiniteVideoGrid = styled(InfiniteVideoGrid)`
...
 `
+
 export const StyledBackgroundPattern = styled(BackgroundPattern)`
   position: absolute;
-  right: 0;
+  top: 73px;
+  right: 80px;
   z-index: ${zIndex.background};
+  transform: scale(1.3);
+  @media screen and (min-width: ${breakpoints.base}) {
```
kdembler: Since we assume we'll always be at minimum `breakpoints.base`, we actually don't need that media query. We can just include `display: none` as a default and it will later get overridden

src/views/BrowseView/BrowseView.style.ts
```diff
@@ -30,12 +35,26 @@ export const StyledInfiniteVideoGrid = styled(InfiniteVideoGrid)`
...
 `
 export const IntersectionTarget = styled.div`
   min-height: 1px;
 `
+
 export const StyledBackgroundPattern = styled(BackgroundPattern)`
   position: absolute;
-  right: 0;
+  top: 73px;
+  right: 80px;
```
kdembler: Will this value ever be used? Below `small` we're not showing anything and later it's get set to `PATTERN_OFFSET.SMALL` I think?

src/views/BrowseView/BrowseView.tsx
```diff
@@ -48,24 +48,26 @@ const BrowseView: React.FC<RouteComponentProps> = () => {
...
-        loading={categoriesLoading}
-        selectedCategoryId={selectedCategoryId}
-        onChange={handleCategoryChange}
-        isAtTop={inView}
-      />
-      <ErrorBoundary fallback={ErrorFallback}>
-        <StyledInfiniteVideoGrid categoryId={selectedCategoryId || undefined} ready={!!selectedCategoryId} />
-      </ErrorBoundary>
-    </Container>
+    <>
```
kdembler: Do we need this `React.Fragment`? We only return one top-level element (`Container`) so we shouldn't need that or am I missing something?

kdembler (commented):
Looks much better now! Left some small comments but overall looks good :)

kdembler (approved):
LGTM!

PR #203, opened 2021-01-08
Add background pattern as a separate component, add it to channel view

mikkio-j:
- extracted bg pattern into separate component
- replaced styled component in channel view


kdembler (approved):
Looks good!

PR #205, opened 2021-01-11
Video mobile desktop

### LINKED ISSUE #144: Fix mobile version of video player
There are quite a few discrepancies, not least of which being the size of the title, viewport margin and the way the playback controls work, 

Screenshot 2020-12-10 at 11 33 43

while this is the design

Screenshot 2020-12-10 at 11 34 05

look at designs here, 

https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=1086%3A4948

but note that there are genuine things we do not want from the design just yet like the three "purple" buttons, the chromecast control and the settings control.

drillprop:
Fix #144 

I've added play/pause on touch functionality, but I kept the play button. I think we still need some indication that the video is playing or it's paused.

image


src/shared/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -102,7 +105,7 @@ const VideoPlayer: React.FC<VideoPlayerProps> = ({
         </PlayOverlay>
       )}
       <div data-vjs-player>
-        <video ref={playerRef} className="video-js" />
+        <video ref={playerRef} className="video-js" onTouchStart={handlePlayOverlayClick} />
```
kdembler: We may need some additional handling for that event. The way it works now if I try to scroll down (I need to touch the video surface to do that), my scrolling will trigger the pause. We should be able to pause only if the user taps, not starts touch movement

src/shared/components/VideoPlayer/VideoPlayer.style.tsx
```diff
@@ -46,6 +46,13 @@ export const Container = styled.div<ContainerProps>`
     height: ${sizes(16)} !important;
     align-items: center;
 
+    @media screen and (max-width: ${breakpoints.small}) {
```
kdembler: I may have failed to communicate that, but we do the mobile-first media queries everywhere in the app. So we first define the default values for the smallest screen (without media query) and then provide overrides via media queries with `min-width`

src/shared/components/VideoPlayer/VideoPlayer.style.tsx
```diff
@@ -101,18 +116,43 @@ export const Container = styled.div<ContainerProps>`
...
+            @media screen and (max-width: ${breakpoints.small}) {
+              position: absolute;
+              top: -5px;
+              content: '';
+              display: initial;
+              width: 14px;
+              height: 14px;
+              background: ${colors.blue[500]};
+              border-radius: 100%;
+              border: 2px solid #fff;
```
kdembler: We should use white from the theme here I think

src/views/VideoView/VideoView.style.tsx
```diff
@@ -82,3 +88,12 @@ export const LicenseContainer = styled.div`
     color: ${colors.gray[500]};
   }
 `
+
+export const TitleText = styled(Text)`
+  @media screen and (max-width: ${breakpoints.medium}) {
+    font-size: 32px;
```
kdembler: I think I'd be best to use values from `theme.typography.sizes` here and below. 32px will be `h3`

kdembler (commented):
Great improvement overall. Looks much better on mobile now. One additional thing that could be useful would be using the `playsinline` video attribute so that on mobile it doesn't go to fullscreen automatically and native browser controls aren't displayed. You'd need to add that in `shared/components/VideoPlayer/videoJsPlayers.ts` starting at line 45:
```ts
    const videoJsOptions: VideoJsPlayerOptions = {
      controls: true,
      // @ts-ignore @types/video.js is outdated and doesn't provide types for some newer video.js features
      playsinline: true,
    }
```

src/shared/components/VideoPlayer/VideoPlayer.tsx
drillprop: I think I'll revert this change because I'm not sure how to fix it in a relatively short time.  I'll just create another issue for this(and for the general mobile UX).

src/shared/components/VideoPlayer/VideoPlayer.style.tsx
```diff
@@ -42,12 +42,18 @@ export const Container = styled.div<ContainerProps>`
...
     font-family: ${typography.fonts.base};
-    background-color: rgba(0, 0, 0, 0.3);
-    height: ${sizes(16)} !important;
+    background: none;
+    margin-top: auto;
+    padding: 0;
+    z-index: ${zIndex.overlay + 1};
     align-items: center;
-
     /* account for progress bar on top */
```
kdembler: I think comment doesn't apply here anymore? We should probably move it along with `padding` prop

kdembler (approved):
Looks good!

PR #206, opened 2021-01-11
Fix horizontal spacing for player and playback controls

### LINKED ISSUE #182: Video playback page browser support (CSS)
When trying Atlas with slightly outdated Chrome version on desktop (78) and Opera Mini on mobile I wasn't able to see the video player on the video playback page (ie. https://play.joystream.org/video/1570).

I found that this is due to the use of `width: min(1250px, 100%)` style for one of the video player containers.

Both browsers didn't support this style (https://caniuse.com/css-math-functions) which resulted in the container having `0` width and `0` height, effectively making the player invisible.

Using math functions like `min`, `max` or `clamp` currently has relatively small browser support (~90% with prefixes), so I think it's worth to consider some alternative that has a higher browser support.

mikkio-j:
- player has full width on all devices
- videos have black bars if ratio doesn't match
- placeholder corresponds to video player

horizontal-spacing-video-player


src/shared/components/VideoPlayer/VideoPlayer.style.tsx
```diff
@@ -31,7 +31,10 @@ const backgroundContainerCss = css`
 
 export const Container = styled.div<ContainerProps>`
   position: relative;
-
+  height: 30vh;
```
kdembler: We shouldn't duplicate these values across two components (this and `PlayerPlaceholder`). I think we can move these styles to `PlayerContainer` and then set `height: 100%` for both this and `PlayerPlaceholder`?

kdembler (commented):
Nice! I think this doesn't fully fix the horizontal spacing thing though:

Screenshot 2021-01-11 at 13 31 30

There shouldn't be any (black) space on left or right. To achieve that we will need to add negative margin to the video player container so that it can negate the global horizontal padding.

src/shared/components/VideoPlayer/VideoPlayer.style.tsx
mikkio-j: Done

kdembler (approved):

PR #208, opened 2021-01-12
Save Interrupted Video State

### LINKED ISSUE #137: Enable saving interrupted videos
After #135 is done, we can add a support to `VideoPlayer`/`VideoView` (not sure which at the moment) to save the interrupted videos. This will probably need to happen periodically (we want to save some data if the user quits the app). Also, we should have some decent policy for when the video isn't marked as interrupted. For example, users may finish watching during last 20s or so and the video probably should be marked as watched.

fcescob:
Fix #137

### DIFF
--- a/src/hooks/usePersonalData/index.ts
+++ b/src/hooks/usePersonalData/index.ts
@@ -1,19 +1,1 @@
-import { useEffect, useRef } from 'react'
-
 export * from './usePersonalData'
-
-type Callback = (...args: unknown[]) => void
-export const useInterval = (cb: Callback, delay = 1000) => {
-  const savedCallback = useRef<Callback>()
-
-  useEffect(() => {
-    savedCallback.current = cb
-  }, [cb])
-
-  useEffect(() => {
-    if (savedCallback.current) {
-      const id = setInterval(() => savedCallback.current!(), delay)
-      return () => clearInterval(id)
-    }
-  }, [delay])
-}
--- a/src/hooks/usePersonalData/index.ts
+++ b/src/hooks/usePersonalData/index.ts
@@ -1,1 +1,19 @@
+import { useEffect, useRef } from 'react'
+
 export * from './usePersonalData'
+
+type Callback = (...args: unknown[]) => void
+export const useInterval = (cb: Callback, delay = 1000) => {
+  const savedCallback = useRef<Callback>()
+
+  useEffect(() => {
+    savedCallback.current = cb
+  }, [cb])
+
+  useEffect(() => {
+    if (savedCallback.current) {
+      const id = setInterval(() => savedCallback.current!(), delay)
+      return () => clearInterval(id)
+    }
+  }, [delay])
+}
--- a/src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
+++ b/src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
@@ -1,13 +1,3 @@
-import {
-  CompletedVideo,
-  COMPLETED_VIDEO,
-  FollowedChannel,
-  InterruptedVideo,
-  INTERRUPTED_VIDEO,
-  PersonalDataClient,
-  WatchedVideo,
-} from './types'
-
 const promisify = <T,>(fn: (...args: unknown[]) => T) => (...args: Parameters<typeof fn>) =>
   new Promise((resolve) => resolve(fn(...args))) as Promise<T>
 
@@ -121,3 +111,58 @@ const localStorageClient: PersonalDataClient = {
 }
 
 export default localStorageClient
+
+type WatchedVideoFields = {
+  id: string
+}
+
+export const INTERRUPTED_VIDEO = 'INTERRUPTED'
+export const COMPLETED_VIDEO = 'COMPLETED'
+
+type InterruptedVideo = WatchedVideoFields & {
+  __typename: typeof INTERRUPTED_VIDEO
+  timestamp: number
+}
+type CompletedVideo = WatchedVideoFields & {
+  __typename: typeof COMPLETED_VIDEO
+  id: string
+}
+export type WatchedVideo = InterruptedVideo | CompletedVideo
+
+export type FollowedChannel = {
+  id: string
+}
+
+export interface PersonalDataClient {
+  // ==== watched videos ====
+
+  // get all the interrupted or completed videos
+  watchedVideos: () => Promise<WatchedVideo[]>
+
+  // get all the interrupted videos
+  interruptedVideos: () => Promise<InterruptedVideo[]>
+
+  // get all the completed videos
+  completedVideos: () => Promise<CompletedVideo[]>
+
+  // get status of one specific watched video
+  watchedVideo: (id: string) => Promise<WatchedVideo | null>
+
+  // mark the video as interrupted or completed
+  setWatchedVideo: (
+    __typename: typeof INTERRUPTED_VIDEO | typeof COMPLETED_VIDEO,
+    id: string,
+    timestamp?: number
+  ) => Promise<void>
+
+  // ==== followed channels ====
+
+  // get all the followed channels
+  followedChannels: () => Promise<FollowedChannel[]>
+
+  // set the channel following status
+  setChannelFollowing: (id: string, follow: boolean) => Promise<void>
+
+  // get the following status of one specific channel
+  isFollowingChannel: (id: string) => Promise<boolean>
+}
--- a/src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
+++ b/src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
@@ -1,3 +1,13 @@
+import {
+  CompletedVideo,
+  COMPLETED_VIDEO,
+  FollowedChannel,
+  InterruptedVideo,
+  INTERRUPTED_VIDEO,
+  PersonalDataClient,
+  WatchedVideo,
+} from './types'
+
 const promisify = <T,>(fn: (...args: unknown[]) => T) => (...args: Parameters<typeof fn>) =>
   new Promise((resolve) => resolve(fn(...args))) as Promise<T>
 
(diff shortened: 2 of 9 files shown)


src/views/VideoView/VideoView.tsx
```diff
@@ -21,17 +21,25 @@ import { GetVideo, GetVideoVariables } from '@/api/queries/__generated__/GetVide
...
+  const videoTimestamp = useMemo(() => {
+    const currentVideo = state.watchedVideos.find((v) => v.id === data?.video?.id)
+    return currentVideo?.__typename === 'INTERRUPTED' ? currentVideo.timestamp : null
+  }, [data?.video?.id, state.watchedVideos])
+
+  const videoRef = useRef<HTMLVideoElement>(null)
   const videoID = data?.video?.id
 
-  const [playing, setPlaying] = useState<boolean>(true)
+  const [playing, setPlaying] = useState<boolean>()
```
kdembler: Let's pass the default value explicitly

src/views/VideoView/VideoView.tsx
```diff
@@ -70,6 +78,24 @@ const VideoView: React.FC<RouteComponentProps> = () => {
     })
   }, [addVideoView, videoID])
 
+  // Save the video timestamp
+  useInterval(() => {
```
kdembler: I think there may be a better approach to this. Now we're passing the video component ref and accessing it's internal state. In general that should be avoided if possible. Instead, we could have some callback (e.g. `onTimeUpdated`) on the video component and just pass some function to it that would update the saved state. It can be debounced so it doesn't save too often.

src/views/VideoView/VideoView.tsx
```diff
@@ -70,6 +78,24 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
+  }, 5000)
+
+  const handleVideoEnd = () => {
+    if (data?.video?.id) {
+      updateWatchedVideos('COMPLETED', data?.video?.id)
+    }
+  }
+  const handleVideoStart = () => {
+    if (videoTimestamp && videoRef.current) {
+      videoRef.current.currentTime = videoTimestamp
```
kdembler: This can be done by adding some prop like `startTime` or something to the video component

kdembler: Shouldn't we add `setPlaying(true)` here?

src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
```diff
@@ -121,3 +111,58 @@ const localStorageClient: PersonalDataClient = {
 }
 
 export default localStorageClient
+
```
kdembler: Any specific reason to move the types here? I think it was easier to read when they were in a separate file

src/views/VideoView/VideoView.tsx
fcescob: That doesn't work because it causes the player to restart if you hit pause, I can pass in `false` as default though

src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
fcescob: This is a workaround, the app won't start otherwise, see rocket chat for the full error logs

src/views/VideoView/VideoView.tsx
kdembler: That's what I meant

src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
kdembler: Seems renaming the file from types.d.ts to types.ts will do the trick. Let's do that

src/shared/components/VideoPlayer/videoJsPlayer.ts
```diff
@@ -150,5 +164,22 @@ export const useVideoJsPlayer: VideoJsPlayerHook = ({
     }
   }, [player, onPause])
 
+  useEffect(() => {
+    if (!player || !onEnd) {
+      return
+    }
+    player.on('ended', onEnd)
```
kdembler, line 171: We should probably add a cleanup for that as well

src/views/VideoView/VideoView.tsx
```diff
@@ -21,17 +22,25 @@ import { GetVideo, GetVideoVariables } from '@/api/queries/__generated__/GetVide
...
 
 const VideoView: React.FC<RouteComponentProps> = () => {
   const { id } = useParams()
   const { loading, data, error } = useQuery<GetVideo, GetVideoVariables>(GET_VIDEO, {
     variables: { id },
   })
   const [addVideoView] = useMutation<AddVideoView, AddVideoViewVariables>(ADD_VIDEO_VIEW)
+  const { state, updateWatchedVideos } = usePersonalData()
 
+  const videoTimestamp = useMemo(() => {
```
kdembler: I'm a bit concerned about that. _In theory_, React docs state that `useMemo` should be used only for performance and not as a semantic guarantee. _In theory_, future version of React could make `useMemo` recalculate the value from time to time. That would cause the video to start from the updated `startTime` again, right?

Speaking of which, now I'm wondering, `state.watchedVideos` is in the deps, won't that be changed during watching the video? When we do `updateWatchedVideos('INTERRUPTED', data.video.id, time)`?

src/views/VideoView/VideoView.tsx
```diff
@@ -21,17 +22,25 @@ import { GetVideo, GetVideoVariables } from '@/api/queries/__generated__/GetVide
...
+    const currentVideo = state.watchedVideos.find((v) => v.id === data?.video?.id)
+    return currentVideo?.__typename === 'INTERRUPTED' ? currentVideo.timestamp : 0
+  }, [data?.video?.id, state.watchedVideos])
+
+  const videoRef = useRef<HTMLVideoElement>(null)
```
kdembler: Do we still need the ref?

src/shared/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -9,13 +9,10 @@ type VideoPlayerProps = {
...
 } & VideoJsConfig
 
-const VideoPlayer: React.FC<VideoPlayerProps> = ({
-  className,
-  autoplay,
-  isInBackground,
-  playing,
-  ...videoJsConfig
-}) => {
+const VideoPlayer: React.ForwardRefRenderFunction<HTMLVideoElement, VideoPlayerProps> = (
```
kdembler, line 12: I don't think we need to pass the ref anymore?

src/shared/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -86,6 +83,17 @@ const VideoPlayer: React.FC<VideoPlayerProps> = ({
     }
   })
 
+  useEffect(() => {
```
kdembler, line 86: Probably not needed anymore

src/hooks/usePersonalData/index.ts
```diff
@@ -1 +1,19 @@
+import { useEffect, useRef } from 'react'
+
 export * from './usePersonalData'
+
+type Callback = (...args: unknown[]) => void
+export const useInterval = (cb: Callback, delay = 1000) => {
```
kdembler: I don't think we need that anymore?

src/views/VideoView/VideoView.tsx
```diff
@@ -21,17 +22,25 @@ import { GetVideo, GetVideoVariables } from '@/api/queries/__generated__/GetVide
...
   const videoID = data?.video?.id
 
-  const [playing, setPlaying] = useState<boolean>(true)
+  const [playing, setPlaying] = useState(false)
```
kdembler: This will basically remove the autoplay, so we need to do something about it. Since we're controlling the playback state outside of the player, we can remove the `autoplay` prop and something like this:

```tsx
const handlePlay = () => {
  setVideoPlaying(true)
}

const handlePause = () => {
  setVideoPlaying(false)
}

// ...

<VideoPlayer
  // ...
  playing={videoPlaying}
  onPlay={handlePlay}
  onPause={handlePause}
/>
```

src/views/VideoView/VideoView.tsx
kdembler: We can just run `useEffect once on mount here and save the timestamp to some state.

src/views/VideoView/VideoView.tsx
```diff
@@ -21,17 +22,28 @@ import { GetVideo, GetVideoVariables } from '@/api/queries/__generated__/GetVide
...
   const { id } = useParams()
   const { loading, data, error } = useQuery<GetVideo, GetVideoVariables>(GET_VIDEO, {
     variables: { id },
   })
   const [addVideoView] = useMutation<AddVideoView, AddVideoViewVariables>(ADD_VIDEO_VIEW)
+  const { state, updateWatchedVideos } = usePersonalData()
+
+  const [startTimestamp, setStartTimestamp] = useState<number>()
+  useEffect(() => {
+    if (startTimestamp) {
```
kdembler: We probably should initialize `startTimestamp` with `null` and here check for  `startTimestamp != null`. Like it is right now, if the initial timestamp is 0, it will get updated once anyway

src/views/VideoView/VideoView.tsx
```diff
@@ -70,6 +82,30 @@ const VideoView: React.FC<RouteComponentProps> = () => {
     })
   }, [addVideoView, videoID])
 
+  // Save the video timestamp
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
kdembler, line 89: Why do we disable that rule? The only missing dependency here is `updateWatchedVideos` I think? And this one should be stable so no trouble in including it?

src/views/VideoView/VideoView.tsx
```diff
@@ -70,6 +82,30 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
+  const handleTimeUpdate = useCallback(
+    debounce((time) => {
+      if (data?.video?.id) {
+        updateWatchedVideos('INTERRUPTED', data.video.id, time)
+      }
+    }, 5000),
+    [data?.video?.id]
+  )
+
+  const handleVideoEnd = () => {
```
kdembler: Now I think of it, we probably should put all of those handlers in `useCallback` (also `handlePlay` and `handlePause`). Otherwise the event handlers will be set/unset on every render.

kdembler (approved):
LGTM

PR #209, opened 2021-01-13
Nav rework

### LINKED ISSUE #162: Rework the top bar/nav menu
Currently we use only top bar for all the navigation. We should rework the app to use the navigation from the design. That means:
- reintroducing the left-side nav menu (we have a component for that (`shared/components/SideNav`), may need some updates) 
- removing the navigation from the top bar
- centring the search bar in the top bar

Design for reference: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=2159%3A0

Some things to take into account:
- depending on how hard this proves to be, it'd be nice for the sidenav to autocollapse in the video view
- the design includes both major and minor (sub) entries - for now we will use only that major ones

drillprop:
Fix #162


src/components/LayoutWithRouting.tsx
```diff
@@ -38,6 +55,7 @@ const LayoutWithRouting: React.FC = () => {
     <>
       <GlobalStyle additionalStyles={globalStyles} />
       <Navbar default />
```
kdembler: Just a thought now. Maybe we could rename the `Navbar` to `Topbar` or something? Now the navigation is in the `Sidenav`, so this way we can avoid some confusion

kdembler (approved):
LGTM!

PR #211, opened 2021-01-14
Update the Browse view

### LINKED ISSUE #160: Update the Browse view
Exactly the same as the current "Browse" screen for the Babylon release:
- title is "Videos"
- a carousel, titled "Featured", between the title and the category selector, which contains a set of featured videos. These videos are not from a single category, which is part of the reason why this carousel lives above the category selector. We can play with allowing these videos to be larger than the videos in the infinite scrollers also. When scrolling, these videos are, like the page title, only visible when at the very top of the page, unlike the category selector. (we currently have this carousel in the home page)
- introduce an "All" meta category which is selected by default.
- videos are ordered by most recently created first.

mikkio-j:
Fix #160


src/mocking/data/raw/categories.json
```diff
@@ -1,4 +1,8 @@
 [
+  {
```
kdembler: `All` shouldn't be a category by itself. It should only exist as part of the category picker

src/mocking/server/data.ts
```diff
@@ -65,12 +78,16 @@ export const createMockData = (server: MirageJSServer) => {
...
       license: licenseEntities[idx % licenseEntities.length],
     })
+  })
 
-    server.create('VideoViewsInfo', {
-      id: video.id,
-      views: video.views,
+  videos
+    .filter((_, idx) => FEATURED_VIDEOS_INDEXES.includes(idx))
+    .map((video) => {
```
kdembler: I guess this `map` was here when you returned the value previously, I think now `forEach` will work fine?

src/shared/components/InfiniteVideoGrid/InfiniteVideoGrid.tsx
```diff
@@ -124,6 +124,16 @@ const InfiniteVideoGrid: React.FC<InfiniteVideoGridProps> = ({
     }
   }, [loading, loadedVideosCount, targetLoadedVideosCount, allVideosLoaded, fetchMore, endCursor, getFetchVariables])
 
+  useEffect(() => {
```
kdembler: Why is this needed? We already have code for fetching data

src/views/BrowseView/BrowseView.tsx
```diff
@@ -62,7 +79,11 @@ const BrowseView: React.FC<RouteComponentProps> = () => {
         isAtTop={inView}
       />
       <ErrorBoundary fallback={ErrorFallback}>
-        <StyledInfiniteVideoGrid categoryId={selectedCategoryId || undefined} ready={!!selectedCategoryId} />
+        {allCategoryId !== selectedCategoryId ? (
```
kdembler: We should only use one `StyledInfiniteVideoGrid` and change its props

src/api/queries/videos.ts
```diff
@@ -85,7 +85,7 @@ export const GET_NEWEST_VIDEOS = gql`
 
 export const GET_FEATURED_VIDEOS = gql`
   query GetFeaturedVideos {
-    featuredVideos(orderBy: createdAt_DESC) {
+    featuredVideos {
```
kdembler: We still need that argument. If mocking code returns incorrect results because of it, we need to change the mocking resolver to strip that argument instead of removing it from the query

kdembler (changes requested):

src/shared/components/CategoryPicker/CategoryPicker.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useState, useRef } from 'react'
+import * as _ from 'lodash'
```
kdembler: We shouldn't use `import *` syntax if not necessary. Doing that will force the full library to get included in the built bundle, increasing its size. If you import only specific members, like `import { cloneDeep } from 'lodash`, Webpack should be able to do treeshaking and only include the required members in the bundle.

src/shared/components/CategoryPicker/CategoryPicker.tsx
```diff
@@ -19,12 +26,15 @@ const CategoryPicker: React.FC<CategoryPickerProps> = ({
   onChange,
   className,
 }) => {
+  const categoriesCopy = categories && _.cloneDeep(categories)
```
kdembler: Let's use array spread to create `displayedCategories`

src/shared/components/InfiniteVideoGrid/InfiniteVideoGrid.tsx
```diff
@@ -71,14 +71,14 @@ const InfiniteVideoGrid: React.FC<InfiniteVideoGridProps> = ({
   )
 
   useEffect(() => {
-    if (ready && !called) {
+    if ((ready && !called) || categoryId === cachedCategoryId) {
```
kdembler: Why this change?

src/views/BrowseView/BrowseView.tsx
```diff
@@ -46,12 +55,20 @@ const BrowseView: React.FC<RouteComponentProps> = () => {
...
 
+  const isAllCategorySelected = selectedCategoryId === 'All'
+  const featuredVideos = featuredVideosData?.featuredVideos.map((featuredVideo) => featuredVideo.video)
+  const hasFeaturedVideosError = featuredVideosError && !featuredVideosLoading
   return (
     <Container>
       <BackgroundPattern />
       <Header variant="hero" ref={headerRef}>
-        Browse
+        Videos
```
kdembler: Since the view name changed, let's update the routing URL and nav link in the navbar. Let's also use the icon from here: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3560%3A1372

src/views/BrowseView/BrowseView.tsx
```diff
@@ -46,12 +55,20 @@ const BrowseView: React.FC<RouteComponentProps> = () => {
...
       </Header>
+      {!hasFeaturedVideosError ? (
+        <VideoGallery title="Featured" loading={featuredVideosLoading} videos={featuredVideos} />
+      ) : (
+        <ErrorFallback error={featuredVideosError} resetError={() => refetchFeaturedVideos()} />
+      )}
       <Text variant="h5">Topics that may interest you</Text>
```
kdembler: We could use some spacing between the gallery and this header

src/views/BrowseView/BrowseView.tsx
```diff
@@ -62,7 +79,10 @@ const BrowseView: React.FC<RouteComponentProps> = () => {
         isAtTop={inView}
       />
       <ErrorBoundary fallback={ErrorFallback}>
-        <StyledInfiniteVideoGrid categoryId={selectedCategoryId || undefined} ready={!!selectedCategoryId} />
+        <StyledInfiniteVideoGrid
+          categoryId={isAllCategorySelected ? undefined : selectedCategoryId || undefined}
```
kdembler: This seems too complicated than needs to be, details during the call

src/shared/components/ToggleButton/ToggleButton.styles.tsx
```diff
@@ -56,7 +56,7 @@ const colorsFromProps = ({ variant, toggled }: ToggleButtonStyleProps) => {
     case 'secondary': {
       styles = css`
         color: ${colors.white};
-        background-color: ${toggled ? colors.blue[500] : colors.transparent};
+        background-color: ${toggled ? colors.blue[500] : colors.black};
```
kdembler: I think those can still be `transparent`

src/views/VideosView/VideosView.tsx
```diff
@@ -34,14 +34,14 @@ const VideosView: React.FC<RouteComponentProps> = () => {
...
+  const topicsRef = useRef<HTMLHeadingElement>(null)
   const { ref: targetRef, inView } = useInView({
     rootMargin: `-${TOP_NAVBAR_HEIGHT - GRID_TOP_PADDING}px 0px 0px`,
   })
   const handleCategoryChange = (categoryId: string | null, scrollTop = true) => {
     setSelectedCategoryId(categoryId)
-    if (headerRef.current && scrollTop) {
-      headerRef.current.scrollIntoView({ block: 'end', inline: 'nearest', behavior: 'smooth' })
+    if (topicsRef.current && scrollTop) {
+      topicsRef.current.scrollIntoView({ block: 'start', inline: 'nearest' })
```
kdembler:
```suggestion
      setTimeout(() => {
        topicsRef.current?.scrollIntoView({ behavior: 'smooth' })
      })
```

PR #212, opened 2021-01-14
Follower count

### LINKED ISSUE #138: Introduce follower count on channel cover area
### Blocked by https://github.com/Joystream/orion/issues/3

As in the title

drillprop:
Fix #138 

I'm not sure if I did it right. Should I add followers to the localstorage?


src/api/client/resolvers.ts
```diff
@@ -29,6 +53,8 @@ export const queryNodeStitchingResolvers = (
     featuredVideos: createResolverWithoutVideoViewsField(queryNodeSchema, 'featuredVideos'),
     search: createResolverWithoutVideoViewsField(queryNodeSchema, 'search'),
```
kdembler: `search` is tricky because it deals with both videos and channels so will require both transforms to work properly. My suggestion is that we can create one function called `createResolverWithTransforms` that takes 3 params (`schema`, `fieldName` and `transforms` array), basically almost the same as what we have in `createResolverWithoutVideoViewsField`. Then here we can just pass an array of transforms for each query so it's easy to do both transform for the `search` query. Makes sense?

src/api/client/transforms/orionFollows.ts
```diff
@@ -0,0 +1,76 @@
...
+        kind: 'Name',
+        value: 'follows',
+      },
+    },
+  ],
+}
+
+export const ORION_FOLLOWS_QUERY_NAME = 'channelFollows'
+
+// Transform a request to expect VideoViewsInfo return type instead of an Int
```
kdembler:
```suggestion
// Transform a request to expect ChannelFollowsInfo return type instead of an Int
```

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -35,6 +35,8 @@ type Channel {
   language: Language
 
   videos: [Video!]!
+
+  follows: Int
```
kdembler, line 39:
```suggestion
  # extended from Orion
  follows: Int
```

src/mocking/server/data.ts
```diff
@@ -16,6 +16,12 @@ import {
 type MirageJSServer = any
 
 export const createMockData = (server: MirageJSServer) => {
+  mockChannels.forEach((channel, idx) => {
```
kdembler: Small thing, but maybe we can do that in `mockChannels.map` below, as part of the same iteration? Like we do for `mockVideos`

src/views/ChannelView/ChannelView.tsx
```diff
@@ -59,6 +61,7 @@ const ChannelView: React.FC<RouteComponentProps> = () => {
           {data?.channel ? (
             <TitleContainer>
               <Title variant="h1">{data.channel.handle}</Title>
+              <SubTitle>{data.channel.follows ? formatNumberShort(data.channel.follows) : 0} Followers</SubTitle>
```
kdembler: We should add a placeholder for that as well, like the title one

kdembler (approved):
Some small stuff, otherwise great job!

PR #218, opened 2021-01-15
refactor infinite grid, add channels view

### LINKED ISSUE #161: Create the Channels view
Exactly the same as current "Browse" screen for the Babylon release, except

- title is "Channels"
- no carousel with featured channels
- no category picker - just an infinite grid of channels

Complexity here lies mostly in the infinite scroller part as it doesn't have any kind of support for channels

kdembler:
Fix #161

Adds a channel view and slightly refactor the infinite grid. The new `useInfiniteGrid` hook can be used to create infinite grids with arbitrary content, not only videos and channels. The grid code was simplified a bit and some bugs were resolved.


src/shared/components/Grid/Grid.tsx
```diff
@@ -61,6 +48,27 @@ const Grid: React.FC<GridProps> = ({
   )
 }
 
+type ContainerProps = Required<Pick<GridProps, 'gap' | 'maxColumns' | 'minWidth' | 'repeat'>>
+
+const maxColumnsCss = ({ maxColumns }: ContainerProps) =>
```
fcescob, line 53: Why is this needed exactly?

fcescob (changes requested):

src/shared/components/Grid/Grid.tsx
kdembler: So that we don't have to use `maxColumns` every time. For channel previews, which don't fill available space, max columns don't make much sense

fcescob (approved):

PR #224, opened 2021-01-19
Follow/unfollow channel

mikkio-j:
- add button for following/unfollowing channels
- followed channels are saved in local storage
- add mock and mutation queries for following

### DIFF
--- a/src/api/queries/channels.ts
+++ b/src/api/queries/channels.ts
@@ -45,3 +45,12 @@ export const GET_CHANNEL = gql`
   }
   ${allChannelFieldsFragment}
 `
+
+export const FOLLOW_CHANNEL = gql`
+  mutation FollowChannel($id: ID!) {
+    followChannel(where: { id: $id }) {
+      ...AllChannelFields
+    }
+  }
+  ${allChannelFieldsFragment}
+`
--- a/src/api/queries/channels.ts
+++ b/src/api/queries/channels.ts
@@ -47,7 +47,7 @@ export const GET_CHANNEL = gql`
 `
 
 export const FOLLOW_CHANNEL = gql`
-  mutation followChannel($channelId: ID!) {
+  mutation FollowChannel($channelId: ID!) {
     followChannel(channelId: $channelId) {
       id
       follows
@@ -55,7 +55,7 @@ export const FOLLOW_CHANNEL = gql`
   }
 `
 export const UNFOLLOW_CHANNEL = gql`
-  mutation unfollowChannel($channelId: ID!) {
+  mutation UnfollowChannel($channelId: ID!) {
     unfollowChannel(channelId: $channelId) {
       id
       follows
--- a/src/api/queries/channels.ts
+++ b/src/api/queries/channels.ts
@@ -48,7 +48,7 @@ export const GET_CHANNEL = gql`
 
 export const FOLLOW_CHANNEL = gql`
   mutation followChannel($channelId: ID!) {
-    followChannel(where: { id: $channelId }) {
+    followChannel(channelId: $channelId) {
       id
       follows
     }
@@ -56,7 +56,7 @@ export const FOLLOW_CHANNEL = gql`
 `
 export const UNFOLLOW_CHANNEL = gql`
   mutation unfollowChannel($channelId: ID!) {
-    unfollowChannel(where: { id: $channelId }) {
+    unfollowChannel(channelId: $channelId) {
       id
       follows
     }
--- a/src/api/queries/channels.ts
+++ b/src/api/queries/channels.ts
@@ -47,10 +47,18 @@ export const GET_CHANNEL = gql`
 `
 
 export const FOLLOW_CHANNEL = gql`
-  mutation FollowChannel($id: ID!) {
-    followChannel(where: { id: $id }) {
-      ...AllChannelFields
+  mutation followChannel($channelId: ID!) {
+    followChannel(where: { id: $channelId }) {
+      id
+      follows
     }
   }
-  ${allChannelFieldsFragment}
+`
+export const UNFOLLOW_CHANNEL = gql`
+  mutation unfollowChannel($channelId: ID!) {
+    unfollowChannel(where: { id: $channelId }) {
+      id
+      follows
+    }
+  }
 `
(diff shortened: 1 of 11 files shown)


src/mocking/server/resolvers.ts
```diff
@@ -115,6 +115,38 @@ export const channelFollowsResolver: QueryResolver<ChannelFollowsArgs> = (obj, a
...
+  if (!channelInfo) {
+    const channelInfo = context.mirageSchema.channelFollowsInfos.create({
+      id: args.channelId,
+      follows: 0,
+    })
+
+    return channelInfo
+  }
+  channelInfo.update({
+    follows: channelInfo.follows - 1,
```
kdembler:
```suggestion
    follows: Math.max(0, channelInfo.follows - 1),
```

src/views/ChannelView/ChannelView.tsx
```diff
@@ -18,18 +20,56 @@ import {
...
+type FollowedChannel = {
+  id: string
+}
+
 const ChannelView: React.FC<RouteComponentProps> = () => {
   const { id } = useParams()
   const { data, loading, error } = useQuery<GetChannel, GetChannelVariables>(GET_CHANNEL, {
     variables: { id },
   })
+  const [followChannel] = useMutation<GetChannelVariables>(FOLLOW_CHANNEL, {
```
kdembler: The generic types here should be specific for this mutation. I think you haven't run the codegen for the changes. Run `yarn queries:codegen` and it should create arg types for these queries.

Also it seems you are using `GetChannelVariables` as the return type which is incorrect

src/views/ChannelView/ChannelView.tsx
```diff
@@ -18,18 +20,56 @@ import {
...
+  const [unfollowChannel] = useMutation<GetChannelVariables>(UNFOLLOW_CHANNEL, {
+    variables: {
+      channelId: id,
+    },
+  })
+  const { updateChannelFollowing } = usePersonalData()
+  const [isFollowing, setFollowing] = useState<boolean>()
+
+  useEffect(() => {
+    const followedChannels = localStorage.getItem('followedChannels')
```
kdembler: You shouldn't ever acces this directly. That's what the `usePersonalData` hook is for. It will also do the parsing for you

src/views/ChannelView/ChannelView.tsx
```diff
@@ -71,6 +111,9 @@ const ChannelView: React.FC<RouteComponentProps> = () => {
               </>
             )}
           </TitleContainer>
+          <StyledButton variant={isFollowing ? 'secondary' : 'primary'} onClick={handleFollow}>
```
kdembler: Some issues here:

1. I know the design says otherwise but I think for the time being this should be centred vertically w.r.t. the channel avatar.
2. Take a look at [this example](https://deploy-preview-224--atlas-development.netlify.app/channel/0d75be46-9e81-4d79-a38f-753fbec0adf6) at `600px` wide screen. There are some issues here:
  2.1. The channel title doesn't collapse forcing the button outside the screen. This was introduced recently. You can fix that by adding `overflow: hidden;` to the `TitleContainer`
  2.2. Even with the previous point out of the way, there needs to be some padding between the title and the button.
3. Once you go below `600px` wide screen, there needs to be some spacing between the button and the followers count

kdembler (changes requested):

src/api/queries/channels.ts
```diff
@@ -45,3 +45,20 @@ export const GET_CHANNEL = gql`
   }
   ${allChannelFieldsFragment}
 `
+
+export const FOLLOW_CHANNEL = gql`
+  mutation followChannel($channelId: ID!) {
```
kdembler: Sorry, just noticed that, we use PascalCase for queries and mutations. So we should just update that and then run codegen again

src/mocking/server/resolvers.ts
```diff
@@ -115,6 +115,38 @@ export const channelFollowsResolver: QueryResolver<ChannelFollowsArgs> = (obj, a
...
+  if (!channelInfo) {
+    const channelInfo = context.mirageSchema.channelFollowsInfos.create({
+      id: args.channelId,
+      follows: 0,
+    })
+
+    return channelInfo
+  }
+  channelInfo.update({
+    follows: Math.max(channelInfo.follows - 1),
```
kdembler: What's the point of that `Math.max`? 😛

src/views/ChannelView/ChannelView.style.tsx
```diff
@@ -216,3 +217,13 @@ export const SubTitlePlaceholder = styled(Placeholder)`
...
   }
 `
+
+export const StyledButton = styled(Button)`
+  align-self: flex-start;
+  margin-top: ${sizes(2)};
+  @media screen and (min-width: ${breakpoints.small}) {
+    margin-left: auto;
+    align-self: center;
+    margin-top: -31px;
```
kdembler: What's the point of this margin? I think we've agreed to the button being centred vertically

src/views/ChannelView/ChannelView.style.tsx
```diff
@@ -158,8 +158,9 @@ export const TitleSection = styled.div`
 `
 export const TitleContainer = styled.div`
   max-width: 100%;
-  @media screen and (min-width: ${breakpoints.medium}) {
-    max-width: 60%;
+  overflow: hidden;
+  @media screen and (min-width: ${breakpoints.small}) {
+    max-width: 53%;
```
kdembler: Why was this change made? This is `max-width`, it should stay the same. If the follow button takes up some space, then `TitleContainer` will just shrink

src/views/ChannelView/ChannelView.style.tsx
```diff
@@ -216,3 +217,13 @@ export const SubTitlePlaceholder = styled(Placeholder)`
     height: ${SUBTITLE_HEIGHT};
   }
 `
+
+export const StyledButton = styled(Button)`
+  align-self: flex-start;
```
kdembler: We don't need that and the one from the media query, that should work out of the box

src/views/ChannelView/ChannelView.style.tsx
kdembler: If we want some padding between the button and the container, this isn't the way to go. Manual value like this is easy to break in the future, for example if the button width changes. What you can do to achieve that spacing is:
a) add `margin-left` other than `auto` to the button, that would describe the spacing and make sure it's positioned to the right of the screen some other way
b) put the button into some container that would define `padding-left`
c) add `padding-right` to the `TitleContainer`, although I think the previous approach may be better

kdembler (changes requested):

PR #226, opened 2021-01-20
Allow linking to specific timestamp

### LINKED ISSUE #136: Allow deep linking to a specific timestamp in the video playback
To allow continuing playback of interrupted videos, we should add on option to deep link to a specific moment in a video. We can use something similar to youtube, e.g. a query param `?time=128s` or something similar. The `VideoView` should take that param and pass it over to the `VideoPlayer`

drillprop:
Fix #136

### DIFF
--- a/src/views/VideoView/VideoView.tsx
+++ b/src/views/VideoView/VideoView.tsx
@@ -23,7 +23,7 @@ import { formatVideoViewsAndDate } from '@/utils/video'
 import { AddVideoView, AddVideoViewVariables } from '@/api/queries/__generated__/AddVideoView'
 
 import { ChannelLink, InfiniteVideoGrid } from '@/components'
-import { usePersonalData } from '@/hooks'
+import { usePersonalData, useRouterQuery } from '@/hooks'
 
 const VideoView: React.FC<RouteComponentProps> = () => {
   const { id } = useParams()
@@ -32,6 +32,7 @@ const VideoView: React.FC<RouteComponentProps> = () => {
   })
   const [addVideoView] = useMutation<AddVideoView, AddVideoViewVariables>(ADD_VIDEO_VIEW)
   const { state, updateWatchedVideos } = usePersonalData()
+  const timestampFromQuery = Number(useRouterQuery('time'))
 
   const [startTimestamp, setStartTimestamp] = useState<number>()
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


src/views/VideoView/VideoView.tsx
```diff
@@ -32,15 +32,23 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
 
   const [startTimestamp, setStartTimestamp] = useState<number>()
   useEffect(() => {
     if (startTimestamp != null) {
       return
     }
     const currentVideo = state.watchedVideos.find((v) => v.id === data?.video?.id)
-    setStartTimestamp(currentVideo?.__typename === 'INTERRUPTED' ? currentVideo.timestamp : 0)
-  }, [data?.video?.id, state.watchedVideos, startTimestamp])
+    const duration = data?.video?.duration ? Number(data?.video?.duration) : 0
```
kdembler: Cannot we do just
```suggestion
    const duration = data?.video?.duration ?? 0
```

src/views/VideoView/VideoView.tsx
```diff
@@ -32,15 +32,23 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
+
+    if (timestampFromQuery && timestampFromQuery < duration) {
```
kdembler: if `timestampFromQuery < duration` is false, we still need to do something here. If we just don't call `setStartTimestamp`, the initial condition in this `useEffect` will never fail so we will just keep calling this whenever the `state.watchedVideos` changes

src/views/VideoView/VideoView.tsx
```diff
@@ -40,15 +40,19 @@ const VideoView: React.FC<RouteComponentProps> = () => {
...
+  }, [data?.video?.id, state.watchedVideos, startTimestamp, data?.video?.duration])
+
+  useEffect(() => {
+    const duration = data?.video?.duration ?? 0
+    if (!timestampFromQuery || timestampFromQuery > duration) {
+      return
     }
-    if (!timestampFromQuery) {
-      setStartTimestamp(currentVideo?.__typename === 'INTERRUPTED' ? currentVideo.timestamp : 0)
+    if (timestampFromQuery < duration) {
```
kdembler: I don't think we need that `if`, right? Since this is already taken care of above

kdembler (approved):

PR #227, opened 2021-01-20
Add RecentSearches to local storage

fcescob:


src/hooks/usePersonalData/localStorageClient/types.ts
```diff
@@ -51,4 +56,12 @@ export interface PersonalDataClient {
...
   // get the following status of one specific channel
   isFollowingChannel: (id: string) => Promise<boolean>
+
+  // === recent searches ===
+
+  // get all recent searches
+  recentSearches: () => Promise<RecentSearch[]>
+
+  // add a recent search
+  setRecentSearch: (id: string, type: 'video' | 'channel') => Promise<void>
```
kdembler, line 66: Maybe this should be called `addRecentSearch`?

src/hooks/usePersonalData/usePersonalData.tsx
```diff
@@ -77,11 +88,25 @@ const usePersonalData = () => {
     },
     [dispatch]
   )
-  return { state, updateWatchedVideos, updateChannelFollowing }
+
+  const updateRecentSearches = useCallback(
```
kdembler, line 92: The purpose of this function isn't to update recent searches though, right? It's a side effect but the main purpose is to add a recent search, so let's make the function name communicate that

PR #228, opened 2021-01-21
Sidenavbar update

### LINKED ISSUE #220: Update sidenav to the latest design
We will need to slightly update our sidenav to match the latest design: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3560%3A1372

Specifically:
- Update styling
  - Only filled icons, active item indicated by darkened background
  - Nav items should take the full height so they are easier to click
  - We can skip the icon label in the collapsed version for now
  - Update the color
- Introduce "Followed channels" section

drillprop:
Fix #220


src/shared/components/SideNavbar/FollowedChannels.style.ts
```diff
@@ -0,0 +1,62 @@
+import { sizes, colors, typography } from '@/shared/theme'
+import styled from '@emotion/styled'
+import ChannelLink from '../../../components/ChannelLink'
```
kdembler: If this imports from `src/components` then it shouldn't be in `src/shared`. `src/shared` in general is meant to host components that could be used outside of Atlas. It's a bit messy at the moment though. For now let's just move this into `src/components`

kdembler: We can move the full sidenavbar there actually

src/shared/components/SideNavbar/FollowedChannels.tsx
```diff
@@ -0,0 +1,65 @@
...
+  expanded: boolean
+  onClick: () => void
+}
+
+const FollowedChannels: React.FC<FollowedChannels> = ({ followedChannels, expanded, onClick }) => {
+  const [isShowingMore, setIsShowingMore] = useState(false)
+
+  const numberOfChannels = followedChannels.length
+  const channelsToSlice = isShowingMore ? numberOfChannels : MAX_CHANNELS
+  const channels = numberOfChannels > MAX_CHANNELS ? followedChannels.slice(0, channelsToSlice) : followedChannels
```
kdembler: I don't think we need that condition here

```suggestion
  const channels = followedChannels.slice(0, channelsToSlice)
```

src/shared/components/SideNavbar/FollowedChannels.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <>
+      <CSSTransition
+        in={expanded}
+        unmountOnExit
+        timeout={parseInt(transitions.timings.loading)}
+        classNames={transitions.names.fade}
+      >
+        <ChannelsTitle variant="h6">{numberOfChannels ? 'Following Channels' : 'No Following Channels'}</ChannelsTitle>
```
kdembler: Let's omit this header altogether if there are no followed channels. Also let's use `Followed Channels` instead of `Following Channels`

src/shared/components/SideNavbar/FollowedChannels.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const MAX_CHANNELS = 4
+
+type FollowedChannels = {
+  followedChannels: FollowedChannel[]
+  expanded: boolean
+  onClick: () => void
+}
+
+const FollowedChannels: React.FC<FollowedChannels> = ({ followedChannels, expanded, onClick }) => {
```
kdembler: We usually add `Props` suffix to all prop types

```suggestion
type FollowedChannelsProps = {
  followedChannels: FollowedChannel[]
  expanded: boolean
  onClick: () => void
}

const FollowedChannels: React.FC<FollowedChannelsProps> = ({ followedChannels, expanded, onClick }) => {
```

kdembler (changes requested):
Great job overall, looks really nice!

There are some issues (not all are introduced by your changes, but are surfacing now and will need a fix):
1. ~The infinite scroller doesn't properly update on channel change~ - fixed by my commit
2. You've used the `ChannelLink` component to display channels in the sidenav, which is good, but we will need to adjust it a bit. At the moment component used `useApolloClient().readFragment()` to get data directly from cache. That means it _will not_ trigger a network request, so if the resource isn't loaded in some other way, it will end up loading forever. This worked fine so far, because the component wasn't used in cases where the data wasn't already there. We probably will need to make a real query there
3. I'm getting this warning
Screenshot 2021-01-25 at 11 24 19

kdembler (approved):

PR #230, opened 2021-01-22
"Continue watching" carousel

### LINKED ISSUE #219: Create "Continue watching" carousel
We should create a component for the "Continue watching" functionality. This will be a carousel composed of videos that were started but not finished (so present in the `usePersonalData().state.watchedVideos` with the correct `__typename`). All of them should have the duration marked with the duration `prop` and on click navigate to the video which will start the playback at the correct time.

Once ready, the component should be added to the home view, as the top most data, right below the cover video.

mikkio-j:
Fix #219

### DIFF
--- a/src/api/queries/__generated__/VideoFields.ts
+++ b/src/api/queries/__generated__/VideoFields.ts
@@ -72,6 +72,4 @@ export interface VideoFields {
   media: VideoFields_media;
   channel: VideoFields_channel;
   license: VideoFields_license;
-  timestamp?: number;
-  progress?:number
 }
--- a/src/api/queries/__generated__/VideoFields.ts
+++ b/src/api/queries/__generated__/VideoFields.ts
@@ -72,4 +72,6 @@ export interface VideoFields {
   media: VideoFields_media;
   channel: VideoFields_channel;
   license: VideoFields_license;
+  timestamp?: number;
+  progress?:number
 }
--- a/src/api/schemas/extendedQueryNode.graphql
+++ b/src/api/schemas/extendedQueryNode.graphql
@@ -273,7 +273,7 @@ type Query {
   video(where: VideoWhereUniqueInput!): Video
 
   # Lookup videos by where params
-  videos(where: VideoWhereInput): [Video]
+  videos(where: VideoWhereInput): [Video!]
 
   # List all videos by given constraints
   videosConnection(first: Int, after: String, where: VideoWhereInput, orderBy: VideoOrderByInput): VideoConnection!
--- a/src/api/schemas/extendedQueryNode.graphql
+++ b/src/api/schemas/extendedQueryNode.graphql
@@ -273,7 +273,7 @@ type Query {
   video(where: VideoWhereUniqueInput!): Video
 
   # Lookup videos by where params
-  videos(where: VideoWhereInput): [Video!]
+  videos(where: VideoWhereInput): [Video!]!
 
   # List all videos by given constraints
   videosConnection(first: Int, after: String, where: VideoWhereInput, orderBy: VideoOrderByInput): VideoConnection!
--- a/src/api/schemas/extendedQueryNode.graphql
+++ b/src/api/schemas/extendedQueryNode.graphql
@@ -213,6 +213,7 @@ input VideoWhereInput {
   channelId_eq: ID
   isCurated_eq: Boolean
   isPublic_eq: Boolean
+  id_in: [ID!]
 }
 
 input VideoWhereUniqueInput {
--- a/src/api/schemas/extendedQueryNode.graphql
+++ b/src/api/schemas/extendedQueryNode.graphql
@@ -272,6 +272,9 @@ type Query {
   # Lookup video by its ID
   video(where: VideoWhereUniqueInput!): Video
 
+  # Lookup videos by where params
+  videos(where: VideoWhereInput): [Video]
+
   # List all videos by given constraints
   videosConnection(first: Int, after: String, where: VideoWhereInput, orderBy: VideoOrderByInput): VideoConnection!
 
--- a/src/components/InterruptedVideosGallery.tsx
+++ b/src/components/InterruptedVideosGallery.tsx
@@ -1,0 +1,42 @@
+import React from 'react'
+import { RouteComponentProps } from '@reach/router'
+
+import { useQuery } from '@apollo/client'
+
+import { usePersonalData } from '@/hooks'
+
+import { GET_VIDEOS_WITH_IDS } from '@/api/queries'
+import { GetVideosWithIds, GetVideosWithIdsVariables } from '@/api/queries/__generated__/GetVideosWithIds'
+
+import { ErrorFallback, VideoGallery } from '@/components'
+
+const InterruptedVideosGallery: React.FC<RouteComponentProps> = () => {
+  const {
+    state: { watchedVideos },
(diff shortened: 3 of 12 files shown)


src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,49 @@
...
+  } = usePersonalData()
+  const interruptedVideosState = watchedVideos.filter((video) => video.__typename === 'INTERRUPTED' && video.timestamp)
+  const interruptedVideosId = interruptedVideosState.map((video) => video.id)
+
+  const {
+    loading: interruptedVideosLoading,
+    data: interruptedVideosData,
+    error: interruptedVideosError,
+    refetch: refetchInterruptedVideos,
+  } = useQuery<{ videos: VideoFields[] }, GetVideosWithIdsVariables>(GET_VIDEOS_WITH_IDS, {
```
kdembler: Why not just use `GetVideosWithIds` for the generic type here?

src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const InterruptedVideosGallery: React.FC<RouteComponentProps> = () => {
+  const {
+    state: { watchedVideos },
+  } = usePersonalData()
+  const interruptedVideosState = watchedVideos.filter((video) => video.__typename === 'INTERRUPTED' && video.timestamp)
+  const interruptedVideosId = interruptedVideosState.map((video) => video.id)
+
+  const {
+    loading: interruptedVideosLoading,
```
kdembler: Usually when there's only one query in a file we don't rename fields returned by `useQuery`

src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,49 @@
...
+    error: interruptedVideosError,
+    refetch: refetchInterruptedVideos,
+  } = useQuery<{ videos: VideoFields[] }, GetVideosWithIdsVariables>(GET_VIDEOS_WITH_IDS, {
+    variables: { ids: interruptedVideosId },
+  })
+
+  const interruptedVideos = interruptedVideosData?.videos.map((video, idx) => ({
+    ...video,
+    // @ts-ignore interruptedVideosState is filtered so it has timestamp
+    timestamp: interruptedVideosState[idx].timestamp,
```
kdembler: There's no guarantee that `interruptedVideosData?.videos` order will match that of `interruptedVideosState` - it can be influenced by ordering, etc. So we will need to do something else. What you could do is create a lookup object like this for example.
```ts
const videoTimestampsMap: Record<string, number> = interruptedVideosState.reduce((acc, video) => {
  acc[video.id] = video.timestamp || 0
  return acc
}, {})
```

src/api/queries/__generated__/VideoFields.ts
```diff
@@ -72,4 +72,6 @@ export interface VideoFields {
   media: VideoFields_media;
   channel: VideoFields_channel;
   license: VideoFields_license;
+  timestamp?: number;
```
kdembler: You should never edit files in `__generated__` manually. If you do, next time someone runs codegen, they will overwrite your changes.

kdembler: What you can do for now is to add the `progress` prop manually to do the `VideoGalleryProps.videos` type

kdembler (changes requested):

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -271,6 +272,9 @@ type Query {
   # Lookup video by its ID
   video(where: VideoWhereUniqueInput!): Video
 
+  # Lookup videos by where params
+  videos(where: VideoWhereInput): [Video!]
```
kdembler:
```suggestion
  videos(where: VideoWhereInput): [Video!]!
```

src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,54 @@
...
+    variables: { ids: interruptedVideosId },
+  })
+
+  const videoTimestampsMap: Record<string, number> = interruptedVideosState.reduce((acc: Lookup, video) => {
+    if (video.__typename === 'INTERRUPTED') {
+      acc[video.id] = video.timestamp || 0
+      return acc
+    }
+    return acc
+  }, {})
```
kdembler: Do we need that `acc: Lookup`? Is `acc` type not figured out correctly? If not, we can instead do something like this:

```suggestion
  const videoTimestampsMap = interruptedVideosState.reduce((acc, video) => {
    if (video.__typename === 'INTERRUPTED') {
      acc[video.id] = video.timestamp || 0
      return acc
    }
    return acc
  }, {} as Record<string, number>)
```

src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+interface Lookup {
+  [key: string]: number
+}
+
+const InterruptedVideosGallery: React.FC<RouteComponentProps> = () => {
+  const {
+    state: { watchedVideos },
+  } = usePersonalData()
+  const interruptedVideosState = watchedVideos.filter((video) => video.__typename === 'INTERRUPTED' && video.timestamp)
```
kdembler: That condition isn't needed anymore since we provide a default now

```suggestion
  const interruptedVideosState = watchedVideos.filter((video) => video.__typename === 'INTERRUPTED')
```

src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const interruptedVideosId = interruptedVideosState.map((video) => video.id)
+
+  const { loading, data, error, refetch } = useQuery<GetVideosWithIds, GetVideosWithIdsVariables>(GET_VIDEOS_WITH_IDS, {
+    variables: { ids: interruptedVideosId },
+  })
+
+  const videoTimestampsMap: Record<string, number> = interruptedVideosState.reduce((acc: Lookup, video) => {
+    if (video.__typename === 'INTERRUPTED') {
+      acc[video.id] = video.timestamp || 0
+      return acc
```
kdembler: We return outside of `if` anyway

src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+    return acc
+  }, {})
+
+  const interruptedVideos = data?.videos?.map((video) => ({
+    ...video,
+    timestamp: videoTimestampsMap[video.id],
```
kdembler: Is this field used anywhere?

src/components/InterruptedVideosGallery.tsx
```diff
@@ -0,0 +1,54 @@
...
+interface Lookup {
+  [key: string]: number
+}
+
+const InterruptedVideosGallery: React.FC<RouteComponentProps> = () => {
+  const {
+    state: { watchedVideos },
+  } = usePersonalData()
+  const interruptedVideosState = watchedVideos.filter((video) => video.__typename === 'INTERRUPTED' && video.timestamp)
+  const interruptedVideosId = interruptedVideosState.map((video) => video.id)
```
kdembler, line 22: We probably should limit number of videos here (not necessarily in this line). Let's say 16.

kdembler (changes requested):

Issue #171, opened 2020-12-15
Add routing transition


kdembler:
As seen here https://www.figma.com/proto/2Q0wavkniGCAlnmvQPzXRf/Joystream-Atlas---Duplicate?node-id=2115%3A23&viewport=1311%2C1101%2C0.1049661710858345&scaling=scale-down we want to implement the following transitions:
- Jump to video v2.
- Page Transition
- Channel Preview
- Discovery v3

They're all basically the same behaviour - on route change fade the current view out and then fade&slide in the new view. We should be able to accomplish this quite easily with route transitions, however we may not want them for every route (especially if the search ends up being an overlay) so it's worth taking into the account.

PR #234, opened 2021-01-26
Add routing transition

mikkio-j:
add routing transition for #171


src/styles/routingTransitions.ts
```diff
@@ -0,0 +1,37 @@
...
+import { TitleSection, VideoSection } from '@/views/ChannelView/ChannelView.style'
+import { InfoContainer } from '@/views/VideoView/VideoView.style'
+import { Container as HomeViewContainer } from '@/views/HomeView'
+import { Container as ChannelsViewContainer } from '@/views/VideosView/VideosView.style'
+
+export const routingTransitions = css`
+  .${transitions.names.fadeAndSlide}-enter {
+    opacity: 0;
+
+    & ${InfoContainer}, ${TitleSection}, ${VideoSection}, ${HomeViewContainer}, ${ChannelsViewContainer} {
```
kdembler: Do we need to manually specify all the containers here? This will be get hard to maintain very quickly

src/views/LayoutWithRouting.tsx
```diff
@@ -55,26 +59,38 @@ const LayoutWithRouting: React.FC = () => {
...
-          <Route default Component={HomeView} />
-          <Route path={routes.video()} Component={VideoView} />
-          <Route path={routes.search()} Component={SearchView} />
-          <Route path={routes.videos()} Component={VideosView} />
-          <Route path={routes.channels()} Component={ChannelsView} />
-          <Route path={routes.channel()} Component={ChannelView} />
-        </Router>
+        <Location>
+          {({ location }) => (
+            <TransitionGroup>
```
kdembler: There are 2 problems here:
1) The transition sometimes looks bad when two views overlap. The transitions in Figma actually fade out the current view and only then start fading in the new one. I think it will look better without the overlap. (if not feasible, we will need to adjust entry/exit timing to alleviate
2) The old view sometimes error out. Go to some channel page, open console and switch to some different view. An exception will be logged

kdembler (changes requested):

src/styles/routingTransitions.ts
```diff
@@ -0,0 +1,30 @@
...
+    & .${transitions.names.slide} {
+      transform: translateY(${sizes(8)});
+    }
+  }
+
+  .${transitions.names.fadeAndSlide}-enter-active {
+    opacity: 1;
+    & .${transitions.names.slide} {
+      transform: translateY(0);
+      transition: ${transitions.timings.loading} ${transitions.easing};
```
kdembler: Let's actually add a new timing specific for routing. `transitions.timings.routing` let's say, and let's try `300ms`. It should be used in this file and also in `timeout` passed to `CSSTransition`

kdembler (commented):
Also, the videos from the `Continue watching` gallery don't fade out on screen change which looks really weird

src/views/LayoutWithRouting.tsx
```diff
@@ -28,16 +31,29 @@ const SIDENAVBAR_ITEMS: NavItemType[] = [
...
+  const location = useLocation()
   const navigate = useNavigate()
+
   useEffect(() => {
-    window.scrollTo(0, 0)
-  }, [pathname])
+    // delay scroll to allow transition to finish first
+    setTimeout(() => {
+      window.scrollTo(0, 0)
+    }, parseInt(transitions.timings.regular))
```
kdembler: This also needs the `routing` timing

src/views/LayoutWithRouting.tsx
```diff
@@ -83,9 +92,21 @@ const RouterWrapper = () => {
...
 const MainContainer = styled.main`
   padding: 0 var(--global-horizontal-padding);
-
+  margin-top: ${TOP_NAVBAR_HEIGHT}px;
   @media screen and (min-width: ${breakpoints.medium}) {
     margin-left: ${SIDENAVBAR_WIDTH}px;
   }
 `
+
+export const Container = styled.div`
```
kdembler: Few points here:
1) If we want this to be shared between views, let's extract it to `components`. Views shouldn't import from other views. We can call it `ViewWrapper`
2) It doesn't need `absolute` position I think, `relative` should be enough. Matter of fact, since `MainContainer` already takes care of padding and margins, I believe `position: relative` is the only style that needs to be here. Given that maybe we don't need that to be a component but rather just make sure all the views have `position: relative` in their top-most container? Or maybe we can export something like `viewWrapperCss` that can be used in all the views? Let's discuss

src/styles/routingTransitions.ts
```diff
@@ -0,0 +1,30 @@
...
+    & .${transitions.names.slide} {
+      transform: translateY(${sizes(8)});
+    }
+  }
+
+  .${transitions.names.fadeAndSlide}-enter-active {
+    opacity: 1;
+    & .${transitions.names.slide} {
+      transform: translateY(0);
+      transition: ${transitions.timings.routing} ${transitions.easing};
```
kdembler: Let's add an additional easing specifically for routing. Let's try `cubic-bezier(0.87, 0, 0.13, 1)` for now

src/views/HomeView.tsx
```diff
@@ -4,7 +4,8 @@ import { ErrorBoundary } from '@sentry/react'
 import { ErrorFallback, CoverVideo, InfiniteVideoGrid, VideoGallery } from '@/components'
 import useVideosConnection from '@/api/hooks/videosConnection'
 import { VideoOrderByInput } from '@/api/queries'
-import InterruptedVideosGallery from '@/components/InterruptedVideosGallery'
+import { InterruptedVideosGallery, ViewWrapper } from '@/components/'
```
kdembler: We already have an import from `@/components` above

src/views/VideoView/VideoView.tsx
```diff
@@ -128,81 +129,83 @@ const VideoView: React.FC = () => {
...
-        <MoreVideosContainer>
-          <MoreVideosHeader>
-            {video ? `More from ${video.channel.handle}` : <Placeholder height={23} width={300} />}
-          </MoreVideosHeader>
-          <InfiniteVideoGrid ready={!loading} channelId={video?.channel.id} showChannel={false} />
-        </MoreVideosContainer>
-      </InfoContainer>
-    </Container>
+    <ViewWrapper>
+      <Container>
```
kdembler: Since we already have a wrapper for the full view here I think we can make the `Container` be actually `styled(ViewWrapper`? This way we can reduce nesting

kdembler: Or we can even call it `StyledViewWrapper` to be consistent

PR #235, opened 2021-01-26
New search exp

fcescob:


src/api/schemas/extendedQueryNode.graphql
```diff
@@ -254,7 +256,10 @@ type Query {
   # Lookup a channel by its ID
   channel(where: ChannelWhereUniqueInput!): Channel
 
-  # List all channel by given constraints
+  # List all channels by given constraints
+  channels(where: ChannelWhereInput): [Channel!]
```
kdembler, line 260: Non-nullable in the query node

```suggestion
  channels(where: ChannelWhereInput): [Channel!]!
```

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -272,6 +277,9 @@ type Query {
   video(where: VideoWhereUniqueInput!): Video
 
   # List all videos by given constraints
+  videos(where: VideoWhereInput): [Video!]
```
kdembler, line 280: Non-nullable in the query node

```suggestion
  videos(where: VideoWhereInput): [Video!]!
```

src/views/LayoutWithRouting.tsx
```diff
@@ -44,6 +53,21 @@ const Route: React.FC<RouteProps> = ({ Component, ...pathProps }) => {
   )
 }
 
+type RoutesProps = Omit<RouterProps, 'component'>
+const Routes: React.FC<RoutesProps> = (props) => (
+  <Router {...props}>
+    <Route default Component={HomeView} />
+    <Route path={routes.video()} Component={VideoView} />
+    <Route path={routes.searchOverlay()} Component={SearchOverlay} />
```
kdembler, line 61: This doesn't seem to be working as expected. The search overlay should exist on top of the currently selected view. That means that both the overlay and previous route should be displayed at the same time. The example desired behaviour is this:
1. I open Videos view, pick a category, scroll a bit
2. I open the search overlay
3. I close the search overlay
4. I'm left with the exact state as it was after step 1 - Videos view shouldn't get unmounted&mounted in the process

src/views/SearchOverlay/SearchOverlay.tsx
```diff
@@ -0,0 +1,81 @@
...
+          return channelData?.channels ? channelData.channels.find((c) => c.id === s.id) : []
+        }
+      }),
+    [channelData, recentSearches, videoData]
+  )
+
+  return (
+    <Container>
+      <Title variant="hero">Recent Searches</Title>
+      {
```
kdembler, line 51: Maybe we could add something for no recent searches?

src/views/SearchOverlay/SearchOverlay.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <SearchesList>
```
kdembler, line 52: We need some placeholders here for when the data is being loaded, currently for some time there will be no results and then they will appear. Let's use some kind of placeholders.

src/views/SearchOverlay/SearchOverlay.style.tsx
```diff
@@ -0,0 +1,79 @@
...
+  width: 88px;
+  height: 88px;
+`
+
+export const SmallChannelPreview: React.FC<ChannelPreview> = ({ handle, avatarPhotoUrl, onClick }) => (
+  <PreviewContainer>
+    <StyledChannelAvatar handle={handle} imageUrl={avatarPhotoUrl} onClick={onClick} />
+    <div onClick={onClick}>
+      <Text variant="h6">{handle}</Text>
+      <Text variant="subtitle2">Channel</Text>
```
kdembler, line 76: Color seems to be off here compared to the design. (contrast is too low)

kdembler, line 76: We need to improve spacing here - the design doesn't specify margins, but our line heights are lower, resulting in vertical squeezing

src/views/SearchOverlay/SearchOverlay.style.tsx
```diff
@@ -0,0 +1,79 @@
...
+type ChannelPreview = {
+  handle: string
+  avatarPhotoUrl: string | null
+  onClick: () => void
+}
+export const SmallVideoPreview: React.FC<VideoPreview> = ({ title, thumbnailUrl, onClick }) => (
+  <PreviewContainer>
+    <VideoImage src={thumbnailUrl} onClick={onClick} />
+    <PreviewTitle variant="h6" onClick={onClick}>
+      {title}
```
kdembler, line 61: Maybe we could included some kind of subtext similar to the channel preview for consistency? Maybe just `Video`?

kdembler (changes requested):
Feels good in general, although we will need to tweak some things. Will do a more thorough look at the code later, let's focus on the experience first

PR #236, opened 2021-01-27
Homeview update

### LINKED ISSUE #158: Update the home view content
As indicated by #30, we should update the home view's content to the following:
1. Continue watching - carousel with interrupted videos (#219)
2. Infinite scroller - one of the following:
a) If there's X videos published at least Y ago from the channels the user follows, then those videos
b) Otherwise, current "Recent videos", only without skipping the ones currently displayed in the carousel (we will get rid of it)

drillprop:
Fix #158


src/api/queries/videos.ts
```diff
@@ -82,6 +82,41 @@ export const GET_NEWEST_VIDEOS = gql`
...
+      edges {
+        node {
+          id
+        }
+      }
+    }
+  }
+`
+
+export const GET_FOLLOWED_CHANNELS_RECENT_VIDEOS = gql`
```
kdembler: Do we need that query? Wouldn't a regular videos connection work here (just with additional args)?

src/api/queries/videos.ts
```diff
@@ -82,6 +82,41 @@ export const GET_NEWEST_VIDEOS = gql`
...
 `
+export const GET_FOLLOWED_CHANNELS_RECENT_VIDEOS_IDS = gql`
+  query GetFollowedChannelsRecentVideosIds($channelIdIn: [ID], $createdAtGte: Date) {
+    videosConnection(where: { channelId_in: $channelIdIn, createdAt_gte: $createdAtGte }) {
+      edges {
+        node {
+          id
+        }
+      }
+    }
```
kdembler: Is the goal here to just get a count of videos? If so, we could use `totalCount` instead

src/components/InfiniteGrids/InfiniteFollowedChannelsVideoGrid.tsx
```diff
@@ -0,0 +1,110 @@
...
+    queryVariables,
+    targetRowsCount,
+    dataAccessor: (rawData) => rawData?.videosConnection,
+    itemsPerRow: videosPerRow,
+  })
+
+  if (error) {
+    throw error
+  }
+  const items = displayedItems.filter((node) => !videosToExclude?.includes(node.id))
```
kdembler: This will be problematic, because it can break the grid behaviour. Hook will always return items to fill the desired amounts of rows, this will get them filtered. I don't think we have a requirement to do that here though? In the future if we need to exclude watched videos, we'll ask the query node to provide something like `ID_not_in` filter so we can exclude that on the query level

src/views/HomeView.tsx
```diff
@@ -4,14 +4,34 @@ import { RouteComponentProps } from '@reach/router'
...
+  GetFollowedChannelsRecentVideosIdsVariables,
+} from '../api/queries/__generated__/GetFollowedChannelsRecentVideosIds'
+import { usePersonalData } from '@/hooks'
+
 const NEWEST_VIDEOS_COUNT = 8
+const MIN_FOLLOWED_CHANNELS_VIDEOS = 16
+
+// last three months
+const NOW = new Date()
+const MIN_DATE_FOLLOWED_CHANNELS_VIDEOS = new Date(NOW.setMonth(NOW.getMonth() - 3))
```
kdembler: This works fine, but we can use `date-fns` to make it a bit clearer. That would become

```ts
const MIN_DATE_FOLLOWED_CHANNELS_VIDEOS = sub(new Date(), { months: 3 })
```

src/views/HomeView.tsx
```diff
@@ -22,10 +42,23 @@ const HomeView: React.FC<RouteComponentProps> = () => {
...
+    variables: {
+      channelIdIn,
+      createdAtGte: MIN_DATE_FOLLOWED_CHANNELS_VIDEOS,
+    },
+  })
 
+  const newestVideos = videosData?.videosConnection.edges.slice(0, NEWEST_VIDEOS_COUNT).map((e) => e.node)
   const hasNewestVideosError = newestVideosError && !newestVideosLoading
-
+  const watchedVideosIds = watchedVideos.map((video) => video.id)
```
kdembler: Like mentioned in the grid comment, I don't think we need that at the moment

src/views/HomeView.tsx
```diff
@@ -35,10 +68,24 @@ const HomeView: React.FC<RouteComponentProps> = () => {
         ) : (
           <ErrorFallback error={newestVideosError} resetError={() => refetchNewestVideos()} />
         )}
-
-        <ErrorBoundary fallback={ErrorFallback}>
-          <StyledInfiniteVideoGrid title="More videos" skipCount={NEWEST_VIDEOS_COUNT} />
-        </ErrorBoundary>
+        {watchedVideosWithoutFollowed && watchedVideosWithoutFollowed.length >= MIN_FOLLOWED_CHANNELS_VIDEOS ? (
```
kdembler: This has one issue now. When we start loading, the header will be `More videos` and will later change to `Followed Channels`. We should probably make the header a placeholder as well till we figure out what to display

src/views/HomeView.tsx
```diff
@@ -35,10 +68,24 @@ const HomeView: React.FC<RouteComponentProps> = () => {
...
+          <ErrorBoundary fallback={ErrorFallback}>
+            <InfiniteFollowedChannelsVideoGrid
+              videosToExclude={watchedVideosIds}
+              title="Followed Channels"
```
kdembler: Let's make the header `Recent videos from followed channels`

src/views/HomeView.tsx
```diff
@@ -35,10 +68,24 @@ const HomeView: React.FC<RouteComponentProps> = () => {
...
+              createdAtGte={MIN_DATE_FOLLOWED_CHANNELS_VIDEOS}
+              channelIdIn={channelIdIn}
+            />
+          </ErrorBoundary>
+        ) : (
+          <ErrorBoundary fallback={ErrorFallback}>
+            <StyledInfiniteVideoGrid
+              title="More videos"
```
kdembler: And this `Recent videos`

src/views/HomeView.tsx
```diff
@@ -35,10 +68,24 @@ const HomeView: React.FC<RouteComponentProps> = () => {
...
+              skipCount={NEWEST_VIDEOS_COUNT}
```
kdembler: As part of this PR we should also remove the `Newest videos` carousel. So we can get rid of `skipCount` then as well

src/mocking/server/resolvers.ts
```diff
@@ -51,16 +56,57 @@ export const videoResolver: QueryResolver<UniqueArgs, VideoFields> = (obj, args,
...
+      })
+      .filter((v) => {
+        if (where?.createdAt_gte) {
+          return new Date(where.createdAt_gte) <= new Date(v.createdAt)
+        }
+        return v
+      })
+      .map((v) => ({
+        ...v,
+        channel: schema.channels.find(v.channelId),
```
kdembler, line 103: Are those necessary?

kdembler (changes requested):

src/mocking/server/resolvers.ts
drillprop: Unfortunately, they are. If I delete these lines I'll get no results when `channelIdIn` and `createdAtGte` are passed as an argument.

src/views/HomeView.tsx
```diff
@@ -1,45 +1,43 @@
...
-
-  const newestVideos = videosConnection?.edges.slice(0, NEWEST_VIDEOS_COUNT).map((e) => e.node)
-
-  const hasNewestVideosError = newestVideosError && !newestVideosLoading
-
+    state: { followedChannels },
+  } = usePersonalData()
+
+  const channelIdIn = followedChannels.map((channel) => channel.id)
+  const { videosConnection, loading } = useVideosConnection({
```
kdembler: Let's handle error on this. We can just do `if (errror) throw error`

src/components/InfiniteGrids/InfiniteVideoGrid.tsx
```diff
@@ -2,15 +2,18 @@ import React, { useCallback, useEffect, useState } from 'react'
...
 import { sizes } from '@/shared/theme'
-import { Grid, Text, VideoPreviewBase } from '@/shared/components'
+import { Grid, Text, VideoPreviewBase, Placeholder } from '@/shared/components'
 import VideoPreview from '@/components/VideoPreviewWithNavigation'
 import { GetVideosConnectionDocument, GetVideosConnectionQuery, GetVideosConnectionQueryVariables } from '@/api/queries'
 import useInfiniteGrid from './useInfiniteGrid'
 
 type InfiniteVideoGridProps = {
   title?: string
+  isTitleLoading?: boolean
```
kdembler: I don't think we need this prop anymore. Let's just display placeholder if `!ready`

src/components/InfiniteGrids/InfiniteVideoGrid.tsx
```diff
@@ -103,9 +124,41 @@ const InfiniteVideoGrid: React.FC<InfiniteVideoGridProps> = ({
 
     setQueryVariables({
       ...(channelId ? { channelId } : {}),
+      ...(channelIdIn ? { channelIdIn } : {}),
```
kdembler, line 125: This looks messy but let's keep that for now, we need to figure out some general solution to this

kdembler (approved):

Issue #231, opened 2021-01-25
Extract queries out of components 


kdembler:
To improve our components readability and reduce their dependence on one specific library, let's try to extract all the `useQuery` into `src/api/hooks`. We can have a hook per query, just like the `useCoverVideo` in `src/api/hooks/coverVideo.ts`. We should extract only the trivial ones, if the code is more complex, like with the infinite scroller, let's leave the query there

PR #242, opened 2021-01-27
Extract queries out of components

mikkio-j:
fix for #231

### DIFF
--- a/src/api/hooks/categories.ts
+++ b/src/api/hooks/categories.ts
@@ -1,0 +1,14 @@
+import { useQuery } from '@apollo/client'
+import { GET_CATEGORIES } from '@/api/queries'
+import { GetCategories } from '@/api/queries/__generated__/GetCategories'
+
+const useCategories = () => {
+  const { loading, data, error } = useQuery<GetCategories>(GET_CATEGORIES)
+  return {
+    loading,
+    data: data?.categories,
+    error,
+  }
+}
+
+export default useCategories
--- a/src/api/hooks/channel.ts
+++ b/src/api/hooks/channel.ts
@@ -1,0 +1,27 @@
+import { useQuery } from '@apollo/client'
+import { GET_CHANNEL } from '@/api/queries/channels'
+import { GetChannel, GetChannelVariables } from '@/api/queries/__generated__/GetChannel'
+
+export const useChannel = (id: string) => {
+  const { data, loading, error } = useQuery<GetChannel, GetChannelVariables>(GET_CHANNEL, {
+    variables: { id },
+  })
+  return {
+    loading,
+    data: data?.channel,
+    error,
+  }
+}
+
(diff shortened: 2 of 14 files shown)


src/api/hooks/categories.ts
```diff
@@ -0,0 +1,14 @@
+import { useQuery } from '@apollo/client'
+import { GET_CATEGORIES } from '@/api/queries'
+import { GetCategories } from '@/api/queries/__generated__/GetCategories'
+
+const useCategories = () => {
+  const { loading, data, error } = useQuery<GetCategories>(GET_CATEGORIES)
```
kdembler, line 6: We should probably pass everything from here, not just selected fields. So something like `const { data, ...rest } = ...`

src/api/hooks/channel.ts
```diff
@@ -0,0 +1,27 @@
...
+    variables: { id },
+  })
+  return {
+    loading,
+    data: data?.channel,
+    error,
+  }
+}
+
+export const useChannelLink = (id: string | undefined) => {
```
kdembler, line 16: `ChannelLink` isn't a resource, so we shouldn't have a separate query for that. The purpose here is to `useChannel` but with some additional options. What you can do is add something like `opts` arg to `useChannel` that will allow specifying things like `fetchPolicy` and `skip`

src/api/hooks/newestVideos.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const useNewestVideos = () => {
+  const { loading, data, error, refetch } = useQuery<GetNewestVideos, GetNewestVideosVariables>(GET_NEWEST_VIDEOS, {
+    variables: { first: NEWEST_VIDEOS_COUNT },
+    notifyOnNetworkStatusChange: true,
+  })
+
+  return {
+    loading,
+    data: data?.videosConnection.edges.slice(0, NEWEST_VIDEOS_COUNT).map((e) => e.node),
```
kdembler, line 15: This should happen outside of hook, the hook is just abstraction over `useQuery`

src/api/hooks/newestVideos.ts
```diff
@@ -0,0 +1,22 @@
+import { useQuery } from '@apollo/client'
+import { GET_NEWEST_VIDEOS } from '@/api/queries'
+import { GetNewestVideos, GetNewestVideosVariables } from '@/api/queries/__generated__/GetNewestVideos'
+
+const NEWEST_VIDEOS_COUNT = 8
+
+const useNewestVideos = () => {
+  const { loading, data, error, refetch } = useQuery<GetNewestVideos, GetNewestVideosVariables>(GET_NEWEST_VIDEOS, {
+    variables: { first: NEWEST_VIDEOS_COUNT },
```
kdembler, line 9: `first` should be passed from outside

kdembler:
Closing in favor of #287

Issue #35, opened 2020-11-24
Move to react-router@6


bedeho:
Will be ready by end of year hopefully.

kdembler:
That's `react-router` if anyone's wondering :)

kdembler:
So it seems they're still working on the final release, but authors themselves encourage using the beta as it should be pretty stable. Let's do that then, and upgrade to `react-router@v6.0.0-beta.0`

PR #265, opened 2021-01-28
Move from reach-router to react-router@6

mikkio-j:
fix for #35


src/views/LayoutWithRouting.tsx
```diff
@@ -28,10 +28,18 @@ const SIDENAVBAR_ITEMS: NavItemType[] = [
   },
 ]
 
+export type RouteComponentProps = {
```
kdembler: Do we even need that now? This was previously enforced by `reach`, I think we can just remove that

src/components/SideNavbar/SideNavbar.tsx
```diff
@@ -91,7 +90,7 @@ const NavItem: React.FC<NavItemProps> = ({ expanded = false, subitems, children,
 
   return (
     <SidebarNavItem>
-      <SidebarNavLink onClick={onClick} to={to} getProps={isActive} expanded={expanded} content={itemName}>
+      <SidebarNavLink onClick={onClick} to={to} expanded={expanded} content={itemName}>
```
kdembler: This isn't highlighted for home view when opening the app. Also when you use the Joystream logo to go home

kdembler (changes requested):

kdembler (approved):

PR #273, opened 2021-01-28
Fix main margin

drillprop:


src/views/LayoutWithRouting.tsx
```diff
@@ -76,7 +76,7 @@ const MainContainer = styled.main`
   padding: 0 var(--global-horizontal-padding);
 
   @media screen and (min-width: ${breakpoints.medium}) {
-    margin-left: ${sizes(14)};
+    margin-left: ${sizes(18)};
```
kdembler: Shouldn't we use `SIDENAVBAR_WIDTH` here?

kdembler (approved):

PR #287, opened 2021-01-29
Extract queries, migrate to graphql code generator

### LINKED ISSUE #216: Migrate to graphql-code-generator
Currently, we use `apollo-tooling` (Apollo Codegen) to generate Typescript typings for all of our queries based on the schema. However, Apollo Codegen seems unmaintained for some time now and there are some issues with it. (generating interfaces instead of types, a lot of types duplication, weird types for some queries/unions). At some point we should switch to https://github.com/dotansimha/graphql-code-generator instead - it's much better maintained and offers much more features. There is a proposal to actually move Apollo Codegen to that library but doesn't seem to be moving forward. More details and possible migration guide here - https://github.com/apollographql/apollo-tooling/issues/2053

### LINKED ISSUE #231: Extract queries out of components

drillprop:
Fix #216 
Fix #231


src/shared/components/VideoPreview/VideoPreview.tsx
```diff
@@ -138,7 +138,7 @@ const VideoPreview: React.FC<VideoPreviewProps> = ({
 
   const metaNode = (
     <MetaText variant="subtitle2" main={main} scalingFactor={scalingFactor}>
-      {formatVideoViewsAndDate(views, createdAt, { fullViews: main })}
+      {views && formatVideoViewsAndDate(views, createdAt, { fullViews: main })}
```
kdembler: This condition can break it a bit, because date should still be displayed even when there's no views info. `formatVideoViewsAndDate` can handle `views` being `null` so we either need to make that `views || null` here or change the signature of `formatVideoViewsAndDate` to accept `undefined`

src/api/hooks/search.ts
```diff
@@ -0,0 +1,14 @@
+import { useSearchQuery, SearchQueryVariables, SearchQuery } from '@/api/queries/__generated__/search.generated'
+import { QueryHookOptions } from '@apollo/client'
+
+type Opts = QueryHookOptions<SearchQuery>
+const useSearch = (variables: SearchQueryVariables, opts?: Opts) => {
+  const { data, ...rest } = useSearchQuery({ ...opts, variables })
+
+  return {
+    data: data?.search,
```
kdembler, line 9: Shouldn't we rename here as well?

src/api/hooks/videosConnection.ts
```diff
@@ -0,0 +1,18 @@
+import { QueryHookOptions } from '@apollo/client'
+import {
+  useGetVideosConnectionQuery,
+  GetVideosConnectionQueryVariables,
+  GetVideosConnectionQuery,
+} from '@/api/queries/__generated__/videos.generated'
+
+type Opts = QueryHookOptions<GetVideosConnectionQuery>
+const useVideosConnection = (variables?: GetVideosConnectionQueryVariables, opts?: Opts) => {
```
kdembler, line 9: I think we will need to provide default sorting here to make sure we get newest videos first

src/mocking/server/data.ts
```diff
@@ -1,6 +1,6 @@
 import { ModelInstance } from 'miragejs/-types'
 import faker from 'faker'
-
+// TODO update types
```
kdembler: What specifically do you mean by that?

kdembler (commented):
Some small comments but feels really solid in general. This will make our codebase much cleaner

src/api/hooks/search.ts
kdembler: Ah I think I know why you left it out. We can keep it this way

kdembler (approved):

PR #289, opened 2021-02-01
use graphql files for documents

kdembler:


src/api/queries/channels.ts
```diff
@@ -1,64 +1 @@
...
-`
-export const UNFOLLOW_CHANNEL = gql`
-  mutation UnfollowChannel($channelId: ID!) {
-    unfollowChannel(channelId: $channelId) {
-      id
-      follows
-    }
-  }
-`
+export * from './__generated__/channels.generated'
```
drillprop, line 1: Do we really need these files? Can we just create one index.ts file?

drillprop (commented):
I'm not sure about these single-line files inside the queries folder. Besides that, looks good to me

PR #292, opened 2021-02-02
Add header text field component

### LINKED ISSUE #243: Create header text field component
#### Description

Let's create editable header text field. See how it should work like - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A21784

#### Requirements

- `HeaderTextField` components added to `src/shared/components`
  - Should expose `ref` to its inner `input` element
  - Should allow passing helper text
  - Should allow passing state/variant (warning/error/default)
  - Should have proper hover state as indicated by the design (only when non-empty)
- Component added to the storybook

mikkio-j:
Fix #243


src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -0,0 +1,80 @@
...
+  cursor: pointer;
+  &:hover {
+    filter: brightness(80%);
+  }
+`
+
+export const WarningText = styled(Text)`
+  color: ${colors.warning};
+  background-color: ${colors.gray[800]};
+  max-width: 600px;
```
drillprop, line 50: Take a look at the design. The width should not expand the whole window. Should be tied to the text length.

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -0,0 +1,80 @@
...
+  padding: ${sizes(2)};
```
drillprop, line 51: I think padding at the bottom should be a little bit higher. Maybe you could also change `line-height`.

src/shared/components/HeaderTextField/HeaderTextField.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    return (
+      <Container>
+        {isInEditMode ? (
+          <StyledInput
+            ref={ref}
+            type="text"
+            value={text}
+            onChange={(e) => setText(e.target.value)}
+            onKeyDown={(e) => handleKeyDown(e)}
```
drillprop:
```suggestion
            onKeyDown={handleKeyDown}
```

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -0,0 +1,80 @@
...
+    color: #fff;
+    text-align: center;
+  }
+  &:after {
+    display: inline-block;
+    content: '';
+    margin-left: ${sizes(4)};
+    transform: rotate(90deg);
+    border: 10px solid #52616b;
+    border-color: transparent #52616b transparent transparent;
```
drillprop:
```suggestion
    border: 10px solid transparent;
    border-right-color:  ${colors.gray[400]};
```

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -0,0 +1,80 @@
...
+  &:before {
+    display: inline-block;
+    content: attr(data-text);
+    font-size: ${typography.sizes.body2};
+    font-weight: ${typography.weights.light};
+    position: absolute;
+    top: ${sizes(4)};
+    width: 200px;
+    padding: 10px;
+    background: #52616b;
```
drillprop:
```suggestion
    background: ${colors.gray[400]};
```

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -0,0 +1,80 @@
...
+  background-color: ${colors.gray[800]};
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
+  width: 100%;
+  max-width: 600px;
+  transition: 0.2s;
+  cursor: pointer;
+  &:hover {
+    filter: brightness(80%);
```
drillprop, line 38: I don't think you need a filter here, `opacity` should also work.

src/shared/components/HeaderTextField/HeaderTextField.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useState } from 'react'
+import { Container, Title, WarningText, StyledTooltip, StyledInput } from './HeaderTextField.style'
+
+type HeaderTextFieldProps = {
```
kdembler: How do you access this component's value from its consumer? You made the input self-controlling (it has its own state and handles the updates on its own). This way we won't be able to build a form with it. `value` and `onChange` should come from outside, via props

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
drillprop: On the second thought, I could be wrong. There will be custom background underneath this title, so, yeah, the filter may be better here.

src/shared/components/HeaderTextField/HeaderTextField.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  component: HeaderTextField,
+}
+
+const Template: Story<HeaderTextFieldProps> = () => {
+  const [text, setText] = useState('Lorem ipsum dolor sit amet')
+  const input = React.useRef<HTMLInputElement>(null)
+
+  return (
+    <HeaderTextField
+      ref={input}
```
kdembler: All the props here (except `value` and `onChange`) should be passed via Storybook Controls. Please take a look at the other stories to see how that's done. Then we can easily check different states by flipping controls in storybook

src/shared/components/HeaderTextField/HeaderTextField.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const handleKeyDown = (e: React.KeyboardEvent) => {
+      if (e.key !== 'Enter' || ref === null) {
+        return
+      }
+      setEditMode(false)
+    }
+
+    return (
+      <Container>
+        {isInEditMode ? (
```
kdembler: There's a noticeable change in text/component size when entering the edit mode. In general, is there some reason to change what we display? i.e. is there anything stopping us to display this as a styled `input` at all times?

kdembler (changes requested):

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const WarningText = styled(Text)`
+  color: ${colors.warning};
+  background-color: ${colors.gray[800]};
+  max-width: 600px;
+  padding: ${sizes(2)} 0;
+  ${fluidRange({ prop: 'fontSize', fromSize: '10px', toSize: '16px' })};
+`
+
+export const StyledTooltip = styled.span`
```
kdembler: oh, also, tooltip isn't in scope for this task, there's another one specifically for creating tooltip element since we'll need that in a few places. So the tooltip will come from outside, from consumer. To not waste your work, you can move what you have here to `shared/components/WIP_Tooltip` so later we can use that as a starting point

src/shared/components/HeaderTextField/HeaderTextField.tsx
```diff
@@ -0,0 +1,47 @@
...
+  value: string
+  warningText?: string
+  errorText?: string
+  onChange: (value: string) => void
+  variant?: Variant
+}
+type ChangeEvent = React.ChangeEvent<HTMLInputElement>
+
+const HeaderTextField = React.forwardRef<HTMLInputElement, HeaderTextFieldProps>(
+  ({ value, warningText, errorText, onChange, variant = 'default' }, ref) => {
```
kdembler: Seems the `ref` isn't passed to the componeent

src/shared/components/HeaderTextField/HeaderTextField.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const inputElement = useRef<HTMLInputElement>(null)
+    useEffect(() => {
+      if (inputElement.current === null) {
+        return
+      }
+      inputElement.current.style.width = value.length + 'ch'
```
kdembler: Like mentioned in the call, we cannot rely on the `value` as it will not be present in the uncontrolled state. Not sure how to solve this tbh

src/shared/components/HeaderTextField/HeaderTextField.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useRef, useEffect } from 'react'
+import { Container, WarningText, StyledInput } from './HeaderTextField.style'
+
+export type Variant = 'default' | 'error' | 'warning'
+
+export type HeaderTextFieldProps = {
+  value: string
+  warningText?: string
+  errorText?: string
+  onChange: (value: string) => void
```
kdembler: Let's keep this consistent and pass the full event

src/shared/components/HeaderTextField/HeaderTextField.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+    variant: {
+      control: {
+        type: 'select',
+        options: ['default', 'error', 'warning'],
+      },
+    },
+  },
+}
+
+type StoryProps = {
```
kdembler: Any reason to specify props explicitly? Doesn't the default controls approach work here?
```
const Template: Story<HeaderTextFieldProps> = (args) => (
```

kdembler (approved):

PR #296, opened 2021-02-04
Refactor channel preview components

eldiegod:


src/components/InfiniteGrids/InfiniteChannelGrid.tsx
```diff
@@ -20,7 +21,7 @@ type InfiniteChannelGridProps = {
 }
 
 const INITIAL_ROWS = 4
-const INITIAL_CHANNELS_PER_ROW = 4
+const INITIAL_CHANNELS_PER_ROW = 7
```
kdembler, line 23: Why this change?

src/shared/components/ChannelPreview/ChannelPreviewBase.style.tsx
```diff
@@ -34,14 +36,19 @@ export const InnerContainer = styled.div<InnerContainerProps>`
...
 
   display: flex;
   flex-direction: column;
   align-items: center;
+  border: 1px solid transparent;
   ${hoverTransition}
 `
 
+export const Anchor = styled.a`
+  all: unset;
```
kdembler, line 50: What does this do? I've never seen this CSS property used before

src/shared/components/ChannelPreview/ChannelPreviewBase.tsx
```diff
@@ -1,37 +1,73 @@
...
+  VideoCountContainer,
+} from './ChannelPreviewBase.style'
 import Placeholder from '../Placeholder'
+import { CSSTransition, SwitchTransition } from 'react-transition-group'
+import { transitions } from '@/shared/theme'
 
 type ChannelPreviewBaseProps = {
-  avatarNode?: React.ReactNode
-  nameNode?: React.ReactNode
+  avatarURL?: string
```
kdembler: Could we rename that to `avatarUrl`? We used `URL` in queries before but at some point we switch to `Url`

src/shared/components/ChannelPreview/ChannelPreviewBase.tsx
```diff
@@ -1,37 +1,73 @@
...
+  name?: string
```
kdembler: Let's make this `handle` to be in sync with query node fields

src/shared/components/ChannelPreview/ChannelPreviewBase.tsx
```diff
@@ -1,37 +1,73 @@
...
 const ChannelPreviewBase: React.FC<ChannelPreviewBaseProps> = ({
-  avatarNode,
-  nameNode,
+  avatarURL,
+  name,
+  videoCount,
+  loading = true,
+  channelHref,
   className,
   animated = false,
```
kdembler: I think we can default to `true` here

kdembler: Or even better, remove that prop altogether and assume we animate if `!loading && channelHref`

src/shared/components/ChannelPreview/ChannelPreviewBase.tsx
```diff
@@ -1,37 +1,73 @@
...
+                {loading ? <Placeholder rounded /> : <StyledAvatar imageUrl={avatarURL} handle={name} />}
+              </AvatarContainer>
+
+              <Info>
+                {loading ? <Placeholder width="140px" height="16px" /> : <NameHeader variant="h6">{name}</NameHeader>}
+                <VideoCountContainer>
+                  {loading ? (
+                    <Placeholder width="140px" height="16px" />
+                  ) : (
+                    <VideoCount variant="subtitle2">{videoCount ? `${videoCount} Uploads` : '⠀'}</VideoCount>
```
kdembler: Let's try to fade that in, as we talked in the chat

kdembler (commented):
Looks really good overall, some small details to adjust, mostly from the code being old

src/components/InfiniteGrids/InfiniteChannelGrid.tsx
eldiegod: @kdembler I wanted to ask about that, why is it a 4, isn't it too few?

src/shared/components/ChannelPreview/ChannelPreviewBase.style.tsx
eldiegod: it resets all of the selected element’s properties; in this case, I use it to strip all the base styles from the `a` tag that we don't need

kdembler: Huh, cool to know :+1:

src/components/InfiniteGrids/InfiniteChannelGrid.tsx
kdembler: This is just an initial value, after the first render this should be changed based on the grid's size: https://github.com/Joystream/atlas/blob/personalisation/src/components/InfiniteGrids/InfiniteChannelGrid.tsx#L77

Does changing this value actually impacts the experience? I can imagine because of that maybe we start fetching too few channels at the first render?

src/shared/components/ChannelPreview/ChannelPreviewBase.tsx
```diff
@@ -1,37 +1,78 @@
...
 
 type ChannelPreviewBaseProps = {
-  avatarNode?: React.ReactNode
-  nameNode?: React.ReactNode
+  avatarUrl?: string
+  handle?: string
+  videoCount?: number
+  channelHref?: string
   className?: string
   animated?: boolean
```
kdembler: We have the correct condition now but I think `animated` prop isn't needed anymore?

eldiegod: Right, forgot about the prop

src/components/InfiniteGrids/InfiniteChannelGrid.tsx
eldiegod: Yes, I noticed something like that so some placeholder animations would be a bit out of sync, and with a bit more INITIAL_CHANNELS_PER_ROW they sync up better

kdembler (approved):
yessirski

PR #297, opened 2021-02-04
Textfield select

### LINKED ISSUE #240: Create text field and select components
#### Description

Let's create basic input components - `TextField` and `Select`. Designs here - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=484%3A1802. Feel free to use some existing library, like Material UI, for inspiration

#### Requirements

- `TextField` and `Select` components added to `src/shared/components`
  - Components share a common visuals so it may be worth it to create something like `InputBase` that holds the styling for border, label, etc.
  - Let's skip variants with icons for now
  - `TextField` should allow passing a helper text (visible in the design in the error state). However, the helper text should be visible regardless of the state - error state only gives it a color. Let's also add a warning state, in which the helper text has warning color, but the input border isn't changed
  - `TextField` should expose `ref` to its inner `input` element
- Both components added to the storybook

drillprop:
Fix #240

### DIFF
--- a/src/shared/components/InputBase/InputBase.styles.ts
+++ b/src/shared/components/InputBase/InputBase.styles.ts
@@ -1,0 +1,103 @@
+import { colors, sizes, transitions, typography } from '@/shared/theme'
+import { css } from '@emotion/react'
+import styled from '@emotion/styled'
+
+type FormGroupProps = {
+  error?: boolean
+  disabled?: boolean
+}
+
+export const FormGroup = styled.div<FormGroupProps>`
+  max-width: 300px;
+  position: relative;
+  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'auto')};
+  input,
+  button {
+    transition: all ${transitions.timings.regular} ${transitions.easing};
+    background: none;
+    padding: ${sizes(3)};
+    font-size: ${typography.sizes.subtitle2};
+  }
+
+  input,
+  button,
+  label {
+    cursor: pointer;
+  }
+
+  ${({ disabled }) =>
+    disabled &&
+    css`
+      input,
+      button,
+      label {
+        pointer-events: none;
+      }
+    `};
+`
+
+export const LabelText = styled.span`
+  position: absolute;
+  top: 12px;
+  left: 9px;
+  padding: 1px ${sizes(1)};
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 100%;
+
+  color: ${colors.gray[400]};
+  font-size: ${typography.sizes.subtitle2};
+
+  background-color: ${colors.black};
+  transition: all ${transitions.timings.regular} ${transitions.easing};
+`
+
+export const labelOnTop = css`
+  transform: translateY(-19px);
+  font-size: ${sizes(3)};
+  color: ${colors.gray[300]};
+`
+
+export const styledinputStates = {
+  default: css`
+    border: 1px solid ${colors.gray[400]};
+  `,
+  activated: css`
+    border: 1px solid ${colors.gray[200]};
+    color: ${colors.gray[200]};
+  `,
+  focused: css`
+    border: 1px solid ${colors.blue[500]};
+    color: ${colors.white};
+  `,
+  filled: css`
+    border: 1px solid ${colors.gray[300]};
+    color: ${colors.gray[300]};
+  `,
+  disabled: css`
+    border: 1px solid ${colors.gray[700]};
+    color: ${colors.gray[700]};
+  `,
+  error: css`
+    border: 1px solid ${colors.error};
+    background: red;
+  `,
+  warning: css``,
+}
+
+type HelperTextProps = {
+  helperTextVariant?: 'default' | 'error' | 'warning'
+}
+
+const helperVariants = {
+  default: colors.gray[400],
+  error: colors.error,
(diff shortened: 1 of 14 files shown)


src/shared/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+      value={selectedItem}
+    />
+  )
+}
+
+export const Default = Template.bind({})
+
+export const WithControlledInput = TemplateWithControlledInput.bind({})
+
+export const Disabled = Template.bind({})
```
mikkio-j: I don't think separate Disabled, Error and Warning is needed since we can achieve that via controls on the Default.

src/shared/components/InputBase/InputBase.styles.ts
```diff
@@ -0,0 +1,103 @@
...
+    css`
+      input,
+      button,
+      label {
+        pointer-events: none;
+      }
+    `};
+`
+
+export const LabelText = styled.span`
```
mikkio-j, line 39: There are visible rectangle corners on transition when label moves up. Maybe `background-color` transition could help. Or pseudoelements that will work as border-top and will shorten and make space when the label moves up.

src/shared/components/TextField/TextField.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+      helperText={isFocused ? 'focused' : 'not focused'}
+    />
+  )
+}
+
+export const Default = Template.bind({})
+
+export const WithControlledInput = TemplateWithControlledInput.bind({})
+
+export const Disabled = Template.bind({})
```
mikkio-j: as in Select - additional template binds

src/shared/components/InputBase/InputBase.styles.ts
```diff
@@ -0,0 +1,104 @@
...
+    transition: all ${transitions.timings.regular} ${transitions.easing};
+    background: none;
+    padding: ${sizes(3)};
+    font-size: ${typography.sizes.subtitle2};
+  }
+
+  input,
+  button,
+  label {
+    cursor: pointer;
```
kdembler: Shouldn't we should use the regular "edit" cursor here?

src/shared/components/TextField/TextField.stories.tsx
```diff
@@ -17,6 +19,17 @@ const TemplateWithControlledInput: Story<TextFieldProps> = (args) => {
   return <TextField {...args} onChange={(e) => setValue(e.target.value)} value={value} />
 }
 
+const TemplateWithUncontrolledInput: Story<TextFieldProps> = (args) => {
```
drillprop, line 25: In uncontrolled component, I'm getting this, but it only happens in stories. I couldn't figure out why.
image
I tested this component in `HomeView` and everything works fine.

src/shared/components/InputBase/InputBase.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { FormGroup, HelperText } from './InputBase.styles'
+
+export type InputBaseProps = {
+  label?: string
+  error?: boolean
+  warning?: boolean
+  helperText?: string
+  disabled?: boolean
```
kdembler, line 9: Let's also include optional `className` here. Then we should pass it to the top-most container. This way we will be later able to style those components with `styled(InputBase)` or `styled(TextField)`. This applies to all your new components

src/shared/components/Select/Select.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export type SelectProps = {
+  onChange?: (changes: UseSelectStateChange<SelectedItem>) => void
+  value?: SelectedItem
+  items: SelectedItem[]
+  placeholder?: string
+} & InputBaseProps
+
+const SelectComponent: React.ForwardRefRenderFunction<HTMLDivElement, SelectProps> = (props, ref) => {
+  const { label = '', items, placeholder = 'Select option', error, warning, value, disabled, onChange } = props
```
kdembler: I think we're missing the `helperText` here. Maybe we extract `...inputBaseProps` for the values that we don't use in this component and then pass that to `InputBase`? Currently I think this will only be `warning` and `helperText`

src/shared/components/Select/Select.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { forwardRef } from 'react'
+import InputBase, { InputBaseProps } from '../InputBase'
+import { SelectButton, SelectMenu, SelectOption, SelectWrapper, StyledLabelText } from './Select.style'
+import { useSelect, UseSelectStateChange } from 'downshift'
+import Icon from '../Icon'
+
+export type SelectedItem = {
+  value: string
+  name: string
+} | null
```
kdembler: I think it may be more clear from consumer perspective if this type isn't nullable by itself. The `onChange` handler could  specify `| null` instead. What do you think?

src/shared/components/TextField/TextField.stories.tsx
kdembler: Seems it's some problem with Storybook - https://github.com/storybookjs/storybook/issues/6471 Let's not worry about it

kdembler (approved):

Issue #244, opened 2021-01-27
Create general dialog component


kdembler:
#### Description

We need a general dialog/modal component that will be later used by other components. Implement according to the design - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3914%3A1856

#### Requirements

- `Dialog` component added to `src/shared/components`
  - Should properly handle viewport changes
  - Should allow specifying (details to figure out):
    - Title (optional)
    - Content
    - Action buttons (optional)
  - Should have two variants: with and without exit button
  - Should feature proper enter/exist transitions as described in the design
- Component added to Storybook
  - Story with just dialog open
  - Story with a button triggering dialog

PR #299, opened 2021-02-05
Add general dialog

mikkio-j:
fix for #244


src/shared/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  handleExit,
+  handlePrimaryButton,
+  handleSecondaryButton,
+}) => {
+  return (
+    <StyledContainer css={dialogTransitions}>
+      {icon || exitButton ? (
+        <StyledHeadRow>
+          {icon && <StyledIcon name={icon} />}
+          {exitButton && <StyledIcon marginLeft={!icon} name="times-white" onClick={handleExit} />}
```
drillprop: It should be wrapped in a button tag possibly  with `aria-label="close dialog"` or/and with `title` attribute with similar content

src/shared/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  StyledSecondaryButton,
+  dialogTransitions,
+} from './Dialog.style'
+
+export type DialogProps = {
+  title?: string
+  content: string
+  primaryButton?: string
+  secondaryButton?: string
+  icon?: 'success' | 'failure' | null
```
drillprop: I think `null`  is redundant here

src/shared/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  exitButton?: boolean
+  handleExit?: () => void
+  handlePrimaryButton?: () => void
```
drillprop:
```suggestion
  handlePrimaryButton?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```

src/shared/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  handleSecondaryButton?: () => void
```
drillprop:
```suggestion
  handleSecondaryButton?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```

src/shared/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+} from './Dialog.style'
+
+export type DialogProps = {
+  title?: string
+  content: string
+  primaryButton?: string
+  secondaryButton?: string
+  icon?: 'success' | 'failure' | null
+  exitButton?: boolean
+  handleExit?: () => void
```
drillprop:
```suggestion
  handleExit?: () => (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```

src/shared/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+}) => {
+  return (
+    <StyledContainer css={dialogTransitions}>
+      {icon || exitButton ? (
+        <StyledHeadRow>
+          {icon && <StyledIcon name={icon} />}
+          {exitButton && <StyledIcon marginLeft={!icon} name="times-white" onClick={handleExit} />}
+        </StyledHeadRow>
+      ) : (
+        ''
```
drillprop: Instead of a blank string, you could just add `null` here

src/shared/theme/transitions.ts
```diff
@@ -1,11 +1,15 @@
 const transitions = {
   easing: 'cubic-bezier(0.165, 0.84, 0.44, 1)',
+  dialogEasing: 'cubic-bezier(0.25, 0.01, 0.25, 1)',
```
drillprop: Not sure if we'll need these values globally. In my opinion, we can just put them inside `Dialog.style.ts`

src/shared/components/Dialog/Dialog.style.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+`
+
+export const StyledSecondaryButton = styled(Button)`
+  background-color: ${colors.transparent};
+  color: ${colors.white};
+  border: 1px solid ${colors.gray[500]};
+  &:hover {
+    background-color: ${colors.gray[500]};
+    border: 1px solid ${colors.gray[500]};
```
drillprop:
```suggestion
```

src/shared/components/Dialog/Dialog.style.ts
```diff
@@ -0,0 +1,93 @@
...
+    color: ${colors.white};
+  }
+  &:active {
+    background-color: ${colors.transparent};
+    border: 1px solid ${colors.gray[500]};
```
drillprop:
```suggestion
```

src/shared/components/Dialog/Dialog.style.ts
```diff
@@ -0,0 +1,93 @@
...
+`
+
+export const StyledTitleText = styled(Text)`
+  margin-bottom: ${sizes(3)};
+`
+
+export const StyledContentText = styled(Text)`
+  color: ${colors.white};
+  font-weight: ${typography.weights.thin};
+  margin-bottom: ${sizes(6)};
```
drillprop: You could also add ` word-wrap: break-word;` but not super sure here

drillprop (commented):
Looks really solid. I like how flexible and configurable it is.

src/shared/components/Dialog/Dialog.tsx
drillprop: also, you could add `cursor: pointer` to this button

src/shared/components/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+        type: 'select',
+        options: ['success', 'failure', null],
+      },
+      defaultValue: 'success',
+    },
+    title: { control: 'text', defaultValue: 'De- spa- Cito' },
+    content: {
+      control: 'text',
+      defaultValue:
+        'Sí, sabes que ya llevo un rato mirándote Tengo que bailar contigo hoy (DY) Vi que tu mirada ya estaba llamándome Muéstrame el camino que yo voy',
```
drillprop: oh, just replace it with `lorem ipsum` please! 😅

src/shared/components/Dialog/Dialog.style.ts
```diff
@@ -0,0 +1,124 @@
...
+  word-wrap: break-word;
```
drillprop:
```suggestion
  word-wrap: break-word;
  line-height: ${sizes(5)};
```
Content and Title definitely needs some air here

src/shared/components/Dialog/Dialog.style.ts
```diff
@@ -0,0 +1,124 @@
...
+  background-color: ${colors.gray[600]};
+  padding: ${sizes(4)};
+  box-shadow: 0px 8px 8px rgba(0, 0, 0, 0.12), 0px 24px 40px rgba(0, 0, 0, 0.16);
+  @media screen and (min-width: ${breakpoints.small}) {
+    padding: ${sizes(6)};
+  }
+`
+
+export const StyledTitleText = styled(Text)`
+  margin-bottom: ${sizes(3)};
```
drillprop:
```suggestion
  margin-bottom: ${sizes(3)};
  line-height: ${sizes(8)};
```

src/shared/components/Dialog/Dialog.style.ts
```diff
@@ -0,0 +1,124 @@
...
+  word-wrap: break-word;
+`
+
+export const StyledHeadRow = styled.div`
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: ${sizes(4)};
+`
+
+export const StyledButtonContainer = styled.div`
```
drillprop: It's a detail, but I think `StyledButtonsContainer` would be a better name

src/shared/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+    document.body.style.setProperty('overflow', 'hidden')
+    return () => {
+      document.body.style.setProperty('overflow', null)
+    }
+  }, [showDialog])
+
+  return (
+    <Portal>
+      <CSSTransition in={showDialog} timeout={250} classNames="modal" unmountOnExit>
+        <StyledBackdrop css={dialogTransitions}>
```
drillprop: Not sure if it's required in this task, but in my opinion, a user should have the ability to close the dialog when he clicks this overlay

drillprop (approved):

src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
```diff
@@ -0,0 +1,74 @@
...
+import { css } from '@emotion/react'
+import { Button } from '@/shared/components'
+import { colors, sizes, breakpoints, zIndex } from '@/shared/theme'
+
+export const StyledBackdrop = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  height: 100vh;
```
kdembler: I think we can just use `bottom: 0;` instead?

src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
```diff
@@ -0,0 +1,74 @@
...
+  background-color: rgba(0, 0, 0, 0.4);
+  z-index: ${zIndex.globalOverlay};
+  @media (min-width: ${breakpoints.medium}) {
+    padding-left: var(--scrollbar-gap-width);
```
kdembler: Why is this needed?

src/components/Portal.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { createPortal } from 'react-dom'
+
+type PortalProps = {
+  portal: React.RefObject<HTMLDivElement>
```
kdembler: It may not be clear what `portal` prop on `Portal` component means. Let's use something like `containerRef`

src/components/Portal.tsx
```diff
@@ -0,0 +1,13 @@
...
+}
+
+const Portal: React.FC<PortalProps> = ({ children, portal }) => {
+  const element = portal.current!
```
kdembler: I'm not sure if we should do this assertion here. I don't think we have certainty that this will not be `null`?

If not, let's just handle this with `if` and `return null` if the container is empty

src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
```diff
@@ -0,0 +1,74 @@
...
+  background-color: ${colors.gray[600]};
+  padding: ${sizes(4)};
+  box-shadow: 0px 8px 8px rgba(0, 0, 0, 0.12), 0px 24px 40px rgba(0, 0, 0, 0.16);
+
+  @media screen and (min-width: ${breakpoints.small}) {
+    padding: ${sizes(6)};
+  }
+`
+
+export const StyledExitButton = styled(Button)`
```
kdembler, line 20: I think this click target may be too big (it's 80x70). We could try reducing the padding on the button a bit

src/components/Dialogs/GeneralDialog/GeneralDialog.tsx
```diff
@@ -0,0 +1,46 @@
...
+    handleOverlayOpen()
+    return () => {
+      handleOverlayClose()
+    }
+  }, [handleOverlayClose, handleOverlayOpen, showDialog])
+  return (
+    overlayContainerRef.current && (
+      <Portal portal={overlayContainerRef}>
+        <CSSTransition in={showDialog} timeout={250} classNames="backdrop" unmountOnExit mountOnEnter>
+          <StyledBackdrop css={dialogTransitions} className="backdrop">
```
kdembler: I know we may have discussed something else, but thinking about it now I don't think the dialog should be a children of the backdrop, doesn't seem to make much semantic sense. Let's make them siblings

src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
kdembler: Btw, just learned you can use `inset: 0;` as a shorthand for all the `top/left/right/bottom` 🤯 🤯

kdembler: nvm, it's supported only by recent Chrome&Firefox :(

src/shared/icons/times-white.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="40" height="40" viewBox="0 0 40 40" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kdembler: Maybe let's just call it for its use, so `close`? Also, what's the difference between this and regular `times`? Do we need both?

src/hooks/useOverlayManager.tsx
```diff
@@ -32,8 +34,9 @@ export const OverlayManagerProvider: React.FC = ({ children }) => {
   return (
     <>
       <Global styles={overlayManagerStyles(scrollbarGap)} />
-      <OverlayManagerContext.Provider value={{ overlayOpen, setOverlayOpen: handleOverlayOpen }}>
+      <OverlayManagerContext.Provider value={{ overlayOpen, setOverlayOpen: handleOverlayOpen, overlayContainerRef }}>
         {children}
+        <div ref={overlayContainerRef}></div>
```
kdembler: This will need some styling to make sure it's not included in the regular page layout. We will need at least:
```css
position: fixed;
top: 0; /* right, bottom, left */
visibility: hidden; /* will need to switch to `visible` when `overlayOpen` */
```

kdembler: Also, if we include `z-index` here, the backdrop/dialog won't have to include it themselves

kdembler (commented):
Looking good, added some comment cause we need to build a strong base here for the future 💪

src/hooks/useOverlayManager.tsx
```diff
@@ -62,5 +79,27 @@ export const useOverlayManager = () => {
     setOverlayOpen(false)
   }, [setOverlayOpen])
 
-  return { handleOverlayOpen, handleOverlayClose }
+  const handleOverlayContainerOpen = useCallback(() => {
+    if (overlayContainerRef.current === null) {
+      return
+    }
+    overlayContainerRef.current.style.visibility = 'visible'
```
kdembler: We shouldn't directly operate on `overlayContainerRef` style, this way it's easy to override styles applied by other components/libraries etc. It's a component that we render ourselves so we can easily do that styling by applying conditional styles.

src/hooks/useOverlayManager.tsx
```diff
@@ -1,17 +1,21 @@
...
 type OverlayManagerContextValue = {
   overlayOpen: boolean
   setOverlayOpen: (value: boolean, scrollbarGap?: number) => void
+  overlayContainerRef: React.RefObject<HTMLDivElement>
 }
 const OverlayManagerContext = React.createContext<OverlayManagerContextValue | undefined>(undefined)
 OverlayManagerContext.displayName = 'OverlayManagerContext'
 
 export const OverlayManagerProvider: React.FC = ({ children }) => {
   const [overlayOpen, setOverlayOpen] = useState(false)
```
kdembler: I think we will need to adjust naming now since this state doesn't actually represent whether there's an overlay opened but rather whether the scroll lock has been applied and I think we need to clear those up. Let's rename this to `scrollLocked` and add `overlayContainerOpened` boolean (we can apply conditional styles based on that, look the other comment).

Let's also clear up the naming for functions. I've originally named the function returned from the hook `handleOverlayOpen` because overlay manager only reacted with applying scroll lock to overlay being opened somewhere else. So my search overlay was calling `handleOverlayOpen` to basically say "hey I've opened an overlay so handle that in whatever way you need to". However, we cannot do the same with `handleOverlayContainerOpen` - this function's purpose isn't to _react_ to overlay container being opened, it's purpose it to actually _open_ the overlay container.

I propose the following names:
```ts
type OverlayManagerContextValue = {
  scrollLocked: boolean
  setScrollLocked: (value: boolean, scrollbarGap?: number) => void
  overlayContainerOpened: boolean
  setOverlayContainerOpened: (value: boolean) => void
  overlayContainerRef: React.RefObject<HTMLDivElement>
}
```

And returned from the hook:
```ts
{
  lockScroll(),
  unlockScroll(),
  openOverlayContainer(),
  closeOverlayContainer(),
  overlayContainerRef
}
```

src/hooks/useOverlayManager.tsx
```diff
@@ -1,39 +1,71 @@
...
 
+  const handleContainerOpened = (value: boolean) => {
+    if (value === overlayContainerOpened) {
+      return
+    }
+    if (value) {
+      setOverlayContainerOpened(true)
+    } else {
+      setOverlayContainerOpened(false)
+    }
```
kdembler: 😅 

```suggestion
    setOverlayContainerOpened(value)
```

src/hooks/useOverlayManager.tsx
```diff
@@ -1,39 +1,71 @@
...
 export const OverlayManagerProvider: React.FC = ({ children }) => {
-  const [overlayOpen, setOverlayOpen] = useState(false)
+  const [scrollLocked, setScrollLocked] = useState(false)
+  const [overlayContainerOpened, setOverlayContainerOpened] = useState(false)
   const [scrollbarGap, setScrollbarGap] = useState(0)
+  const overlayContainerRef = useRef<HTMLDivElement>(null)
 
-  const handleOverlayOpen = (value: boolean, scrollbarGap?: number) => {
-    if (value === overlayOpen) {
+  const handleScrollLocked = (value: boolean, scrollbarGap?: number) => {
```
kdembler: I just noticed this, but there's actually a bug here (introduced by me):

`lockScroll` (old `handleOverlayOpen`) in the hook is created using `useCallback`. That's because it's a dependency to `useEffect` in `SearchOverlayView`, otherwise that `useEffect` would keep getting updated. *But*, those `useCallback`s actually have dependencies of their own (for example this function) which actually aren't memoized with `useCallback` so the `useEffect` will potentially get reexecuted when there's no need to. We need to put this function and `handleContainerOpened` into `useCallback` as well

src/hooks/useOverlayManager.tsx
```diff
@@ -1,39 +1,67 @@
...
+        setScrollLocked(true)
+        setScrollbarGap(scrollbarGap || 0)
+        disableBodyScroll(document.body, { reserveScrollBarGap: true })
+      } else {
+        setScrollLocked(false)
+        setScrollbarGap(0)
+        enableBodyScroll(document.body)
+      }
+    },
+    [scrollLocked]
```
kdembler: This still doesn't fix the issue I've described. Think about what happens with this as a dependency. Let's say we open the search overlay:

1) Overlay gets mounted, it contains this code:
```ts
  useEffect(() => {
    lockScroll()

    return () => {
      unlockScroll()
    }
  }, [lockScroll, unlockScroll])
```
So on mount the overlay will call `lockScroll()` from hook, that in turn will call `setScrollLocked` which is this function.

2) This function will change the value of `scrollLocked` among other things

3) Because `scrollLocked` is defined as a dependency here, `handleScrollLocked` will get updated with a new instance of the function

4) `lockScroll` in the hook has `handleScrollLocked` as a dependency so because of 3) it will get updated

5) the `useEffect` above has `lockScroll` from the hook as a dependency so because of 4) it will get updated

In summary, once the search overlay is mounted, we will call `lockScroll()`, `unlockScroll()` and then `lockScroll()` again or am I wrong?

We will need to remove that dependency here, I guess we can just drop that `if` at the beginning

kdembler (approved):
Good work!

PR #301, opened 2021-02-05
Snackbar

### LINKED ISSUE #247: Create snackbar component/service
#### Description

We need some way to indicate success and snackbars are great for that. See https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3119%3A3264

#### Requirements

- `Snackbar` component added to `src/shared/components`
  - Just the snackbar component, without logic to control it
  - Should have different variants (success/error should do)
- Some kind of service/hook allowing the use of snackbars added
  - Something like this would be great:
  ```tsx
  const displaySnackbar = useSnackbar()
  
  const onClick = () => {
    displaySnackbar({type: 'success', content: 'This is text', timeout: 3000})
  }
  ```
  - Take a look at https://github.com/iamhosseindhv/notistack, if we can use that, that's great, otherwise we should use it as an inspiration
- Component&service added to the Storybook
  - There should be a story with a snackbar permanently open, where we can see how it looks like with different props
  - There should be a story where we can use a button to open snackbars via service

drillprop:
Fix #247


src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { CSSTransition } from 'react-transition-group'
+
+export type DisplaySnackbarArgs = {
+  time?: number
+  variant?: 'success' | 'error' | 'info'
+  buttonText?: string
+  message: string
+}
+
+const snackbarTransitions = css`
```
mikkio-j: maybe it's better to move transition style to `Snackbar.style` to keep hook file cleaner. Or is there a reason why it's here?

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -0,0 +1,39 @@
...
+import styled from '@emotion/styled'
+
+const SNACKBAR_POSITION = 34
+
+export const SnackbarWrapper = styled.div`
+  background-color: ${colors.gray[800]};
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: ${sizes(3)} ${sizes(3)} ${sizes(3)} ${sizes(6)};
```
mikkio-j:
```suggestion
  padding: ${sizes(3)};
  padding-left: ${sizes(6)};
```

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -0,0 +1,39 @@
+import { colors, sizes, typography, zIndex } from '@/shared/theme'
+import styled from '@emotion/styled'
+
+const SNACKBAR_POSITION = 34
```
mikkio-j, line 5: maybe we should make it less on mobile? What do you think?

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -0,0 +1,78 @@
...
+  displaySnackbar: (args: DisplaySnackbarArgs) => void
+  closeSnackbar: () => void
+}
+
+const SnackbarContext = createContext<SnackbarContextValue | undefined>(undefined)
+SnackbarContext.displayName = 'SnackbarContext'
+
+export const SnackbarProvider: React.FC = ({ children }) => {
+  const [isVisible, setIsVisible] = useState(false)
+  const [snackbarOpts, setsnackbarOpts] = useState<DisplaySnackbarArgs | null>(null)
```
kdembler: Should be `setSnackbarOpts`

src/shared/components/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,35 @@
...
+  onClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+}
+
+const SnackbarComponent: React.ForwardRefRenderFunction<HTMLDivElement, SnackbarProps> = (
+  { variant = 'info', message, buttonText = 'Ok', onClick },
+  ref
+) => {
+  return (
+    <>
+      <Global styles={snackbarTransitions} />
```
kdembler: Maybe it'd be a better idea to include that as part of the Snackbar context? I'm not sure what are performance implications of mounting/unmounting `Global` continuously.

src/shared/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+import Button from '../Button'
+import { SnackbarProvider, useSnackbar, DisplaySnackbarArgs } from '@/hooks/useSnackbar/useSnackbar'
+
+export default {
+  title: 'Shared/Snackbar',
+  component: Snackbar,
+  args: {
+    message: 'Hello from snackbar',
+  },
+  decorators: [
```
kdembler, line 13: :+1: I had no idea that existed, will be useful

kdembler (commented):
Looking good, just some minor stuff

mikkio-j (approved):

kdembler (approved):

PR #302, opened 2021-02-05
Tooltip

### LINKED ISSUE #248: Create a tooltip component
#### Description

We should create a tooltip component as seen in the design - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A22070

#### Requirements

- `Tooltip` component added to `src/shared/components`
  - Should allow adding a tooltip to any component. Look at https://material-ui.com/components/tooltips/ - we could use the same API
- Component added to the storybook. The story should have examples of tooltips over different components, e.g. text, button

mikkio-j:
Fix #248


src/shared/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import Tooltip, { TooltipProps } from './Tooltip'
+import { Text, Button } from '@/shared/components'
+import { Regular as ChannelPreview } from '../ChannelPreview/ChannelPreview.stories'
```
kdembler: I think we shouldn't import from other stories, that way we make them dependent on each other. If we need some component, let's import it directly

src/shared/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  title: 'Shared/Tooltip',
+  component: Tooltip,
+  argTypes: {
+    text: { control: 'text', defaultValue: 'Lorem ipsum dolor sit amet!' },
+    arrowDisabled: { control: 'boolean', defaultValue: false },
+  },
+} as Meta
+
+const TextTooltip: Story<TooltipProps> = ({ text, arrowDisabled }) => (
+  <Tooltip text={text} arrowDisabled={arrowDisabled}>
```
kdembler: If we use this instead, we don't need to update the story if we add some new props:

```suggestion
const TextTooltip: Story<TooltipProps> = (args) => (
  <Tooltip {...args}>
```

src/shared/components/Tooltip/Tooltip.style.ts
```diff
@@ -0,0 +1,50 @@
...
+import { colors, sizes, typography, zIndex, transitions } from '@/shared/theme'
+
+type TooltipProps = {
+  isActive?: boolean
+  arrowDisabled?: boolean
+}
+
+export const StyledTooltip = styled.div<TooltipProps>`
+  position: relative;
+  width: fit-content;
```
kdembler: Do we need that? I see the support for `fit-content` isn't great, although we could use it. Could give me some context why we need that?

src/shared/components/Tooltip/Tooltip.style.ts
```diff
@@ -0,0 +1,50 @@
...
+  &:before {
```
kdembler: Let's use `::before` (with double colon), that's the newer syntax for pseudo-elements

src/shared/components/Tooltip/Tooltip.style.ts
```diff
@@ -0,0 +1,50 @@
...
+    min-width: 100px;
+    max-width: 200px;
+    opacity: ${({ isActive }) => (isActive ? 1 : 0)};
+    font-size: ${typography.sizes.caption};
+    font-weight: ${typography.weights.thin};
+    padding: ${sizes(2.5)} ${sizes(2)};
+    background: ${colors.gray[400]};
+    color: ${colors.white};
+    z-index: ${zIndex.nearOverlay};
+    transition: 100ms;
```
kdembler: Let's always explicitly define the transitioned prop, duration and easing

src/shared/components/Tooltip/Tooltip.style.ts
mikkio-j: Tooltip takes whole horizontal space, it does not take children's width into account. I wanted avoid that. Alternatively I could use `display: inline-block` instead

src/shared/components/Tooltip/Tooltip.style.ts
```diff
@@ -0,0 +1,49 @@
...
+    min-width: 100px;
+    max-width: 200px;
+    opacity: ${({ isActive }) => (isActive ? 1 : 0)};
+    font-size: ${typography.sizes.caption};
+    font-weight: ${typography.weights.thin};
+    padding: ${sizes(2.5)} ${sizes(2)};
+    background: ${colors.gray[400]};
+    color: ${colors.white};
+    z-index: ${zIndex.nearOverlay};
+    transition: transform 100ms ${transitions.easing}, opacity 100ms ${transitions.easing};
```
drillprop, line 32:
```suggestion
    transition: transform 100ms ${transitions.easing}, opacity 100ms ${transitions.easing};
    line-height: 16px; 
```
I know that in the design there is no multiline tooltip, but it just feels better with slightly increased line-height,

src/shared/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+import Tooltip, { TooltipProps } from './Tooltip'
+import { Text, Button, ChannelPreview } from '@/shared/components'
+import { Meta, Story } from '@storybook/react'
+
+export default {
+  title: 'Shared/Tooltip',
+  component: Tooltip,
+  argTypes: {
+    text: { control: 'text', defaultValue: 'Lorem ipsum dolor sit amet!' },
+    arrowDisabled: { control: 'boolean', defaultValue: false },
```
drillprop:
```suggestion
    text: { defaultValue: 'Lorem ipsum dolor sit amet!' },
    arrowDisabled: { defaultValue: false },
```
I checked it. You don't really need that - storybook handles it

mikkio-j: Good to know! 👍

PR #303, opened 2021-02-05
new search experience 

kdembler:


src/components/SideNavbar/SideNavbar.style.tsx
```diff
@@ -26,7 +25,7 @@ export const SidebarNav = styled.nav<ExpandableElementProps>`
   top: 0;
   bottom: 0;
   z-index: ${zIndex.header};
-  width: ${({ expanded }) => (expanded ? EXPANDED_SIDENAVBAR_WIDTH : 0)}px;
+  width: ${({ expanded }) => (expanded ? `${EXPANDED_SIDENAVBAR_WIDTH}px` : 'var(--sidenav-collapsed-width)')};
```
drillprop, line 28: Not 100% sure, but could we use CSS variable for `EXPANDED_SIDENAVBAR_WIDTH` as well?

src/components/TopNavbar/TopNavbar.tsx
```diff
@@ -10,35 +10,72 @@ import {
...
   }
 
   const handleCancel = () => {
-    setSearch('')
+    setSearchQuery('')
     setIsFocused(false)
+
+    // navigate to overlaid view or home on searchbar close
+    const overlaidLocation = locationState?.overlaidLocation || { pathname: routes.index() }
+    console.log({ overlaidLocation })
```
drillprop:
```suggestion
```

src/components/ChannelGallery.tsx
```diff
@@ -10,22 +10,31 @@ type ChannelGalleryProps = {
...
+
   return (
     <Gallery title={title} itemWidth={220} exactWidth={true} paddingLeft={sizes(2, true)} paddingTop={sizes(2, true)}>
       {loading
         ? Array.from({ length: PLACEHOLDERS_COUNT }).map((_, idx) => (
             <ChannelPreviewBase key={`channel-placeholder-${idx}`} />
           ))
-        : channels?.map((channel) => <StyledChannelPreview id={channel.id} key={channel.id} />)}
+        : channels?.map((channel) => (
+            <StyledChannelPreview id={channel.id} key={channel.id} onClick={() => handleClick(channel.id)} />
```
drillprop:
```suggestion
        : channels?.map(({ id }) => (
            <StyledChannelPreview id={id} key={id} onClick={() => handleClick({ id })} />
```

src/components/ChannelGrid.tsx
```diff
@@ -11,13 +11,20 @@ const StyledChannelPreview = styled(ChannelPreview)`
...
+    if (onChannelClick) {
+      onChannelClick(id)
+    }
+  }
+
   return (
     <Grid {...gridProps}>
       {channels.map((c) => (
-        <StyledChannelPreview key={c.id} id={c.id} />
+        <StyledChannelPreview key={c.id} id={c.id} onClick={() => handleClick(c.id)} />
```
drillprop:
```suggestion
        <StyledChannelPreview key={id} id={id} onClick={() => handleClick({ id })} />
```

src/views/SearchOverlayView/SearchOverlayView.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { TOP_NAVBAR_HEIGHT } from '@/components/TopNavbar'
+import { QUERY_PARAMS } from '@/config/routes'
+
+const SearchOverlayView: React.FC = () => {
+  const searchQuery = useRouterQuery(QUERY_PARAMS.SEARCH)
+
+  // TODO we should probably handle that in one place for all the overlays
+  // prevent body scroll
+  useEffect(() => {
+    document.body.style.setProperty('overflow', 'hidden')
```
drillprop: Not sure how it looks on another OS but It introduced the so-called [jumping scrollbar](https://aykevl.nl/2014/09/fix-jumping-scrollbar). In my opinion, it's not looking good. Layout shifts a bit when the overlay is showing up. The easiest solution is to just keep the scrollbar always visible, I guess.

kdembler: Damn, I was afraid this would happen. We need some fix for this since we will be introducing more overlays soon. Let me look into possible solutions

src/components/TopNavbar/TopNavbar.tsx
kdembler: wtf, these are commits in my fork:
Screenshot 2021-02-09 at 09 27 34

PR didn't update I guess?

kdembler: (There is commit removing that comment already)

src/components/SideNavbar/SideNavbar.style.tsx
kdembler: I think we could, also for the `NAVBAR_LEFT_PADDING`, but I'm not sure if we should. I introduced this change out of necessity, I think we shouldn't try to introduce it everywhere at this point. Let's try the limited usage for where it's needed and later embrace it fully if we decide it's the right direction

src/views/SearchOverlayView/SearchOverlayView.tsx
mikkio-j: Another simple solution would be adding `overflow-y: overlay;` in body.  Firefox doesn't support that though.

kdembler: Firefox marks it as deprecated and it's not standardized so that means we cannot really use it :(

drillprop (approved):
LGTM

mikkio-j (approved):

PR #305, opened 2021-02-09
Text area

### LINKED ISSUE #246: Create a text area component
#### Description

We'll need a text area component to edit stuff like descriptions. Reference here - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A26329

#### Requirements

- `TextArea` component added to `src/shared/components`
  - Should just be the text area, without any headers
  - Should expose `ref` to its inner `input` element
  - Should allow specifying helper text and warning/error state (same behaviour as described in #240)
  - Should allow passing `maxLength` which specifies the number of characters. The component should display that number in the bottom-left corner (look [design](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3220%3A6478)) only if `currentLength >= maxLength * 0.8`. The component shouldn't do any kind of validation by itself, this will be done outside 
  - Should feature proper hover state (only when non-empty)
- Component added to Storybook

drillprop:
Fix #246

### DIFF
--- a/src/shared/components/TextArea/TextArea.style.ts
+++ b/src/shared/components/TextArea/TextArea.style.ts
@@ -13,7 +13,7 @@ export const StyledTextArea = styled.textarea`
   font-size: ${typography.sizes.subtitle2};
   border: none;
   border-bottom: 1px solid ${colors.gray[900]};
-  transition: background ${transitions.timings.regular} ${transitions.easing};
+  transition: all ${transitions.timings.loading} ${transitions.easing};
   ::placeholder {
     color: ${colors.gray[400]};
   }
--- a/src/shared/components/TextArea/TextArea.style.ts
+++ b/src/shared/components/TextArea/TextArea.style.ts
@@ -1,0 +1,60 @@
+import { colors, sizes, transitions, typography } from '@/shared/theme'
+import styled from '@emotion/styled'
+
+export const TextAreaWrapper = styled.div`
+  max-width: 300px;
+`
+
+export const StyledTextArea = styled.textarea`
+  background: ${colors.black};
+  width: 100%;
+  resize: none;
+  color: ${colors.gray[300]};
+  font-size: ${typography.sizes.subtitle2};
+  border: none;
+  border-bottom: 1px solid ${colors.gray[900]};
+  transition: background ${transitions.timings.regular} ${transitions.easing};
+  ::placeholder {
+    color: ${colors.gray[400]};
+  }
+  &:hover {
+    background: ${colors.gray[900]};
+  }
+  :focus {
+    border-bottom: 1px solid ${colors.blue[500]};
+  }
+  :placeholder-shown {
+    border-bottom: 1px solid ${colors.gray[200]};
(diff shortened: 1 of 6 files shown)


mikkio-j (commented):
Looks good to me 👍

src/shared/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    if (!maxLength) {
+      return
+    }
+
+    const warningLength = maxLength * 0.8
+    const currentLength = e.target.value.length
+
+    setCharactersCount(0)
```
kdembler: Why do we set it to 0 here?

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -0,0 +1,61 @@
+import { colors, sizes, transitions, typography } from '@/shared/theme'
+import styled from '@emotion/styled'
+
+export const TextAreaWrapper = styled.div`
+  max-width: 300px;
+`
+
+export const StyledTextArea = styled.textarea`
+  background: ${colors.black};
```
kdembler: Looked into the design and turns out it has to be transparent. Look here for example: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A18143

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -0,0 +1,61 @@
...
+  color: ${colors.gray[300]};
+  font-size: ${typography.sizes.subtitle2};
+  border: none;
+  border-bottom: 1px solid ${colors.gray[900]};
+  transition: all ${transitions.timings.loading} ${transitions.easing};
+  ::placeholder {
+    color: ${colors.gray[400]};
+  }
+  &:hover {
+    background: ${colors.gray[900]};
```
kdembler: I think we will need to keep it at the `transparent` background at all times. Otherwise, we'd need some padding within the text area. With the colored background you can clearly see the area's boundary which looks visually unsexy without padding

kdembler (commented):
Just some small tweaks

src/shared/components/TextArea/TextArea.tsx
drillprop: Because If a user writes more than 80% of allowed characters and then erases some of them, the counter will not update and won't disappear - will be stuck on `maxLength * 0.8`. 
Couldn't find a better solution for this.

kdembler: Ah I see now, we display the count if `charactersCount !== 0`. What if we changed that condition to something like `charactersWarning === 'warning' || charactersWarning === 'error'`? Then here in the handler, we can just always call `setCharactersCount` and apply the warning state. I think it could be better readable this way

kdembler (approved):

Issue #121, opened 2020-12-09
Rework VideoPreview/ChannelPreview


kdembler:
Currently there are some issues with `VideoPreview` and `ChannelPreview` we didn't foresee when doing initial development:
- Because those live in `shared` and don't accept passing React nodes, we cannot do proper linking from inside. That means #36 is blocked - we can only pass click handlers to the component, but doing basic navigation using `a` elements isn't possible
- It's hard to do proper transitions. Since we use separate components for loading previews and ready previews, we're not able to to easily transition between those two states.

Current components have the feature of being able to display partial placeholders (for example showing all the data when view counts are still loading). We don't use that feature in any way though and it looks like it won't be needed. Getting rid of that and just fully embracing loading/ready state should help with solving that

kdembler:
### Currently

`src/shared/components/VideoPreviewBase`:

- contains layout and accepts 5 nodes as props
- if a node is missing, a placeholder is displayed in its place

`src/shared/components/VideoPreview`:

- accepts all the details about video&channel
- creates nodes based on the details and passes them to the VideoPreviewBase
- accepts `onClick` and `onChannelClick`
- will display "fade-in" whenever it first appears - even when the data is already loaded

`src/components/VideoPreviewWithNavigation`:

- accepts all the details and passed them to `VideoPreview`
- defines click handlers to provide navigation

Data is fetched by consumer and all the fields passed manually to the `VideoPreview`.

### How we could do it

`src/shared/components/VideoPreviewBase`:

- accepts all the video&channel details
- contains layout
- accepts `loading` props, handles it with proper transitions (only when switching between states)
- accepts (something like) `videoHref` and `channelHref` props that will be used to construct links
- possibly still provides `onClick` and `onChannelClick` to allow some additional logic (navigation done through regular `a` though)

`src/components/VideoPreview`:

- fetches data (tries to read from cache) and passes it to `VideoPreviewBase`
- doesn't require any video details passed, only `id`

PR #307, opened 2021-02-10
Refactor video preview

eldiegod:
Second part of the refactor of #121


kdembler (changes requested):
First, pure experience comments, I'll get to the code later:

1. Video previews now have a smaller font size in general compared to the version before changes.

2. After channels are finished loading in the channel gallery, the placeholders are still displayed: 
Screenshot 2021-02-11 at 09 11 19

3. Take a look at `api/client/cache.ts` - we will need the same thing for `video` as we have for `channel` - otherwise in some cases Apollo won't be able to resolve the result directly from cache I think but will try to do a fetch

4. The placeholders now also have the hover animated effect, this shouldn't be the case

5. Take a look at the following usecase in the mocking app:
    a) I search for `aut qui` - the videos gallery should have 3 results
    b) I change the query to `aut qu` and hit enter - new results are being fetched 
    c) Placeholders appear in the videos carousel, but the old videos are still visible for a moment - this looks unsexy. We will need to find a way to alleviate that. As I understand this will happen because we try gracefully transitioning between ready previews and placeholders and it works both ways. One thing we could do is allow passing some kind of key to the video gallery that would be included in `key` when we iterate over the results. If we pass the query as that key then we can be sure that each separate results will not try to reuse the previous components. That's just an idea though, maybe you come up with something else

eldiegod:
@kdembler addressed everything except the font point, it is caused by the placeholders that don't get replaced because we use the same keys thing; wasn't too sure what to do atm

eldiegod:
@kdembler it worked as we expected !

src/components/ChannelGallery.tsx
```diff
@@ -15,24 +15,19 @@ type ChannelGalleryProps = {
...
   if (!loading && channels?.length === 0) {
     return null
   }
 
-  const handleClick = (id: string) => {
-    if (onChannelClick) {
-      onChannelClick(id)
-    }
-  }
+  const handleClick = (id?: string) => () => id && onChannelClick && onChannelClick(id)
```
kdembler: I see why it returns a function now, but I think we'd need to adjust the naming a bit. We haven't used this pattern before (not saying we shouldn't in general), so I think we will need to somehow distinct a general handler and a handler factory like here so it's easy to read the code. First thing that comes to my mind is something like `createClickHandler`. What do you think? I'm open to suggestions

src/components/ChannelPreview.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import ChannelPreviewBase from '../shared/components/ChannelPreview/ChannelPreviewBase'
+import ChannelPreviewBase from '../shared/components/ChannelPreviewBase/ChannelPreviewBase'
```
kdembler:
```suggestion
import { ChannelPreviewBase } from '@/shared/components'
```

src/components/VideoGallery.tsx
```diff
@@ -68,48 +65,29 @@ const VideoGallery: React.FC<VideoGalleryProps> = ({ title, videos, loading, onV
...
-              }}
-            />
-          ))}
+      {[...videos, ...placeholderItems]?.map((video, idx) => (
+        <StyledVideoPreview
+          id={video.id}
+          progress={video?.progress}
+          key={idx}
+          onCoverResize={onCoverResize}
+          onClick={() => {
```
kdembler: Let's use the same approach as in `ChannelGallery`, whatever it ends up being

src/components/VideoPreview.tsx
```diff
@@ -0,0 +1,55 @@
+import { useVideo } from '@/api/hooks'
```
kdembler, line 2: Let's sort the imports a bit. External imports first, so `React` at the top

src/components/VideoPreview.tsx
```diff
@@ -0,0 +1,55 @@
...
+import routes from '@/config/routes'
+import React from 'react'
+import VideoPreviewBase from '../shared/components/VideoPreviewBase/VideoPreviewBase'
```
kdembler:
```suggestion
import { VideoPreviewBase } from '@/shared/components'
```

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -48,6 +155,10 @@ export const CoverHoverOverlay = styled.div`
   align-items: center;
 `
 
+export const InnerContainer = styled.div`
+  display: contents;
```
kdembler: I haven't seen that property used before? Could you briefly explain what it does and why we need it?

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -15,12 +18,116 @@ type ChannelProps = {
...
+const fadeIn = keyframes`
+  0% {
+    opacity: 0
+  }
+  100% {
+    opacity: 1
+  }
+`
+
+export const fadeInAnimation = css`
```
kdembler: Is this still needed? My guess was that this won't be needed anymore now that we use CSS transition

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -1,11 +1,14 @@
...
 import styled from '@emotion/styled'
+import { css, keyframes } from '@emotion/react'
+import { breakpoints, colors, sizes, transitions, typography } from '@/shared/theme'
 import { fluidRange } from 'polished'
-import { colors, sizes, transitions, typography } from '../../theme'
 import Avatar from '../Avatar'
 import Icon from '../Icon'
 import Text from '../Text'
-import { HOVER_BORDER_SIZE, fadeInAnimation } from './VideoPreviewBase.styles'
+import { Link } from 'react-router-dom'
```
kdembler: Let's try to sort the imports - external libraries first

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -0,0 +1,220 @@
...
+              </TextContainer>
+            </InfoContainer>
+          </InnerContainer>
+        </CSSTransition>
+      </SwitchTransition>
+    </Container>
+  )
+}
+
+const SpacedPlaceholder = styled(Placeholder)`
```
kdembler: Those should be imported from the `.style` file as well

src/shared/components/VideoPreviewBase/index.tsx
```diff
@@ -0,0 +1,3 @@
+import VideoPreviewBase, { MIN_VIDEO_PREVIEW_WIDTH } from './VideoPreviewBase'
+export { VideoPreviewBase, MIN_VIDEO_PREVIEW_WIDTH }
```
kdembler: Now that we export one component from here, let's use only default export for the component and named export only for the additional variable

src/components/VideoPreview.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type VideoPreviewProps = {
+  id?: string
+  showChannel?: boolean
```
kdembler: Maybe to make this a little less coupled with the `VideoPreviewBaseProps`, we could extract those "meta" props to a separate type? That type would contain the `showChannel`, `showMeta`, `main` and the event handlers. With that we can say that `type VideoPreviewProps = { ... } & VideoPreviewBaseMetaProps`. Then in the component we also can do `({ id, ..., ...metaProps })` and pass `metaProps` directly to `VideoPreviewBase`

src/components/ChannelGallery.tsx
eldiegod: Yeah I think createClickHandler is a good idea here

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
eldiegod: Yes, it doesn't look as good without both anims

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
eldiegod: It makes it a "phantom" container whose children act like direct children of the parent; in this case, so the anim can work properly and at the same time not to mess up with the parent `Container` flex styles

src/components/VideoPreview.tsx
eldiegod: Yeah I think this is much better, the repetition was bothering me a bit, then there is the progress prop that is an inbetween

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
eldiegod: In the end, removed the CSSTransition in favor of the CSS animation as it was causing too much trouble with the complex layout and its constraints

src/components/VideoPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+  showMeta?: boolean
+  main?: boolean
+  progress?: number
+  className?: string
+}
+
+const VideoPreview: React.FC<VideoPreviewProps & VideoPreviewBaseMetaProps> = ({
```
kdembler: Let's do `& VideoPreviewBaseMetaProps` when declaring the type instead, this way if we export the props type at some point we'll get the same thing as the component accepts

src/components/VideoPreview.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import { useVideo } from '@/api/hooks'
+import routes from '@/config/routes'
+import VideoPreviewBase, { VideoPreviewBaseMetaProps } from '@/shared/components/VideoPreviewBase/VideoPreviewBase'
+
+type VideoPreviewProps = {
+  id?: string
+  showChannel?: boolean
```
kdembler: It seems this and 2 following props here can be omitted since they're already part of the meta props

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -0,0 +1,217 @@
...
+  const clickable = (!!onClick || !!videoHref) && !isLoading
+  const channelClickable = (!!onChannelClick || !!channelHref) && !isLoading
+  const handleChannelClick = (e: React.MouseEvent<HTMLElement>) => {
+    if (!onChannelClick) {
+      return
+    }
+    e.stopPropagation()
+    onChannelClick(e)
+  }
+  const handleAnchorClick = (href?: string) => (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
```
kdembler: Like in the carousels, let's indicate a difference between those handlers (this is a higher order function) and regular handlers (which are just functions)

kdembler (approved):

PR #310, opened 2021-02-10
Avatar update

### LINKED ISSUE #250: Update the avatar component
#### Description

Currently we use `src/shared/components/Avatar` for all channel avatars. Let's update it so we can use it in the channel create/update flows as well. Related design - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A9266

#### Requirements

- Code to generate default avatar based on initials removed, person silhouette (look design) displayed when there's no avatar
- `editable` prop added
  - If prop enabled and there's no avatar, should show the `Add Avatar` state (look design)
  - If prop enabled and there's an avatar, should show the `Edit` state **on hover** (https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A22288)

drillprop:
Fix #250


src/shared/components/Avatar/Avatar.tsx
```diff
@@ -1,40 +1,52 @@
...
+  onEditClick,
+  className,
+  editable,
+}) => {
+  const isEditable = !loading && editable && size !== 'default'
   return (
     <Container onClick={onClick} size={size} className={className}>
       <StyledTransitionGroup>
+        {isEditable && (
+          <EditButton size={size} withAvatar={!!imageUrl} onClick={onEditClick}>
```
kdembler: Do you think `onEditClick` should stop event bubbling? Otherwise when I click the edit button, both handlers will be called

src/assets/avatar-silhouette.svg
```diff
@@ -0,0 +1,11 @@
...
+  <defs/>
+  <mask id="a" width="136" height="136" x="0" y="0" maskUnits="userSpaceOnUse">
+    <circle cx="68" cy="68" r="68" fill="white"/>
+  </mask>
+  <g mask="url(#a)">
+    <circle cx="68" cy="68" r="68" fill="#0E1113"/>
+    <circle cx="68" cy="60" r="28" fill="#1F2529"/>
+    <circle cx="68" cy="164" r="68" fill="#1F2529"/>
+  </g>
+</svg>
```
kdembler: Let's make sure there's a newline at the file end, also for the camera svg

src/assets/avatar-silhouette.svg
```diff
@@ -0,0 +1,11 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 136 136">
+  <defs/>
+  <mask id="a" width="136" height="136" x="0" y="0" maskUnits="userSpaceOnUse">
+    <circle cx="68" cy="68" r="68" fill="white"/>
+  </mask>
+  <g mask="url(#a)">
```
mikkio-j, line 6: Just the idea. Previously we had some issues with `clip-path` property with SVGs that are over `CoverVideo`. In `SearchOverlay` this SVG can be present. Might be worth checking if it behaves properly with `mask` property as well. What do you think?

drillprop: I moved around this SVG and it seems to be working.
image

kdembler (approved):

Issue #311, opened 2021-02-10
Allow removing interrupted videos


kdembler:
One part of the design we've missed includes a button to remove particular video from the "Continue watching" carousel. We should add that. This will require:
a) adding a remove button (as seen [here](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=1006%3A19103)) to the `VideoPreview` (should be 2 optional props: show the button and click callback)
b) making the remove button actually remove the interrupted video from the local storage

PR #313, opened 2021-02-11
Remove button to interrupted videos

mikkio-j:
fix for #311

### DIFF
--- a/src/components/VideoGallery.tsx
+++ b/src/components/VideoGallery.tsx
@@ -22,7 +22,7 @@ type VideoGalleryProps = {
   videos?: VideoFieldsWithProgress[]
   loading?: boolean
   removeButton?: boolean
-  removeInterruptedVideos?: (id: string) => void
+  onRemoveButtonClick?: (id: string) => void
   onVideoClick?: (id: string) => void
 }
 
@@ -48,7 +48,7 @@ const VideoGallery: React.FC<VideoGalleryProps> = ({
   loading,
   onVideoClick,
   removeButton,
-  removeInterruptedVideos,
+  onRemoveButtonClick,
 }) => {
   const [coverHeight, setCoverHeight] = useState<number>()
   const onCoverResize = useCallback((_, imgHeight) => {
@@ -96,9 +96,9 @@ const VideoGallery: React.FC<VideoGalleryProps> = ({
               key={video.id}
               onCoverResize={onCoverResize}
               removeButton={removeButton}
-              handleRemoveInterruptedVideo={() => {
-                if (removeInterruptedVideos) {
-                  removeInterruptedVideos(video.id)
+              onRemoveButtonClick={() => {
+                if (onRemoveButtonClick) {
+                  onRemoveButtonClick(video.id)
                 }
               }}
               onClick={() => {
--- a/src/components/VideoGallery.tsx
+++ b/src/components/VideoGallery.tsx
@@ -1,5 +1,4 @@
 import React, { useState, useMemo, useCallback } from 'react'
-import { usePersonalData } from '@/hooks'
 import { css } from '@emotion/core'
 import styled from '@emotion/styled'
 
@@ -23,6 +22,7 @@ type VideoGalleryProps = {
   videos?: VideoFieldsWithProgress[]
   loading?: boolean
   removeButton?: boolean
+  removeInterruptedVideos?: (id: string) => void
   onVideoClick?: (id: string) => void
 }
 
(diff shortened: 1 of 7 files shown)


src/components/VideoGallery.tsx
```diff
@@ -15,6 +15,8 @@ import { VideoFieldsFragment } from '@/api/queries'
 
 interface VideoFieldsWithProgress extends VideoFieldsFragment {
   progress?: number
+  removeButton?: boolean
```
kdembler: These shouldn't be props on `Video` itself, they don't represent anything about the video data, like `progress`. Instead, we can specify them explicitly in `VideoGalleryProps`. `handleRemove` can be a general handler that's called with the video ID and that will be handled by interrupted video gallery

src/components/VideoGallery.tsx
```diff
@@ -15,6 +15,8 @@ import { VideoFieldsFragment } from '@/api/queries'
...
+  handleRemove?: () => Promise<undefined>
```
kdembler: We should use `Promise<void>` instead, for rationale behind check out https://stackoverflow.com/questions/58885485/why-does-typescript-have-both-void-and-undefined

In this specific case it shouldn't make any difference but let's use `void` for consistency with other cases

src/shared/components/VideoPreview/VideoPreview.styles.tsx
```diff
@@ -56,7 +57,29 @@ export const CoverIcon = styled(Icon)`
   color: ${colors.white};
 `
 
+export const RemoveButton = styled(Button)`
+  position: absolute;
+  top: ${sizes(2)};
+  right: ${sizes(2)};
+  width: 39px;
```
kdembler: Take a look at the inspector and the final results:

1. The button actually doesn't correspond to the size defined here:
Screenshot 2021-02-11 at 12 51 13

2. This is the sizing of that button:
Screenshot 2021-02-11 at 12 51 06

Currently, this `width` isn't respected at all, you can set it to `10px` and the button will still be `42px` wide. That's because it has a horizontal padding that will not be collapsed until you set something like `max-width`. It's only coincidence that `padding*2 + border*2` is pretty close to `39px`.

To alleviate, we can set `padding: 0;` as we want to set the size manually anyway. Let's also make it `40x40` so it's square

src/shared/components/VideoPreview/VideoPreview.styles.tsx
```diff
@@ -56,7 +57,29 @@ export const CoverIcon = styled(Icon)`
...
+  height: 40px;
+  background-color: ${colors.blue};
+  color: ${colors.white};
+  transition: all ${transitions.timings.regular} ${transitions.easing};
+  cursor: pointer;
+  ${CoverIcon} {
+    width: 20px;
```
kdembler: Gotta be careful when setting explicit size for `svg` elements. If you look at the inspector, the `svg` itself has the correct size, but the `span` that contains it is actually `20x22`, it grows additional 2px to the bottom. This actually causes the icon to be vertically misaligned (by 1px but still :P). That's because by default `svg` will have `display: inline;` which makes it's size calculation not straight forward so it's container will be slightly bigger. You can fix that by adding `display: block;` to these styles

src/shared/components/VideoPreview/VideoPreview.styles.tsx
```diff
@@ -56,7 +57,29 @@ export const CoverIcon = styled(Icon)`
...
+    height: 20px;
+    transform: translateY(0);
```
kdembler: Why use `CoverIcon` as styling base? Because of that we need to override its styles which don't have anything to do with this specific icon.  If you use just `Icon` instead, we probably will only need to add `color` and won't need to override `transform`

src/shared/components/VideoPreview/VideoPreview.tsx
```diff
@@ -104,10 +105,9 @@ const VideoPreview: React.FC<VideoPreviewProps> = ({
     if (!onClick) {
       return
     }
-    if (handleRemove) {
-      e.stopPropagation()
-      handleRemove()
-    }
+
+    e.stopPropagation()
+    updateWatchedVideos('REMOVED', id)
```
kdembler: That's not what I meant. Matter of fact, we cannot do that, this component is in `shared`, which means it cannot interact with `usePersonalData` which is Atlas-specific. This component should be purely presentational, it needs to handle all interaction via props and callback. Previous approach was ok, we just shouldn't pass callback via `Video` type - `handleRemove` should be a prop on `VideoGallery` as well

src/shared/components/VideoPreview/VideoPreview.tsx
```diff
@@ -35,6 +35,8 @@ type VideoPreviewProps = {
   duration?: number
   // video watch progress in percent (0-100)
   progress?: number
+  removeButton?: boolean
+  handleRemoveInterruptedVideo?: () => void
```
kdembler: Let's make it consistent with the rest of the handlers and name it `onRemoveButtonClick` (we can also pass the event as the rest of the handlers.

This also has a nice benefit that `VideoPreview` doesn't have to know about `interruptedVideo` - it just has a remove button and it can be also reused for other purposes

src/components/VideoGallery.tsx
```diff
@@ -21,6 +21,8 @@ type VideoGalleryProps = {
   title?: string
   videos?: VideoFieldsWithProgress[]
   loading?: boolean
+  removeButton?: boolean
+  removeInterruptedVideos?: (id: string) => void
```
kdembler: Likewise, let's make it `onRemoveButtonClick`

PR #315, opened 2021-02-11
Pagination component

### LINKED ISSUE #253: Create pagination component
#### Description

We will need a pagination component as seen here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3198%3A4911

#### Requirements

- `Pagination ` component added to `src/shared/components`

drillprop:
Fix #253


src/shared/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <PaginationWrapper>
+      {startingIdx > 1 && (
+        <StyledLink isChevron={true} to={`${url}/${prevPage}`}>
+          <Icon name="chevron-left" />
+        </StyledLink>
+      )}
+      {Array.from({ length: totalPages })
+        .map((_, i) => (
+          <StyledLink key={i + startingIdx} to={`${url}/${i + startingIdx}`}>
+            {i + startingIdx}
```
mikkio-j: First displayed page is 0 - I know it feels natural but I believe we should display 1 as a first page in the component

src/shared/components/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+import { MemoryRouter, Route, useLocation } from 'react-router-dom'
```
mikkio-j:
```suggestion
import { MemoryRouter } from 'react-router-dom'
```

src/shared/components/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+import Pagination from './Pagination'
+
+const Router: React.FC = (children) => {
```
mikkio-j: We don't need that since you are using decorator

mikkio-j (approved):

kdembler (changes requested):
I'll get to the code later, looked at the storybook so far:
1) I'm not sure we should involve routing in page change. I don't see that as a benefit and it certainly complicates the data flow from this component to the consumer. I'd rather see that as a simple `value/onChange` component. What do you think?
2) The right arrow doesn't seem to work, whenever I click it, the 2nd page is chosen, regardless of what is currently selected
3) If I click the last element, the items should "scroll" to reveal the next items. For example, if there are 10 pages, 5 are visible at the beginning and I select the 5th page, the list should probably show pages 2-6 so I'm able to select the next one

src/shared/components/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+import { MemoryRouter } from 'react-router-dom'
+import Pagination from './Pagination'
+
+export default {
+  title: 'Shared/Pagination',
+  component: Pagination,
+  argTypes: {
+    take: {
```
kdembler: This wasn't updated

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useState } from 'react'
+import Icon from '../Icon'
+import { PaginationWrapper, PaginationButton } from './Pagination.style'
+
+type PaginationProps = {
+  itemsForPage?: number
```
kdembler: Let's use `itemsPerPage` or `itemsOnPage`, `for` doesn't seem correct to me

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const Pagination: React.FC<PaginationProps> = ({
+  itemsPerPage = 0,
+  totalCount = 0,
+  currentPage = 1,
+  maxPaginationLinks = 5,
+  onClick,
+}) => {
+  const [page, setPage] = useState(currentPage)
```
kdembler: Let's just rely on value from outside, we probably don't need internal state here

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,72 @@
...
+  totalCount?: number
+  currentPage?: number
+  maxPaginationLinks?: number
+  onClick?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+}
+
+const Pagination: React.FC<PaginationProps> = ({
+  itemsPerPage = 0,
+  totalCount = 0,
+  currentPage = 1,
```
kdembler: Let's use 1-based indexing only for display, the values should be 0-based

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from 'react'
+import Icon from '../Icon'
+import { PaginationWrapper, PaginationButton } from './Pagination.style'
+
+type PaginationProps = {
+  itemsPerPage?: number
+  totalCount?: number
+  currentPage?: number
+  maxPaginationLinks?: number
+  onClick?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```
kdembler: Let's make this `onChange` / `onPageChange` and pass the new page number to the consumer

kdembler (approved):

Issue #245, opened 2021-01-27
Create a multistepper component


kdembler:
#### Description

We'll need a dialog-based multistepper component that allows guiding user through sequential steps. Reference - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3119%3A2454

#### Requirements

- `Multistepper` component added to `src/shared/components`
  - Should let the component parent control the current step
  - Left to figure out whether component itself should handle content updates or whether that's parent's responsibility
- Component added to Storybook (with some example logic to go through steps)

PR #317, opened 2021-02-12
Multistepper

mikkio-j:
fix for #245

### DIFF
--- a/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
+++ b/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
@@ -17,9 +17,6 @@ export const StyledContainer = styled.div`
   @media screen and (min-width: ${breakpoints.small}) {
     padding: var(--dialog-padding-large);
   }
-  &.wide {
-    max-width: 700px;
-  }
 `
 
 export const StyledExitButton = styled(Button)`
@@ -30,29 +27,5 @@ export const StyledExitButton = styled(Button)`
   border: none;
   &:hover {
     background-color: ${colors.transparent};
-  }
-`
-
-export const dialogTransitions = css`
-  &.dialog-enter {
-    opacity: 0;
-    transform: scale(0.88);
-  }
-
-  &.dialog-enter-active {
-    opacity: 1;
-    transform: scale(1);
-    transition: 150ms cubic-bezier(0.25, 0.01, 0.25, 1);
-  }
-
-  &.dialog-exit {
-    opacity: 1;
-    transform: scale(1);
-  }
-
-  &.dialog-exit-active {
-    opacity: 0;
-    transform: scale(0.88);
-    transition: 100ms cubic-bezier(0.25, 0.01, 0.25, 1);
   }
 `
--- a/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
+++ b/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
@@ -3,13 +3,17 @@ import { css } from '@emotion/react'
 import { Button } from '@/shared/components'
 import { colors, sizes, breakpoints } from '@/shared/theme'
 
-export const StyledContainer = styled.div`
+type ContainerProps = {
+  wide?: boolean
+}
+
+export const StyledContainer = styled.div<ContainerProps>`
   position: relative;
   width: 90%;
-  max-width: ${sizes(110)};
+  max-width: ${({ wide }) => (wide ? '700px' : '440px')};
   min-height: 150px;
   margin: ${sizes(32)} auto;
-  background-color: ${colors.gray[600]};
+  background-color: ${colors.gray[700]};
   padding: ${sizes(4)};
   box-shadow: 0px 8px 8px rgba(0, 0, 0, 0.12), 0px 24px 40px rgba(0, 0, 0, 0.16);
   @media screen and (min-width: ${breakpoints.small}) {
--- a/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
+++ b/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
@@ -1,22 +1,20 @@
 import styled from '@emotion/styled'
-import { css } from '@emotion/react'
 import { Button } from '@/shared/components'
 import { colors, sizes, breakpoints } from '@/shared/theme'
 
 export const StyledContainer = styled.div`
-  --dialog-padding-small: ${sizes(4)};
-  --dialog-padding-large: ${sizes(6)};
+  --dialog-padding: ${sizes(4)};
+  @media screen and (min-width: ${breakpoints.small}) {
+    --dialog-padding: ${sizes(6)};
+  }
   position: relative;
   width: 90%;
   max-width: 440px;
   min-height: 150px;
   margin: ${sizes(32)} auto;
   background-color: ${colors.gray[600]};
-  padding: var(--dialog-padding-small);
+  padding: var(--dialog-padding);
   box-shadow: 0px 8px 8px rgba(0, 0, 0, 0.12), 0px 24px 40px rgba(0, 0, 0, 0.16);
-  @media screen and (min-width: ${breakpoints.small}) {
-    padding: var(--dialog-padding-large);
-  }
 `
 
 export const StyledExitButton = styled(Button)`
--- a/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
+++ b/src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
@@ -3,21 +3,22 @@ import { css } from '@emotion/react'
 import { Button } from '@/shared/components'
 import { colors, sizes, breakpoints } from '@/shared/theme'
 
-type ContainerProps = {
-  wide?: boolean
-}
-
-export const StyledContainer = styled.div<ContainerProps>`
+export const StyledContainer = styled.div`
+  --dialog-padding-small: ${sizes(4)};
+  --dialog-padding-large: ${sizes(6)};
   position: relative;
   width: 90%;
-  max-width: ${({ wide }) => (wide ? '700px' : '440px')};
+  max-width: 440px;
   min-height: 150px;
   margin: ${sizes(32)} auto;
-  background-color: ${colors.gray[700]};
-  padding: ${sizes(4)};
+  background-color: ${colors.gray[600]};
+  padding: var(--dialog-padding-small);
   box-shadow: 0px 8px 8px rgba(0, 0, 0, 0.12), 0px 24px 40px rgba(0, 0, 0, 0.16);
   @media screen and (min-width: ${breakpoints.small}) {
-    padding: ${sizes(6)};
+    padding: var(--dialog-padding-large);
+  }
+  &.wide {
+    max-width: 700px;
   }
 `
 
--- a/src/components/Dialogs/Multistepper/Multistepper.stories.tsx
+++ b/src/components/Dialogs/Multistepper/Multistepper.stories.tsx
@@ -28,8 +28,8 @@ const Element: React.FC<ElementType> = ({ whichStep, handleStep, currentStep }) 
     <>
       <h1>Hello! This is a {whichStep} step!</h1>
       <div style={{ display: 'flex', justifyContent: 'space-between' }}>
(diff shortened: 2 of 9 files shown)


src/components/Dialogs/Multistepper/Multistepper.style.ts
```diff
@@ -0,0 +1,81 @@
+import styled from '@emotion/styled'
+import { Text, Button } from '@/shared/components'
+import { colors, sizes, breakpoints, typography } from '@/shared/theme'
+
+type CircleProps = {
+  isFilled?: boolean
+}
+
+type isActiveProp = {
```
drillprop:
```suggestion
type StyledStepInfoProps  = {
```
I know that you used it two times, but still - I think we should stick to types written in uppercase. Also, in my opinion, this is a more appropriate name here

src/components/Dialogs/Multistepper/Multistepper.style.ts
```diff
@@ -0,0 +1,81 @@
...
+      flex-shrink: 1;
+    }
+  }
+`
+
+export const StyledStepsInfoContainer = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+`
+export const StyledStepInfo = styled.div<isActiveProp>`
```
drillprop:
```suggestion
export const StyledStepInfo = styled.div<StyledStepInfoProps>`
```

src/components/Dialogs/Multistepper/Multistepper.style.ts
```diff
@@ -0,0 +1,81 @@
...
+  justify-content: center;
+  align-items: center;
+  width: 32px;
+  height: 32px;
+  border-radius: 100%;
+  border: 1px solid ${colors.gray[400]};
+  background-color: ${({ isFilled }) => (isFilled ? colors.gray[400] : 'transparent')};
+  color: ${({ isFilled }) => (isFilled ? colors.white : colors.gray[300])};
+`
+export const StyledStepInfoText = styled(Text)<isActiveProp>`
```
drillprop:
```suggestion
export const StyledStepInfoText = styled(Text)<StyledStepInfoProps>`
```

src/components/Dialogs/Multistepper/Multistepper.style.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+`
+
+export const StyledStepsInfoContainer = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+`
+export const StyledStepInfo = styled.div<isActiveProp>`
+  display: ${({ isActive }) => (isActive ? 'flex' : 'none')};
+  align-items: center;
```
drillprop, line 50: Small detail but you could add at least `margin-bottom` here or refactor it a bit.  It doesn't look great on all sizes. 

image

drillprop: Check also when `wide` is set to false

src/components/Dialogs/Multistepper/Multistepper.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+  decorators: [
+    (Story) => (
+      <OverlayManagerProvider>
+        <Story />
+      </OverlayManagerProvider>
+    ),
+  ],
+} as Meta
+type ElementType = {
+  whichStep: string
```
drillprop: You could also name it just `step`

src/components/Dialogs/Multistepper/Multistepper.style.ts
```diff
@@ -0,0 +1,81 @@
...
+  isActive?: boolean
+}
+
+export const StyledHeader = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  min-height: 80px;
+  border-bottom: 1px solid ${colors.gray[500]};
+  margin: -${sizes(4)} -${sizes(4)} 0;
```
drillprop: I'm a little confused why these margins and paddings are set like this. Can you do it without negative values?

mikkio-j: The general dialog has this padding because  almost every dialog that will come in the future will have this padding as well, and only few of them will have borders inside. Either I could remove padding from the general dialog and add custom paddings for every dialog specifically or add padding to the general with negative margins only for those dialogs that have border inside. Went for the second approach - thought it's less work in a long run

drillprop (approved):

src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
```diff
@@ -3,13 +3,17 @@ import { css } from '@emotion/react'
...
+
+export const StyledContainer = styled.div<ContainerProps>`
   position: relative;
   width: 90%;
-  max-width: ${sizes(110)};
+  max-width: ${({ wide }) => (wide ? '700px' : '440px')};
   min-height: 150px;
   margin: ${sizes(32)} auto;
-  background-color: ${colors.gray[600]};
+  background-color: ${colors.gray[700]};
```
kdembler: Plot twist: multistepper uses gray-700 but a general dailog gray-600. Must be a mistake on designers part. Let's use the `600` for everything

src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
```diff
@@ -3,13 +3,17 @@ import { css } from '@emotion/react'
...
-export const StyledContainer = styled.div`
+type ContainerProps = {
+  wide?: boolean
+}
+
+export const StyledContainer = styled.div<ContainerProps>`
   position: relative;
   width: 90%;
-  max-width: ${sizes(110)};
+  max-width: ${({ wide }) => (wide ? '700px' : '440px')};
```
kdembler: Since I guess `wide` will be a case specific to the multistepper, I think we can just style the dialog directly inside the multistepper instead of adding a prop here. This is an issue at the moment though, `className` isn't passed to the `StyledContainer` so you cannot really style it at the moment. I had to fix that as well when working on the image crop dialog.

src/components/Dialogs/Multistepper/Multistepper.tsx
```diff
@@ -0,0 +1,62 @@
...
+                <StyledCircle isFilled={isActive || isCompleted}>
+                  {isCompleted ? <Icon name="check" /> : idx + 1}
+                </StyledCircle>
+                <StyledStepInfoText isActive={isActive}>{step.title}</StyledStepInfoText>
+                {isLast ? null : <hr />}
+              </StyledStepInfo>
+            )
+          })}
+        </StyledStepsInfoContainer>
+        {exitButton && (
```
kdembler: Doesn't the base dialog already have this?

src/components/Dialogs/Multistepper/Multistepper.tsx
```diff
@@ -0,0 +1,62 @@
...
+  StyledStepInfo,
+  StyledCircle,
+  StyledStepInfoText,
+  StyledExitButton,
+} from './Multistepper.style'
+import { Icon } from '@/shared/components'
+
+type Step = {
+  title: string
+  element: JSX.Element
```
kdembler: Let's use `React.ReactNode` for the type, that's what we use in other places

src/components/Dialogs/Multistepper/Multistepper.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+type MultistepperProps = {
+  steps: Step[]
+  currentStep?: number
```
kdembler: `currentStep` suggests the prop is of `Step` type - let's use something else, for example `currentStepIdx`

src/components/Dialogs/Multistepper/Multistepper.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import Dialog from '../GeneralDialog/GeneralDialog'
+import { DialogProps } from '@/components/Dialogs/GeneralDialog/GeneralDialog'
```
kdembler: Let's add that to the import above

src/components/Dialogs/Multistepper/Multistepper.style.ts
kdembler: We probably need those margins but I'm a bit concerned with the coupling here, now we will need to remember to update those values if we update the base dialog. Maybe in the base dialog we could try specifying a responsive CSS var, something like `--dialog-padding` and use it both in dialog and here? Let me know your thoughts, I can also share more details on how this could work.

One thing though, do we need a vertical margin/padding as well? I think we could only use the horizontal ones to stretch the border 100%?

kdembler (changes requested):

src/components/Dialogs/Multistepper/Multistepper.tsx
mikkio-j: Positioning differs. `exitButton` in `Multistepper` is aligned with steps' indicators while in base dialog it's positioned with top/right. That's why I decided to overwrite that. Tried using base dialog button at first but didn't come up with something that looks good

src/components/Dialogs/GeneralDialog/GeneralDialog.style.ts
```diff
@@ -4,16 +4,18 @@ import { Button } from '@/shared/components'
 import { colors, sizes, breakpoints } from '@/shared/theme'
 
 export const StyledContainer = styled.div`
+  --dialog-padding-small: ${sizes(4)};
+  --dialog-padding-large: ${sizes(6)};
```
kdembler: We can actually simplify this! This lets us share the values easily but one problem still persists - whatever consumes this data still has to know what breakpoint to use. So there's still some unnecessary coupling between the components - there's some knowledge that you need to have.

One of the awesome things CSS vars allow is actually this: instead of the styles being responsive and changing which value they use, the value itself can be responsive:

```suggestion
  --dialog-padding: ${sizes(4)};
  @media screen and (min-width: ${breakpoints.small}) {
    --dialog-padding: ${sizes(6)};
  }
```

Then in any consumer we don't have to include the media query but just consume the value which will change automatically

src/components/Dialogs/Multistepper/Multistepper.style.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export const StyledHeader = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-start;
+
+  border-bottom: 1px solid ${colors.gray[500]};
+  margin: 0 calc(-1 * var(--dialog-padding-small));
+  padding: 0 var(--dialog-padding-small);
+  padding-bottom: var(--dialog-padding-small);
```
kdembler: Why is this based on dialog padding? I think this component can specify this padding by itself

src/components/Dialogs/Multistepper/Multistepper.style.ts
```diff
@@ -0,0 +1,94 @@
...
+  @media screen and (min-width: ${breakpoints.small}) {
+    margin: 0 calc(-1 * var(--dialog-padding-large));
+    padding: 0 var(--dialog-padding-large);
+    padding-bottom: var(--dialog-padding-large);
+  }
+  hr {
+    display: none;
+    @media screen and (min-width: 800px) {
```
kdembler: Let's use one of breakpoints from the theme here

src/components/Dialogs/Multistepper/Multistepper.tsx
kdembler: Hmm yeah I see. That seems like an inconsistency on designers part. I also had to make some tweaks to the exit button for my image crop. I suggest we use exit button from the base dialog for now even if it doesn't look ideal. Hopefully my changes help, if they don't, we will see something's off during integration and can tweak that then.

src/hooks/useOverlayManager.tsx
```diff
@@ -54,7 +55,11 @@ export const OverlayManagerProvider: React.FC = ({ children }) => {
         }}
       >
         {children}
-        <StyledOverlayContainer ref={overlayContainerRef} isOpened={overlayContainerOpened}></StyledOverlayContainer>
+        <StyledOverlayContainer
+          ref={overlayContainerRef}
+          isOpened={overlayContainerOpened}
+          css={transitionStyles}
```
kdembler: Let's add those to the `Global` above instead

kdembler: (`styles` prop can be an array)

kdembler (approved):

PR #319, opened 2021-02-15
Channel cover

### LINKED ISSUE #251: Extract channel cover to a component
#### Description

We should extract the channel cover we currently use in the channel view to a separate component so it can be reused for publishing views. Designs here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A22494

#### Requirements

- `ChannelCover` component created in `src/components/shared` based off of the code from `ChannelView`
- `editable` prop added
  - If prop enabled, should show the `Click to edit` state (look design) **on hover**

drillprop:
Fix #251


src/shared/components/ChannelCover/ChannelCover.style.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  svg {
+    padding: 10px;
+    margin-right: 12px;
+    border-radius: 100%;
+    width: 40px;
+    border: 1px solid ${colors.white};
+    fill: ${colors.white};
+  }
+  span {
```
mikkio-j: I believe this needs some position tweaks in mobile view - text and svg are not centered

deploy-preview-319--atlas-storybook netlify app__path=_story_shared-textarea--default

src/shared/components/ChannelCover/ChannelCover.style.ts
```diff
@@ -0,0 +1,226 @@
...
+  z-index: 2;
+  max-width: 100%;
+`
+
+export const Title = styled(Text)`
+  line-height: 1;
+  padding: ${sizes(1)} ${sizes(2)} ${sizes(2)};
+  ${fluidRange({ prop: 'fontSize', fromSize: '32px', toSize: '40px' })};
+  background-color: ${colors.gray[800]};
+  text-overflow: ellipsis;
```
mikkio-j: There is a line break instead of `ellipsis`. Maybe try adding `height` property

src/shared/components/ChannelCover/ChannelCover.tsx
```diff
@@ -0,0 +1,154 @@
...
+          <TransitionGroup>
+            <CSSTransition
+              key={showBgPattern ? 'pattern' : 'cover'}
+              timeout={parseInt(transitions.timings.loading)}
+              classNames={transitions.names.fade}
+            >
+              {showBgPattern ? (
+                <BackgroundPattern />
+              ) : (
+                <CoverImage editable={editable && overlayVisible} src={channel?.coverPhotoUrl!} />
```
mikkio-j: I believe it's better to solve it somehow without this assertion - something like adding union type with `null` in `CoverImageProps` and then do conditional styling in `CoverImage`. What do you think?

mikkio-j (approved):

src/shared/components/ChannelCover/ChannelCover.tsx
```diff
@@ -0,0 +1,77 @@
...
+  Media,
+  MediaWrapper,
+  RemoveCoverButton,
+} from './ChannelCover.style'
+
+type BasicChannelCoverProps = {
+  coverPhotoUrl?: string | null
+}
+type EditableProps =
+  | {
```
kdembler: I think we can simplify this and don't do this union, just specify `editable` as boolean and the handlers, I don't see much value coming from this

src/shared/components/ChannelCover/ChannelCover.tsx
```diff
@@ -0,0 +1,77 @@
...
+      editable?: false
+      handleEditCover?: never
+      handleRemoveCover?: never
```
kdembler: Let's adjust the naming 

```suggestion
      onCoverEditClick?: never
      onCoverRemoveClick?: never
```

src/shared/components/ChannelCover/ChannelCover.style.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  background-repeat: no-repeat;
+  background-position: center;
+  background-attachment: local;
+  background-size: cover;
+
+  background-image: linear-gradient(0deg, black 0%, rgba(0, 0, 0, 0.8) 30%, rgba(0, 0, 0, 0) 100%),
+    url(${({ src }) => src});
+  transition: opacity ${transitions.timings.loading} ${transitions.easing};
+  opacity: ${({ editable }) => (editable ? 0.6 : 1)};
```
kdembler: Do we need that opacity change here? If we don't change the opacity of the image itself but just put some gradient/overlay on top of it, then we could much simplify the CSS. We could get rid of mouseenter/mouseleave event handlers and just use `:hover` for `EditableControls` styles. Makes sense?

src/shared/components/ChannelCover/ChannelCover.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <MediaWrapper>
+      {editable && (
+        <EditableControls
+          withImage={!!coverPhotoUrl}
+          onMouseEnter={() => setoverlayVisible(true)}
+          onMouseLeave={() => setoverlayVisible(false)}
+        >
+          <EditCoverButton onClick={handleEditCover}>
+            <Icon name="camera" />
```
kdembler: Compare what's in Figma:
Screenshot 2021-02-18 at 12 12 11

With what we have:
Screenshot 2021-02-18 at 12 12 15

There's a couple of issues:
1. There are problems with the camera icon itself - it looks worse in our version
2. The border is of different width
3. Our circle is actually an ellipse, it's 40x38

kdembler (changes requested):

kdembler:
Forgot to put that in the review comment:

Also take a look at the channel view before and after the changes (first image is from `sumer`):

Screenshot 2021-02-18 at 12 04 47
Screenshot 2021-02-18 at 12 04 42

On mobile, the layout is different and different portion of the image is visible

Screenshot 2021-02-18 at 12 05 23
Screenshot 2021-02-18 at 12 05 32

On desktop, different portion of the image is visible and the gradient shadow is different.

We're extracting the component out of the view, but there shouldn't be any visual changes to the view itself

src/shared/components/ChannelCover/ChannelCover.tsx
```diff
@@ -0,0 +1,70 @@
...
+  onCoverRemoveClick,
+}) => {
+  return (
+    <CoverWrapper>
+      <MediaWrapper>
+        {editable && (
+          <EditableControls>
+            <EditCoverButton onClick={onCoverEditClick}>
+              <EditIconWrapper>
+                <Icon className="desktop" name="camera" />
```
kdembler: Let's try to avoid using `className` like this. This creates a coupling that cannot be typechecked. Instead, we can do a `StyledIcon` that will accept a prop

src/shared/icons/gear.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24">
+  <defs/>
+  <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-miterlimit="10" stroke-width="2" d="M23 12a2 2 0 00-2-2h-1.26a7.97 7.97 0 00-.85-2.06l.89-.89a2 2 0 10-2.83-2.83l-.9.9A7.97 7.97 0 0014 4.25V3a2 2 0 00-4 0v1.26c-.73.2-1.42.48-2.06.85l-.89-.89a2 2 0 10-2.83 2.83l.9.9A7.97 7.97 0 004.25 10H3a2 2 0 000 4h1.26c.2.73.48 1.42.85 2.06l-.89.89a2 2 0 102.83 2.83l.9-.9c.63.38 1.32.67 2.05.86V21a2 2 0 004 0v-1.26a7.97 7.97 0 002.06-.85l.89.89a2 2 0 102.83-2.83l-.9-.9c.38-.63.67-1.32.86-2.05H21a2 2 0 002-2z"/>
+  <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-miterlimit="10" stroke-width="2" d="M12 15a3 3 0 100-6 3 3 0 000 6z"/>
+</svg>
```
kdembler: Missing newline at the end

kdembler (commented):
Looks good! Just some small tweaks

Issue #252, opened 2021-01-27
Create checkbox component


kdembler:
#### Description

We will need a checkbox component. Design here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=489%3A1407

#### Requirements

- `Checkbox ` component added to `src/shared/components`
  - We can skip the error state for now
  - We can skip the label for now
- Component added to the Storybook

PR #320, opened 2021-02-15
Checkbox component

eldiegod:
[ #252](https://github.com/Joystream/atlas/issues/252)

### DIFF
--- a/src/shared/components/Checkbox/Checkbox.styles.tsx
+++ b/src/shared/components/Checkbox/Checkbox.styles.tsx
@@ -1,27 +1,28 @@
-import { sizes, colors } from '@/shared/theme'
+import { sizes, colors, transitions } from '@/shared/theme'
 import { css } from '@emotion/react'
 import styled from '@emotion/styled/'
 
 export const Container = styled.div<CheckboxStateProps>`
   position: relative;
   width: max-content;
-  padding: ${theme.sizes(2)};
-  margin: -${theme.sizes(2)};
+  padding: ${sizes(2)};
+  margin: -${sizes(2)};
   cursor: pointer;
-  border-radius: 9999px;
-  color: ${theme.colors.gray[300]};
+  border-radius: 100%;
+  color: ${colors.gray[300]};
+  transition: background ${transitions.timings.loading} ${transitions.easing};
   ${({ isFocused, isHovered, selected, disabled }) =>
     (isFocused || isHovered) &&
-    `background: ${theme.colors.gray[800]};
-    ${selected && `background: ${theme.colors.blue[900]};`}
+    `background: ${colors.gray[800]};
+    ${selected && `background: ${colors.blue[900]};`}
     ${disabled && `background: transparent;`}`}
 `
 
 const selectedStyles = (props: CheckboxStateProps) =>
   props.selected
     ? css`
-        background-color: ${props.selected ? theme.colors.blue[500] : 'transparent'};
-        border: 1px solid ${theme.colors.blue[500]};
+        background-color: ${props.selected ? colors.blue[500] : 'transparent'};
+        border: 1px solid ${colors.blue[500]};
       `
     : null
 const disabledStyles = (props: CheckboxStateProps) =>
@@ -32,9 +33,9 @@ const disabledStyles = (props: CheckboxStateProps) =>
         `,
         props.selected &&
           css`
-            background-color: ${props.disabled ? theme.colors.gray[700] : 'transparent'};
-            border: 1px solid ${theme.colors.gray[700]};
-            color: ${theme.colors.gray[400]};
+            background-color: ${props.disabled ? colors.gray[700] : 'transparent'};
+            border: 1px solid ${colors.gray[700]};
+            color: ${colors.gray[400]};
           `,
       ]
     : null
@@ -42,11 +43,11 @@ const errorStyles = (props: CheckboxStateProps) =>
   props.error
     ? [
         css`
-          border: 1px solid ${theme.colors.error};
+          border: 1px solid ${colors.error};
         `,
         props.selected &&
           css`
-            background-color: ${theme.colors.error};
+            background-color: ${colors.error};
           `,
       ]
     : null
@@ -59,13 +60,13 @@ export type CheckboxStateProps = {
 }
 export const InnerContainer = styled.div<CheckboxStateProps>`
   transition: all 0.125s ease;
-  color: ${theme.colors.white};
-  border: 1px solid ${theme.colors.gray[300]};
+  color: ${colors.white};
+  border: 1px solid ${colors.gray[300]};
   ${selectedStyles}
   ${errorStyles}
   ${disabledStyles}
   &:active {
-    border: 1px solid ${theme.colors.gray[100]};
+    border: 1px solid ${colors.gray[100]};
   }
 `
 
--- a/src/shared/components/Checkbox/Checkbox.styles.tsx
+++ b/src/shared/components/Checkbox/Checkbox.styles.tsx
@@ -11,11 +11,11 @@ export const Container = styled.div<CheckboxStateProps>`
   border-radius: 100%;
   color: ${colors.gray[300]};
   transition: background ${transitions.timings.loading} ${transitions.easing};
-  ${({ isFocused, isHovered, selected, disabled }) =>
-    (isFocused || isHovered) &&
-    `background: ${colors.gray[800]};
-    ${selected && `background: ${colors.blue[900]};`}
-    ${disabled && `background: transparent;`}`}
+  ${({ isFocused, selected, disabled }) =>
+    isFocused && !disabled && `background: ${selected ? colors.blue[900] : colors.gray[800]};`}
+  &:hover {
+    ${({ selected, disabled }) => !disabled && `background: ${selected ? colors.blue[900] : colors.gray[800]};`}
+  }
 `
 
 const selectedStyles = (props: CheckboxStateProps) =>
@@ -56,7 +56,6 @@ export type CheckboxStateProps = {
   disabled: boolean
   error: boolean
   isFocused: boolean
-  isHovered: boolean
 }
 export const InnerContainer = styled.div<CheckboxStateProps>`
   transition: all 0.125s ease;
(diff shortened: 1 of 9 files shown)


eldiegod:
We need a hover effect for the error state of the component from design

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -0,0 +1,90 @@
...
+  margin: -${theme.sizes(2)};
+  cursor: pointer;
+  border-radius: 9999px;
+  color: ${theme.colors.gray[300]};
+  ${({ isFocused, isHovered, selected, disabled }) =>
+    (isFocused || isHovered) &&
+    `background: ${theme.colors.gray[800]};
+    ${selected && `background: ${theme.colors.blue[900]};`}
+    ${disabled && `background: transparent;`}`}
+`
```
drillprop, line 23: I think you could add some basic transition for the `hover` state.
For example:
```js 
transition: background ${transitions.timings.loading} ${transitions.easing};
```

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { css } from '@emotion/react'
+import styled from '@emotion/styled/'
+
+export const Container = styled.div<CheckboxStateProps>`
+  position: relative;
+  width: max-content;
+  padding: ${theme.sizes(2)};
+  margin: -${theme.sizes(2)};
+  cursor: pointer;
+  border-radius: 9999px;
```
drillprop: In my opinion, more intuitive is to set `border-radius` to just `100%`

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -0,0 +1,90 @@
+import theme from '@/shared/theme'
```
drillprop:
```suggestion
import { sizes, colors } from '@/shared/theme'
```
This is just for consistency, we import in this manner everywhere in the application

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
+// this rule gives trouble on forward refed components
+/* eslint-disable react/display-name */
```
drillprop: Not 100% sure but, I think for debugging reasons we shouldn't disable this. Check the last lines of the `Button` component how it's done.

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
...
+  onFocus?: (event: React.FocusEvent<HTMLDivElement>) => void
+  onBlur?: (event: React.FocusEvent<HTMLDivElement>) => void
+}
+
+export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
+  ({ value, indeterminate, onChange, disabled = false, onFocus, onBlur, error = false, className, ...props }, ref) => {
+    const isIndeterminate = indeterminate || false
+    const isSelected = !!value
+    const [isHovered, setIsHovered] = React.useState<boolean>(false)
+    const [isFocused, setIsFocused] = React.useState<boolean>(false)
```
drillprop:
```suggestion
    const [isHovered, setIsHovered] = useState(false)
    const [isFocused, setIsFocused] = useState(false)
```
You don't need that.

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
+// this rule gives trouble on forward refed components
+/* eslint-disable react/display-name */
+import React from 'react'
```
drillprop:
```suggestion
import React, { useState } from 'react'
```
again, just for consistency

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const [isFocused, setIsFocused] = React.useState<boolean>(false)
+
+    return (
+      <Container
+        selected={value}
+        disabled={disabled}
+        isFocused={isFocused}
+        isHovered={isHovered}
+        error={error}
+        onMouseEnter={() => {
```
drillprop: Not super sure, but are these onMouseEnter and onMouseLeave required? Could we just use CSS here?

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
...
+            }}
+            onFocus={(e) => {
+              if (!disabled) {
+                setIsFocused(true)
+                if (onFocus) {
+                  onFocus(e)
+                }
+              }
+            }}
+            onBlur={(e) => {
```
drillprop: In my opinion, you should create handlers for `onBlur`, `onFocus`, and `onChange`

src/shared/components/Checkbox/Checkbox.styles.tsx
eldiegod: yup, much better

src/shared/components/Checkbox/Checkbox.tsx
eldiegod: ooh i see

src/shared/components/Checkbox/Checkbox.tsx
eldiegod: Added them in case it was needed for more complex cases / renderProps but don't think it will be the case, I will keep it to css

eldiegod: But actually, it's cleaner this way coz otherwise in the styles we would have to extract the Container Hover/Focused css into a separate function to be reused

drillprop: Still think it's much better to remove them. I managed to change it by myself and I think it works well. 
```js
export const Container = styled.div<CheckboxStateProps>`
  position: relative;
  width: max-content;
  padding: ${sizes(2)};
  margin: -${sizes(2)};
  cursor: pointer;
  border-radius: 100%;
  color: ${colors.gray[300]};
  transition: background ${transitions.timings.loading} ${transitions.easing};
  ${({ isFocused, selected, disabled }) =>
    isFocused &&
    `background: ${colors.gray[800]};
    ${selected && `background: ${colors.blue[900]};`}
    ${disabled && `background: transparent;`}`}

  ${({ disabled }) =>
    !disabled &&
    css`
      :hover {
        background: ${colors.gray[800]};
      }
    `};
`
```

drillprop: Still think it's much better to remove `onMouseEnter` and `onMouseLeave`. 
```js

File: src/shared/components/Checkbox/Checkbox.tsx [hunk 9]
```diff
@@ -0,0 +1,86 @@
...
+      if (!disabled) {
+        setIsFocused(false)
+        if (onBlur) {
+          onBlur(e)
+        }
+      }
+    }
+    return (
+      <Container
+        onMouseEnter={onMouseEnter}
```
drillprop: Still think it's much better to remove `onMouseEnter` and `onMouseLeave`. You can just update your `Container` like this
```js
export const Container = styled.div<CheckboxStateProps>`
  position: relative;
  width: max-content;
  padding: ${sizes(2)};
  margin: -${sizes(2)};
  cursor: pointer;
  border-radius: 100%;
  color: ${colors.gray[300]};
  transition: background ${transitions.timings.loading} ${transitions.easing};
  ${({ isFocused, selected, disabled }) =>
    isFocused &&
    `background: ${colors.gray[800]};
    ${selected && `background: ${colors.blue[900]};`}
    ${disabled && `background: transparent;`}`}

  ${({ disabled }) =>
    !disabled &&
    css`
      :hover {
        background: ${colors.gray[800]};
      }
    `};
`
```
And of course, remove `onMouseEnter` and `onMouseLeave` handlers.

kdembler: +1, I totally agree, let's try to do that with just CSS

We could also get rid of focus/blur handlers if we used `:focus-within`, but the support isn't great so let's postpone that for now

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -0,0 +1,91 @@
+import { sizes, colors, transitions } from '@/shared/theme'
+import { css } from '@emotion/react'
+import styled from '@emotion/styled/'
+
+export const Container = styled.div<CheckboxStateProps>`
+  position: relative;
+  width: max-content;
+  padding: ${sizes(2)};
+  margin: -${sizes(2)};
```
kdembler, line 10: Why do we use that negative margin?

src/shared/components/Checkbox/Checkbox.tsx
eldiegod: Aight, did pretty much that but with a tweak

src/shared/components/Checkbox/Checkbox.styles.tsx
eldiegod: A negative margin, in this case, makes it so the checkbox click area is a little bit bigger since its a small component makes it easier to click

src/shared/icons/checkbox/icons-16/Check.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="12" height="11" viewBox="0 0 12 11" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kdembler: I don't think this SVG is actually used anywhere? It wasn't added to `icons/index.ts` and in the component you use the old `check` icon

kdembler (commented):
I'm looking at the issue now and I'm not sure why I've mentioned we can skip the label. Must be a mistake on my part, as we certainly will need the label, as can be seen [here](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4814%3A7618). So the question is, should we let the scope on this PR change or just add the label as a new issue?

src/shared/icons/checkbox/icons-16/Check.svg
eldiegod: I probably forgot to replace the old one along the way, fixed now 👍

eldiegod: I probably forgot to replace the old one along the way, fixed now 👍

kdembler (approved):

Issue #254, opened 2021-01-27
Create menu component


kdembler:
#### Description

We will need a menu component that will allow user to select contextual action. Design here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A6065 (this is a concrete menu, the component should be generic)

#### Requirements

- `Menu` component added to `src/shared/components`
  - Entries will need to include both the text and the icon
- Component added to the Storybook with some example story

PR #321, opened 2021-02-16
Context menu

mikkio-j:
#254


src/hooks/useContextMenu.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [isActive, setMenuActive] = useState(false)
+
+  useEffect(() => {
+    document.addEventListener('click', () => setMenuActive(false), true)
+    return () => {
+      document.removeEventListener('click', () => setMenuActive(false), true)
+    }
+  })
+
+  const openContextMenu = useCallback((event) => {
```
drillprop:
```suggestion
  const openContextMenu = useCallback((event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
```

src/shared/components/ContextMenu/ContextMenu.style.ts
```diff
@@ -0,0 +1,47 @@
...
+  right: ${({ position }) => !position.left && position.x}px;
+  opacity: ${({ isActive }) => (isActive ? '1' : '0')};
+  transform: scale(${({ isActive }) => (isActive ? '1' : '0.88')});
+  background-color: ${colors.gray[800]};
+  width: 200px;
+  color: ${colors.white};
+  padding: ${sizes(2)};
+  z-index: ${zIndex.globalOverlay};
+  transition: opacity 200ms ${transitions.easing}, transform 200ms ${transitions.easing};
+`
```
drillprop, line 21:
```suggestion
  word-break: break-all;
`
```
Also, there is one more minor issue(and I think it occurs also in Snackbar).
Look at the size of the Icon.
image

drillprop (commented):
Looks really neat, but I found one issue - If you open one context menu, and click once more on the button triggering a new dialog, transition animation doesn't occur. 
Not sure if it's easy to fix.

src/hooks/useContextMenu.tsx
drillprop: Sorry, was wrong. 
```suggestion
  const openContextMenu = useCallback((event:  React.MouseEvent) => {
```

src/shared/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import { Portal } from '@/components'
+import { useOverlayManager } from '@/hooks'
+import { CSSTransition } from 'react-transition-group'
+import { StyledContainer, StyledMenuItem, StyledIcon, StyledText, menuTransitions } from './ContextMenu.style'
+import * as Icons from '../../icons'
```
drillprop:
```suggestion
import type { IconType } from '../Icon'
```

src/shared/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type MenuItemProps = {
+  iconName: Icons.IconType
```
drillprop:
```suggestion
  iconName: IconType
```
should work

drillprop (approved):
Nicely done 👍

src/hooks/useContextMenu.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const useContextMenu = () => {
+  const [position, setPosition] = useState({ x: 0, y: 0, left: true })
+  const [isActive, setMenuActive] = useState(false)
+
+  useEffect(() => {
+    document.addEventListener('click', () => setMenuActive(false), true)
+    return () => {
+      document.removeEventListener('click', () => setMenuActive(false), true)
+    }
+  })
```
kdembler: We need a dependency array here (an empty one), otherwise the hook will run on every render

src/hooks/useContextMenu.tsx
```diff
@@ -0,0 +1,34 @@
+import { useState, useCallback, useEffect } from 'react'
+
+export const useContextMenu = () => {
+  const [position, setPosition] = useState({ x: 0, y: 0, left: true })
+  const [isActive, setMenuActive] = useState(false)
+
+  useEffect(() => {
+    document.addEventListener('click', () => setMenuActive(false), true)
+    return () => {
+      document.removeEventListener('click', () => setMenuActive(false), true)
```
kdembler: I think (95% sure) this will not work as intended. The reason is that `removeEventListener` should receive the same _function instance_ (so basically pointer to the same location in memory) to properly remove the event. However, you pass inline function to both `addEventListener` and `removeEventListener`. And since functions defined inline will never be the same (you cannot compare functions by their body, only by reference), you will try to remove a function that wasn't previously added as a listener, which will likely do nothing.

```js
> const a = () => console.log()
> const b = () => console.log()
> a === b
// false
```

src/hooks/useContextMenu.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+  })
+
+  const openContextMenu = useCallback((event: React.MouseEvent) => {
+    const clickPositionFromRight = document.body.clientWidth - event.pageX
+    if (clickPositionFromRight > 200) {
```
kdembler: Ideally we'd avoid this magic number here. If this is supposed to represent width of the menu maybe we could pass it as an argument to the hook?

src/shared/components/ContextMenu/ContextMenu.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+    contextMenuOpts: rightContextMenuOpts,
+  } = useContextMenu()
+  return (
+    <>
+      <div style={{ display: 'flex', justifyContent: 'space-between' }}>
+        <Button onClick={leftOpenContextMenu}>Open menu on the left side</Button>
+        <Button onClick={rightOpenContextMenu}>Open menu on the right side</Button>
+      </div>
+      <div>
+        <ContextMenu contextMenuOpts={leftContextMenuOpts} {...args}>
```
kdembler: What's the difference between this and the other menu? I know they're titled left and right but other than that I think they are the same?

kdembler: I guess I just don't fully understand

kdembler (commented):
Left some comments, but looks solid! I really like the way it behaves and looks

src/shared/components/ContextMenu/ContextMenu.stories.tsx
mikkio-j: The only purpose of two menus in stories is to represent how it behaves when there is more than one instance of it in view, e.g. view with video tiles, where every tile has its own context menu and we are jumping from one menu to another, then we can observe enter/exit transition between them.

I could remove it and leave only one menu in stories. What do you think?

kdembler: Ah ok, it just gave me the impression they should be somehow different but that makes sense

src/hooks/useContextMenu.tsx
```diff
@@ -0,0 +1,35 @@
+import { useState, useCallback, useEffect } from 'react'
+
+export const useContextMenu = () => {
+  const [position, setPosition] = useState({ x: 0, y: 0, left: true })
+  const [isActive, setMenuActive] = useState(false)
+
+  useEffect(() => {
+    const closeContextMenu = () => setMenuActive(false)
+    document.addEventListener('click', closeContextMenu, true)
```
kdembler: I actually thought about this under shower today, maybe we could postpone adding the click handler until needed? Imagine this scenario: we display grid of 40 video previews, each of those has its own `ContextMenu` so on mount we create 40 click handlers, one for each. Shouldn't be too big of a problem I guess, but not ideal.

What if instead we added the click handler only once the menu gets opened? You can actually pass `{once: true}` to `addEventListener` so we wouldn't need to clean it up later. How does that sound?

mikkio-j: Hmmm, would be better that way, sounds good - let's do that

kdembler (approved):

PR #322, opened 2021-02-16
Form Field

### LINKED ISSUE #255: Create form field component
#### Description

We will need a form field component to make form composition easier. As seen here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A14619

#### Requirements

- `FormField ` component added to `src/shared/components`
  - Should allow passing in the input (probably via `children`)
  - Should allow passing a title (required)
  - Should allow passing a description (optional)
- Component added to the Storybook with some example inputs

drillprop:
Fix #255


src/shared/components/FormField/FormField.style.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const FormFieldWrapper = styled.div<FormFieldWrapperProps>`
+  margin-top: ${sizes(2)};
+  width: ${({ fullWidth }) => (fullWidth ? '100%' : '350px')};
+`
+
+export const FormFieldTitle = styled(Text)`
+  margin: 0;
+  line-height: ${sizes(5)};
```
mikkio-j, line 12:
```suggestion
  line-height: ${sizes(5)};
  word-wrap: break-word;
```

src/shared/components/FormField/FormField.style.ts
```diff
@@ -0,0 +1,28 @@
...
+`
+
+export const FormFieldDescription = styled(Text)`
+  margin-top: ${sizes(2)};
+  font-size: ${typography.sizes.subtitle2};
+  line-height: ${sizes(5)};
```
mikkio-j:
```suggestion
  line-height: ${sizes(5)};
  word-wrap: break-word;
```

src/shared/components/FormField/FormField.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import { ChildrenWrapper, FormFieldDescription, FormFieldTitle, FormFieldWrapper } from './FormField.style'
+
+export type FormFieldProps = {
+  title: string
+  description?: string | string[]
```
mikkio-j, line 6: Just an idea - we could always pass an array of strings so there would be no need for ternary operator or type guard in the component - `description` could be mapped even if it has only one string. What do you think? I would still consider that interface friendly

drillprop: I like how it is set up right now. It's more intuitive to pass a string here in the first place. 

I would think about passing an array of strings as an extra option. Still not sure if I should keep it. I just got an idea that can be useful sometimes.

mikkio-j (approved):

src/shared/components/FormField/FormField.style.ts
```diff
@@ -0,0 +1,30 @@
...
+  word-wrap: break-word;
+`
+
+export const FormFieldDescription = styled(Text)`
+  margin-top: ${sizes(2)};
+  font-size: ${typography.sizes.subtitle2};
+  color: ${colors.gray[300]};
+  line-height: ${sizes(5)};
+  word-wrap: break-word;
```
kdembler, line 23: One thing I see in Figma in video upload flow is that the description in general doesn't span as much width as the input itself. Something around 80/90% I think? https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3220%3A1730 Maybe we could add that?

kdembler (commented):
All good, just one small remark

src/shared/components/FormField/FormField.style.ts
drillprop: Had to change the input elements max-widths(and widths) as well. I decided to set the global variable for that. Not sure if it's a good idea - I can always revert that.

kdembler (approved):

Issue #256, opened 2021-01-27
Create datepicker component


kdembler:
#### Description

We will need some kind of datepicker component to allow users to input dates. As seen here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3220%3A1802

#### Requirements

- `Datepicker ` component added to `src/shared/components`
  - For the time being, we can keep it simple and just make it a text field with formatted parts for dd/mm/yyyy
- Component added to the Storybook

PR #325, opened 2021-02-17
Datepicker

mikkio-j:
#256


src/shared/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { forwardRef, useState } from 'react'
+import NumberFormat from 'react-number-format'
+import { parse, isValid } from 'date-fns'
+import TextField from '../TextField'
+
+export type DatepickerProps = {
+  required?: boolean
```
kdembler, line 7: This component currently isn't really usable - there's no way to access its value from outside - the consumer needs to be able to read that date somehow. I guess we could add some `onChange` that would be called as part of the `onBlur` handler. Possibly the validation would be done outside as well. `onChange` would return either a valid or an invalid date, and the form validation would set the error based on that?

src/shared/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,44 @@
...
+  error?: boolean
+  disabled?: boolean
```
kdembler, line 9: Let's also pass the `helperText`

kdembler (changes requested):

src/shared/components/Datepicker/Datepicker.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  },
+} as Meta
+
+const Template: Story<DatepickerProps> = (args) => {
+  const ref = useRef<HTMLInputElement | null>(null)
+  const [date, setDate] = useState<Date>()
+  const [validationError, setValidationError] = useState(false)
+
+  const handleInputChange: (event: React.ChangeEvent<HTMLInputElement>) => void = (e) => {
+    const date = parse(e.target.value, 'dd/MM/yyyy', new Date())
```
kdembler: I think this specific line should live inside the component actually and the `date` should be passed as an argument to the `onChange`. Right now, we're coupling how the datepicker works to all the consumer. Let's say we want to support a different date format later on and we'd need to update all the consumers. If we parse, inside the component, then we only need to update the datepicker itself

src/shared/components/Datepicker/Datepicker.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const handleDateValidation = () => {
+    if (!date) {
+      return
+    }
+    setValidationError(!isValid(date))
+  }
+
+  return (
+    <>
+      <Datepicker {...args} ref={ref} onChange={handleChange} onBlur={handleDateValidation} error={validationError} />
```
kdembler, line 30:
```suggestion
      <Datepicker {...args} ref={ref} onChange={handleChange} onBlur={handleDateValidation} error={validationError} />
      <span>{date ? (isValid(date) ? formatISO(date) : 'Invalid date') : 'No date'}</span>
```

src/shared/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { parse } from 'date-fns'
+import NumberFormat from 'react-number-format'
+import TextField from '../TextField'
+
+export type DatepickerProps = {
+  required?: boolean
+  error?: boolean
+  disabled?: boolean
+  helperText?: string
+  onChange: (date: Date) => void
```
kdembler: I've used the line I proposed for the story to test it out and noticed that if you type anything and then delete it, the result will be forever the `Invalid Date`. This is a small problem because if a user types something inside that field by mistake and then deletes it, the validation will keep thinking that there's an error there. I suggest we make the onChange have `data: Date | null` and in `handleChange` we call it with `null` if there's no value in the text field. What do you think?

kdembler (approved):

Issue #257, opened 2021-01-27
Create playground view


kdembler:
#### Description

Let's add some view that we can use to test out stuff internally

#### Requirements

- `PlaygroundView ` added

PR #326, opened 2021-02-17
Playground view

eldiegod:
#257


eldiegod:
Not sure if should be adding something else in here

kdembler (approved):
That's all we need for now, more stuff will come

Issue #327, opened 2021-02-17
Create circular progress component


kdembler:
#### Description

We will need a component to indicate progress, as seen here: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A10258

#### Requirements

- `CircularProgress` component added to `src/shared/components`
  - Should allow passing `progress` prop and transition smoothly when it changes
- Component added to the storybook.

PR #330, opened 2021-02-18
Circular Progress Bard Component

eldiegod:
#327

### DIFF
--- a/src/shared/components/CircularProgressbar/CircularProgressbar.tsx
+++ b/src/shared/components/CircularProgressbar/CircularProgressbar.tsx
@@ -1,0 +1,63 @@
+import * as React from 'react'
+import { SVG, Background, Trail, StyledPath } from './CircularProgressbar.style'
+
+export const VIEWBOX_WIDTH = 100
+export const VIEWBOX_HEIGHT = 100
+export const VIEWBOX_HEIGHT_HALF = 50
+export const VIEWBOX_CENTER_X = 50
+export const VIEWBOX_CENTER_Y = 50
+
+export type CircularProgressbarProps = {
+  value: number
+  circleRatio: number
+  counterClockwise: boolean
+  maxValue: number
+  minValue: number
+  strokeWidth: number
+  background: boolean
+  backgroundPadding: number
+}
+
+const CircularProgressbar: React.FC<CircularProgressbarProps> = ({
+  value,
+  background = false,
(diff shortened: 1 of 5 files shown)


src/shared/components/CircularProgressbar/CircularProgressbar.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const VIEWBOX_WIDTH = 100
+export const VIEWBOX_HEIGHT = 100
+export const VIEWBOX_HEIGHT_HALF = 50
+export const VIEWBOX_CENTER_X = 50
+export const VIEWBOX_CENTER_Y = 50
+
+export type CircularProgressbarProps = {
+  value: number
+  circleRatio: number
```
kdembler, line 12: What does this represent?

src/shared/components/CircularProgressbar/CircularProgressbar.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from 'react'
+import { SVG, Background, Trail, StyledPath } from './CircularProgressbar.style'
+
+export const VIEWBOX_WIDTH = 100
+export const VIEWBOX_HEIGHT = 100
+export const VIEWBOX_HEIGHT_HALF = 50
```
kdembler:
```suggestion
export const VIEWBOX_HEIGHT_HALF = VIEWBOX_HEIGHT / 2
```

src/shared/components/CircularProgressbar/CircularProgressbar.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const VIEWBOX_CENTER_X = 50
```
kdembler: Aren't those `height(width) / 2` as well?

eldiegod: Yup it is

src/shared/components/CircularProgressbar/CircularProgressbar.tsx
eldiegod: Is how much of the circle the progress bar will cover so 1 for the whole circle or maybe 0.5 for half the circle so at the max value the progressbar will only cover half the circle

kdembler (approved):
Looks real nice!

Issue #285, opened 2021-01-28
Create action bar component


kdembler:
#### Description

Some of the views will reuse an action bar to allow users to save changes. Design:
- first variant - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A6212
- second variant - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3220%3A7007

#### Requirements

- `ActionBar` component added to `src/shared/components`
  - Should enable both design variants (or we do them as separate components)
- Component added to the storybook.

PR #332, opened 2021-02-19
Action bar

mikkio-j:
#285


drillprop (commented):
Not sure how much should we follow the design. The mobile views from the design look much different. It is always horizontal, not vertical like you did here and some content is hidden - not sure if you should change that.  Also, I found that the cancel button looks different in the upload videos view(has a border) 
https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3220%3A16599 


image

image
image

Besides that, the code looks solid. I don't have anything to add. 👍

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+`
+
+export const StyledPrimaryText = styled(Text)`
+  color: ${colors.white};
+  font-family: ${typography.fonts.headers};
+  font-size: ${typography.sizes.h5};
+  font-weight: ${typography.weights.bold};
+  text-align: right;
+  letter-spacing: ${sizes(0.5)};
```
kdembler: Why was this set? Doesn't look very good:

Screenshot 2021-02-22 at 12 20 12

src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,60 @@
...
+  detailsTextIcon,
+  secondaryButtonIcon,
+  onClickPrimaryButton,
+  onClickSecondaryButton,
+}) => {
+  return (
+    <StyledActionBarContainer>
+      <StyledInfoContainer>
+        <StyledPrimaryText>{primaryText}</StyledPrimaryText>
+        <StyledSecondaryText>{secondaryText}</StyledSecondaryText>
```
kdembler, line 44: This should be centred vertically but isn't:
Screenshot 2021-02-22 at 12 20 57

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -0,0 +1,98 @@
+import styled from '@emotion/styled'
+import { colors, sizes, typography, breakpoints } from '@/shared/theme'
+import { Text, Button } from '@/shared/components'
+
+export const StyledActionBarContainer = styled.div`
+  display: flex;
+  width: 100%;
+  background-color: ${colors.gray[800]};
+  padding: ${sizes(3)} ${sizes(8)};
```
kdembler: It seems there are some inconsistencies regarding what padding should be used. Different values are used in the Component/Action bar and upload video view. I'll try to clear that up with designers today, but one thing we'll need to adjust for sure is horizontal padding on mobile - it's too big, let's use `16px` on mobile

src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const ActionBar: React.FC<ActionBarProps> = ({
+  primaryText,
+  secondaryText,
+  primaryButtonText,
+  secondaryButtonText,
+  detailsText,
+  detailsTextIcon,
+  secondaryButtonIcon,
+  onClickPrimaryButton,
```
kdembler: So far we've only used `on...Click`, for example `onEditClick`, `onChannelClick`, let's stay consistent with that naming

src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,60 @@
...
+  primaryButtonText?: string
+  detailsText?: string
+  detailsTextIcon: IconType
+  secondaryButtonText?: string
+  secondaryButtonIcon?: IconType
+  onClickPrimaryButton?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+  onClickSecondaryButton?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+}
+
+const ActionBar: React.FC<ActionBarProps> = ({
```
kdembler, line 28: This works fine as a generic action bar but I think it would be beneficial to also add a component that's specifically targeting the transaction use case, as we will be reusing that for few views. It would use this one inside it. So for example, instead of `primaryText`, it would accept the number of joy and add the `Fee: ...` itself, it'd specify the `secondaryText` internally, etc. The props that aren't consistent will still need to be passed.

This way we will be able to avoid some duplication across views and easier time if we want to update the `secondaryText` for example.

For the text, let's use `Every change in the blockchain state requires paying a nominal fee.`

kdembler (changes requested):

src/shared/components/ActionBar/ActionBarTransaction.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import ActionBar, { ActionBarProps } from './ActionBar'
+
+type ActionBarTransactionProps = {
+  fee: number
+} & ActionBarProps
```
kdembler:
```suggestion
} & Omit<ActionBarProps, 'primaryText' | 'secondaryText'>
```

src/shared/components/ActionBar/ActionBar.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+    primaryButtonText: 'Edit the channel',
+    secondaryButtonText: 'Cancel',
+  },
+} as Meta
+
+const Template: Story<ActionBarProps> = (args) => <ActionBar {...args} />
+
+export const Default = Template.bind({})
+
+const TransactionTemplate: Story<ActionBarProps> = (args) => <ActionBarTransaction {...args} fee={1} />
```
kdembler: Let's use `ActionBarTransactionProps` here so we get proper controls in the story. Then we can also get rid of constant `fee={1}`

kdembler (approved):

PR #333, opened 2021-02-19
Dismissible message

drillprop:


mikkio-j (commented):
do not have any suggestions on that one really, good job 👍

src/hooks/usePersonalData/localStorageClient/localStorageClient.tsx
```diff
@@ -107,14 +108,28 @@ const setRecentSearch = async (id: string, type: 'video' | 'channel') => {
 
   writeToLocalStorage('recentSearches', newRecentSearches)
 }
+
+const dismissedMessages = promisify(() => readFromLocalStorage<DismissedMessage[]>('dismissedMessages') ?? [])
+const setDismissedMessages = async (id: string, add = true) => {
```
kdembler: If we want to use this pattern of one function, we will need to adjust the naming. it says `setDimissiedMessages` so  suggests you use it to set and array, however it accepts an id and boolean as args and modifies state of only one message. so `setDimissiedMessage` I guess works better

kdembler (commented):
Just a small naming thing

src/shared/components/DismissibleMessage/DismissibleMessage.style.ts
```diff
@@ -0,0 +1,46 @@
+import { colors, sizes } from '@/shared/theme'
+import styled from '@emotion/styled'
+import Text from '../Text'
+
+export const MessageWrapper = styled.div`
+  position: relative;
+  margin: ${sizes(6)} 0;
```
kdembler:
```suggestion
```

We probably shouldn't have a margin on the message component itself, as this will vary based on where its used. The view can supply that itself

kdembler (approved):

PR #336, opened 2021-02-22
Create Tabs stories

### LINKED ISSUE #334: Add Tabs component to Storybook
Like in the title, also make sure that the visuals match what's in Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3198%3A6107

drillprop:
Fix #334

### DIFF
--- a/src/shared/components/Tabs/Tabs.styles.tsx
+++ b/src/shared/components/Tabs/Tabs.styles.tsx
@@ -46,7 +46,7 @@ export const BackgroundGradient = styled.div<BackgroundGradientProps>`
   bottom: 0;
   left: ${(props) => (props.direction === 'prev' ? 0 : 'auto')};
   right: ${(props) => (props.direction === 'next' ? 0 : 'auto')};
-  width: 30%;
+  width: 20%;
   z-index: ${zIndex.overlay};
   background-image: linear-gradient(${(props) => (props.direction === 'prev' ? 270 : 90)}deg, transparent, black);
   pointer-events: none;
--- a/src/shared/components/Tabs/Tabs.styles.tsx
+++ b/src/shared/components/Tabs/Tabs.styles.tsx
@@ -24,6 +24,7 @@ export const TabsGroup = styled.div`
   ::-webkit-scrollbar {
     display: none;
   }
+  scrollbar-width: none;
 `
 
 export const Tab = styled.div<TabProps>`
--- a/src/shared/components/Tabs/Tabs.styles.tsx
+++ b/src/shared/components/Tabs/Tabs.styles.tsx
@@ -1,16 +1,34 @@
 import styled from '@emotion/styled'
-import { colors } from '@/shared/theme'
-
-export const TabsGroup = styled.div`
-  display: flex;
-`
+import { colors, zIndex } from '@/shared/theme'
 
 type TabProps = {
   selected: boolean
 }
 
+type BackgroundGradientProps = {
+  direction: 'prev' | 'next'
+}
+
+export const TAB_WIDTH = 120
+
+export const TabsWrapper = styled.div`
+  position: relative;
+  width: 100%;
+`
+
+export const TabsGroup = styled.div`
+  display: flex;
+  position: relative;
+  scroll-behavior: smooth;
+  overflow: auto;
+  ::-webkit-scrollbar {
+    display: none;
+  }
+`
+
 export const Tab = styled.div<TabProps>`
-  width: 120px;
+  width: ${TAB_WIDTH}px;
+  min-width: ${TAB_WIDTH}px;
   padding: 22px 0;
   font-size: 14px;
   color: ${(props) => (props.selected ? colors.white : colors.gray[300])};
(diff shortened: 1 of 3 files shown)


src/shared/components/Tabs/Tabs.styles.tsx
```diff
@@ -1,16 +1,34 @@
...
+  width: 100%;
+`
+
+export const TabsGroup = styled.div`
+  display: flex;
+  position: relative;
+  scroll-behavior: smooth;
+  overflow: auto;
+  ::-webkit-scrollbar {
+    display: none;
```
mikkio-j, line 25: scrollbar in Firefox is still visible, so maybe you could add something like `scrollbar-width: none;` there

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -1,29 +1,64 @@
...
+              setSelected(idx)
+            }}
+            key={`${tab}-${idx}`}
+            selected={selected === idx}
+          >
+            <span>{tab}</span>
+          </Tab>
+        ))}
+      </TabsGroup>
+      {isRightGradientVisible && <BackgroundGradient direction="next" />}
```
kdembler: I see we only display a right shadow which isn't ideal. If we use a shadow as an indication of "some more content is there" then we will need it on both sides once you scroll. It also should disappear when you scroll all the way to the right

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -1,29 +1,112 @@
...
+      if (!tabsRef.current || !width) {
+        return
+      }
+      const isScrollBarVisible = tabsRef.current.scrollWidth > width
+      if (isScrollBarVisible) {
+        setCenter(width / 3)
+      } else {
+        setCenter(0)
+      }
+    },
```
kdembler: Could you explain a bit what's happening here? I don't understand

kdembler: I guess I don't get what `center` is supposed to represent

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -1,29 +1,112 @@
...
+  })
+
+  useEffect(() => {
+    const tabsGroup = tabsRef.current
+    if (!tabsGroup) {
+      return
+    }
+    const currentItemOffsetleft = selected > 0 && TAB_WIDTH * selected
```
kdembler: I think we could simplify the flow here if we move the `selected > 0` condition to the `if` above and return if it's `< 1`

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -1,29 +1,112 @@
...
+        timeout={100}
+        classNames={transitions.names.fade}
+        unmountOnExit
+      >
+        <BackgroundGradient direction="prev" />
+      </CSSTransition>
+      <TabsGroup ref={tabsRef}>
+        {tabs.map((tab, idx) => (
+          <Tab
+            onClick={() => {
```
kdembler: I know this is carried from the old code, but maybe we could elevate that function to a named handler to make it more readable? We can use the `createClickHandler` pattern

kdembler (commented):
Works great. Left some small comments and asked for clarification

src/shared/components/Tabs/Tabs.tsx
drillprop: This variable here basically means "the desired position of the visible middle Tab" 😅 . I used that value to move scroll a bit if the user clicks on a certain tab - `scrollLeft` will change. 
Maybe I'll rename that to just `middleTabPosition`

kdembler: `middleTabPosition` certainly sounds better, we should try to make variable names as descriptive as possible. Now that I know what that is supposed to represent, why do we do `setCenter(width / 3)`?

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -1,29 +1,112 @@
...
+    },
+  })
+
+  useEffect(() => {
+    const tabsGroup = tabsRef.current
+    if (!tabsGroup || selected < 1) {
+      return
+    }
+    const currentItemOffsetleft = TAB_WIDTH * selected
+    if (currentItemOffsetleft) {
```
kdembler: I guess we don't need this condition anymore

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -1,29 +1,95 @@
...
+    const { scrollLeft, clientWidth, scrollWidth } = tabsGroup
+
+    const middleTabPosition = clientWidth / 2 - TAB_WIDTH / 2
+    const currentItemOffsetleft = TAB_WIDTH * selected
+
+    tabsGroup.scrollLeft = currentItemOffsetleft - middleTabPosition
+
+    const touchHandler = throttle(() => {
+      setShadowsVisible({
+        left: scrollLeft > SCROLL_SHADOW_OFFSET,
```
kdembler: `scrollLeft` here will always be 0 because that's what you create handler with

kdembler (changes requested):
Shadows not working currently

kdembler (approved):

Issue #264, opened 2021-01-28
Introduce form validation logic


kdembler:
#### Description

Since we'll be creating a bunch of forms for the publishing, we'll need some solution for form validation. I suggest using https://react-hook-form.com/

#### Requirements

- Some example code added to the playground illustrating form validation
- Possible some extracted logic to be reused, if any

PR #342, opened 2021-02-23
Form validation

mikkio-j:
#264


src/views/PlaygroundView.tsx
```diff
@@ -9,12 +10,12 @@ export const PlaygroundView = () => {
...
       <Text variant="h2">Internal testing view</Text>
       <LinksContainer>
-        <Link to="./first">First</Link>
+        <Link to="./form">Form Validation</Link>
         <Link to="./second">Second</Link>
         <Link to="./third">Third</Link>
       </LinksContainer>
       <Routes>
-        <Route key="first" path="/first" element={<p>First</p>} />
+        <Route key="form" path="/form" element={<PlaygroundForm />} />
```
drillprop: Should we name that differently? Maybe `PlaygroundValidationForm` or `PlaygroundValidation`?
Not sure how this playground will look in the future, but I think there will be more `form` components there.

src/views/Playgrounds/PlaygroundForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+        <FormField title="Description">
+          <Textarea
+            name="textarea"
+            ref={register(textFieldValidation('Description', 3, 20))}
+            error={!!errors.textarea}
+            helperText={errors.textarea?.message}
+          />
+        </FormField>
+
+        <Button>Submit</Button>
```
drillprop: You could clear the form after submit or add a reset button form.

src/views/Playgrounds/PlaygroundForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+              value={checkboxValue}
+              onChange={setCheckboxValue}
+              error={!!errors.check}
+            />
+            <Text>My video features a paid promotion material</Text>
+          </StyledCheckboxContainer>
+        </FormField>
+        <FormField title="Description">
+          <Textarea
+            name="textarea"
```
drillprop:
```suggestion
            name="textarea"
            maxLength={20}
```

Small thing, but you can pass also `maxLength` to `TextArea`. ;)

image

src/views/Playgrounds/PlaygroundForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+type Inputs = {
+  title: string
+  videoSelect: string
+  header: string
+  check: string
+  textarea: string
+}
+
+const PlaygroundForm = () => {
+  const { register, handleSubmit, control, setValue, errors } = useForm<Inputs>({
```
drillprop: Im not super experienced with `react-hook-form` but I think we could omit the `Inputs` type and set `defaultValue` like this: 
```js
    defaultValues: {
      title: '',
      videoSelect: '',
      header: '',
      textarea: '',
    },
```
IMO this will be little bit readable and intuitive

src/views/Playgrounds/PlaygroundForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+    shouldFocusError: false,
+  })
+  const [selectedItem, setSelectedItem] = useState<SelectedItem | null>(null)
+  const [checkboxValue, setCheckboxValue] = useState(false)
```
drillprop: Can we get rid of these two guys? 
Although I tried myself to do this and failed with that. But still think that we could remove at least the checkbox handler - maybe we should do some changes in Checkbox itself

src/views/Playgrounds/PlaygroundForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+import styled from '@emotion/styled'
+
+const items: SelectedItem[] = [
+  { name: 'Public (Anyone can see this video', value: 'public' },
+  { name: 'Private', value: 'private' },
+]
+
+type Inputs = {
+  title: string
+  videoSelect: string
```
drillprop:
```suggestion
  selectedVideo: string
```
or something similar

src/views/Playgrounds/PlaygroundForm.tsx
kdembler: :+1: Most certainly, we shouldn't need to keep any state on our own. Take a look here: https://react-hook-form.com/get-started/#IntegratingControlledInputs

Or is there something stopping us from doing that?

src/views/Playgrounds/PlaygroundForm.tsx
mikkio-j: Yup, was aware of that

kdembler (commented):
Solid work. :+1: for generic validation function. Noticed one thing though: If I submit an empty form and get all the errors and then later pick an option from the select, the helper text stays there:
Screenshot 2021-02-24 at 10 50 01

Not sure why that happens but ideally we'd have that fixed

kdembler (approved):

PR #343, opened 2021-02-23
Add ability to save video drafts

### LINKED ISSUE #266: Add ability to save video drafts
#### Description

As part of the publishing flows, we want to give users and ability so save draft videos. That means videos that have some details provided but others not. We need a way to save them

#### Requirements

- Some logic to add/modify/delete draft videos added
  - We can assume this lives in local storage
  - Should be separate from the `usePersonalData`
- Example added to the playground

drillprop:
Fix #266


src/hooks/useDrafts/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { promisify } from '@/utils/data'
+import { readFromLocalStorage, writeToLocalStorage } from '@/utils/localStorage'
+
+export const draft = async (id: string) => {
```
eldiegod: This could possibly be called `getDraft()` for clarity

src/hooks/useDrafts/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+  const currentDrafts = await drafts()
+  return currentDrafts.find((d) => d.id === id) ?? null
+}
+
+export const drafts = promisify(() => readFromLocalStorage<any[]>('drafts') || [])
```
eldiegod: Same as above could be called `getDrafts()`

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const getSingleDraft = async (draftId: string) => {
+    const singleDraft = await draft(draftId)
+    return singleDraft
+  }
+
+  useEffect(() => {
+    getInitialDrafts()
+  }, [])
+
+  const createOrSaveDraft = async (draftId: string) => {
```
eldiegod: I think `createOrUpdateDraft` here would be more consistent / make more sense

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,45 @@
...
+    clearDrafts()
+    await getInitialDrafts()
+  }
+
+  return {
+    drafts: draftsState,
+    createOrSaveDraft,
+    getSingleDraft,
+    discardDraft,
+    discardAllDrafts,
```
eldiegod:
```suggestion
   getDraft: getSingleDraft,
    removeDraft: discardDraft,
    removeAllDrafts: discardAllDrafts,
```

This could be something like this so the names from the utils don't clash and so we get a nicer API on the hook that's what we are going to be consuming most of the time

eldiegod (commented):
Very clean PR; added some naming suggestions 🚀

eldiegod (approved):
Awesome, approved !

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,45 @@
+import { useEffect, useState } from 'react'
+import { addOrUpdateDraft, clearDrafts, getDraft, getDrafts, removeDraft } from './utils'
+
+type CommonDraftProps = { id: string; recentlyUpdatedAt: string }
```
kdembler: I think `updatedAt` is enough, that's also the same name query node uses

kdembler: My guess it the idea is to allow to store any kind of draft here, not just a video draft, right? If so, we probably should add some field here that will tell us what kind of drafts this is

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,45 @@
...
+    await removeDraft(draftId)
+    await getInitialDrafts()
+  }
+
+  const discardAllDrafts = async () => {
+    clearDrafts()
+    await getInitialDrafts()
+  }
+
+  return {
```
kdembler, line 104: If we save different types of drafts, we will need to provide a way to access video-specific drafts here

src/hooks/useDrafts/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { promisify } from '@/utils/data'
+import { readFromLocalStorage, writeToLocalStorage } from '@/utils/localStorage'
+
+export const getDraft = async (id: string) => {
+  const currentDrafts = await getDrafts()
+  return currentDrafts.find((d) => d.id === id) ?? null
+}
+
+export const getDrafts = promisify(() => readFromLocalStorage<any[]>('drafts') || [])
```
kdembler: We will need to figure out some way to type those, we should avoid `any`

kdembler (changes requested):

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react'
+import { addDraft, clearDrafts, getDraft, getDrafts, removeDraft, updateDraft } from './utils'
+
+export type CommonDraftProps = {
+  id: string
+  updatedAt?: string
+}
+
+export type DraftType = 'videos' | 'channels'
```
kdembler: I suggest using `'video' | 'channel'`, it's a type of a single draft

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export type VideoDraft = {
+  type: 'videos'
+  title?: string
+  description?: string
+  publicness?: 'public' | 'private'
+  publishedBefore?: string
+  marketing?: boolean
+  contentRating?: 'all' | 'mature'
```
kdembler: We should use the same field names as in the schema. So:
- `publicness` -> `isPublic`
- `marketing` -> `hasMarketing`
- `contentRating` -> `isExplicit`
- `publishedBefore` -> `publishedBeforeJoystream`

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+} & CommonDraftProps
+
+export type ChannelDraft = {
```
kdembler: I'm not sure if we should include this here. We will not be using channel drafts for this release. General drafts handling is fine, but this will be dead, unused code for the foreseeable future so I think we should just add `VideoDraft` now

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react'
+import { addDraft, clearDrafts, getDraft, getDrafts, removeDraft, updateDraft } from './utils'
+
+export type CommonDraftProps = {
+  id: string
+  updatedAt?: string
```
kdembler: Should this be nullable? I think we will always have some date, right?

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+} & CommonDraftProps
+
+export type Draft = VideoDraft | ChannelDraft
+
+type DraftState = {
+  videos: VideoDraft[]
+  channels: ChannelDraft[]
+}
+
+type ContextType = {
```
kdembler: Let's call this `DraftsContextValue`

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+  draftsState: DraftState
+  getInitialDrafts: () => Promise<void>
+}
+
+const DraftsContext = React.createContext<undefined | ContextType>(undefined)
+DraftsContext.displayName = 'DraftsContext'
+
+export const DraftsProvider: React.FC = ({ children }) => {
+  const [draftsState, setDraftsState] = useState<ContextType['draftsState']>({
```
kdembler:
```suggestion
  const [draftsState, setDraftsState] = useState<DraftState>({
```

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+      const currentDrafts = await getDrafts()
+      const videos = currentDrafts.filter((draft): draft is VideoDraft => draft.type === 'videos')
+      const channels = currentDrafts.filter((draft): draft is ChannelDraft => draft.type === 'channels')
+      setDraftsState({
+        videos,
+        channels,
+      })
+    } catch (error) {
+      return Promise.reject(error)
+    }
```
kdembler: Not catching the error will have the same effect (promise rejection) so I don't think this is needed here

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export const useDrafts = <T extends DraftType>(type: T) => {
+  const { draftsState, getInitialDrafts } = useContextDrafts()
+
+  const getSingleDraft = async (draftId: string) => {
+    const draft = (await getDraft(draftId)) as GetDraftType<T>
+    return draft
+  }
+
+  const updateSingleDraft = async (draftId: string, draftProps: Omit<Draft, 'id' | 'updatedAt' | 'type'>) => {
```
kdembler: Let's try to extract this `Omit` type since we're reusing it. We can call it `RawDraft` or something

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+  if (ctx === undefined) {
+    throw new Error('useDrafts must be used within a DraftsProvider')
+  }
+  return ctx
+}
+
+export const useDrafts = <T extends DraftType>(type: T) => {
+  const { draftsState, getInitialDrafts } = useContextDrafts()
+
+  const getSingleDraft = async (draftId: string) => {
```
kdembler: Let's put all those functions in `useCallback` so they can be safely used within hooks

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+const DraftsContext = React.createContext<undefined | ContextType>(undefined)
+DraftsContext.displayName = 'DraftsContext'
+
+export const DraftsProvider: React.FC = ({ children }) => {
+  const [draftsState, setDraftsState] = useState<ContextType['draftsState']>({
+    channels: [],
+    videos: [],
+  })
+
+  const getInitialDrafts = useCallback(async () => {
```
kdembler: Not sure if `getInitialDrafts` is a good name. It makes sense within this context provider but we also use it in the hook to refetch the data, so at that point the drafts are no longer initial. Maybe just `fetchDrafts` or something along those lines?

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const { draftsState, getInitialDrafts } = useContextDrafts()
+
+  const getSingleDraft = async (draftId: string) => {
+    const draft = (await getDraft(draftId)) as GetDraftType<T>
+    return draft
+  }
+
+  const updateSingleDraft = async (draftId: string, draftProps: Omit<Draft, 'id' | 'updatedAt' | 'type'>) => {
+    const updatedDraft = await updateDraft(draftId, draftProps)
+    getInitialDrafts()
```
kdembler: We don't await here, but we do in the next function. Is that intentional? If yes, why? If not, let's be consistent about it

src/hooks/useDrafts/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  const newDrafts = currentDrafts.map((draft) =>
+    draft.id === draftId ? { ...draft, ...draftProps, updatedAt } : draft
+  )
+  writeToLocalStorage('drafts', newDrafts)
+  return newDrafts.find((draft) => draft.id === draftId)
+}
+
+export const removeDraft = async (id: string) => {
+  const currentDrafts = await getDrafts()
+  const newDrafts = [...currentDrafts.filter((draft) => draft.id !== id)]
```
kdembler: Why do we create a new array here? `filter` already returns a copy

kdembler (changes requested):
Really like this updated version, good job with typings. Some changes will be needed

kdembler (approved):

PR #344, opened 2021-02-23
add ImageCropDialog component 

### LINKED ISSUE #249: Create a crop&position dialog
#### Description

For a number of assets, user will need to be able to specify desired size and position of an image. This will apply to:
- channel avatars (https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A23999)
- channel covers (this isn't included in the design)
- video thumbnails (https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3851%3A45246)

#### Requirements

- `ImageCropModal` added to `src/shared/components`
  -  (As I imagine it) Should trigger browser file select on mount (if it proves problematic we can try a different approach)
  - Should allow for both circular and 16:9 cropping
  - Should highlight currently "selected" image part and show the other parts of the image greyed out
  - Once user is done, the component should call some callback with information about the selected image's part (I'm not sure what that should be. `x` and `y` in pixels from the upper left corner + some `zoom`?)
  - Feel free to look for some library offering this functionality, as long as it can support the desired usecases
- Component added to the storybook. The story should feature a button triggering file select + crop

### LINKED ISSUE #268: Add image cropping logic
#### Description

As part of the publishing flows, there will be some places where we want to give the user ability to crop&position their images, for example pick the visible part of the avatar image. Since for the time being metadata specify only the file location and not any other information, the cropping needs to be done in Atlas so the cropped file can be uploaded.

#### Requirements

- Logic to crop&position image based on some parameters (this needs to be in sync of what `ImageCropModal` https://github.com/Joystream/atlas/issues/249 works with) added
  - Should allow accessing the cropped file as a blob
- Example added to the playground

kdembler:
Resolves #249
Resolves #268


src/components/Dialogs/ImageCropDialog/cropper.ts
```diff
@@ -0,0 +1,113 @@
...
+  // initialize
+  useEffect(() => {
+    if (!imageEl) {
+      return
+    }
+
+    const handleZoomEvent = (event: Cropper.ZoomEvent<HTMLImageElement>) => {
+      const { ratio, oldRatio } = event.detail
+      console.log({ ratio, oldRatio })
+      // TODO fix min zoom
```
kdembler: This still needs adjustment, it's not very straightforward though. Will fix tomorrow

src/components/Dialogs/ActionDialog/ActionDialog.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
+import BaseDialog, { BaseDialogProps } from '../BaseDialog'
+import { ActionsContainer, ButtonsContainer, AdditionalActionsContainer } from './ActionDialog.style'
+import { Button } from '@/shared/components'
+
+export type ActionDialogProps = {
+  additionalActionsNode?: React.ReactNode
+  primaryButtonText?: string
```
kdembler, line 8: Buttons will probably need to be nodes instead, I'll need `disabled` as well and that seems like passing too many props.

eldiegod (commented):
Very cool compo 🥇
I noticed a couple of things from the storybook:

- Clicking on an upload area and then clicking on cancel when the dialog comes up leads to the dialog coming up with no image and throwing an error after using the zoom.
-  Shouldn't the magnifying glass icons be clickable and increment/decrease the zoom by clicking on them?
- I noticed something weird where after uploading an image and trying to use the zoom, there is a range of the slider where the image doesn't seem affected by zoom changes; it varies by img; it probably has to do with small images that cant be zoomed in more, not sure

src/components/Dialogs/ActionDialog/ActionDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const hasAnyAction = additionalActionsNode || primaryButtonText || secondaryButtonText
+
+  return (
+    <BaseDialog {...baseDialogProps}>
+      {children}
+      {hasAnyAction && (
+        <ActionsContainer>
+          {additionalActionsNode && <AdditionalActionsContainer>{additionalActionsNode}</AdditionalActionsContainer>}
+          <ButtonsContainer>
+            {primaryButtonText && <Button onClick={handlePrimaryButtonClick}>{primaryButtonText}</Button>}
```
drillprop:
```suggestion
            {primaryButtonText && <Button onClick={onPrimaryButtonClick}>{primaryButtonText}</Button>}
```

src/components/Dialogs/ActionDialog/ActionDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (onPrimaryButtonClick) {
+      onPrimaryButtonClick(e)
+    }
+  }
+
+  const handleSecondaryButtonClick = (e: React.MouseEvent) => {
+    if (onSecondaryButtonClick) {
+      onSecondaryButtonClick(e)
+    }
+  }
```
drillprop:
```suggestion
```

src/components/Dialogs/ActionDialog/ActionDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <BaseDialog {...baseDialogProps}>
+      {children}
+      {hasAnyAction && (
+        <ActionsContainer>
+          {additionalActionsNode && <AdditionalActionsContainer>{additionalActionsNode}</AdditionalActionsContainer>}
+          <ButtonsContainer>
+            {primaryButtonText && <Button onClick={handlePrimaryButtonClick}>{primaryButtonText}</Button>}
+            {secondaryButtonText && (
+              <Button variant="secondary" onClick={handleSecondaryButtonClick}>
```
drillprop:
```suggestion
              <Button variant="secondary" onClick={onSecondaryButtonClick}>
```
Should work, right?

src/components/Dialogs/ActionDialog/ActionDialog.tsx
kdembler: Oh yeah, we don't need handlers here. Will get rid of them

drillprop (commented):
Solid work. Zoom has to be tweaked a bit. Is it possible to disable it when the image is smaller than the cropping area? 
Also sometimes you need to move a bit slider(like 10% or 20%) first to start zooming. Not sure if it's fixable.

kdembler:
Fixed the issues with the zoom, should handle all image sizes properly now.

@eldiegod regarding the zoom icons, I don't think they're intended as buttons in the design, nothing seems to point to that. Let's leave it out for now, we can introduce them later if needed

Haven't managed to fix the file cancel on Firefox yet, will try to figure something out

kdembler:
@eldiegod @drillprop Changed the flow to the following:
1. Once there's a click (on the avatar for example), the file picker will be triggered without showing the dialog. This had to be done with the imperative handle and calling a function directly via `ref` unfortunately. I haven't figured out any other way to do this.
2. Then, only if user picked a file the dialog will be opened. If no file was selected, nothing will happen

eldiegod (approved):
Looks solid and the storybook worked as expected as well 👌

Issue #278, opened 2021-01-28
Update the VideoPreview component


kdembler:
#### Description

We will need to update the Video Preview component so it enables all the publishing use cases. This includes:
- Support for different states (unlisted/draft)
- Support for no thumbnail (some default)
- Needs to have an editing mode
  - Displays different icon on hover
  - Can be selectable
  - Has a menu

Relevant designs here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3978%3A4436

**This will be a bit complex so details will be specified in a team call**

PR #345, opened 2021-02-23
Video preview publishing component

eldiegod:
Issue #278


src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -37,6 +48,23 @@ export type VideoPreviewBaseMetaProps = {
...
+      onSelectClick: (value: boolean) => void
+    }
+  | {
+      publisherMode?: false | undefined
+      videoPublishState?: undefined | 'default'
+      selected?: undefined
+      isAnyVideoSelected?: undefined
+      onSelectClick?: undefined
+    }
+
```
eldiegod: Used to avoid impossible states that might lead to bugs i.e. `publisherMode=false` while at the same time `videoPublishState='draft'`

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -202,12 +261,33 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
                     <SpacedPlaceholder height={main ? 16 : 12} width={main ? '40%' : '80%'} />
                   ) : createdAt ? (
                     <MetaText variant="subtitle2" main={main} scalingFactor={scalingFactor}>
-                      {formatVideoViewsAndDate(views ?? null, createdAt, { fullViews: main })}
+                      {/* TODO: draft one should not be createdAt ?? */}
+                      {videoPublishState === 'draft'
+                        ? `Last updated ${formatDateAgo(createdAt)}`
```
eldiegod: @kdembler I think here we need something else from the query node that tells us when the video was last updated, not sure if it already exists

eldiegod: Nvm I think this all will be saved in localStorage

kdembler: Yeah, `last updated` is displayed only for drafts and we will get that from local storage

kdembler (commented):
Looking only at the Storybook atm, will review the code later.

Having said that, great job, I didn't expect it to go that smoothly. Seems to work very well. I've found some small issues:

1. The publisher menu is displayed for a loading tile as well, I guess we should hide it until we have the data:
Screenshot 2021-02-24 at 12 51 42

2. This is probably because of how complex these interactions are: When at least one tile is selected, the rest should:
    a) have a checkbox visible at all times (that's done)
    b) not display the "edit" icon on hover, and allow selecting by clicking anywhere on the cover (that's missing) as seen [here](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3978%3A4883) (let me know if it's not clear what I mean here)

kdembler:
Also seems like the "other tile selected" cover should be darker:
Screenshot 2021-02-24 at 13 11 18

eldiegod:
Pushed some changes and I believe I addressed all the points + added some hover states for the context menu Icon

src/shared/components/VideoPreviewBase/VideoPreview.stories.tsx
```diff
@@ -1,25 +1,63 @@
...
+import { OverlayManagerProvider } from '@/hooks'
 
 export default {
   title: 'Shared/VideoPreview',
   component: VideoPreviewBase,
   argTypes: {
     createdAt: { control: 'date' },
     progress: { control: { type: 'range', min: 0, max: 100, step: 1 } },
+    channelAvatarUrl: { table: { disable: true } },
+    thumbnailUrl: { table: { disable: true } },
```
kdembler: Why are we disabling those? I see how it doesn't make sense for the drafts for example, but it'd still be beneficial to be able to check different thumbnails on a regular preview

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -43,6 +43,9 @@ const clickableAnimation = (clickable: boolean) =>
         ${CoverIcon} {
           transform: translateY(0);
         }
+        ${CoverIcon} {
```
kdembler: Seems like a duplicate of what's above?

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -148,12 +168,33 @@ export const CoverHoverOverlay = styled.div`
 export const CoverIcon = styled(Icon)`
   transform: translateY(40px);
   transition: all ${transitions.timings.regular} ${transitions.easing};
-  width: 54px;
-  height: 54px;
   color: ${colors.white};
 `
-
-export const CoverPlayIcon = ({ ...props }) => <CoverIcon name="play-outline" {...props} />
+export const CoverPlayIcon = ({ ...props }) => (
```
kdembler, line 170: It seems the regular play icon was lost somewhere along the way :P 

Screenshot 2021-02-25 at 13 13 54

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -220,3 +275,35 @@ export const CoverPlaceholder = styled(Placeholder)`
...
+  align-items: center;
+  justify-content: center;
+  width: 40px;
+  height: 40px;
+  cursor: pointer;
+  position: relative;
+  border-radius: 100%;
+  transition: all ${transitions.timings.regular} ${transitions.easing};
+
+  path:not([fill='none']) {
```
kdembler: That's an unusual selector, what's happening here?

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -197,17 +261,39 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
...
                 ))}
               {showMeta && (
-                <MetaContainer main={main}>
+                <MetaContainer noMarginTop={showChannel === false} main={main}>
                   {isLoading ? (
                     <SpacedPlaceholder height={main ? 16 : 12} width={main ? '40%' : '80%'} />
                   ) : createdAt ? (
                     <MetaText variant="subtitle2" main={main} scalingFactor={scalingFactor}>
-                      {formatVideoViewsAndDate(views ?? null, createdAt, { fullViews: main })}
+                      {/* TODO: instead of createdAt add the lastUpdate from useDraft hook */}
```
kdembler: I think we can just leave it out as it is. We can assume that `createdAt` in case of a draft means `updatedAt`

```suggestion
```

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -197,17 +261,39 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
...
                   ) : null}
                 </MetaContainer>
               )}
             </TextContainer>
+            {publisherMode && !isLoading && (
+              <ContextMenuContainer>
+                <KebabMenuIconContainer onClick={(e) => openContextMenu(e, 200)}>
+                  <KebabMenuIcon />
+                </KebabMenuIconContainer>
+                {/* TODO: menu to be integrated with the video publishing functions */}
```
kdembler: Maybe we should just provide a callback for all of those in the props? Then the consumer can take care of this by itself

src/shared/icons/index.ts
```diff
@@ -49,13 +55,19 @@ const icons = [
   'chevron-left',
   'check',
   'dash',
+  'kebab-menu',
+  'pencil',
+  'pencil2',
```
kdembler: Maybe let's try to make the name describe the difference? It will be easy to get lost in what `pencil` vs `pencil2` is. Maybe something like `big-pencil` and `small-pencil` 😆 At least it will give us _some_ info I guess :D

kdembler (commented):
Mostly minor stuff, looking really good

src/shared/components/VideoPreviewBase/VideoPreview.stories.tsx
eldiegod: makes sense

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
eldiegod: Yes, I forgot to delete that one

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
eldiegod: Uhh, was assuming `isAnyVideoSelected` was a `boolean` but in reality, it is `boolean | undefined`

src/shared/icons/index.ts
eldiegod: haha will try to figure out something 👌

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
eldiegod: This is to target the path's tags inside the svg's so they transition nicely and to change the color of the icon on hover

eldiegod: but actually just realised I dont need the `:not([fill='none']) ` part on the first one

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
eldiegod: Added them but I'm not too happy with it atm, I think I'll end up lifting this component later to VideoPreview or something like that

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -99,49 +143,78 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
...
-    e.stopPropagation()
     onChannelClick(e)
   }
   const createAnchorClickHandler = (href?: string) => (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
     if (!href) {
       e.preventDefault()
     }
   }
+  const handleCoverHoverOverlayClick = () => {
+    isAnyVideoSelected && onSelectClick?.(!isSelected)
```
kdembler: Seems we need to stop propagation of the event here, otherwise clicking the cover will also trigger the regular `onClick`. This also applies to the checkbox in the regular state, without any other video selected

eldiegod: Added a case in there to handle onClick instead, think it makes more sense to have these together

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -197,17 +274,43 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
                   </Anchor>
                 ))}
               {showMeta && (
-                <MetaContainer main={main}>
+                <MetaContainer noMarginTop={showChannel === false} main={main}>
```
kdembler:
```suggestion
                <MetaContainer noMarginTop={!showChannel} main={main}>
```

Shouldn't really be a problem in this case, but in general this should be a safer construct (for example to handle `undefined` as well)

kdembler (approved):
Great work on this! Will probably need some small refinements to make it fully useable but I believe this can be done further along the way. LGTM!

Issue #340, opened 2021-02-22
Update the button component


kdembler:
Recently, the design for buttons was updated to improve sizing, etc.

We should update our button component with accordance to the design - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5002%3A2808

#### Requirements
- Should use the new size naming - `large`, `medium` (default), `small`
- Should use the correct positions and sizes for icons for each button size
- Should resemble what's in Figma
- Should make sure the `ToggleButton` component looks the same as it did before the change

PR #346, opened 2021-02-24
Button update

mikkio-j:
#340


src/shared/components/Button/Button.style.ts
```diff
@@ -103,25 +103,16 @@ const sizeFromProps = ({ size = 'regular', full, hasText }: ButtonStyleProps) =>
...
-          color: ${colors.gray[200]};
-          background-color: ${colors.gray[400]};
-          border-color: ${colors.gray[400]};
-        }
-        &:active {
-          color: ${colors.gray[200]};
-          background-color: ${colors.gray[400]};
-          border-color: ${colors.gray[400]};
-        }
+        filter: brightness(50%);
```
kdembler: This is a nice shortcut, but it will not produce the values specified by the design, I think we should implement what's in there.

src/shared/components/Button/Button.style.ts
```diff
@@ -103,25 +103,16 @@ const sizeFromProps = ({ size = 'regular', full, hasText }: ButtonStyleProps) =>
...
+        pointer-events: none;
       `
     : null
 
 export const StyledIcon = styled(Icon)<IconStyleProps>`
   flex-shrink: 0;
+  width: ${({ size }) => (size === 'large' ? '24px' : '14px')};
```
kdembler: Couple points here:
1. I know it's not specified in Figma yet but designers mentioned to me that the medium button will use a `20px` icon. Forgot to mention that, sorry
2. Since the values here are repeated a couple of times I suggest we define them as variables so we don't need to do too much digging if we decide to change them

src/shared/components/Button/Button.tsx
```diff
@@ -46,7 +46,7 @@ const ButtonComponent: React.ForwardRefRenderFunction<HTMLButtonElement, ButtonP
       size={size}
       ref={ref}
     >
-      {icon && <StyledIcon disabled={disabled} name={icon} />}
+      {icon && <StyledIcon disabled={disabled} name={icon} size={size} />}
       {children && <span>{children}</span>}
```
kdembler, line 50: There are some issues with typography. The font sizes don't match what's currently in Figma. Also the line heights are wrong producing wrong button sizes. This isn't necessarily a fault in your work here, but rather some deeper problems. I fixed some of those in my image crop dialog. I think I'll open a separate PR to make sure typography matches exactly what's in the design, hopefully that helps

src/shared/icons/block.svg
```diff
@@ -1 +1,10 @@
-<svg width="16" height="16" fill="none" xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" r="6" stroke="#7B8A95" stroke-width="2" stroke-linecap="round"/><path d="M4 4l8 8" stroke="currentColor" stroke-width="2" stroke-linecap="round"/></svg>
\ No newline at end of file
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kdembler: Not sure if you're aware of this tool: https://jakearchibald.github.io/svgomg/

It's a really great way to clean up SVGs and reduce their size. I usually copy the full SVG content and paste it into "Paste markup". Then I tick the `prettify` option and copy markup. In this case the result is only 40% size of the original

Let's put this icon through it

kdembler (changes requested):
Since we basically changed the default button size from `large` to `medium`, we will need to make sure the app still looks fine. I've found a couple of places that will need fixes:

Screenshot 2021-02-24 at 13 21 28

Screenshot 2021-02-24 at 13 21 35
(carousel right button)

Screenshot 2021-02-24 at 13 36 48

Let's try to produce output as close as possible to what we had before those changes

src/shared/components/Button/Button.style.ts
```diff
@@ -6,29 +6,33 @@ import { colors, typography } from '../../theme'
...
-        color: ${colors.blue[500]};
+        color: ${colors.white};
         &:hover {
-          color: ${colors.blue[300]};
+          background-color: ${colors.gray[800]};
+          border-radius: ${circleRadius};
         }
         &:active {
-          color: ${colors.blue[700]};
+          background-color: transparent;
```
kdembler: Those are the ones I've mentioned were getting updated in the design, take a look at Figma

src/shared/components/Button/Button.style.ts
```diff
@@ -72,53 +75,116 @@ const colorsFromProps = ({ variant }: ButtonStyleProps) => {
...
 
-const sizeFromProps = ({ size = 'regular', full, hasText }: ButtonStyleProps) => {
-  let padding, fontSize
+const sizeFromProps = ({ size = 'medium', full, hasText }: ButtonStyleProps) => {
+  let padding, fontSize, buttonSizeWithoutText
   switch (size) {
-    case 'smaller': {
-      padding = '10px'
+    case 'small': {
+      padding = '9.5px 12px'
```
kdembler: I don't think we need this additional `.5px`? The button seems to have the correct size (`32px`) with `9px` vertical padding

src/shared/components/Button/Button.style.ts
```diff
@@ -72,53 +75,116 @@ const colorsFromProps = ({ variant }: ButtonStyleProps) => {
...
+      padding = hasText ? `15px 20px` : '14px'
       fontSize = typography.sizes.button.large
+      buttonSizeWithoutText = '48px'
       break
     }
   }
   return css`
-    width: ${full ? '100%' : ''};
+    width: ${!hasText ? buttonSizeWithoutText : full ? '100%' : ''};
+    height: ${!hasText ? buttonSizeWithoutText : 'auto'};
```
kdembler: Why do we need to set the height explicitly? I think we should be fine with just the padding?

src/shared/components/Button/Button.style.ts
```diff
@@ -72,53 +75,116 @@ const colorsFromProps = ({ variant }: ButtonStyleProps) => {
...
+}
+const iconSizeFromProps = ({ size: sizeProp = 'medium', hasText }: IconStyleProps) => {
+  let size, margin
+  if (!hasText) {
+    size = 'auto'
+  } else {
+    switch (sizeProp) {
+      case 'small': {
+        size = typography.sizes.icon.small
+        margin = '-9.5px 0'
```
kdembler: We should put those values in some variables since they're used in more than one place. This way it will be easy to miss this when changing the paddings

kdembler (changes requested):
Some small improvements needed

src/shared/components/Button/Button.style.ts
mikkio-j: It's here to keep 1:1 ratio for buttons with just the icon

src/shared/components/Button/Button.style.ts
```diff
@@ -127,12 +197,26 @@ export const StyledIcon = styled(Icon)<IconStyleProps>`
     margin-left: 10px;
   }
   filter: ${(props) => (props.disabled ? 'brightness(0.7)' : null)};
+  ${iconSizeFromProps}
 `
 export const StyledButton = styled.button<ButtonStyleProps>`
+  --size-without-text-small: 32px;
```
kdembler: I don't think we necessarily need to put those in CSS vars. I meant storing them in regular JS vars. I don't think those values will be used outside of this component, so we don't need to worry about sharing

kdembler (commented):
Just one last code comment added. Also, I've noticed one of the icon buttons is incorrectly sized:

Screenshot 2021-03-05 at 14 21 29

FYI, the design assumes the icon inside the button will always be 16x16 (there may be some edge cases).

Food for thought: do you think it'd make sense to have icon button (just icon) as a separate component? I've noticed we have some cases to target specifically this situation. Not saying we should do that here, just something to think about

kdembler:
And yeah, one thing the designers changed again is the sizing of icons inside the button. The previous version used `24px`, `20px` and `16px`. The final approach is to always use `16x16` inside the buttons, regardless its size. Sorry you need to change that again 😞

src/shared/components/Button/Button.style.ts
```diff
@@ -72,67 +75,112 @@ const colorsFromProps = ({ variant }: ButtonStyleProps) => {
...
 
-const sizeFromProps = ({ size = 'regular', full, hasText }: ButtonStyleProps) => {
-  let padding, fontSize
+const sizeFromProps = ({ size = 'medium', full, hasText }: ButtonStyleProps) => {
+  let padding, fontSize, buttonSizeWithoutText
   switch (size) {
-    case 'smaller': {
-      padding = '10px'
+    case 'small': {
+      padding = 'var(--vertical-padding-small) var(--horizontal-padding-small)'
```
kdembler: Shouldn't this have `hasText ?` as well?

src/shared/components/Button/Button.style.ts
```diff
@@ -72,67 +75,112 @@ const colorsFromProps = ({ variant }: ButtonStyleProps) => {
...
-      padding = hasText ? `14px 20px` : '14px'
+      padding = hasText ? `var(--vertical-padding-large) var(--horizontal-padding-large)` : '0'
       fontSize = typography.sizes.button.large
+      buttonSizeWithoutText = '48px'
       break
     }
   }
   return css`
-    width: ${full ? '100%' : ''};
+    width: ${!hasText ? buttonSizeWithoutText : full ? '100%' : ''};
```
kdembler, line 103: Probably worth adding `min-width` for when we specify the width explicitly, so the buttons don't get squeezed

kdembler (commented):
Just last tweaks

kdembler (changes requested):
Was just about to merge but checked the app and there are problems again with the icon buttons:

Screenshot 2021-03-08 at 18 08 36
Screenshot 2021-03-08 at 18 08 28

PR #347, opened 2021-02-24
Add label to checkbox

### LINKED ISSUE #338: Add label to the checkbox component
Currently `Checkbox` component doesn't offer a way to use a label with it. We should add it. It should use the native `label` element and toggle the checkbox on click

drillprop:
Fix #338


src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -1,7 +1,24 @@
...
+
+type CheckboxLabelProps = {
+  disabled?: boolean
+}
+
+export const CheckboxLabel = styled.label<CheckboxLabelProps>`
+  padding: ${sizes(2)} 0;
+  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
+  align-items: center;
+  display: flex;
```
kdembler:
```suggestion
  display: inline-flex;
```

Only the area with text should be clickable

kdembler: Once you make that change, the stories will be slightly broken so may need to put them in some container or something

kdembler (commented):
Looks good

kdembler (approved):

Issue #272, opened 2021-01-28
Add video file metadata parsing logic


kdembler:
#### Description

As part of creating a video, some metadata about the file needs to be provided. That includes things like size in pixels, size in bytes, encoding, etc. To allow uploading videos through Atlas, we will need to get that info in the app

#### Requirements

- Some logic to extract video file metadata out of a selected file
  - Should extract all the necessary fields
  - Should be as lightweight as possible. One road we can take is something like https://github.com/ffmpegwasm/ffmpeg.wasm - this is a giant piece of software though with much much more functionality than we require. If possible, we should use something that does only the probing part
- Some example added to the playground

PR #349, opened 2021-02-25
GetVideoMetadata utility fn

eldiegod:
Issue #272


src/hooks/useVideoMetadata.ts
```diff
@@ -0,0 +1,42 @@
...
+      // Video should now be loaded but we can add a second check
+      if (videoEl.readyState >= 3) {
+        const height = videoEl.videoHeight
+        const width = videoEl.videoWidth
+        const duration = videoEl.duration
+        setMetadata({ type, sizeInBytes, height, width, duration })
+      }
+    }
+
+    if (canPlay) {
```
kdembler: `canPlay` means whether the browser supports the given file? If so, maybe we could add some warning/error if it's `false`?

src/hooks/useVideoMetadata.ts
```diff
@@ -0,0 +1,42 @@
...
+  duration: number
+}
+export const useVideoMetadata = () => {
+  const [file, setFile] = useState<File>()
+  const [metadata, setMetadata] = useState<VideoMetaData>()
+  useEffect(() => {
+    const videoEl = document.createElement('video')
+    const type = file?.type ?? ''
+    const sizeInBytes = file?.size ?? 0
+    const URL = window.URL || window.webkitURL
```
kdembler: I think we can just safely use `URL` without any fallbacks. It's at 96% support https://caniuse.com/?search=url and we're using other much more cutting edge stuff anyway

kdembler: To clarify, it should be just a global and doesn't need defining

src/hooks/useVideoMetadata.ts
```diff
@@ -0,0 +1,42 @@
...
+  type: string
+  sizeInBytes: number
+  height: number
+  width: number
+  duration: number
+}
+export const useVideoMetadata = () => {
+  const [file, setFile] = useState<File>()
+  const [metadata, setMetadata] = useState<VideoMetaData>()
+  useEffect(() => {
```
kdembler: I'm thinking that maybe we could actually make this synchronous? Right now you need to pass the file to the hook and possibly wait a render to get the data back. This will probably make some flows harder, for example when we want to validate the file (minimum resolution) in the upload flow. Maybe this doesn't have to be a hook but just an async function?

src/views/PlaygroundView.tsx
```diff
@@ -39,3 +40,26 @@ const LinksContainer = styled.div`
 `
 
 export default PlaygroundView
+
+const VideoMetadata = () => {
```
kdembler: Maybe let's put this in a separate file like the `PlaygroundValidationForm` so it doesn't get messy?

src/hooks/useVideoMetadata.ts
```diff
@@ -0,0 +1,42 @@
+import { useEffect, useState } from 'react'
+
+export type VideoMetaData = {
+  type: string
```
kdembler: Maybe let's call it `mimeType` for clarity

src/views/PlaygroundView.tsx
```diff
@@ -39,3 +40,26 @@ const LinksContainer = styled.div`
...
+  const { setFile, metadata } = useVideoMetadata()
+  console.log({ metadata })
+  return (
+    <div>
+      <input
+        type="file"
+        accept="video/*"
```
kdembler:
```suggestion
        accept="video/*,.mkv"
```

It seems `.mkv` files don't have a MIME type? Something along those lines

kdembler (commented):
Smooth and elegant, I really like we went the native way

src/hooks/useVideoMetadata.ts
eldiegod: The problem is we need to wait for the listener loadeddata as the video doesn't get processed inmediatly

src/hooks/useVideoMetadata.ts
eldiegod: added some

eldiegod:
No longer a hook now an async fn for easier file validation 
@kdembler

src/utils/video.ts
```diff
@@ -10,3 +10,41 @@ export const formatVideoViewsAndDate = (
...
+export type VideoMetadata = {
+  mimeType: string
+  sizeInBytes: number
+  height: number
+  width: number
+  duration: number
+}
+export const getVideoMetadata = async (
+  file?: File
+): Promise<{ metadata: VideoMetadata; error?: undefined } | { metadata?: undefined; error: Error }> => {
```
kdembler: I don't think we need error as a return value. Promises provide a native `reject` functionality and that can carry the error by itself

src/utils/video.ts
```diff
@@ -10,3 +10,41 @@ export const formatVideoViewsAndDate = (
...
+
+export type VideoMetadata = {
+  mimeType: string
+  sizeInBytes: number
+  height: number
+  width: number
+  duration: number
+}
+export const getVideoMetadata = async (
+  file?: File
```
kdembler: Any reason to allow an `undefined` file here?

src/utils/video.ts
```diff
@@ -10,3 +10,41 @@ export const formatVideoViewsAndDate = (
...
+): Promise<{ metadata: VideoMetadata; error?: undefined } | { metadata?: undefined; error: Error }> => {
+  const videoEl = document.createElement('video')
+  const mimeType = file?.type ?? ''
+  const sizeInBytes = file?.size ?? 0
+  const Url = URL
```
kdembler: Let's remove this line and just use `URL` directly

src/utils/video.ts
```diff
@@ -10,3 +10,41 @@ export const formatVideoViewsAndDate = (
...
+  const canPlay = videoEl.canPlayType(mimeType)
+
+  return new Promise((resolve) => {
+    const handleLoadeddata = (e: Event) => {
+      // Video should now be loaded but we can add a second check
+      if (videoEl.readyState >= 3) {
```
kdembler, line 30: If it's in some other state, let's reject the promise, because I guess otherwise it will be pending forever?

src/utils/video.ts
```diff
@@ -10,3 +10,41 @@ export const formatVideoViewsAndDate = (
...
+        const duration = videoEl.duration
+        resolve({ metadata: { mimeType, sizeInBytes, height, width, duration } })
+        videoEl.removeEventListener('loadeddata', handleLoadeddata)
+      }
+    }
+
+    if (canPlay) {
+      const fileURL = Url.createObjectURL(file)
+      videoEl.src = fileURL
+      videoEl.addEventListener('loadeddata', handleLoadeddata)
```
kdembler: FYI, you could use `{once: true}` to not have to remove the listener later manually, but this works as well.

src/utils/video.ts
```diff
@@ -10,3 +10,41 @@ export const formatVideoViewsAndDate = (
...
+  file?: File
+): Promise<{ metadata: VideoMetadata; error?: undefined } | { metadata?: undefined; error: Error }> => {
+  const videoEl = document.createElement('video')
+  const mimeType = file?.type ?? ''
+  const sizeInBytes = file?.size ?? 0
+  const Url = URL
+  const canPlay = videoEl.canPlayType(mimeType)
+
+  return new Promise((resolve) => {
+    const handleLoadeddata = (e: Event) => {
```
kdembler: Let's do proper camel case with `handleLoadedData`

kdembler (changes requested):

src/views/PlaygroundView.tsx
kdembler: @eldiegod this wasn't resolved

src/utils/video.ts
eldiegod: I like this better

src/utils/video.ts
eldiegod: Yes, the file we get from the input is of type `File | undefined` so this way it's easier to type/use

src/utils/video.ts
eldiegod: true, it was mostly so we could destructure `const { metadata, error } = await ...` for easier consuming API, what you think?

eldiegod:
storybook failed to deploy for some reason

src/utils/video.ts
kdembler: But that's handling the input. The controller should check input's value and based on that either call this function or not. It doesn't make sense to call this function with `undefined`

src/utils/video.ts
```diff
@@ -10,3 +10,39 @@ export const formatVideoViewsAndDate = (
...
 }
+
+export type VideoMetadata = {
+  mimeType: string
+  sizeInBytes: number
+  height: number
+  width: number
+  duration: number
+}
+export const getVideoMetadata = async (file?: File): Promise<{ metadata: VideoMetadata }> => {
```
kdembler: I don't think there's a point in wrapping the data in additional object, let's just return `Promise<VideoMetadata>`

src/utils/video.ts
eldiegod: True, that makes sense

kdembler (approved):
Will need a conflict resolved

eldiegod:
Solved @kdembler !

kdembler (approved):

Issue #339, opened 2021-02-22
Add radio button component to Storybook


kdembler:
We currently have a `RadioButton` component but it was created some time ago. We should:
- Add it to storybook
- Make sure it looks like in the design
- Change it to use the `label` element
- Make it usable with a group of options. Something like `RadioButtonGroup` that would manage changing values of options inside?

PR #350, opened 2021-02-26
Radio button

mikkio-j:
#339

### DIFF
--- a/src/shared/components/RadioButton/RadioButton.style.ts
+++ b/src/shared/components/RadioButton/RadioButton.style.ts
@@ -69,10 +69,10 @@ export const StyledInput = styled.div<RadioButtonStyleProps>`
   justify-content: center;
   align-items: center;
   background-clip: content-box;
-  padding: 4px;
+  padding: 2px;
   box-sizing: border-box;
-  width: 24px;
-  height: 24px;
+  width: 18px;
+  height: 18px;
   &::before {
     content: '';
     top: -8px;
--- a/src/shared/components/RadioButton/RadioButton.style.ts
+++ b/src/shared/components/RadioButton/RadioButton.style.ts
@@ -6,13 +6,13 @@ import { colors, sizes, typography, transitions } from '../../theme'
 export type RadioButtonStyleProps = Partial<{
   error: boolean
   disabled: boolean
-  clickable: boolean
   checked: boolean
 }>
 
 export const Input = styled.input`
   margin: auto;
   opacity: 0;
+  cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
 `
 
 const colorFromProps = ({ error, checked, disabled }: RadioButtonStyleProps) => {
@@ -93,7 +93,7 @@ export const Label = styled.label<RadioButtonStyleProps>`
   display: inline-flex;
   align-items: center;
   margin-bottom: ${sizes(4)};
-  cursor: ${(props) => (props.disabled ? 'not-allowed' : props.clickable ? 'pointer' : 'auto')};
+  cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
   & > span:nth-of-type(1) {
     margin: 8px;
   }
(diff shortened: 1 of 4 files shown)


eldiegod:
Was looking at the storybook, not 100% sure but doesn't it have to match the [figma](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=489%3A1407) design?

src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -2,30 +2,31 @@ import React from 'react'
...
+  selected: string | number
   label: string
+  onClick: (e: React.MouseEvent<HTMLElement>) => void
 }> &
   Omit<RadioButtonStyleProps, 'clickable'> &
-  React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>
+  React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLInputElement>, HTMLInputElement> &
+  Omit<React.InputHTMLAttributes<HTMLInputElement>, 'onClick'>
 
 const RadioButton: React.FC<RadioButtonProps> = ({
```
drillprop: I think we should also pass ref here.

drillprop (commented):
Found few things which I think we could improve here.
- Font size is definitely too big, should be 14px like in a checkbox. 
- Also I think the label should be using only one line
- Not sure if we need all these positions: `top, bottom, right, left`. We omit them in `Checkbox
- Don't like how the disabled button work right now - you should add `not-allowed` cursor. Also, it should be prioritized. 
- We could add some CSS transitions on hover and on click

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -67,6 +70,9 @@ export const StyledInput = styled.div<RadioButtonStyleProps>`
   align-items: center;
   background-clip: content-box;
   padding: 4px;
+  box-sizing: border-box;
+  width: 24px;
+  height: 24px;
```
eldiegod:
```suggestion
  width: 18px;
  height: 18px;
```

when selecting the component on figma it looks like its 24px but its misleading as it seems to have some padding in there they actually are 18px

src/views/Playgrounds/PlaygroundValidationForm.tsx
```diff
@@ -117,6 +119,38 @@ const PlaygroundValidationForm = () => {
...
+                <RadioButton
+                  value="mature"
+                  label="Mature"
+                  onChange={(e) => {
+                    clearErrors('radioGroup')
+                    setValue('radioGroup', e.currentTarget.value)
+                  }}
+                  selected={props.value}
+                  error={!!errors.radioGroup}
+                />
```
eldiegod: Checking on the playground, not sure if this might be an issue with the radio buttons themselves or more with the useForm hook but if I click on both options a couple times it stops changing the values

eldiegod (changes requested):
Component and functionality look good to me just some small things

eldiegod (changes requested):
Ah, looking great, really like how it looks, just a couple last minor things I noticed:

1) Don't know why might be but on the playground neither the label nor the radio button get the cursor as pointer
2) On storybook the label has cursor pointer as it should but the center of the input doesn't have it

eldiegod (approved):
Working great 🏅

src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -1,35 +1,29 @@
 import React from 'react'
-import { Label, Input, StyledInput, RadioButtonStyleProps } from './RadioButton.style'
+import { Label, Input, StyledInput, StyledLabelText, RadioButtonStyleProps } from './RadioButton.style'
 
 type RadioButtonProps = Partial<{
+  selected: string | number
```
kdembler: Maybe we could call this something like `selectedValue`? This may suggest it's a boolean, like `isSelected`

kdembler (approved):

Issue #335, opened 2021-02-20
Implement the channel create/edit view layout


kdembler:
### Blocked by #264, #249 and #328

Extracted from #281. Before doing the full integration, let's do a layout + form validation first, without the web3 stuff. This means implementing both "New channel" and "Settings view" which are almost identical.

- Settings view (channel edit) - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A27218:
  - we skip the `Verification` and `Member` sections
- New channel view - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A8273:
  - Note there won't be inline cover edit as seen [here](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A15032). Instead we will do that with the crop dialog
- Both should be responsive, as seen in the Settings Figma page

Issue #269, opened 2021-01-28
Schema related changes


kdembler:
#### Description

As part of publishing, some new information will need to be provided by the query node. 

As we change the schema to allow channel avatars, etc. to be hosted by the storage providers, we will need to add support for that everywhere in Atlas.

#### Requirements

- Our schema updated based on https://gist.github.com/ondratra/f0b261381027578156cc1f12912aef5d
  - Should only include changes we will need, i.e. don't copy the entire schema
  - Both videos and channels should have only one category, like in our current schema. This is an error in the linked schema file
  - In general, we should make sure everything works with the new schema, some changes may be required (i.e. to the video player since `Asset` (old `JoystreamMediaLocation`) definition changed
  - The mocking backend will need to be updated for everything to work properly (feel free to ping me on that)
- All images that can be hosted by storage nodes (channel avatar, channelCover, video thumbnail) properly displayed in Atlas
  - It may make sense to create some common component for displaying an image coming from a storage node

PR #354, opened 2021-03-03
Create/Edit Channel layout

mikkio-j:
#335


src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -1,16 +1,22 @@
...
+
 export const TextAreaWrapper = styled.div`
   width: 100%;
 `
 
-export const StyledTextArea = styled.textarea`
+export const StyledTextArea = styled.textarea<TextAreaProps>`
   background: transparent;
   width: 100%;
+  height: 200px;
```
drillprop: What if we want to use a higher text-area? I know that is not the scope of this issue but maybe we can use the text-area `rows` attribute here

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -1,16 +1,22 @@
 import { colors, sizes, transitions, typography } from '@/shared/theme'
 import styled from '@emotion/styled'
 
+type TextAreaProps = {
+  lineHeight?: number
```
drillprop: Do we really need that? I think we could set some default constant value for the line-height

src/views/PlaygroundView.tsx
```diff
@@ -1,25 +1,25 @@
...
-      <Text variant="h2">Internal testing view</Text>
+    <>
+      {/* <Text variant="h2">Internal testing view</Text>
       <LinksContainer>
         <Link to="./validation-form">Validation Form</Link>
-        <Link to="./second">Second</Link>
+        <Link to="./channel-edit">Channel Edit</Link>
         <Link to="./third">Third</Link>
-      </LinksContainer>
+      </LinksContainer> */}
```
drillprop: Not sure why you disable this. Everything seems to be fine here, right?

src/views/Playgrounds/PlaygroundChannelCreate.tsx
```diff
@@ -0,0 +1,208 @@
...
+type Inputs = {
+  channelName?: string
+  description?: string
+  selectedPublicness?: string
+  selectedLanguage?: string
+  avatar: string | null
+  cover: string | null
+}
+
+const PlaygroundChannelEdit = () => {
```
drillprop:
```suggestion
const PlaygroundChannelCreate = () => {
```

src/views/Playgrounds/PlaygroundChannelCreate.tsx
```diff
@@ -0,0 +1,208 @@
...
+          name="cover"
+          control={control}
+          render={() => <ImageCropDialog imageType="cover" onConfirm={handleCoverConfirm} ref={coverDialogRef} />}
+        />
+      </form>
+    </ViewWrapper>
+  )
+}
+
+export default PlaygroundChannelEdit
```
drillprop:
```suggestion
export default PlaygroundChannelCreate
```

src/views/Playgrounds/PlaygroundChannelEdit.tsx
```diff
@@ -0,0 +1,236 @@
...
+          </StyledFormField>
+          <StyledActionBarTransaction
+            fee={1}
+            primaryButtonText="Publish Changes"
+            secondaryButtonText="Cancel"
+            onCancelClick={(e) => {
+              e.preventDefault()
+              reset()
+            }}
+          />
```
drillprop: Small thing. If you pick values inside select inputs and then click cancel - selected values will not update. I think it might be an issue with the select input itself.

drillprop (commented):
Again, besides the mirage thing, looks solid. Well done

eldiegod:
Gotta solve some conflicts over here

eldiegod (changes requested):
LGTM overall just a visual thing I noticed; could we add a transition to when we hover the video cover it feels a bit rough?

drillprop:
Also, it would be nice to hide tooltips if the user click inputs or the text-area

src/views/Playgrounds/PlaygroundChannelEditCreate.tsx
```diff
@@ -0,0 +1,238 @@
...
+                ref={register(textFieldValidation('Description', 3, 160))}
+                maxLength={160}
+                error={!!errors.description}
+                helperText={errors.description?.message}
+              />
+            </Tooltip>
+          </FormField>
+          <StyledFormField
+            title="Change Language"
+            description="Channel language is the main language the content you publish on your channel."
```
kdembler:
```suggestion
            description="Channel language is the main language of the content you publish on your channel"
```

src/views/Playgrounds/PlaygroundChannelEditCreate.tsx
```diff
@@ -0,0 +1,238 @@
...
+                  error={!!errors.selectedLanguage && !value}
+                  helperText={errors.selectedLanguage?.message}
+                />
+              )}
+            />
+          </StyledFormField>
+
+          <StyledFormField
+            title="Publicness"
+            description="Channel language is the main language the content you publish on your channel. We use it to provide users feed they look for. This text is FPO."
```
kdembler:
```suggestion
            description="Publicness dictates whether your channel will be displayed in users' feeds"
```

src/views/Playgrounds/PlaygroundChannelEditCreate.tsx
```diff
@@ -0,0 +1,238 @@
...
+                  }}
+                  error={!!errors.selectedPublicness && !value}
+                  helperText={errors.selectedPublicness?.message}
+                />
+              )}
+            />
+          </StyledFormField>
+          <StyledActionBarTransaction
+            fee={1}
+            primaryButtonText="Publish Changes"
```
kdembler: This should be `Create a Channel` when not editing

src/views/Playgrounds/PlaygroundChannelEditCreate.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+  const handleSubmit = useHandleSubmit((data) => {
+    console.log(data)
+  })
+  const handleAvatarConfirm = (blob: Blob, url: string) => {
+    setValue('avatar', url)
+    setAvatarImageUrl(url)
+  }
+  const handleCoverConfirm = (blob: Blob, url: string) => {
+    setValue('cover', url)
```
kdembler: I think we will need to save the blob as well, URL may not be enough to upload the file since we would need to read them again to get the data

src/views/Playgrounds/PlaygroundChannelEditCreate.tsx
```diff
@@ -0,0 +1,238 @@
...
+  StyledActionBarTransaction,
+  StyledAvatar,
+  InnerFormContainer,
+} from './PlaygroundChannelEditCreate.style'
+import { Header, SubTitle, SubTitlePlaceholder, TitlePlaceholder } from '../ChannelView/ChannelView.style'
+
+import { requiredValidation, textFieldValidation } from './formValidationOptions'
+import { formatNumberShort } from '@/utils/number'
+
+const languages: SelectedItem[] = [
```
kdembler: We will need to populate this with some more languages. Also, the key should probably be the ISO 639-1 identifier of the language as that's what's in the schema. I suggest to find some list of the most popular languages and put the first 20 or so here. We can revisit later

src/views/Playgrounds/PlaygroundChannelEditCreate.tsx
```diff
@@ -0,0 +1,238 @@
...
+  { name: 'English', value: 'english' },
+  { name: 'Polish', value: 'polish' },
+  { name: 'Spanish', value: 'spanish' },
+]
+
+const publicnessItems: SelectedItem[] = [
+  { name: 'Public (Anyone can see this video', value: 'public' },
```
kdembler:
```suggestion
  { name: 'Public (Channel will appear in feeds)', value: 'public' },
```

src/views/Playgrounds/PlaygroundChannelEditCreate.tsx
```diff
@@ -0,0 +1,238 @@
...
+  { name: 'Private', value: 'private' },
```
kdembler: Better value here would probably be `Unlisted` since the channel will be always public (it's in the blockchain after all). Also the value should be `true/false`. The corresponding field is `isPublic`

kdembler (changes requested):
Will take a look at the code tomorrow, for now just looking at the app. Some observations:
1) I think we should actually put this at a proper route already, to see how it looks as a separate view. We can wait till #359 is merged to get the basic `/studio` routing. My guess is the route will look like `/studio/channel/{id}` or `/studio/channel/new`. Based on that we will know whether it's edit or a new channel
2) Are you rebased on latest button changes? It seems the `Cancel` and `Publish` buttons have different heights
3) The `Publish` button doesn't have a `pointer` cursor
4) We should pre-populate value for both selects. In case of editing a channel we should obviously display the channel's values. When a new channel is created we can use defaults: `Public` and `English`
5) The bottom action bar should only be visible when:
    a) it's a new channel
    b) there were some changes made

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+                avatarDialogRef.current?.open()
+              }}
+              editable
+            />
+
+            <TitleContainer>
+              <Controller
+                name="channelName"
+                control={control}
+                rules={textFieldValidation('Channel name', 3, 20)}
```
kdembler: I think we will need to be a tad more liberal here, I suggest maybe like 40 characters max

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+        <InnerFormContainer>
+          <FormField title="Description">
+            <Tooltip text="Click to edit channel description" offsetY={-50}>
+              <StyledTextarea
+                name="description"
+                placeholder="Add description"
+                spellcheck={false}
+                rows={200}
+                ref={register(textFieldValidation('Description', 3, 160))}
```
kdembler, line 193: This will make it required, right? I think we can make the description optional

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+  StyledTitleSection,
+  TitleContainer,
+  StyledFormField,
+  StyledTextarea,
+  StyledActionBarTransaction,
+  StyledAvatar,
+  InnerFormContainer,
+} from './CreateEditChannelView.style'
+import { Header, SubTitle, SubTitlePlaceholder, TitlePlaceholder } from '@/views/consumer/ChannelView/ChannelView.style'
+import { requiredValidation, textFieldValidation } from '@/views/playground/Playgrounds/formValidationOptions'
```
kdembler: We certainly shouldn't rely on anything from playground outside of it. If it has something we need, let's elevate it higher

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+import { formatNumberShort } from '@/utils/number'
+
+const languages: SelectedItem[] = [
```
kdembler: This looks like a much better list! I have two suggestions though:
1. Let's sort it alphabetically so it's easier to find your desired language (we can keep English at the top as the default option)
2. Maybe let's elevate this somewhere above to `config/` or elsewhere so this file is easier to read (we shouldn't use `SelectedItem` type there though)

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+type CreateEditChannelViewProps = {
+  newChannel?: boolean
+}
+
+const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChannel }) => {
+  // TODO Add hook for fetching currently active channel
+  const id = newChannel ? null : '1494'
+
+  const { channel, loading, error } = useChannel(id || '')
```
kdembler: I don't think it makes much sense to do a call fetching channel with `{id: ''}` - this will always be `null`. We should skip this call when dealing with a new channel

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+      isPublic: isPublic[0],
+    },
+  })
+
+  useEffect(() => {
+    if (loading || newChannel) {
+      return
+    }
+    setValue('avatar', { blob: null, url: channel?.avatarPhotoUrl })
+    setValue('cover', { blob: null, url: channel?.coverPhotoUrl })
```
kdembler: This will get slightly more complicated once we do #269 - then we won't necessarily have a url for an asset. Or resolving that URL can be async, depending on what we do. I guess we can leave it for now, just FYI

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+    setValue('channelName', channel?.handle)
+    // TODO Add setValue for description, language and publicness after Channel fields update
```
kdembler: Is this blocked atm?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+  }, [loading, newChannel, setValue, channel?.avatarPhotoUrl, channel?.coverPhotoUrl, channel?.handle])
+
+  useEffect(() => {
+    setAvatarImageUrl(getValues('avatar')?.url)
+    setCoverImageUrl(getValues('cover')?.url)
+  }, [getValues, loading])
```
kdembler: Do we need a separate `useEffect` for this? Cannot those be set at the same time as `setValue` calls are made?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  const avatarDialogRef = useRef<ImageCropDialogImperativeHandle>(null)
+  const coverDialogRef = useRef<ImageCropDialogImperativeHandle>(null)
+
+  const [avatarImageUrl, setAvatarImageUrl] = useState<string | null>(null)
+  const [coverImageUrl, setCoverImageUrl] = useState<string | null>(null)
+  const [isActionBarActive, setActionBarActive] = useState(false)
```
kdembler, line 78: Let's move those above `useEffect`s, we use some state inside them

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+                  onChange={(e) => onChange(e.selectedItem)}
+                  error={!!errors.isPublic && !value}
+                  helperText={(errors.isPublic as FieldError)?.message}
+                />
+              )}
+            />
+          </StyledFormField>
+          <StyledActionBarTransaction
+            fee={FEE}
+            isActive={isDirty || isActionBarActive || newChannel}
```
kdembler, line 242: Would it be a lot of work to show this only if there are some changes to be submitted? For example, if I remove the last character of channel name and input it a gain, the form will be dirty even though there are no changes to actually publish

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+            primaryButtonText={newChannel ? 'Create a Channel' : 'Publish Changes'}
+            secondaryButtonText="Cancel"
+            onCancelClick={(e) => {
+              e.preventDefault()
+              // For reseting selects field, avatar and cover
+              window.location.reload()
```
kdembler: This was fine for the playground view, but the final thing shouldn't do this. The question then is, what should happen when we click that. Truth be told, I'm not sure, maybe let's just keep it empty for now and I ask designers

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  return (
+    <ViewWrapper>
+      <form onSubmit={handleSubmit}>
+        <Header>
+          <ChannelCover
+            coverPhotoUrl={coverImageUrl}
+            onCoverEditClick={(e) => {
+              e.preventDefault()
```
kdembler, line 144: Why is this needed?

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -63,6 +63,12 @@ export const StyledButtonsContainer = styled.div`
   flex-direction: row;
   align-items: center;
   justify-content: flex-end;
+  button:hover {
+    cursor: pointer;
```
kdembler: We're using a Button component for those buttons right? Shouldn't they have that applied already?

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -63,6 +63,12 @@ export const StyledButtonsContainer = styled.div`
...
+  }
+  button:nth-of-type(1) {
```
kdembler:
```suggestion
  > *:not(:last-child) {
```

This is a minor change but:
- by using `> *` we make sure the margin works for whatever is here, for example if we want to make the cancel button a link, this will work, `button` will not catch that
- with `nth-of-type(1)` we will add margin only to one of the buttons, if we add more of them, we'd need to update the styles. with `:not(:last-child)` we can target all items except the last one which is what I believe we want to do

Let me know if something here doesn't make sense

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -31,13 +31,16 @@ export const StyledInput = styled.input<StyledInputProps>`
...
+  max-width: calc(100vw - 150px);
+  width: ${({ widthSize }) => (widthSize ? `calc(${widthSize}ch + ${sizes(2.5)})` : '100%')};
   height: ${sizes(13)};
   &:hover {
     filter: brightness(80%);
   }
   @media screen and (min-width: ${breakpoints.small}) {
+    max-width: calc(100vw - 200px);
+  }
+  @media screen and (min-width: ${breakpoints.medium}) {
```
kdembler: Could you explain whats going on here? Seems there's a lot of arbitrary values - `150px`, `200px`, `sizes(2.5)`

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -11,6 +11,7 @@ export const StyledTextArea = styled.textarea`
   resize: none;
   color: ${colors.gray[300]};
   font-size: ${typography.sizes.subtitle2};
+  line-height: 2;
```
kdembler: Per design text here is `Body 1`, so we should use values from the theme that correspond to it
```suggestion
  font-size: ${typography.sizes.body1};
  line-height: ${typography.lineHeights.body1};
```

src/views/studio/CreateEditChannelView/CreateEditChannelView.style.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const TitleContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+`
+
+export const InnerFormContainer = styled.div`
+  width: 100%;
+  max-width: 760px;
```
kdembler: I'm a bit unsure about this and the `margin-left` below but I think those may be affected by the max view width we discussed in the chat so let's see how it goes

src/views/studio/CreateEditChannelView/CreateEditChannelView.style.tsx
```diff
@@ -0,0 +1,65 @@
...
+  margin-top: 50px;
+  padding-bottom: 200px;
```
kdembler: I think this may be a bit too much on mobile

kdembler (changes requested):
It's great to see all these coming together. Did a deeper dive this time and found some things we can improve, nothing major though!

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
mikkio-j: Hook form captures click and triggers submission

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
mikkio-j: `HeaderTextField` changes it's width depends on its input's length, so it was little bit tricky to style for all devices and different input's length. Changed that now to more cleaner option, I believe

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -17,6 +17,13 @@ export const StyledActionBarContainer = styled.div`
   }
 `
 
+export const StyledInnerContainer = styled.div`
+  display: flex;
+  width: 100%;
+  max-width: var(--max-inner-width);
```
eldiegod: This will have to be removed after merging #385 that handles that for all studio routes

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -17,6 +17,13 @@ export const StyledActionBarContainer = styled.div`
...
+  margin: 0 auto;
```
eldiegod: same as above

src/views/LayoutWithRouting.tsx
```diff
@@ -128,6 +128,7 @@ const RouterWrapper = () => {
 }
 
 const MainContainer = styled.main`
+  --max-inner-width: 1440px;
```
eldiegod: Same same as above #385

src/views/studio/CreateEditChannelView/CreateEditChannelView.style.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const TitleContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+`
+
+export const InnerFormContainer = styled.div`
+  width: 100%;
+  max-width: var(--max-inner-width);
```
eldiegod: Not sure if you would still need this one but this variable will still exists after #385 is merged

eldiegod (approved):
Looks good to me; added some comments on some things that we will need to be removed as we merge the `Layout for studio routes` PR.

As a side note, the channel cover img hovering effect turning darker could use a transition

src/api/queries/__generated__/channels.generated.tsx
```diff
@@ -105,6 +107,11 @@ export const BasicChannelFieldsFragmentDoc = gql`
     handle
     avatarPhotoUrl
     createdAt
+    description
```
kdembler: This breaks the consumer views in mocked version since we're not returning this info

kdembler (commented):
It seems if I have my cursor below the description text area, the `click anywhere to edit cover image` is still displayed. I think we will need to restrict the hover area for that since it produces not ideal experience:

https://user-images.githubusercontent.com/12646744/111434877-5de57280-8700-11eb-865b-a0fecdc07a1a.mov

src/shared/components/HeaderTextField/HeaderTextField.style.tsx
```diff
@@ -19,6 +19,13 @@ export const Container = styled.div`
 `
 
 export const StyledInput = styled.input<StyledInputProps>`
+  --input-max-width: 220px;
```
kdembler: I think we're trying to restrict width of this component too much. Take a look what happens across screen widths:

Screenshot 2021-03-17 at 15 45 25

Screenshot 2021-03-17 at 15 47 59

I feel this could be wider to offer better support to longer channel names. Look into this please

src/views/studio/CreateEditChannelView/CreateEditChannelView.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+  width: 100%;
+  margin-top: 50px;
+  padding-bottom: 100px;
+  @media screen and (min-width: ${breakpoints.medium}) {
+    padding-bottom: 200px;
+  }
+`
+
+export const StyledFormField = styled(FormField)`
+  margin-top: 40px;
```
kdembler: I don't think this should be `40px`. If you look at the design, `40px` is indeed spacing between end of one select to the header of the another but our components also have a reserved space for the helper text which will make this spacing actually bigger. I think the value used in the component is correct.

src/views/studio/CreateEditChannelView/CreateEditChannelView.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const InnerFormContainer = styled.div`
+  width: 100%;
+  margin-top: 50px;
+  padding-bottom: 100px;
+  @media screen and (min-width: ${breakpoints.medium}) {
+    padding-bottom: 200px;
+  }
+`
+
+export const StyledFormField = styled(FormField)`
```
kdembler: I believe these fields should be consistent across views so if some changes are needed to those, we should just put the styles into the component so each view doesn't have to style those separately

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -17,6 +18,11 @@ export const StyledActionBarContainer = styled.div`
   }
 `
 
+export const StyledInnerContainer = styled(StudioContainer)`
```
kdembler, line 21: Not related to this component specifically, but I think we could adjust the breakpoints here a bit:
Screenshot 2021-03-17 at 15 59 11

It seems there's enough space to display the `Every change...` next to the fee instead of below it

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,271 @@
...
+      cover: { blob: null, url: coverPhotoUrl },
+      channelName: handle,
+      description: description || '',
+      isPublic: isPublic ? isPublicSelect[0] : isPublicSelect[1],
+      selectedLanguage: foundLanguage || languages[0],
+    })
+  }, [channel, getValues, loading, newChannel, reset])
+
+  useEffect(() => {
+    setAvatarImageUrl(watch().avatar.url)
```
kdembler: Why do we need those? We already call those setters in `handleAvatarConfirm` and `handleCoverConfirm`

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -0,0 +1,271 @@
...
+    }
+    const { avatarPhotoUrl, coverPhotoUrl, handle, description, isPublic, language } = channel
+    const foundLanguage = languages.find(({ name }) => name === language?.name)
+    reset({
+      ...getValues(),
+      avatar: { blob: null, url: avatarPhotoUrl },
+      cover: { blob: null, url: coverPhotoUrl },
+      channelName: handle,
+      description: description || '',
+      isPublic: isPublic ? isPublicSelect[0] : isPublicSelect[1],
```
kdembler: I think this can be just `isPublic: isPublic`?

kdembler (commented):
Some final tweaks

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
kdembler: This hook will be called a lot of times

kdembler (approved):
Good job!

Issue #337, opened 2021-02-22
Create the file drop component


kdembler:
### Description

As seen in the [video publishing view](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4814%3A6921), we will need some way to accept a number of files from the user. The component would need to allow selecting the first file, do a transition as in the design and then allow selecting the second file.

### Requirements
- `MultiFileSelect` component added to `src/shared/components`
  - Should include both the file select area and the stepper indicating progress. Maybe it makes sense to actually do the file select as a general separate component and then include that and the stepper in `MultiFileSelect`. I suggest to start with the latter and possibly extract later.
  - Should include some way to access the files from the consumer's perspective.
  - Should allow both regular file select (via browser file select prompt) and file drag-and-drop on the area. I suggest using some existing library for that as it can be a lot of work to handle all the interactions ourselves. When choosing the library, please take into the account how well maintained it is and how customizable it is.
  - Should allow specifying a description and allowed file type for each file (step)
- Component added to the Storybook

PR #355, opened 2021-03-03
File drop

drillprop:
#337


drillprop:
Converted to a draft, need to fix styles for smaller devices and some other minor issues. Not ready for review yet.

src/shared/components/MultiFileSelect/MultiFileSelect.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <Story />
+      </OverlayManagerProvider>
+    ),
+  ],
+} as Meta
+
+const Template: Story<MultiFileSelectProps> = (args) => {
+  const [files, setFiles] = useState<FileState>({
+    video: null,
+    image: null,
```
mikkio-j, line 23: After confirming thumbnail and then clicking and confirming again to change and choose new thumbnail - it doesn't change. Have to click for the third time to apply thumbnail change

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,152 @@
...
+type ProgressBarProps = {
+  progress?: number
+}
+
+export const DragAndDropArea = styled.div<DragAndDropAreaProps>`
+  position: relative;
+
+  background-color: ${darken(0.16, colors.gray[600])};
+  cursor: pointer;
+  width: 320px;
```
mikkio-j: I believe percentage value would be better for small devices. Or does it have to be fixed?

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+`
+export const UploadButton = styled(Button)`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  svg {
+    margin-right: ${sizes(2)};
+    position: relative;
+    top: 0.125em;
```
mikkio-j: Hopefully, we won't have to do that after merging #346

src/shared/components/FileDrop/FileDrop.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const onDropAccepted: DropzoneOptions['onDropAccepted'] = useCallback(
+    (acceptedFiles) => {
+      const [file] = acceptedFiles
+      onUploadFile(file)
+    },
+    [onUploadFile]
+  )
+
+  const onDropRejected: DropzoneOptions['onDropRejected'] = useCallback(() => {
+    setError('Wrong file type!')
```
mikkio-j: Does it reject file based only on type? Or e.g. size as well? Maybe we should be able to customise error message

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,152 @@
...
+export const ButtonsGroup = styled.div`
+  display: none;
+  @media screen and (min-width: ${breakpoints.small}) {
+    margin-top: ${sizes(8)};
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+`
+export const UploadButton = styled(Button)`
```
mikkio-j: this one has default cursor on hover

mikkio-j (commented):
Really like the flow of it! Good job on that one! 👍

src/shared/components/MultiFileSelect/MultiFileSelect.stories.tsx
drillprop: My intention was to have the ability to readjust **already uploaded image** after a click on the thumbnail.  I think it's more natural. Right now it doesn't work well, because file select also appears. I'll fix it soon

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,149 @@
...
+  progress?: number
+}
+
+export const DragAndDropArea = styled.div<DragAndDropAreaProps>`
+  position: relative;
+
+  background-color: ${darken(0.16, colors.gray[600])};
+  cursor: pointer;
+  width: 100%;
+  height: calc(100vw * 0.5625);
```
kdembler: This operates on the assumption that there won't be anything else in the page layout but I'm not sure this will always be true. It can potentially lead to the aspect ratio being distorted. Couldn't we use the same approach here that we use for the video previews? (i.e. `padding-top: 56.25%`)?

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,149 @@
...
+export const DismissButton = styled.button`
+  cursor: pointer;
+  border: none;
+  background: none;
+  color: white;
+  margin-top: ${sizes(1)};
+  margin-left: 10px;
+`
+
+export const InnerContainer = styled.div`
```
kdembler: This has both `height: 100%` and `margin-top` and results in this: 
Screenshot 2021-03-09 at 12 39 29

This element overflows the container and actually makes the full area clickable slightly below it

kdembler: You can either account for the margin in height or use flexbox to position this

src/shared/components/FileDrop/FileDrop.tsx
```diff
@@ -0,0 +1,110 @@
...
+  ProgressBar,
+  StyledIcon,
+  Thumbnail,
+  Title,
+  UploadButton,
+} from './FileDrop.style'
+import { Step } from '../MultiFileSelect'
+
+export type FileDropProps = {
+  step: Step
```
kdembler: I don't think we should use `Step` here. This is a lower level of abstraction, `FileDrop` by itself shouldn't bother with steps. We also create strange dependency tree where `FileDrop` depends on `MultiFileSelect` (for `Step`) and `MultiFileSelect` depends on `FileDrop`.

We can just call this `fileType` or something which should be clearer to the consumer and we can define that type right here

src/shared/components/FileDrop/FileDrop.tsx
```diff
@@ -0,0 +1,110 @@
...
+      const [file] = acceptedFiles
+      onUploadFile(file)
+    },
+    [onUploadFile]
+  )
+
+  const { getRootProps, getInputProps, isDragAccept, isFileDialogActive } = useDropzone({
+    onDropAccepted,
+    onDropRejected,
+    accept: step + '/*',
```
kdembler: I suggest we pass this explicitly from the consumer, we may want to use something more sophisticated, for example `video/*,.mkv`

src/shared/components/FileDrop/FileDrop.tsx
```diff
@@ -0,0 +1,110 @@
...
+    maxSize,
+  })
+
+  const handleReAdjustThumbnail = (e: React.MouseEvent<HTMLImageElement, MouseEvent>) => {
+    e.stopPropagation()
+    onReAdjustThumbnail?.()
+  }
+
+  return (
+    <DragAndDropArea {...getRootProps()} isDragAccept={isDragAccept} isFileDialogActive={isFileDialogActive}>
```
kdembler: Could we make only the button clickable instead of the full area? I think it may kinda defeat the purpose of the button this way

kdembler: What do you think?

src/shared/components/FileStep/FileStep.style.ts
```diff
@@ -0,0 +1,114 @@
...
+export const FileName = styled(Text)`
+  display: block;
+  font-family: ${typography.fonts.headers};
+  font-size: ${typography.sizes.caption};
+  width: 180px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`
+
+export const DeleteButton = styled.button`
```
kdembler: We should use the `Button` component for that. After the recent merge this is one of the variants from the design. If you click on the `Button` node in Figma you can see the desired props for the component

src/shared/components/FileStep/FileStep.style.ts
```diff
@@ -0,0 +1,114 @@
...
+import styled from '@emotion/styled'
+import CircularProgressbar from '../CircularProgressbar'
+import Text from '../Text'
+
+type StepProps = {
+  active?: boolean
+  disabled?: boolean
+}
+
+export const StepWrapper = styled.div<StepProps>`
```
kdembler, line 12: It seems there are some issues with sizing here:

1) The height of the step changes once I pick the file. It should be fixed
Screenshot 2021-03-09 at 13 13 17
Screenshot 2021-03-09 at 13 13 04

2) Seems we're not matching the height from the design, let's try to do that

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export const DragAndDropArea = styled.div<DragAndDropAreaProps>`
+  position: relative;
+
+  background-color: ${darken(0.16, colors.gray[600])};
+  cursor: pointer;
+  width: 100%;
+  height: calc(100vw * 0.5625);
+  max-width: 640px;
+  max-height: 400px;
```
kdembler: To add to the point above, this isn't 16:9 ratio and I think it should be

src/shared/components/MultiFileSelect/MultiFileSelect.stories.tsx
kdembler: What's the status on that as I see it's not marked as resolved?

src/shared/components/FileStep/FileStep.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Step } from '../MultiFileSelect'
+
+export type FileStepProps = {
+  stepNumber: number
+  active: boolean
+  onDelete: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+  step: Step
+  fileName?: string
+  thumbnail?: string
+  onChangeStep?: (step: Step) => void
```
kdembler: I think this more of a `onSelect` right? The consumer ultimately decides if the step is changed

src/shared/components/MultiFileSelect/MultiFileSelect.tsx
```diff
@@ -0,0 +1,153 @@
...
+        setProgress(0)
+        setStep('image')
+      }
+    }, 5)
+
+    return () => clearTimeout(timeout)
+  }, [isLoading, progress])
+
+  const handleUploadFile = (file: File) => {
+    if (file.type.includes('video')) {
```
kdembler: We can use `step` instead to determine what to set, right?

src/shared/components/MultiFileSelect/MultiFileSelect.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const handleChangeStep = (step: Step) => {
+    setStep(step)
+  }
+
+  const handleDeleteFile = (fileType: 'video' | 'image') => {
+    onChangeFiles({ ...files, [fileType]: null })
+    setIsLoading(false)
+    setProgress(0)
+    if (fileType === 'image') {
+      onCropImage?.('')
```
kdembler: What does this empty string mean? I think it'll be more clear to use `null` in this case

src/shared/components/MultiFileSelect/MultiFileSelect.tsx
```diff
@@ -0,0 +1,153 @@
...
+import Icon from '../Icon'
+import { MultiFileSelectContainer, StepDivider, StepsContainer } from './MultiFileSelect.style'
+
+export type FileState = {
+  video: File | null
+  image: File | null
+}
+
+export type MultiFileSelectProps = {
+  onChangeFiles: React.Dispatch<React.SetStateAction<FileState>>
```
kdembler: Why not just `(fileState: FileState) => void`? Less verbose and actually more clear on what happens I think

kdembler (changes requested):
Damn, this looks nice! Will need some adjustments but overall really good job :)

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  background-color: ${darken(0.16, colors.gray[600])};
+  width: 100%;
+  height: 0;
+  padding-top: 90%;
+  display: flex;
+  justify-content: center;
+
+  transition: all ${transitions.timings.routing} ${transitions.easing};
+  border: 1px dashed
```
kdembler: According to the design the border should be inside the image, not around it:

Screenshot 2021-03-11 at 10 29 08

You could probably use some `::after` element to achieve that

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,161 @@
...
+  height: 100%;
+`
+
+export const DragAndDropArea = styled.div<DragAndDropAreaProps>`
+  position: relative;
+
+  background-color: ${darken(0.16, colors.gray[600])};
+  width: 100%;
+  height: 0;
+  padding-top: 90%;
```
kdembler: I think last time we talked about this we agreed to always keep the 16:9 ratio (even on mobile). What's the rationale to go for a more square-ish box on mobile?

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,161 @@
...
+  width: ${sizes(12)};
+  @media screen and (min-width: ${breakpoints.small}) {
+    width: ${sizes(18)};
+  }
+`
+
+export const Title = styled(Text)`
+  line-height: 1.2;
+  margin-top: ${sizes(2)};
+  font-size: ${typography.sizes.h5};
```
kdembler: Both `line-height` and `font-size` should be already set by the component right?

src/shared/components/FileDrop/FileDrop.style.ts
```diff
@@ -0,0 +1,161 @@
...
+  @media screen and (min-width: ${breakpoints.small}) {
+    margin-top: ${sizes(4)};
+  }
+`
+
+export const Paragraph = styled(Text)`
+  line-height: ${sizes(5)};
+  font-size: ${typography.sizes.subtitle2};
```
kdembler: Same here

src/shared/components/FileDrop/FileDrop.tsx
```diff
@@ -0,0 +1,120 @@
...
+} from './FileDrop.style'
+
+export type FileDropProps = {
+  fileType: string
+  accept: string
+  onUploadFile: (file: File) => void
+  icon: 'video-dnd' | 'image-dnd'
+  title: string
+  paragraph: string
+  thumbnail?: string | null
```
kdembler: Let's call this `thumbnailUrl`

src/shared/components/FileDrop/FileDrop.tsx
```diff
@@ -0,0 +1,120 @@
...
+  Paragraph,
+  ProgressBar,
+  StyledIcon,
+  Thumbnail,
+  Title,
+  UploadButton,
+} from './FileDrop.style'
+
+export type FileDropProps = {
+  fileType: string
```
kdembler: So I'm rethinking the approach to the supported types a bit. I see two options:
1) We can do what we do now, but maybe just with added typechecking to this prop
2) We could actually make `FileDrop` aware of supported types. Then based on the type we can assign `accept` and `icon` internally in the component.

What do you think?

src/shared/components/FileDrop/FileDrop.tsx
```diff
@@ -0,0 +1,120 @@
...
+                </ButtonsGroup>
+              </InnerContainer>
+            </CSSTransition>
+          </SwitchTransition>
+        )}
+        {error && (
+          <ErrorContainer onClick={(e) => e.stopPropagation()}>
+            <ErrorIcon name="error-second" />
+            <ErrorText variant="body2">{error}</ErrorText>
+            <DismissButton variant="tertiary" icon={'close'} onClick={() => onError?.('')}></DismissButton>
```
kdembler:
```suggestion
            <DismissButton variant="tertiary" icon={'close'} onClick={() => onError?.(null)}></DismissButton>
```

src/shared/components/FileStep/FileStep.tsx
```diff
@@ -0,0 +1,76 @@
...
+  Thumbnail,
+  StyledProgress,
+} from './FileStep.style'
+import { Step } from '../MultiFileSelect'
+
+export type FileStepProps = {
+  stepNumber: number
+  active: boolean
+  onDelete: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+  step: Step
```
kdembler: You removed `Step` from `FileDrop` but this has the same problem. We have bi-directional dependency here. We import `Step` from `MultiFileSelect` but in `MultiFileSelect` we import `FileStep`. In this case it makes much more sense to just define `Step` here instead

src/shared/components/MultiFileSelect/MultiFileSelect.style.ts
```diff
@@ -0,0 +1,40 @@
+import { breakpoints, colors, sizes } from '@/shared/theme'
+import styled from '@emotion/styled'
+
+export const MultiFileSelectContainer = styled.div`
+  width: 100%;
+  max-width: 640px;
+  @media screen and (min-width: ${breakpoints.small}) {
+    width: 640px;
+  }
```
kdembler: Why do we do this? Shouldn't the above styles already take care of this? Right now, if the screen is slightly above `600px`, the component will overflow the screen, we shouldn't have that

src/shared/components/MultiFileSelect/MultiFileSelect.tsx
```diff
@@ -0,0 +1,154 @@
...
+    if (files.image) {
+      dialogRef.current?.open(files.image)
+    }
+  }
+
+  const handleChangeStep = (step: Step) => {
+    setStep(step)
+  }
+
+  const handleDeleteFile = (fileType: 'video' | 'image') => {
```
kdembler: Let's also use `FileType` here

src/shared/components/MultiFileSelect/MultiFileSelect.tsx
```diff
@@ -0,0 +1,154 @@
...
+        onReAdjustThumbnail={handleReAdjustThumbnail}
+        progress={progress}
+        fileType={step}
+        icon={step === 'video' ? 'video-dnd' : 'image-dnd'}
+        title={step === 'video' ? 'Select Video File' : 'Add Thumbnail Image'}
+        thumbnail={croppedImageUrl}
+        paragraph={
+          step === 'video'
+            ? `16:9 Ratio preferred. 4K, 1440p, 1080p or 720p. This is example FPO data only.`
+            : `Accepting JPG, PNG formats. Can't exceed 20MB (Example data).`
```
kdembler: My comments weren't applied here

kdembler (changes requested):
The image doesn't seem to fill the area fully, can you look into why this happens? (~1px gap on top and bottom, zoom in):
Screenshot 2021-03-11 at 10 23 55

src/shared/components/FileStep/FileStep.tsx
kdembler: I had an additional comment here but it didn't submit somehow, trying to recreate:

What we call `Step` actually isn't a step in any way. It's just a type of the file this step is about. Notion of step exists only in context of the `MultiFileSelect`. Let's do the following:
- Let's define `type FileType = 'video' | 'channel` in `src/types`. This type will be useful for other parts of the app as well (like the upload manager)
- Let's define the following prop on both `FileStep` and `FileDrop`: `type: FileType`
- We can still use something called step in `MultiFileSelect` but it should be of `FileType` type probably

src/shared/components/FileDrop/FileDrop.style.ts
drillprop: There is not a lot of space to keep every element there. If I set padding-top back to 56,25 it will look something like this: 
image
Which I consider not ideal. I could lower even more margins, font-sizes and button padding, but it still looks kinda bad. 

I was thinking about three options here
1. I could remove button on mobile and allow mobile users to click on the droparea. I did it previously  
image
2. Keep the present version with a squarish box
3. Remove something else on mobile view, maybe subtitle or icon?

What do you think? For me, the best option is the first one

src/shared/components/FileDrop/FileDrop.tsx
drillprop: The last one seems better to me. I'll do that

src/shared/components/FileDrop/FileDrop.style.ts
kdembler: For 1 we'd need some logic based on the screen size though, right? Let's do the easiest thing for now (my guess is 2 since it's already done) and I'll bring this up with designers

src/shared/components/MultiFileSelect/MultiFileSelect.tsx
```diff
@@ -0,0 +1,151 @@
...
+        onUploadFile={handleUploadFile}
+        onReAdjustThumbnail={handleReAdjustThumbnail}
+        progress={progress}
+        fileType={step}
+        title={step === 'video' ? 'Select Video File' : 'Add Thumbnail Image'}
+        thumbnailUrl={croppedImageUrl}
+        paragraph={
+          step === 'video'
+            ? `16:9 Ratio preferred. 4K, 1440p, 1080p or 720p. This is example FPO data only.`
+            : `Accepting JPG, PNG formats. Can't exceed 20MB (Example data).`
```
kdembler: Let's try this:

```suggestion
            ? `Minimum 480p. Accepts any format supported by your browser.`
            : `Maximum 10MB. Accepts any format supported by your browser.`
```

src/shared/components/FileDrop/FileDrop.style.ts
kdembler: @drillprop I think this hasn't been fixed, the area is in focused stat even with the cursor below it. I actually think `height: 100%` may not be needed here

kdembler (approved):

PR #358, opened 2021-03-03
add fallback for failed video thumbnails #351

### LINKED ISSUE #351: Provide fallback for failed video thumbnails
As part of Sumer, we created a fallback gradient that will be displayed in the video preview in the case where user didn't provide a thumbnail image. It would be nice to be able to do this in cases when the image is no longer available and fails to loa d as well.

Couple of possible approaches to try out:
1. `onError` event on the `img`
2. HTML `picture`
3. HTML `object`
4. Something else?

0xVolodya:
closes #351 
Hi @kdembler, is this what was wanted?


kdembler (approved):
Hey @nadalfederer, thanks for this fix, seems to work perfectly!

PR #359, opened 2021-03-04
Publishing entrypoint

### LINKED ISSUE #276: Create publishing entrypoint
#### Description

Users will need some way to enter the publishing part of Atlas. Let's add a button in the main sidenav for that. Design here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3893%3A180

#### Requirements

- Button added to the sidenav that redirects users to `/studio`
  - Studio should be a separate module that's **hot-loaded**. We will probably require a lot of dependencies for studio that aren't needed anywhere else. So we should make sure there's proper codesplitting and we only load necessary bundles when Studio is loaded
  - Dismissible message about publishing tools (as in the design) added

drillprop:
Fix #276


src/components/SideNavbar/SideNavbar.style.tsx
```diff
@@ -56,13 +56,44 @@ export const SidebarNavList = styled.ul`
...
+export const ButtonGroup = styled.div`
+  width: ${EXPANDED_SIDENAVBAR_WIDTH}px;
+  margin-bottom: ${sizes(10)};
+  flex-grow: 1;
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+`
+
+export const ButtonLink = styled(Link)`
```
kdembler: I'm not sure this will fly. We probably should add support for our regular button to be a link. This way we don't need to style it here, we get all the hover/pressed states and we don't need to update it once the design changes

kdembler: Let's wait till the buttons PR is merged and work this out

src/components/SideNavbar/SideNavbar.tsx
```diff
@@ -33,6 +35,8 @@ type SidenavProps = {
 }
 
 const SideNavbar: React.FC<SidenavProps> = ({ items }) => {
+  const isStudio = useMatch('/studio')
```
kdembler: Let's use value from `routes` here and in the `ButtonLink`

src/views/LayoutWithRouting.tsx
```diff
@@ -1,10 +1,9 @@
-import React, { useEffect, useState } from 'react'
+import React, { Suspense, useEffect, useState } from 'react'
```
kdembler: Why do we import `Suspense` here?

src/views/LayoutWithRouting.tsx
```diff
@@ -33,13 +33,18 @@ const SIDENAVBAR_ITEMS: NavItemType[] = [
   },
 ]
 
+const StudioView = loadable(() => import('@/studio/StudioView'))
```
kdembler: What happens when the view takes some time to load? We won't see this now but if the bundle size grows, it can take non-instant amount of time to fetch the bundle. We probably need some kind of fallback for when that happens

kdembler (changes requested):

src/views/LayoutWithRouting.tsx
kdembler: @drillprop I see you've added a fallback both into `loadable` and `Suspense`. Do we need both though? Which of those will get displayed? Or will they both get screen time?

drillprop: Fixed, I did it by mistake. I was experimenting with both and forgot to remove Suspense. `fallback` from loadable should be enough.

kdembler: :+1:

kdembler (commented):
I think what also makes sense to do in this PR is to organise the views a bit. We'll adding a bunch of new ones for publishing, I think it makes sense to organise them in directories. For example `views/consumer` and `views/publisher`.

Maybe we could also split routing somehow? Define the consumer and publisher routes separately in some routing-only components? What do you think about that?

In general, we should try to lay down any groundwork for using the separate, studio, part of the app

src/components/SideNavbar/SideNavbar.style.tsx
kdembler: @drillprop merged now, you should be able to use the new buttons for this

src/shared/components/Button/Button.style.tsx
```diff
@@ -185,6 +187,32 @@ export const StyledButton = styled.button<ButtonStyleProps>`
...
+  &:hover {
+    cursor: ${(props) => (!props.disabled && props.clickable ? 'pointer' : '')};
+  }
+
+  ${colorsFromProps};
+  ${sizeFromProps};
+  ${disabled};
+`
+
+export const StyledLink = styled(LinkWithButtonProps)`
```
drillprop: Unfortunately, I'm still getting this: 
``` React does not recognize the `hasText` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `hastext` instead. If you accidentally passed it from a parent component, remove it from the DOM element.``` 
Trying to find a solution for this. Maybe you have a suggestion or idea on how to fix it.

Comment (drillprop) [reply src/shared/components/Button/Button.style.tsx]: Fixed with https://emotion.sh/docs/styled#customizing-prop-forwarding

File: src/shared/components/Button/Button.tsx [hunk 6]
```diff
@@ -23,21 +26,21 @@ const ButtonComponent: React.ForwardRefRenderFunction<HTMLButtonElement, ButtonP
...
   const clickable = !!onClick
   const hasText = !!children
-  const handleClick = (e: React.MouseEvent<HTMLButtonElement>) => {
-    if (!onClick) return
-    onClick(e)
-  }
 
   return (
     <StyledButton
+      as={as}
```
kdembler: Maybe we just don't pass `as` as a prop but instead just hardcode here `as={to ? Link : undefined}`? Then it's even easier to use the button for navigation

src/components/SideNavbar/SideNavbar.style.ts
```diff
@@ -56,13 +57,29 @@ export const SidebarNavList = styled.ul`
...
+  width: ${EXPANDED_SIDENAVBAR_WIDTH}px;
+  margin-bottom: ${sizes(10)};
+  flex-grow: 1;
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+`
+
+export const ButtonLink = styled(Button)`
+  padding: 14px ${NAVBAR_LEFT_PADDING}px;
```
kdembler: Doesn't the default button padding work here?

src/components/SideNavbar/SideNavbar.style.ts
```diff
@@ -56,13 +57,29 @@ export const SidebarNavList = styled.ul`
...
+  margin: 0 ${NAVBAR_LEFT_PADDING}px;
+  border: 1px solid ${colors.gray[400]};
+  font-family: ${typography.fonts.headers};
```
kdembler: Those should be already set by button comp as well?

drillprop: The border needs to stay, the `tertiary` variant has no border, on the other hand `secondary` variant has a different `active` state.
image
Also, I found that the `secondary` variant has a black background. Should we change it to transparent?

kdembler: That's a good point. So per the design this button is actually `secondary`. And seems `secondary` should actually be transparent. We can change that, but we'll need to make sure it doesn't break anything (I think only category picker can be affected by this)

drillprop: We also use the `secondary` variant for unfollow button. Setting it to transparent would be really bad here.
image
Should I just set the `secondary` button to transparent and update styles for unfollow button?

kdembler: Hmm yeah good catch. I suggest that we ignore this for now and just keep your custom `border` for the tertiary button. I'll bring this up with the designers

kdembler (approved):

Issue #282, opened 2021-01-28
Implement the my videos view


kdembler:
#### Description

Implement the my videos view, according to the design https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A2721

#### Requirements

- Matches the design and all its functionalities
  - To simplify, let's include all drafts first in the grid
  - No `Private` tab
- Scales on mobile devices properly

PR #360, opened 2021-03-04
My videos view

eldiegod:
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A2721
issues: #282 / #365


src/api/hooks/video.ts
```diff
@@ -25,8 +26,11 @@ export const useVideo = (id: string, opts?: VideoOpts) => {
...
 export const useVideos = (variables?: GetVideosQueryVariables, opts?: VideosOpts) => {
   const { data, ...rest } = useGetVideosQuery({ ...opts, variables })
+  // Only way to get the video count for pagination as of now
+  const { data: connectionData } = useGetVideosConnectionQuery({ ...opts, variables })
   return {
     videos: data?.videos,
+    totalCount: connectionData?.videosConnection.totalCount,
     ...rest,
   }
 }
```
eldiegod: I think this works to get the videos `totalCount` for now @kdembler

eldiegod: Alternatively we could even have a `GetVideosConnectionTotalCount` query that would be the same as `GetVideosConnection` but only fetch totalCount for this use case

kdembler: We have this currently
```graphql
query GetChannelVideoCount($where: VideoWhereInput) {
  videosConnection(first: 0, where: $where) {
    totalCount
  }
}
```

Basically the same as you're describing. We can probably rename that and reuse in this case

src/api/client/cache.ts
```diff
@@ -15,12 +15,38 @@ const cache = new InMemoryCache({
...
 
           // only for counting videos in HomeView
-          if (args?.where.channelId_in && !args?.first) {
+          if (args?.where?.channelId_in && !args?.first) {
             return `${createdAtGte}:${channelIdIn}`
           }
 
           return `${channelId}:${categoryId}:${channelIdIn}:${createdAtGte}`
         }),
+        videos: {
```
kdembler, line 28: Why aren't we using the helper from Apollo client?

src/App.tsx
```diff
@@ -14,7 +14,9 @@ export default function App() {
     <ApolloProvider client={apolloClient}>
       <PersonalDataProvider>
         <OverlayManagerProvider>
-          <LayoutWithRouting />
+          <DraftsProvider>
```
kdembler: This provider should probably be wrapping studio routes, we don't need that in the regular consumer experience. You will need to rebase off the latest changes

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -209,7 +209,7 @@ input VideoWhereInput {
   createdAt_gte: Date
   isCurated_eq: Boolean
   isPublic_eq: Boolean
-  id_in: [ID!]
+  id_in: [ID]
```
kdembler: Why this change?

src/components/VideoPreview.tsx
```diff
@@ -1,33 +1,49 @@
...
   id?: string
   progress?: number
+  isLoading?: boolean
   className?: string
-} & VideoPreviewBaseMetaProps
+} & VideoPreviewBaseMetaProps &
+  VideoPreviewPublisherProps
 
-const VideoPreview: React.FC<VideoPreviewProps> = ({ id, progress, className, ...metaProps }) => {
+const VideoPreview: React.FC<VideoPreviewProps> = ({ id, className, isLoading = false, ...metaProps }) => {
```
kdembler: To continue on the fact that we probably shouldn't use drafts outside of publisher screens - we probably should create a separate component for publishing video preview. Then we don't need to include all the additional logic on the consumer side

src/components/VideoPreview.tsx
```diff
@@ -1,33 +1,49 @@
...
+      videoHref={videoHref}
       channelHref={id ? routes.channel(video?.channel.id) : undefined}
-      progress={progress}
-      isLoading={isLoading}
+      isLoading={_isLoading}
       className={className}
-      // TODO: handle menu contextMenuCallbacks
-      // contextMenuCallbacks={{}}
+      onCopyVideoURLClick={
+        isDraft ? undefined : () => navigator.clipboard.writeText(videoHref ? location.origin + videoHref : '')
```
kdembler: Let's extract pasting into clipboard into a `utils/` method

src/components/VideoPreview.tsx
```diff
@@ -1,33 +1,49 @@
...
+  const [loadOnMount, setloadOnMount] = useState(true)
+  const { drafts } = useDrafts('video')
   const { video, loading } = useVideo(id ?? '', { fetchPolicy: 'cache-first', skip: !id })
-  const isLoading = loading || id === undefined
+
+  useEffect(() => {
+    setloadOnMount(false)
+  }, [])
+
+  const _isLoading = loading || id === undefined || isLoading || loadOnMount
```
kdembler: Let's try to be more descriptive about variable names. Without looking at their definitions, it's not possible to tell what's a difference between `isLoading` and `_isLoading`

kdembler: Unless necessary in the context, let's try to avoid direct comparison to `undefined`. We can do `!id` instead

src/components/VideoPreview.tsx
```diff
@@ -1,33 +1,49 @@
...
   progress?: number
+  isLoading?: boolean
   className?: string
-} & VideoPreviewBaseMetaProps
+} & VideoPreviewBaseMetaProps &
+  VideoPreviewPublisherProps
 
-const VideoPreview: React.FC<VideoPreviewProps> = ({ id, progress, className, ...metaProps }) => {
+const VideoPreview: React.FC<VideoPreviewProps> = ({ id, className, isLoading = false, ...metaProps }) => {
+  const [loadOnMount, setloadOnMount] = useState(true)
```
kdembler: Improper camelCase on the setter

src/config/misc.ts
```diff
@@ -1 +1 @@
-export const MOCKED_SERVER_LOAD_DELAY = 500
+export const MOCKED_SERVER_LOAD_DELAY = 300
```
kdembler: I don't think this is actually used anymore? Should be safe to delete

src/config/routes.ts
```diff
@@ -14,8 +13,10 @@ export default {
     return basePath
   },
   channel: (id = ':id') => `/channel/${id}`,
-  videos: () => '/videos',
   channels: () => '/channels',
+  video: (id = ':id') => `/video/${id}`,
+  videos: () => '/videos',
+  myVideos: () => '/my-videos',
```
kdembler: This needs to be under `/studio` router, will need rebase. Also probably `/studio/videos` should be enough

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -39,13 +40,18 @@ export const DraftsProvider: React.FC = ({ children }) => {
...
+      const currentDrafts = await getDrafts()
+      const videos = currentDrafts
+        .filter((draft): draft is VideoDraft => draft.type === 'video')
+        .filter((draft) => (channelId ? draft.channelId === channelId : true))
+      setDraftsState({
+        videos,
+      })
+    },
+    [setDraftsState]
+  )
```
kdembler: `draftsState` is global and can be reused between views (and they can specify different args)

⬆️ should you give you a hint why there's a problem here

src/shared/components/Glider/Glider.tsx
```diff
@@ -86,7 +86,7 @@ export function useGlider<T extends HTMLElement>({
 function useEventListener<K extends keyof GliderEventMap>(
   element: HTMLElement | undefined | null,
   event: K,
-  listener: (event: GliderEvent<GliderEventMap[K]>) => void = () => {}
+  listener: (event: GliderEvent<GliderEventMap[K]>) => void = () => ({})
```
kdembler: Why was this modified? I don't think Glider is used in any way here?

src/views/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,222 @@
...
+  TabsContainer,
+  ViewContainer,
+} from './MyVideos.styles'
+
+const testChannelId = '100' // staging test channel id
+const TABS = ['All Videos', 'Published', 'Drafts', 'Unlisted']
+const INITIAL_VIDEOS_PER_ROW = 4
+// not yet doable
+// TODO: adjust action bar to the real fee
+// TODO: Unlisted videos
```
kdembler: We can actually do that now, `Video` contains `isPublic` field, that's what tells us whether a video is unlisted

src/views/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,222 @@
...
+// TODO: on edit video callbacks
+// TODO: on delete video callbacks
+// TODO: dynamic channels (not hardcoded)
+// TODO: No videos screen (this was deleted from figma?)
+// doable
+// TODO: how is the action bar supossed to work?
+export const MyVideosView = () => {
+  const [videosPerRow, setVideosPerRow] = useState(INITIAL_VIDEOS_PER_ROW)
+  // Drafts calls can run into race conditions
```
kdembler: Why is that? Possibly fixed by my other comment for `useDrafts`. We cannot have race conditions

kdembler (changes requested):
Reviewed some parts so far. Unfortunately this will need some changes because of the design updates so will do a full review once those are ready.

#### What we will need to change:

The videos won't have checkboxes anymore. They will only have a small icon button on hover. This has a couple of implications:
- No need for "other videos selected" state
- No need for "selected state"
- No need for action bar in this screen - we should probably revert all changes related to it

Since those `VideoPreview` changes have a big impact on this screen I suggest we do them as part of this PR.

src/api/client/cache.ts
eldiegod: I'm not sure if they forgot it or what but on their function the read part is missing and coz of that it doesn't work properly, odd

src/api/schemas/extendedQueryNode.graphql
eldiegod: for consistency as `channelId_in: [ID]` above didn't have the `!`

src/shared/components/Glider/Glider.tsx
eldiegod: I noticed this line was producing an annoying lint warning for every PR as empty arrow functions are not allowed

src/api/schemas/extendedQueryNode.graphql
kdembler: They both should have it, check out https://hydra.joystream.org/graphql

src/api/client/cache.ts
kdembler: They didn't forget it :P It's the same as with cursor based pagination, the full set is returned always by design and you need to slice it yourself. I think we already discussed why this is

src/shared/components/Glider/Glider.tsx
kdembler: I see. This however changes the signature of the function and now it will always return a truthy `{}` value, and I'm not sure if this does or doesn't have any implications. I suggest to skip it here as it's not part of the scope. Post-release we should do a general cleanup of all the eslint warning and we can take care of it then

src/api/client/cache.ts
eldiegod: I don't think I quite get it yet, maybe you can explain the idea on a quick call

src/components/VideoPreview.tsx
```diff
@@ -1,36 +1,97 @@
...
+    VideoPreviewPublisherProps,
+  'publisherMode' | 'videoPublishState'
+>
+export const VideoPreviewPublisher: React.FC<VideoPreviewWPublisherProps> = ({
+  id,
+  className,
+  isLoading = false,
+  isDraft,
+  ...metaProps
+}) => {
```
eldiegod: @kdembler  Added the new component here but wasn't too sure about if it was the best way for the reusability of the logic

src/api/queries/videos.graphql
```diff
@@ -37,6 +37,7 @@ fragment VideoFields on Video {
   duration
   thumbnailUrl
   createdAt
+  isPublic
```
kdembler, line 40: This isn't returned from the mocked data which breaks the mocked version

eldiegod: Added it but I think there might be a bug on the mocked filtering by isPublic, would be good to have some unlisted videos on staging to test on as well not sure if possible

src/components/InfiniteGrids/InfiniteVideoGrid.tsx
```diff
@@ -109,11 +109,12 @@ const InfiniteVideoGrid: React.FC<InfiniteVideoGridProps> = ({
     createdAtGte,
   ])
 
+  const serializedVariables = JSON.stringify(queryVariables)
   const placeholderItems = Array.from({ length: placeholdersCount }, () => ({ id: undefined }))
   const gridContent = (
     <>
       {[...displayedItems, ...placeholderItems]?.map((video, idx) => (
-        <VideoPreview id={video.id} key={idx} showChannel={showChannel} />
+        <VideoPreview id={video.id} key={idx + serializedVariables} showChannel={showChannel} />
```
kdembler: What's the purpose of this?

src/components/VideoPreview.tsx
```diff
@@ -1,36 +1,95 @@
...
 }
 
 export default VideoPreview
+
+export type VideoPreviewWPublisherProps = Omit<
+  {
+    id?: string
+    progress?: number
+    isLoading?: boolean
+    className?: string
```
kdembler: I think those are already included in `VideoPreviewBaseProps`

src/components/VideoPreview.tsx
```diff
@@ -1,36 +1,95 @@
...
+  const { video, loading } = useVideo(id ?? '', { fetchPolicy: 'cache-first', skip: !id })
+  useEffect(() => {
+    setLoadOnMount(false)
+  }, [])
+
+  // a bit complex loading state but couldn't come up with a better name
+  const internalIsLoadingState = loading || !id || isLoading || loadOnMount
+  const videoHref = id ? routes.video(id) : undefined
+
+  return { video, internalIsLoadingState, videoHref }
```
kdembler: I think I'm a bit lost on what's happening here. Why do we need anything other than `const { video, loading } = useVideo(id ?? '', { fetchPolicy: 'cache-first', skip: !id })`?

src/shared/components/VideoPreviewBase/PullUp.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+type PullUpProps = StateProps & {
+  onClick: (e: React.MouseEvent<HTMLElement>) => void
+}
+
+export const PullUp: React.FC<PullUpProps> = ({ disabled, onClick, ...props }) => (
+  <Tooltip above text={disabled ? 'Video already added to edit list' : 'Open in new tab'}>
+    <Container disabled={disabled} onClick={onClick}>
+      <BringUpIcon name="bring-up" {...props} />
+    </Container>
```
kdembler: We should use the `Button` component here, it should also contain all the necessary styles for things like hover and disabled

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -212,18 +197,27 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
...
                     )}
                     {!!duration && <CoverDurationOverlay>{formatDurationShort(duration)}</CoverDurationOverlay>}
                     <CoverHoverOverlay onClick={handleCoverHoverOverlayClick}>
-                      {publisherMode && checkboxNode}
-                      {!isAnyVideoSelected && hoverIconNode}
+                      {publisherMode && (
+                        <CoverTopLeftContainer>
+                          <PullUp
+                            // set to true when video is already on the snackbar
+                            disabled={false}
```
kdembler: We won't access the video tabs here directly so this needs to come into this component as a prop. We can hardcode that prop as `disabled` inside my videos for now

src/views/studio/MyVideosView/AddVideoView.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <div>
+          <Button icon="video-camera">Upload video</Button>
+        </div>
+      </InnerContainerView>
+      <WellIllustration />
+    </ContainerView>
+  )
+}
+
+const ContainerView = styled.div`
```
kdembler: This needs some responsive styling to handle mobile screens

src/views/studio/MyVideosView/AddVideoView.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react'
+import styled from '@emotion/styled'
+import bgIllustration, { ReactComponent as WellIllustration } from '@/assets/well-blue.svg'
+import { Button, Text } from '@/shared/components'
+import { sizes, colors } from '@/shared/theme'
+import { css } from '@emotion/react'
+
+// for when there is absolutely no videos available
+export const AddVideoView = () => {
```
kdembler: This should be typed with `React.FC`

src/views/studio/MyVideosView/AddVideoView.tsx
```diff
@@ -0,0 +1,92 @@
...
+const MessageView = styled.div`
+  display: flex;
+  flex-direction: column;
+  text-align: left;
+  margin-top: 48px;
+  margin-bottom: ${sizes(8)};
+`
+
+// for tabs
+export const AddVideo = () => {
```
kdembler: This won't really work for empty tabs since adding a video doesn't really guarantee that `unlisted` or `drafts` will fill up. For empty tabs maybe we should just show something like `Nothing to show`?

src/views/studio/MyVideosView/AddVideoView.tsx
kdembler: Also, maybe let's call this `EmptyVideosView` or `NoVideosView` to not mistake this with soon to be created add videos action sheet

src/views/studio/MyVideosView/MyVideos.styles.ts
```diff
@@ -0,0 +1,26 @@
+import styled from '@emotion/styled'
+import { colors, sizes } from '@/shared/theme'
+import { DismissibleMessage, Text } from '@/shared/components'
+
+export const ViewContainer = styled.div`
+  padding-top: ${sizes(8)};
+`
+export const StyledText = styled(Text)``
```
kdembler: I don't think we need this

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+const testChannelId = '100' // staging test channel id
+const TABS = ['All Videos', 'Published', 'Drafts', 'Unlisted'] as const
+const INITIAL_VIDEOS_PER_ROW = 4
+// not yet doable
+// TODO: on edit video callbacks
+// TODO: on delete video callbacks
+// TODO: dynamic channels (not hardcoded)
+export const MyVideosView = () => {
+  const [videosPerRow, setVideosPerRow] = useState(INITIAL_VIDEOS_PER_ROW)
+  const [hasAnyVideos, sethasAnyVideos] = useState<boolean>()
```
kdembler: improper camel case for setter

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+  // Drafts calls can run into race conditions
+  const { drafts, removeDraft, removeAllDrafts, addDraft } = useDrafts('video', testChannelId)
+  const [currentTab, setCurrentTab] = useState(0)
+  const { currentPage, setCurrentPage } = usePagination(currentTab)
+  const videosPerPage = 4 * videosPerRow
```
kdembler: Let's elevate this 4 into some constant so it's clear what this represents

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+    {
+      limit: videosPerPage,
+      offset: videosPerPage * (currentPage - 1),
+      where: {
+        channelId_eq: testChannelId,
+        isPublic_eq,
+      },
+    },
+    {
+      // notifyOnNetworkStatusChange: true,
```
kdembler: Should this be here?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+          {((isDraftTab && drafts.length === 0) || (!isDraftTab && totalCount === 0 && !isLoading)) && <AddVideo />}
+          <PaginationContainer>
+            <Pagination
+              onChangePage={(page) => {
+                setCurrentPage(page)
+                currentTabName !== 'Drafts' && FetchMoreVideos(page)
+              }}
+              onMouseEnterPage={(page) => {
+                currentTabName !== 'Drafts' && FetchMoreVideos(page)
+              }}
```
kdembler: I see you tried prefetching here but I'm not sure if we should do that right now. I agree it's a good user experience and something we should have but it would need to be done properly and I don't think it will be this simple. To keep this PR contained, I suggest we keep it out.

https://user-images.githubusercontent.com/12646744/111625231-e8a39b80-87ec-11eb-939e-e7b6fa6c734b.mov

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+                    key={idx + '-' + currentTabName + '-' + currentPage}
+                    id={draft.id}
+                    showChannel={false}
+                    isDraft
+                    onEditVideoClick={() => ({})}
+                    onDeleteVideoClick={() => {
+                      removeDraft(draft.id)
+                    }}
+                  />
+                ))}
```
kdembler: Let's try to extract it into some variable like `gridContent`, it makes reading the layout too hard

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+        offset: videosPerPage * (page - 1),
+        limit: videosPerPage,
+      },
+    })
+  // console.log({ videos, totalCount, hasAnyVideos, isLoading, isPublic_eq })
+  return (
+    <ViewContainer>
+      <StyledText variant="h2">My Videos</StyledText>
+
+      {hasAnyVideos === false ? (
```
kdembler: Do we need that as a state? Wouldn't something like this work here?
```suggestion
      {totalCount === 0 ? (
```

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const FetchMoreVideos = (page: number) =>
+    fetchMore({
+      variables: {
+        // substract 1 coz offset index starts at 0
+        offset: videosPerPage * (page - 1),
+        limit: videosPerPage,
+      },
+    })
+  // console.log({ videos, totalCount, hasAnyVideos, isLoading, isPublic_eq })
```
kdembler: For removal

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+      }
+    }
+    createDrafts()
+    return () => {
+      removeAllDrafts()
+    }
+  }, [])
+
+  const isDraftTab = currentTabName === 'Drafts'
+  const isLoading = loading || (videos?.length === 0 && (totalCount ?? 0) > 0)
```
kdembler: When would that second condition kick in?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const placeholderItems = Array.from({ length: isLoading ? videosPerPage : 0 }, () => ({
+    id: undefined,
+    progress: undefined,
+  }))
+  const videosWPlaceholders = [...(videos || []), ...placeholderItems]
```
kdembler: Let's use the full `With` for sake of clarity

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+      {hasAnyVideos === false ? (
+        <AddVideoView />
+      ) : (
+        <>
+          <TabsContainer>
+            <Tabs initialIndex={0} tabs={[...TABS]} onSelectTab={setCurrentTab} />
+          </TabsContainer>
+          {isDraftTab && (
+            // Should this really be dismissable?
```
kdembler: What do you mean?

src/views/studio/MyVideosView/MyVideosView.tsx
kdembler: I guess we would also need a condition that we're in the first tab

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+              description={
+                'This mean you can only access one on the device you used to create it. Clearing your browser history will delete all your drafts.'
+              }
+            />
+          )}
+          <Grid maxColumns={null} onResize={handleOnResizeGrid}>
+            {!isDraftTab &&
+              videosWPlaceholders
+                // this makes for a smoother transition between pages
+                .slice(0, videosPerPage)
```
kdembler: Won't this be done by the hook already?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+                    key={idx + '-' + currentTabName + '-' + currentPage}
+                    id={video.id}
+                    showChannel={false}
+                    isLoading={loading}
+                  />
+                ))}
+            {isDraftTab &&
+              drafts
+                // pagination slice
+                .slice(videosPerPage * (currentPage - 1), (currentPage - 1) * videosPerPage + videosPerPage)
```
kdembler: Maybe we should just keep the page 0-indexed and only add 1 to it when passing to the paginating component? Seems there's a lot of places where we subtract 1

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const isDraftTab = currentTabName === 'Drafts'
+  const isLoading = loading || (videos?.length === 0 && (totalCount ?? 0) > 0)
+  const placeholderItems = Array.from({ length: isLoading ? videosPerPage : 0 }, () => ({
+    id: undefined,
+    progress: undefined,
+  }))
+  const videosWPlaceholders = [...(videos || []), ...placeholderItems]
+  const handleOnResizeGrid = (sizes: number[]) => setVideosPerRow(sizes.length)
+
+  const FetchMoreVideos = (page: number) =>
```
kdembler: Function name shouldn't be upper case

src/views/studio/MyVideosView/MyVideosView.tsx
kdembler: (take a look at the request count in bottom right)

kdembler (changes requested):

src/components/InfiniteGrids/InfiniteVideoGrid.tsx
eldiegod: This fixes issue #365 and it allows the videoPreviews to show its animations and new data properly when the queryVariables change; i.e. category change

src/components/VideoPreview.tsx
eldiegod: Trueee

src/components/VideoPreview.tsx
eldiegod: Yes the loadOnMount part is just so we allow for the component to always show the animations when it is mounted as sometimes we already have the data and we don't need it to load but it just looks nice i.e issue #365 

The `internalIsLoadingState` similarly takes into account all the possible scenarios in which the video could be considered `loading` be it the query hasn't completed, it has just mounted so we would like to animate it like if it was loading, we didn't pass an ID or we passed the isLoading prop equal to true for whatever external reason

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: At the end we would have to do the same in the pagination component onChange / onMouseEnter i.e

```
          <PaginationContainer>
            <Pagination
              onChangePage={(page) => {
                setCurrentPage(page - 1)
                currentTabName !== 'Drafts' && fetchMoreVideos(page - 1)
              }}
              onMouseEnterPage={(page) => {
                currentTabName !== 'Drafts' && fetchMoreVideos(page - 1)
              }}
              page={currentPage + 1}
              itemsPerPage={videosPerPage}
              totalCount={isDraftTab ? drafts.length : totalCount}
            ></Pagination>
          </PaginationContainer>
```

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: I was thinking that it could be positives to the warning being always there as a reminder of the draft being only on local storage, but just a thought

src/views/studio/MyVideosView/MyVideosView.tsx
kdembler: Hmm I think the `Pagination` component should actually be 0-indexed as well. I think we had discussion about this, right @drillprop ?

src/views/studio/MyVideosView/MyVideosView.tsx
kdembler: I think this is what the designers envisioned, let's keep it like this for now

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: Yes I tried solving this but it would need a memorized throttle fn I think, yes def tricky

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: yeah was my thought as well, even tho slightly confusing at first could be the best solution

src/components/InfiniteGrids/InfiniteVideoGrid.tsx
kdembler: I feel like this is fixing the symptom instead of the problem. This also results in a different behaviour. What #365 describes is that on `master`, the video preview could transition between two videos without a problem. Right now it cannot, so by this change in the video grid we're basically forcing it to go into a loading state in between. I think the problem here is that currently the video preview cannot transition between two videos. If it could, like it does on `master`, my guess is that we also wouldn't need the `loadOnMount` since the previews could do a transition without load in between. What are your thoughts?

eldiegod: I don't think it's a huge deal, what this does is actually makes it behave a little more like it was before as the `key` on master equals the `video.id` and we went for `idx` to make it looks better with the placeholders and adding the queries is more an in-between solution.

To the why the videoPreview cant switch from one video to the other without loadOnMount it's because we replaced the css animations with CSSSwitchTransitions which also have a key and they only trigger when we first pass a `loading=true` I haven't seen a way around that yet other than to make it load on the first time it's rendered and then set is to false on the useEffect like the loadOnMount thing

src/views/studio/MyVideosView/AddVideoView.tsx
eldiegod: Added some, it needs tweak but its no use until we have the final illustration

src/views/studio/MyVideosView/AddVideoView.tsx
eldiegod: Screenshot 2021-03-18 at 18 54 44

What about this? I think it's useful to have the Upload video button in there as it might be something the user is looking for anyways, and it also doesn't imply adding a video will fill in the tab, thoughts?

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: yeah went for a similar condition

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: When calling `fetchMore()` on the pagination it sets videos to `[]` while it gets the new data

src/shared/components/VideoPreviewBase/PullUp.tsx
eldiegod: ~~Isn't the button component too different for this?~~ nvm it worked

src/views/studio/MyVideosView/MyVideosView.tsx
drillprop: @eldiegod if I'm not wrong you can delete both `-1` from slice and everything should work the same. 
Tried by myself. Let me know if I'm missed something

drillprop (approved):

src/views/studio/MyVideosView/MyVideosView.tsx
drillprop: @eldiegod if I'm not wrong you can delete both -1 from slice and everything should work the same.
Tried by myself. Let me know if I'm missed something

eldiegod:
@kdembler I think all we talked about should be addressed in the last commits, as a note: 
-kept the hook for shared logic as I think it's better than repeating the logic around.
-found a fix for the issue you mentioned with pagination, but I'm not 100% sure it's a good one, as always it's tricky stuff; check `cache.ts` and `video.ts` for that, maybe you have a better idea lmk.

src/api/hooks/video.ts
```diff
@@ -25,8 +26,21 @@ export const useVideo = (id: string, opts?: VideoOpts) => {
...
+  // Only way to get the video count for pagination as of now
+  const { data: connectionData } = useGetVideoCountQuery({ ...opts, variables })
   return {
     videos: data?.videos,
+    totalCount: connectionData?.videosConnection.totalCount,
+    ...rest,
+  }
+}
+
+export const useVideosOffsetLimitPagination = (variables?: GetVideosQueryVariables, opts?: VideosOpts) => {
```
kdembler: Maybe we call this `useVideosPage` so it's clear that the hook will actually do the slicing for you?

src/shared/components/VideoPreviewBase/PullUp.tsx
```diff
@@ -0,0 +1,45 @@
...
+  &:hover {
+    ${({ disabled }) => !disabled && `background: ${transparentize(0.8, colors.blue[100])};`}
+  }
+  &:active {
+    ${({ disabled }) => !disabled && `background: ${transparentize(0.8, colors.blue[300])};`}
+    path {
+      ${({ disabled }) => !disabled && `stroke: ${colors.gray[100]};`}
+    }
+  }
+`
```
kdembler: Not sure if I made that comment and it was lost or not, but do we actually need those styles? Shouldn't they be defined in the button already?

src/views/studio/MyVideosView/EmptyVideosView.tsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react'
+import styled from '@emotion/styled'
+import bgIllustration, { ReactComponent as WellIllustration } from '@/assets/well-blue.svg'
+import { Button, Text } from '@/shared/components'
+import { sizes, colors, breakpoints } from '@/shared/theme'
+import { css } from '@emotion/react'
+
+// for when there is absolutely no videos available
+export const EmptyVideosView: React.FC = () => {
```
kdembler, line 9: I may still have some comments to make here or for the empty tab views but let's get till we get the illustration and then we can finalise this thing

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,180 @@
...
+            />
+          ))}
+    </>
+  )
+
+  return (
+    <StudioContainer>
+      <ViewContainer>
+        <Text variant="h2">My Videos</Text>
+        {hasAnyVideo ? (
```
kdembler: I've made a comment previously about this. I don't think we need any state for this. We should be fine with just a computed condition. I think the following would work?
```ts
!totalCount && !isLoading && !drafts.length && currentTabName === 'All Videos'
```

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,180 @@
...
+                isPullupDisabled={false}
+                onEditVideoClick={() => ({})}
+                onDeleteVideoClick={() => {
+                  removeDraft(draft.id)
+                }}
+              />
+            ))
+        : videosWithPlaceholders.map((video, idx) => (
+            <VideoPreviewPublisher
+              key={idx + '-' + currentTabName}
```
kdembler: Do we need to include `currentTabName`? If we skip it, previews would be able to transition between tabs, right? I think that's a good thing, wdyt?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const gridContent = (
+    <>
+      {isDraftTab
+        ? drafts
+            // pagination slice
+            .slice(videosPerPage * currentPage, currentPage * videosPerPage + videosPerPage)
+            .map((draft, idx) => (
+              <VideoPreviewPublisher
+                key={idx + '-' + currentTabName + '-' + currentPage}
```
kdembler: Same as the other comment about the `currentTabName` (although maybe it's different because those are drafts).

Also, should we include the `currentPage` in the key? If we do then the components won't get reused right?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,180 @@
...
+                  removeDraft(draft.id)
+                }}
+              />
+            ))
+        : videosWithPlaceholders.map((video, idx) => (
+            <VideoPreviewPublisher
+              key={idx + '-' + currentTabName}
+              id={video.id}
+              showChannel={false}
+              isLoading={loading}
```
kdembler: Do we need that prop? I think the `id` for empty items will be undefined, so the component will enter the loading state anyway? I think this prop may not be needed at all

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,180 @@
...
+            {((isDraftTab && drafts.length === 0) || (!isDraftTab && totalCount === 0 && !isLoading)) && (
+              <EmptyVideos />
+            )}
+            <PaginationContainer>
+              <Pagination
+                onChangePage={handleChangePage}
+                page={currentPage}
+                itemsPerPage={videosPerPage}
+                totalCount={isDraftTab ? drafts.length : totalCount}
+              ></Pagination>
```
```suggestion
              />
```

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: No, this was in there for the previous solution not needed anymore

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: same as above, I remove these

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: yeah I actually added it but forgot to remove the old one for some reason 🤔

src/shared/components/VideoPreviewBase/PullUp.tsx
eldiegod: Updated them with the new transparent color variables but yup, we very needed them, otherwise, it doesn't look like in figma

eldiegod:
Waiting for illustration from designer

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,166 @@
...
+    id: undefined,
+    progress: undefined,
+  }))
+  const videosWithPlaceholders = [...(videos || []), ...placeholderItems]
+  const handleOnResizeGrid = (sizes: number[]) => setVideosPerRow(sizes.length)
+  const hasNoVideos = currentTabName === 'All Videos' && totalCount === 0 && drafts.length === 0
+
+  const handleChangePage = (page: number) => {
+    setCurrentPage(page)
+    currentTabName !== 'Drafts' &&
```
```suggestion
    !isDraftTab &&
```

src/components/VideoPreview.tsx
```diff
@@ -1,36 +1,78 @@
...
+import { copyToClipboard } from '@/utils/broswer'
 
-type VideoPreviewProps = {
+export type VideoPreviewProps = {
   id?: string
-  progress?: number
-  className?: string
-} & VideoPreviewBaseMetaProps
+} & VideoPreviewBaseMetaProps &
+  Pick<VideoPreviewBaseProps, 'progress' | 'isLoading' | 'className'>
```
kdembler: I don't think we need the `isLoading` prop on those two components anymore, right? We can probably also remove it from the `useVideoSharedLogic`

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -0,0 +1,166 @@
...
+export const MyVideosView = () => {
+  const [videosPerRow, setVideosPerRow] = useState(INITIAL_VIDEOS_PER_ROW)
+  // Drafts calls can run into race conditions
+  const { drafts, removeDraft, removeAllDrafts, addDraft } = useDrafts('video', testChannelId)
+  const [currentTab, setCurrentTab] = useState(0)
+  const { currentPage, setCurrentPage } = usePagination(currentTab)
+  const videosPerPage = ROWS_AMOUNT * videosPerRow
+  const currentTabName = TABS[currentTab]
+  const isPublic_eq = getPublicness(currentTabName)
+  const { loading, videos, totalCount, fetchMore } = useVideosPage({
```
kdembler: We should also get `error` here and throw if it's defined

src/components/VideoPreview.tsx
```diff
@@ -1,36 +1,78 @@
...
+      {...metaProps}
+    />
+  )
+}
+
+const useVideoSharedLogic = (id?: string, isLoading?: boolean) => {
+  const { video, loading } = useVideo(id ?? '', { skip: !id })
+  const internalIsLoadingState = loading || !id || isLoading
+  const videoHref = id ? routes.video(id) : undefined
+  return { video, internalIsLoadingState, videoHref }
```
kdembler: Maybe let's just return `internalIsLoadingState` as loading from here? Name can be a bit misleading

eldiegod:
@kdembler merged the PR and tweaked the read fn a bit, looks good

kdembler (approved):
90 commits, 90 comments PR, that's something. Great job 🥇

PR #361, opened 2021-03-04
fix: video is not clickable after introducing "add publisher mode to VideoPreview"

0xVolodya:
I think it's supposed to be not clickable only on the publisher mode.
Peek 2021-03-05 01-41

Peek 2021-03-05 01-42


src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -158,7 +158,7 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
     onChannelClick(e)
   }
   const createAnchorClickHandler = (href?: string) => (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
-    if (!href) {
+    if (!href || publisherMode) {
```
eldiegod:
```suggestion
    if (!href || isAnyVideoSelected) {
```

The anchor should still work when in publisher mode, tho we want to prevent it from navigating when any videos are being selected since it enters a "selecting videos" mode, and clicking a video cover means we want to select/deselect it

eldiegod (changes requested):
Hey great, thanks for fixing the issue, it's just what needed to be done + a tiny suggestion 🏅

kdembler:
There also seems to be some conflict introduced by my latest merge so it will need to get resolved :pray:

0xVolodya:
fixed

kdembler (approved):

kdembler:
Great, thanks for you contribution @nadalfederer!

Issue #267, opened 2021-01-28
Add ability to save videos being uploaded


kdembler:
#### Description

As part of the publishing flows, users will upload videos which we need to remember. Video upload can begin only after the metadata has been saved in the blockchain. Since upload can take some time, there's a possibility of error/user closing the app. We need to be able to persist videos that have pending uploads

#### Requirements

- Some logic to add/modify/delete upload metadata
  - We can assume this lives in local storage
  - Should be separate from the `usePersonalData`
  - Should save the video ID, assigned storage provider and hashes of all the files
- Example added to the playground

PR #362, opened 2021-03-05
Save uploading files metadata

mikkio-j:
#267


src/components/Dialogs/ImageCropDialog/ImageCropDialog.stories.tsx
```diff
@@ -30,15 +31,15 @@ const RegularTemplate: Story<ImageCropDialogProps> = () => {
...
     setAvatarImageUrl(url)
   }
 
-  const handleThumbnailConfirm = (blob: Blob, url: string) => {
+  const handleThumbnailConfirm = (blob: Blob, url: string, cropData: CropData) => {
     setThumbnailImageUrl(url)
   }
 
-  const handleCoverConfirm = (blob: Blob, url: string) => {
+  const handleCoverConfirm = (blob: Blob, url: string, cropData: CropData) => {
```
drillprop: I guess, these `cropData`  will be needed later, right?

drillprop (commented):
Didn't find much here, so good work. Don't know what you should do with the mirage. Besides that, looks good to me!

src/components/Dialogs/ImageCropDialog/ImageCropDialog.stories.tsx
mikkio-j: Yup, it's gonna take its part in hashing logic for image-type assets

src/mocking/server/index.ts
```diff
@@ -29,55 +29,55 @@ import {
 } from '@/config/urls'
 import { MOCKED_SERVER_LOAD_DELAY } from '@/config/misc'
```
eldiegod: Why is everything commented out here?

eldiegod (approved):
Didn't find any issue; looks solid to me
also, the playground works as expected

src/mocking/server/index.ts
mikkio-j: Sorry, forgot to add the comment for that. This is the reason why this PR is in draft currently. Mirage's trying to capture blob request from `CropImageDialog` (used in playground), causing an error - so to use cropping freely inside the app I had to disable Mirage - it's one of the issues we've discussed today.

eldiegod: Yeah, figure it out it was the mirage issue, then alright besides that looks good to me

src/components/Dialogs/ImageCropDialog/cropper.ts
```diff
@@ -44,6 +44,16 @@ const CANVAS_OPTS_PER_TYPE: Record<CropperImageType, Cropper.GetCroppedCanvasOpt
   },
 }
 
+export type CropData = {
```
kdembler: I see this is basically the same type as the one in Cropper. Couldn't we use `Cropper.Data` instead? Then we don't have to redefine it here. What would make sense is reexporting it as our own so that other parts of the app don't depend on the Cropper directly. So for example `export type CropData = Cropper.Data`. I also suggest to elevate this type for example to `src/types` since it will be used by the logic as well

src/hooks/useUploadingFilesData/useUploadingFilesData.tsx
```diff
@@ -0,0 +1,111 @@
...
+  updateUploadingFileStatus as updateUploadingFileStatusFn,
+  addUploadingFileData as addUploadingFileDataFn,
+  removeUploadingFileData as removeUploadingFileDataFn,
+  clearUploadingFilesData as clearUploadingFilesDataFn,
+} from './utils'
+import { CropData } from '@/components'
+
+type UploadingFileType = 'video' | 'thumbnail' | 'cover' | 'avatar'
+export type StatusType = 'completed' | 'notCompleted'
+export type UploadingFile = {
```
kdembler, line 15: We will also need some way to reference the parent object - channel or video

src/views/Playgrounds/PlaygroundUploadingFilesData.tsx
```diff
@@ -0,0 +1,107 @@
...
+    addUploadingFileData,
+    updateUploadingFileStatus,
+    removeUploadingFileData,
+    clearUploadingFilesData,
+  } = useUploadingFilesData()
+  const [error, seterror] = useState<Error>()
+  const [avatarImageUrl, setAvatarImageUrl] = useState<string | null>(null)
+  const avatarDialogRef = useRef<ImageCropDialogImperativeHandle>(null)
+
+  const handleAvatarConfirm = (blob: Blob, url: string, cropData: CropData) => {
```
kdembler: Do you think we could also demonstrate restoring cropped image from the crop data here? So once we have the crop data saved, user should provide the same file and we should get the same cropped image without any additional user interaction. This will probably require some additional functionality in the cropper logic but should be useful later on

kdembler (commented):
Looking good overall, just some minor changes and one suggestion. This will probably require some updates once we actually start interacting with uploaded assets but works for now :+1:

src/hooks/useUploadingFilesData/useUploadingFilesData.tsx
```diff
@@ -0,0 +1,117 @@
...
+  updateUploadingFileStatus as updateUploadingFileStatusFn,
+  addUploadingFileData as addUploadingFileDataFn,
+  removeUploadingFileData as removeUploadingFileDataFn,
+  clearUploadingFilesData as clearUploadingFilesDataFn,
+} from './utils'
+import { CropData } from '@/components'
+
+type ParentObjectType = 'video' | 'channel'
+type UploadingFileType = 'video' | 'thumbnail' | 'cover' | 'avatar'
+export type StatusType = 'completed' | 'notCompleted'
```
kdembler: Let's make this `completed` | `inProgress` | `error`

src/hooks/useUploadingFilesData/useUploadingFilesData.tsx
```diff
@@ -0,0 +1,117 @@
...
+export type UploadingFile = {
```
kdembler, line 15: We will need few more fields here:
- some info on when this was scheduled for upload, maybe `updatedAt`?
- `size`
- we'll probably also want some metadata, for example `1920x1080`. for now maybe let's just add `metadata?: string`

src/components/Dialogs/ImageCropDialog/cropper.ts
```diff
@@ -44,10 +46,13 @@ const CANVAS_OPTS_PER_TYPE: Record<CropperImageType, Cropper.GetCroppedCanvasOpt
...
 }
 
-export const useCropper = ({ imageEl, imageType }: UseCropperOpts) => {
+export type CropBoxData = Cropper.CropBoxData
+
+export const useCropper = ({ imageEl, imageType, fileName }: UseCropperOpts) => {
   const [cropper, setCropper] = useState<Cropper | null>(null)
   const [currentZoom, setCurrentZoom] = useState(0)
   const [zoomRange, setZoomRange] = useState<[number, number]>([0, 1])
+  const { uploadingFilesData } = useUploadingFilesData()
```
kdembler: We cannot really rely on this here. A couple of reasons:
- We create a weird link - `ImageCropDialog` now relies on `useUploadingFilesData` and will actually throw an error if used outside of `UploadingFilesDataProvider`. This doesn't make much sense since one doesn't have anything to do with the other. We need to separate those concerns 
- Consider what's happening in the `useEffect` now. It has `uploadingFilesData` as a dependency. So whenever the data about files being uploaded changes, the hook will re-run, creating a new cropper instance. For example, let's say I started uploading a video. It's processing in the background and now I want to upload another one. So I pick a thumbnail and start cropping it. If during my cropping, one of the files from the previous videos finishes uploading, my cropping flow will be interrupted and reset. That's not a great user experience

Actually, maybe let's try figuring this restoration flow later on. It will not be that simple, the image for cropping would need to be displayed somewhere, we will probably need a custom hook for that

kdembler (changes requested):

src/components/Dialogs/ImageCropDialog/cropper.ts
```diff
@@ -44,6 +44,8 @@ const CANVAS_OPTS_PER_TYPE: Record<CropperImageType, Cropper.GetCroppedCanvasOpt
   },
 }
 
+export type CropBoxData = Cropper.CropBoxData
```
kdembler: Sorry for a late comment on this:

Maybe we actually elevate this to `src/types`? It will be used for some logic as well. Let's call it `ImageCropData`

kdembler: Wait did I make this comment already after all? I see it's in the `src/types` but also here 😅

src/views/playground/Playgrounds/PlaygroundUploadingFilesData.tsx
```diff
@@ -0,0 +1,109 @@
+import React, { useState, useRef } from 'react'
+import { useUploadingFilesData } from '@/hooks'
+import styled from '@emotion/styled'
+import ImageCropDialog, { ImageCropDialogImperativeHandle } from '@/components/Dialogs/ImageCropDialog/ImageCropDialog'
```
kdembler: I think we should actually export this directly from `@/components`. That's carried over but could you clean that up?

src/components/Dialogs/ImageCropDialog/cropper.ts
mikkio-j: Yup, forgot to clean that up after fiddling with restoring crop feature

kdembler (approved):

PR #364, opened 2021-03-06
fix infinite loop when double click on progress bar

### LINKED ISSUE #314: Video player enters a play/pause toggle infinite loop
When dragging the progress bar around a couple of times letting it go then repeating and then triple-clicking on the progress bar the video player enters a play/pause toggle infinite loop.

Notes:
Could only repro so far if the video was playing and not paused.
Making the player fullscreen also triggered the bug in some cases.
Looks like some specific videos are more prone to it

Video link reference:
https://www.loom.com/share/9bf8d30b01b54a95af38d6a874e19969

0xVolodya:
closes #314
These two properties doing nothing except for this bug as I figured out.
how to reproduce: double click on the progress bar
Peek 2021-03-06 23-30

### DIFF
--- a/src/views/VideoView/VideoView.tsx
+++ b/src/views/VideoView/VideoView.tsx
@@ -102,13 +102,6 @@ const VideoView: React.FC = () => {
     }
   }, [video?.id, handleTimeUpdate, updateWatchedVideos])
 
-  const handlePlay = useCallback(() => {
-    setPlaying(true)
-  }, [])
-  const handlePause = useCallback(() => {
-    setPlaying(false)
-  }, [])
-
   if (error) {
     throw error
   }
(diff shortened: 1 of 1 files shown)


kdembler (changes requested):
Thanks for that!

The fix for this may not be that simple though, we actually need those two handlers. The reason is that we're keeping the `playing` state inside this component. So this component will tell the `VideoPlayer` whether it should be playing or not. However, if you click on the play/pause button, our component doesn't know anything about that - we have those handlers here so that `playing` state can react to play/pause click. If we remove them, the `playing` will remain `true` even if the video was paused.

What's interesting, this isn't obvious at first. I discovered this by accident - if you pause the video it will automatically unpause itself after few seconds. That's a strange behaviour though so I had a look why that happens. In theory, the player should only respond to changes in `playing` and shouldn't do anything if it remains `true`. However, it looks like there's a bug in `VideoPlayer` component itself. We have a bunch of `useEffect` there but none of them have a dependency array. Because of that they will run on every render instead only when the value changes. I think this may be the real culprit here. Maybe try adding dependency arrays to those and see if this still persists?

kdembler:
Closing due to no response, fixed in https://github.com/Joystream/atlas/pull/392

PR #366, opened 2021-03-08
replace mirage.js with msw.js mocking

kdembler:
The mocking code contains a couple of `@ts-ignore` because it was too much work to handle all the typescript cases. In general, the type safety should be much higher than that of previous code anyway


.env
```diff
@@ -1,5 +1,6 @@
 # This file is commited. Do not store secrets here
 REACT_APP_ENV=staging
+#REACT_APP_ENV=development
```
drillprop, line 3: should we uncomment this? Or at least update scripts `"dev": "cross-env REACT_APP_ENV=development yarn start"`.

drillprop (commented):
Wow, that's a lot of code,  great work 🎉 Didn't find anything, in general looks good to me.

.env
drillprop: Also, I just noticed that we don't have `cross-env` anymore, so it should be installed for this

kdembler: I'm not sure tbh. Do you think we should use `development` env by default?

drillprop: Okay, we could just leave this as it is.

eldiegod: I like the `"dev": "cross-env REACT_APP_ENV=development yarn start"` bit, commenting uncommenting would be more annoying imo

eldiegod: also harder to figure out for new people looking at the repo

public/mockServiceWorker.js
```diff
@@ -0,0 +1,310 @@
...
+}
+
+function respondWithMock(clientMessage) {
+  return new Response(clientMessage.payload.body, {
+    ...clientMessage.payload,
+    headers: clientMessage.payload.headers,
+  })
+}
+
+function uuidv4() {
```
eldiegod, line 304: This could be pretty useful in the utils folder :)

src/mocking/accessors/pagination.ts
```diff
@@ -0,0 +1,73 @@
...
+  after?: string | null
+}
+
+type OffsetLimitPaginatedDataArgs = {
+  offset?: number
+  limit?: number
+}
+
+const indexToCursor = (idx: number): string => btoa(idx.toString())
+const cursorToIndex = (cursor: string): number => +atob(cursor)
```
eldiegod, line 27: Can u talk a bit about whats going on here? especially the `+atob(cursor)` part

src/mocking/data/mockVideos.ts
```diff
@@ -1,20 +1,42 @@
...
+    category: mockCategories[idx % mockCategories.length],
+    license: mockLicenses[idx % mockLicenses.length],
+  } as MockVideo
 })
 
+const coverMockVideo: MockVideo = {
+  ...rawCoverVideo.video,
+  createdAt: new Date(rawCoverVideo.video.createdAt),
+  channel: coverMockChannel,
+  // @ts-ignore __typename not matching, whatever
```
eldiegod, line 32: I know that feeling 😂

mikkio-j (commented):
Wow, great work on that one 🔥 

It seems that the bug with not clickable videos thumbnails is back though.

kdembler:
That's because the branch wasn't rebased, should be fine now

eldiegod (approved):
Looks good to me overall just a couple questions 👍

public/mockServiceWorker.js
kdembler: This is an autogenerated file so let's leave it as it is, if we need this functionality later we can copy it

src/mocking/accessors/pagination.ts
kdembler: Sure, `atob` is a function to decode a base64-encoded string. The `+` is an operator that allows you to convert string to number.
```js
const originalIndex = 3
const encodedIndexStr = btoa(originalIndex.toString()) // "Mw=="
const decodedIndexStr = atob(encodedIndexStr) // "3"
const index = +decodedIndexStr // 3
```

src/mocking/data/mockVideos.ts
kdembler: Haha yeah, it was just too much to handle everything, I think I already pushed my Typescript abilities to their limit anyway :P

src/mocking/accessors/pagination.ts
eldiegod: oh pretty cool

.env
kdembler: Let's leave it as it is for now

Issue #328, opened 2021-02-17
Create checkout component


kdembler:
#### Description

As seen in the new channel view, we will need a checkout component that can be used to indicate to user what info they need to provide before proceeding: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A10258

#### Requirements

- `Checkout` component added to `src/shared/components`
  - Should allow passing the list of steps
  - Should allow passing a callback that will be called whenever a step is clicked
  - Should allow passing info on the current progress on these steps (they don't have to be sequential)
- Component added to the storybook.

PR #375, opened 2021-03-11
Checkout component

eldiegod:
Issue: #328 
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3079%3A10258


src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react'
+import styled from '@emotion/styled'
+import { colors } from '@/shared/theme'
+import { Icon } from '..'
+
+export const Container = styled.div`
+  padding: 24px;
+  width: 312px;
+  background-color: #181c20;
```
mikkio-j:
```suggestion
  background-color: ${colors.gray[800]};
```

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const StepState = styled.div<StepStateProps>`
+  width: 24px;
+  height: 24px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border-radius: 100%;
+  ${({ completed }) => [
+    completed &&
+      `background-color: #4038FF;
```
mikkio-j:
```suggestion
      `background-color: ${colors.blue[500]};
```

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+      border: 2px solid transparent;`,
+    completed === false && `border: 2px solid #7B8A95`,
```
mikkio-j:
```suggestion
    completed === false && `border: 2px solid ${colors.gray[300]}`,
```

mikkio-j (changes requested):
Nice 👍 Really like the look of it. Just some minor color changes.

Also I noticed that there is a tertiary button with chevron icon for minimizing/maximizing checkout. I believe it wasn't there when you started working on this 🤷 , but it's a  cool feature

checkout

eldiegod:
> Nice 👍 Really like the look of it. Just some minor color changes.
> 
> Also I noticed that there is a tertiary button with chevron icon for minimazing/maximizing checkout. I believe it wasn't there when you started working on this 🤷 , but it's a cool feature
> 
> checkout

@mikkio-j totally forgot to replace the hardcoded colors thanks 🤓

also, I'm not sure about the minimizing thing @kdembler are we keeping it at the end, I believe u mentioned we were going to skip it?

kdembler:
In general I think it's not 100% clear where we'll go with this component but current design indicates the ability to minimize so if it's already there, let's keep it

eldiegod:
Internal decision was to skip the minimizing feature for now

mikkio-j (approved):
Looks good! 🎉

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+      `background-color: ${colors.blue[500]};
+      border: 2px solid transparent;`,
+    completed === false && `border: 2px solid ${colors.gray[300]}`,
+  ]};
+`
+
+export const Step = styled.div`
+  cursor: pointer;
+  display: flex;
+  gap: 12px;
```
drillprop: This line you can remove, not super sure, but I think it does nothing.

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+`
+
+export const ProgressbarContainer = styled.div`
+  width: 24px;
+  height: 24px;
+`
+
+export const StepsProgressContainer = styled.div`
+  display: flex;
+  gap: 12px;
```
drillprop: `gap` for flexbox isn't well supported, so maybe we could change it - https://caniuse.com/?search=gap 
ls it possible to replace it with some margins and maybe padding with no effort? What do you think?

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 32px;
+`
+
+export const StepsContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 14px;
```
drillprop: Same here

drillprop (commented):
Added just some small comments about `gap` property. Besides that, it looks really good. Great job!

src/shared/components/Checkout/Checkout.styles.tsx
drillprop: On the second thought, maybe I'm wrong here. Users will have to use Chrome or Firefox anyway to do all the interactions with Polkadot, so maybe it's not an issue at all.

src/shared/components/Checkout/Checkout.styles.tsx
eldiegod: Maybe not clear in storybook but if the step name is too long and it wraps this will add a bit of a gap between the text and the chevron

src/shared/components/Checkout/Checkout.styles.tsx
eldiegod: Yeah, I think it might not be worth replacing it because of what you mention; the user already needs to be on a modern browser, and if that's not the case, we would have to change it in several more places where is being used

drillprop (approved):

src/shared/components/Checkout/Checkout.styles.tsx
kdembler: I agree with Bartosz here, the support is pretty poor (69%). Currently our target is +/- 90% so we should be mindful of that. I don't think we actually use `gap` for flexbox anywhere (we use it for grid but that's much better supported).

kdembler: I also don't think our users need to be on a _very_ modern browser, as far as I'm aware our least supported feature atm is resize observer so it's not cutting edge

kdembler (changes requested):

eldiegod:
Fair points, I do hope the `gap` api is widely supported soon because its so nice 😞

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,87 @@
...
+`
+
+export const ProgressbarContainer = styled.div`
+  width: 24px;
+  height: 24px;
+`
+
+export const StepsProgressContainer = styled.div`
+  display: flex;
+  gap: 12px;
```
drillprop: You forgot about this one :P

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -0,0 +1,87 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 32px;
+`
+
+export const StepsContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 14px;
```
drillprop: ...and this one

eldiegod: You right, wonder why the search didnt bring them out for me ty 👨‍💻

drillprop (approved):
👌

src/shared/components/Checkout/Checkout.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const stepsCompletedNumber = steps.filter(({ completed }) => completed).length
+  return (
+    <Container>
+      <TopRowContainer>
+        <Text variant="h6">Progress</Text>
+        <StepsProgressContainer>
+          <ProgressbarContainer>
+            <CircularProgressbar value={stepsCompletedNumber} maxValue={steps.length}></CircularProgressbar>
+          </ProgressbarContainer>
+          <Text variant="body2">
```
kdembler: Just noticed this but the progress circle actually moves based on the progress:

https://user-images.githubusercontent.com/12646744/111629868-07f0f780-87f2-11eb-8434-9562e03a0d5b.mov

You can fix this by attaching
```css
min-width: 25px;
text-align: right;
```

to this text

eldiegod: Right, added the styles doesn't seem to move anymore

kdembler (approved):

Issue #279, opened 2021-01-28
Create publishing topbar


kdembler:
#### Description

Similarly to https://github.com/Joystream/atlas/issues/275, we will need a Studio specific topbar. Design can be found [here](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3107%3A443) and [here](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3893%3A4)

#### Requirements

- `StudioTopbar` component added to `src/components`
- `TopNavbar` renamed to `Topbar` (since it doesn't do any navigation)
- If makes sense, common parts of both extracted to `TopbarBase` in `src/components/shared`
- Menu implemented according to the design
  - Some parts may be detailed later, ping @kdembler 
- New topbar intergrated
  - Should be displayed instead of the regular one when the path starts with `/studio`

PR #377, opened 2021-03-12
Publishing Topbar

mikkio-j:
Issue: #279 
Figma: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5349%3A101363](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5349%3A101363)


src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,186 @@
...
+  isActive?: boolean
+}
+
+export const StyledTopbarBase = styled((props: TopbarBaseProps) => <TopbarBase {...props} />)`
+  display: flex;
+  justify-content: space-between;
+  @media screen and (min-width: ${breakpoints.small}) {
+    display: flex;
+    justify-content: space-between;
+  }
```
eldiegod: Is this repetition needed?

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,186 @@
...
+  ${StyledAvatar} {
+    margin-left: 0;
+  }
+`
+
+export const MemberInfoContainer = styled.div`
+  display: flex;
+  align-items: center;
+  margin: calc(-1 * 12px);
```
eldiegod:
```suggestion
  margin: -12px;
```
isn't this enough here?

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+export const MemberTextContainer = styled.div`
+  display: flex;
+  align-items: center;
+`
+
+export const MemberTitleText = styled(Text)`
+  background-color: ${colors.gray[400]};
+  color: ${colors.white};
+  font-size: 10px;
```
eldiegod: I think this should be `body1` not 100% sure

eldiegod (changes requested):
Ahh this looks so good great stuff!

The only thing is UX wise we might want to not close the dropdown when clicking inside it when this click is not on a button. If I misclick it closes which is not the best; tho I think it might be a bit tricky to implement.

src/assets/full-logo-studio.svg
```diff
@@ -0,0 +1 @@
...a2.9 2.9 0 01-.92-.14 1.8 1.8 0 01-.73-.45c-.2-.2-.37-.47-.48-.79a3.53 3.53 0 01-.18-1.17zm7.49-1.19c0-.51.05-1 .17-1.45.12-.46.3-.86.52-1.2a2.53 2.53 0 012.13-1.1 2.4 2.4 0 012.28 1.33V6.5h1.27V16h-1.27v-1.16a2.5 2.5 0 01-.9.99c-.37.23-.82.34-1.35.34-.48 0-.9-.1-1.26-.3-.36-.19-.65-.46-.89-.8a3.64 3.64 0 01-.53-1.18 6.01 6.01 0 01-.17-1.46zm1.32 1.1a1.54 1.54 0 00.53 1.18 1.97 1.97 0 001.3.45c.63 0 1.11-.2 1.45-.6.34-.4.5-.91.5-1.53v-1.2c0-.62-.16-1.12-.5-1.52-.34-.41-.82-.61-1.45-.61a2.08 2.08 0 00-1.3.45c-.16.14-.3.31-.39.51-.1.2-.14.42-.14.66v2.22zm7.2-5.73V6.48h1.31V7.8h-1.31zm.02 8.2V8.86h1.26V16h-1.25zm3.07-3.57c0-.54.06-1.04.2-1.5.13-.47.32-.86.59-1.18.27-.33.6-.59 1-.78.4-.18.88-.28 1.42-.28.54 0 1.01.1 1.41.28.4.19.74.45 1 .78.27.32.46.71.6 1.18.13.46.2.96.2 1.5 0 .55-.07 1.06-.2 1.51-.14.46-.33.85-.6 1.18-.26.33-.6.58-1 .77-.4.19-.88.28-1.41.28-.54 0-1.02-.1-1.42-.28-.4-.19-.73-.44-1-.77a3.33 3.33 0 01-.6-1.18c-.13-.45-.19-.96-.19-1.5zm1.33 1.1a1.46 1.46 0 00.56 1.18 2.03 2.03 0 001.32.45 2.03 2.03 0 001.31-.44 1.58 1.58 0 00.56-1.18v-2.22a1.58 1.58 0 00-.56-1.18 2.1 2.1 0 00-2.04-.33c-.22.09-.42.2-.6.34-.16.14-.3.31-.4.51-.1.2-.15.42-.15.66v2.22z" fill="#7B8A95"/></svg>
```
drillprop: Not sure if we should create a separate SVG just for the studio logo. We could use a regular logo and attach `studio` to it, right?

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,186 @@
...
+  active?: boolean
+}
+
+type NavDrawerProps = {
+  active?: boolean
+}
+
+type DrawerButtonProps = {
+  isActive?: boolean
+}
```
drillprop: `active`, `active` and then `isActive` 😅  To avoid repetition, maybe you could replace these with something like `CommonTopbarProps`

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }
+  svg {
+    margin-left: auto;
+    margin-right: 12px;
+  }
+`
+
+export const StyledAvatar = styled(Avatar)`
+  width: 42px;
+  height: 42px;
```
drillprop: It's not in the scope of this task, but maybe we could add these sizes as a variant to the Avatar component? 
Just as a sidenote.

src/components/StudioTopbar/StudioTopbar.tsx
```diff
@@ -0,0 +1,166 @@
...
+  name: 'Mikael Cowan',
+  avatar: 'https://picsum.photos/300/300',
+  channels: [
+    { name: 'Wild Crypto Fan16', avatar: 'https://picsum.photos/200/300' },
+    { name: 'Mild Crypto Skeptic', avatar: 'https://picsum.photos/200/300' },
+    { name: 'Average Cryptocurrency Enjoyer', avatar: 'https://picsum.photos/200/300' },
+  ],
+}
+
+const StudioTopbar: React.FC = () => {
```
drillprop: I know that's an early stage for this, but member, avatar, and channels should be passed via props.

drillprop (commented):
Overall great work. Looks really neat. I think we could make some improvements to mobile view. Right now you just hiding completely right side. 
https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5762%3A426
image

src/components/StudioTopbar/StudioTopbar.tsx
mikkio-j: `StudioTopbar` is used in `LayoutWithRouting`, so that would mean that we're going to fetch data in routing component and pass it to top bar. I'm not sure if it's a good idea. I believe that we're going to use hook for getting member and active channels so maybe it's better to use that in top bar component - in that case we don't need to pass it  via props. What do you think?

eldiegod: Yes that's right this is very likely going to be a hook so not a prob imo

drillprop: Okay, that makes sense.

drillprop (approved):
LGTM! 🎉 
Why it's still a draft?

mikkio-j:
> LGTM! 🎉
> Why it's still a draft?

It has hardcoded data for active member and channel, so it'll have to wait for:
1. save active member/channel hook
2. membership queries

src/components/StudioTopbar/StudioTopbar.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (
+      channelInfoRef.current?.contains(event.target as Node) ||
+      drawerButtonRef.current?.contains(event.target as Node)
+    ) {
+      return
+    }
+    if (drawerRef.current && !drawerRef.current.contains(event.target as Node)) {
+      setDrawerActive(false)
+    }
+  }
```
eldiegod: This could be moved inside the useEffect where it is used because the effect might not be able to remove the listener as this function callback will probably change signatures before it gets to the cleanup; another way to do it is wrapping the fn on useCallback

eldiegod (approved):
Beautiful 🎖

I would say let's merge these and add tasks to integrate the blocked parts when they become unblocked? @kdembler

kdembler (commented):
Haven't looked at the code yet , just the visuals, but looks really good! Some things I've noticed:

1. I think we could have bigger max width for the channel name on bigger screens, seems like the wrapping is happening too early
Screenshot 2021-03-17 at 12 51 37

2. I see there are couple of variations for the design, I guess it wasn't finalized. I think the member name when displayed as a subtext could have a bit more contrast, it's really hard to notice now. I did some testing and `gray[300]` seems to work nicely without any opacity. Let's try it out?

Screenshot 2021-03-17 at 12 56 04

3. Safari seems to have some problems with it. It seems you're using `row-gap` and `column-gap` with `flex` for spacing, which has very low support (not supported by safari at all). We need to either use `flex` with some other spacing mechanism or just use `grid`:
Screenshot 2021-03-17 at 12 57 20

4. I think clicking `New channel` should close the overlay, same as changing the channel

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,180 @@
...
+import styled from '@emotion/styled'
+import { Link } from 'react-router-dom'
+import { breakpoints, colors, sizes, transitions, typography, zIndex } from '@/shared/theme'
+import { TopbarBase, Avatar, Text, Button, TOP_NAVBAR_HEIGHT, TopbarBaseProps } from '@/shared/components'
+
+type CommonStudioTopbarProps = {
+  isActive?: boolean
+}
+
+export const StyledTopbarBase = styled((props: TopbarBaseProps) => <TopbarBase {...props} />)`
```
kdembler: any reason not to just do `styled(TopbarBase)`?

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,180 @@
...
+      display: none;
+      @media screen and (min-width: ${breakpoints.medium}) {
+        display: flex;
+      }
+    }
+  }
+`
+
+export const AddVideoButton = styled(Button)`
+  display: none;
```
kdembler: Do we need to hide it? It seems there's enough place for it there and it'd be nice for the video upload to be always handy

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,180 @@
...
+    display: flex;
+  }
+`
+
+export const MemberInfoContainer = styled.div`
+  display: flex;
+  align-items: center;
+  margin: calc(-1 * ${sizes(3)});
+  margin-bottom: 0;
+  border-bottom: 1px solid rgba(255, 255, 255, 0.18);
```
kdembler: Let's try to use one of `colors.transparentWhite` for this. The closest we have defined is `32`, not sure if it will work, if not we can just define `18` there

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,180 @@
...
+  opacity: 0.6;
+  margin-left: ${sizes(3)};
+`
+
+export const StyledLink = styled(Link)`
+  text-decoration: none;
+`
+
+export const LogoutButton = styled(Button)`
+  background-color: transparent;
```
kdembler: Should be coming from the component after recent merge

src/components/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -0,0 +1,180 @@
...
+  cursor: pointer;
```
kdembler: Button should have that set already right?

src/components/StudioTopbar/StudioTopbar.tsx
```diff
@@ -0,0 +1,198 @@
...
+  channels: [
+    { name: 'Wild Crypto Fan16', avatar: 'https://picsum.photos/201/300' },
+    { name: 'Mild Crypto Skeptic', avatar: 'https://picsum.photos/202/300' },
+    { name: 'Average Cryptocurrency Enjoyer', avatar: 'https://picsum.photos/203/300' },
+  ],
+}
+
+const StudioTopbar: React.FC = () => {
+  const [isDrawerActive, setDrawerActive] = useState(false)
+  // TODO Change that to use hook for saving/getting currently active channel
```
kdembler: I guess this can be integrated now?

src/components/StudioTopbar/StudioTopbar.tsx
```diff
@@ -0,0 +1,198 @@
...
+      }
+      if (drawerRef.current && !drawerRef.current.contains(event.target as Node)) {
+        setDrawerActive(false)
+      }
+    }
+    document.addEventListener('click', handleClickOutside, true)
+    return () => {
+      document.removeEventListener('click', handleClickOutside, true)
+    }
+  })
```
kdembler: This will run on every render since it doesn't have a dependency array, I don't think that's desired?

src/components/StudioTopbar/StudioTopbar.tsx
```diff
@@ -0,0 +1,198 @@
...
+        {channels.map((channel) => (
+          <ChannelInfo
+            key={channel.name}
+            channel={channel}
+            member={memberName}
+            active={channel.name === currentChannel.name}
+            onClick={() => onCurrentChannelChange(channel)}
+          />
+        ))}
+        <StyledLink to="channel/new" onClick={handleClose}>
```
kdembler: This should come from `routes`

kdembler (commented):
Some minor stuff 👍

src/components/StudioTopbar/StudioTopbar.style.tsx
mikkio-j: Got an error: `Error: You are trying to create a styled element with an undefined component. You may have forgotten to import it.`

Went back to `styled(TopbarBase)` but had to use relative import path for that component instead of `'@/shared/components'`

src/components/StudioTopbar/StudioTopbar.tsx
mikkio-j: Yup, but we still need membership queries to make it work

src/components/StudioTopbar/StudioTopbar.style.tsx
kdembler: Yep, that's a known issue

src/components/StudioTopbar/StudioTopbar.tsx
```diff
@@ -0,0 +1,200 @@
...
+      }
+      if (drawerRef.current && !drawerRef.current.contains(event.target as Node)) {
+        setDrawerActive(false)
+      }
+    }
+    document.addEventListener('click', handleClickOutside, true)
+    return () => {
+      document.removeEventListener('click', handleClickOutside, true)
+    }
+  }, [])
```
kdembler: I think we potentially could improve this. This will add an event listener on mount and react on every page click. So regardless of whether the overlay is open or not, we will constantly call `setDrawerActive(false)` on every page click. What we could is add event listener only once the overlay is opened. Then we could also simplify the logic here I think. Not sure this will work 1:1, but consider this:
```suggestion
  useEffect(() => {
    if (!isDrawerActive) {
      return
    }

    const handleClickOutside = (event: Event) => {
      if (drawerRef.current && !drawerRef.current.contains(event.target as Node)) {
        // stop propagation so drawer doesn't get triggered again on button click
        event.stopPropagation()
        setDrawerActive(false)
      }
    }
    document.addEventListener('click', handleClickOutside, true)
    return () => {
      document.removeEventListener('click', handleClickOutside, true)
    }
  }, [isDrawerActive])
```

I think this also allows us to get rid of `channelInfoRef` and `drawerButtonRef`. Let me know if that works

src/components/Topbar/Topbar.tsx
```diff
@@ -1,15 +1,8 @@
...
-  NavigationContainer,
-  SearchbarContainer,
-  ShortLogo,
-  StyledSearchbar,
-} from './TopNavbar.style'
+import { TopbarBase } from '@/shared/components'
+import { SearchbarContainer, StyledSearchbar } from './Topbar.style'
 import { RoutingState } from '@/types/routing'
 
 const TopNavbar: React.FC = () => {
```
kdembler: I think if we have `StudioTopbar`, we probably should call this `ViewerTopbar`. Also let's make sure the component name matches the filename

src/components/Topbar/index.ts
```diff
@@ -0,0 +1,2 @@
+import TopNavbar from './Topbar'
```
kdembler: Let's also update those names

src/config/routes.ts
```diff
@@ -18,11 +18,12 @@ export default {
   channels: () => '/channels',
   playground: () => '/playground',
   studio: () => `/studio`,
+  studioNewChannel: () => '/studio/channel/new',
```
kdembler, line 21: This should probably go into `studioRoutes`?

kdembler: Oh wait it's already there?

kdembler (approved):
Great work. Merging this one, but we will need some followup due to the most recent design updates. Created https://github.com/Joystream/atlas/issues/391 to tackle those

Issue #367, opened 2021-03-09
Add ability to save currently selected channel and member


kdembler:
In the publishing flows, we'll need to be able to tell which channel is currently selected by the user with ability to change them. Same for the membership but without changing (just logout). This should be saved in local storage so the choices are persisted. Should work similarly to the drafts storage

PR #378, opened 2021-03-15
Save active channel and member

mikkio-j:
#367


src/hooks/useMember/useMember.tsx
```diff
@@ -0,0 +1,91 @@
...
+    fetchMember()
+  }, [fetchMember])
+
+  return {
+    getMember: memberState,
+    addMember,
+    addChannel,
+    setActiveChannel,
+    removeMember,
+  }
```
eldiegod: Do we also need a remove channel?

src/hooks/useMember/useMember.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return updatedMember
+  }, [])
+
+  const removeMember = useCallback(async () => {
+    removeMemberFn()
+    fetchMember()
+  }, [fetchMember])
+
+  return {
+    getMember: memberState,
```
eldiegod: Is this a function? if not, should it be just `member`?

eldiegod (changes requested):
Hey did a first check and it looks good but it's a bit hard to tests everything I think we could add a playground for it or something like that 🤔

Tho, if you think it would take too long then forget it.

src/hooks/useMember/useMember.tsx
mikkio-j: turned out, it's not in the scope at the moment

src/hooks/useMember/useMember.tsx
```diff
@@ -0,0 +1,103 @@
...
+  removeMember as removeMemberFn,
+  addChannel as addChannelFn,
+} from './utils'
+
+export type Member = {
+  id: string
+  handle: string
+  avatarUrl: string
+  channels: Channel[]
+  activeChannel: Channel
```
kdembler: I'm not sure if we actually should save anything except the IDs. It will probably make more sense to just fetch all the rest. Let's discuss in the call

src/hooks/useMember/useMember.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react'
+import {
+  getMember as getMemberFn,
+  setMember as setMemberFn,
+  setActiveChannel as setActiveChannelFn,
+  removeMember as removeMemberFn,
+} from './utils'
+
+export type Member = {
```
drillprop: To avoid confusion I think we could name this differently, maybe `ActiveMember`?  Probably we will have something like `Member`  after we update queries and generate new types.

src/hooks/useMember/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const setMember = async (member: Member) => {
+  writeToLocalStorage('member', member)
+  const newMember = await getMember()
+  return newMember
+}
+
+export const setActiveChannel = async (channelId: string) => {
+  const member = await getMember()
+  if (!member?.id) {
```
drillprop:
```suggestion
  if (!member) {
```
Will also work, right?

src/views/playground/Playgrounds/PlaygroundMemberChannel.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { member, setMember, removeMember, setActiveChannel } = useMember()
+
+  const [selected, setSelected] = useState<string | number>('')
+  const [memberString, setMemberString] = useState('')
+
+  const handleActiveChannelChange: (e: React.MouseEvent<HTMLElement>) => void = (e) => {
+    const element = e.currentTarget as HTMLInputElement
+    setSelected(element.value)
+    setActiveChannel(element.value)
+  }
```
drillprop:
```suggestion
  const handleActiveChannelChange = (e: React.MouseEvent<HTMLInputElement>) => {
    const element = e.currentTarget
    setSelected(element.value)
    setActiveChannel(element.value)
  }
```
Not sure why you did that. I think what I suggested here will also work

src/hooks/useMember/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+    throw new Error('Member must be setted first.')
```
drillprop: also, I think we shouldn't throw error here. We could just `return`

drillprop (commented):
Found some small things to consider. Otherwise, it works well

drillprop (approved):
LGTM!

src/hooks/useMember/utils.ts
```diff
@@ -0,0 +1,26 @@
+import { promisify } from '@/utils/data'
+import { readFromLocalStorage, writeToLocalStorage } from '@/utils/localStorage'
+import { ActiveMember } from './useMember'
+
+export const getMember = promisify(() => readFromLocalStorage<ActiveMember>('activeMember') || null)
+
+export const setMember = async (member: ActiveMember) => {
+  writeToLocalStorage('activeMember', member)
+  const newMember = await getMember()
```
eldiegod: Is this `await getMember()` fetch needed? or can we return member?

src/hooks/useMember/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const setActiveChannel = async (channelId: string) => {
+  const member = await getMember()
+  if (!member) {
+    return
+  }
+  const updatedMember = { ...member, activeChannelId: channelId }
+  writeToLocalStorage('activeMember', updatedMember)
+  const newMember = await getMember()
```
eldiegod: Same here? updatedMember should be fine(?)

eldiegod (approved):
Works goood

kdembler (approved):
Looks good, sharing the same comments as Diego

src/hooks/useMember/useMember.tsx
```diff
@@ -0,0 +1,78 @@
...
+import {
+  getMember as getMemberFn,
+  setMember as setMemberFn,
+  setActiveChannel as setActiveChannelFn,
+  removeMember as removeMemberFn,
+} from './utils'
+
+export type ActiveMember = {
+  id: string
+  activeChannelId: string
```
kdembler: I actually think we should make those nullable. For example when you log out as a member we will need to clear those. Our hook will allow this as well, either by exposing additional functions to unset or just allow empty values in setters

src/hooks/useMember/useMember.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react'
+import {
+  getMember as getMemberFn,
+  setMember as setMemberFn,
+  setActiveChannel as setActiveChannelFn,
+  removeMember as removeMemberFn,
+} from './utils'
+
+export type ActiveMember = {
```
kdembler: Just came to me - we should also keep the active account address, we will need that to sign the transactions (you can have a number of accounts in your extension and we need to pick one that we want the user to sign with). This also creates a question - what should be a structure of this. The hierarchy is the following `account -> member -> channel` so if we want to include an address as well, the current structure may need to change. What do you think? Sorry, I haven't thought of this before

kdembler (approved):

PR #380, opened 2021-03-15
Button fixes

### LINKED ISSUE #376: Button fixes
Several button issues has been introduced recently:
- search close icon not visible entirely
Screenshot 2021-03-11 at 16 30 00

- remove button on `VideoPreview` and category picker buttons both have `min-width` which they shouldn't have
Screenshot 2021-03-11 at 16 30 20
Screenshot 2021-03-11 at 16 31 27

- cover video play button too small
Screenshot 2021-03-11 at 16 34 25

compare to `master`:
Screenshot 2021-03-11 at 16 34 30

mikkio-j:
#376

### DIFF
--- a/src/shared/components/Searchbar/Searchbar.style.tsx
+++ b/src/shared/components/Searchbar/Searchbar.style.tsx
@@ -32,16 +32,10 @@ export const CancelButton = styled(Button)`
   position: absolute;
   right: 0;
   border: none;
-  padding: 14px ${sizes(3)};
   color: ${colors.white};
   :focus,
   :hover {
     color: ${colors.white};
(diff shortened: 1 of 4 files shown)


src/shared/components/Searchbar/Searchbar.style.tsx
```diff
@@ -32,17 +32,11 @@ export const CancelButton = styled(Button)`
   position: absolute;
```
kdembler, line 32: Maybe we could add some margin-right here? Seems like the button is too closely hugging the right border

kdembler (approved):

PR #385, opened 2021-03-16
Layout for studio routes

eldiegod:
Adds a max-width to the views according to designs, might be tweaked later.

### DIFF
--- a/src/views/studio/StudioRouter.tsx
+++ b/src/views/studio/StudioRouter.tsx
@@ -1,8 +1,22 @@
+import styled from '@emotion/styled'
 import React from 'react'
 import { Routes } from 'react-router'
 
 const StudioRouter = () => {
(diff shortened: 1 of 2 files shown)


src/views/studio/StudioRouter.tsx
```diff
@@ -1,8 +1,21 @@
+import styled from '@emotion/styled'
 import React from 'react'
 import { Routes } from 'react-router'
 
 const StudioRouter = () => {
-  return <Routes>{/* routes for studio here */}</Routes>
+  return (
+    <>
```
kdembler: Why the fragment here?

eldiegod: forgot to remove

kdembler (approved):

PR #388, opened 2021-03-17
Update ImageCropDialog with latest design

### LINKED ISSUE #373: Update ImageCropDialog with latest design
Some design changes has been made to the `ImageCropDialog`:
- zoom in/out icons should be updated and should now be clickable to allow adjustment of zoom (@eldiegod 😛)
- updated overlay color and overlay border color

drillprop:
Fix #373
Figma https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5654%3A131190


src/components/Dialogs/ImageCropDialog/ImageCropDialog.style.ts
```diff
@@ -14,6 +14,9 @@ const roundedCropperCss = css`
     border-radius: 50%;
   }
 `
+export const StyledZoomButton = styled(Button)`
+  cursor: pointer;
```
kdembler: Doesn't button already have this style?

src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
```diff
@@ -83,15 +84,15 @@ const ImageCropDialogComponent: React.ForwardRefRenderFunction<
 
   const zoomControlNode = (
     <ZoomControl>
-      <Icon name="zoom-out" />
+      <StyledZoomButton icon="zoom-out" variant="tertiary" onClick={() => handleZoomChange(currentZoom - 0.1)} />
```
kdembler: These values may not work the best here. What works best will depend on the picture really, take a look:

https://user-images.githubusercontent.com/12646744/111460165-f2110300-871b-11eb-9775-de56734e6cf4.mov

The change should be more gradual here. Maybe you can use `zoomStep` for that?

src/components/Dialogs/ImageCropDialog/ImageCropDialog.style.ts
drillprop: nope. Maybe we should update the Button as well at some point

src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
drillprop: I wasn't aware of this, tested only on one image. Should be good now.

src/components/Dialogs/ImageCropDialog/ImageCropDialog.style.ts
drillprop: Wait, my mistake, you were right.  I just found that if you add `onClick`, `pointer` will be added

src/components/Dialogs/ImageCropDialog/ImageCropDialog.style.ts
```diff
@@ -68,13 +68,13 @@ export const CropContainer = styled.div<{ rounded?: boolean }>`
       top: 0;
       border-radius: ${({ rounded }) => (rounded ? '50%' : '0')};
 
-      box-shadow: inset 0 0 0 2px rgba(187, 231, 255, 0.48);
+      box-shadow: inset 0 0 0 2px rgba(256, 256, 256, 0.32);
```
kdembler: As far as I'm aware RGB values go up to `255` :P

On a more serious note, those semi-transparent layers have been define as part of the colors. Maybe to keep this tidy we should introduce those into `theme.colors`. Then we could use those both here and below for `.cropper-modal`

kdembler: You can find those here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4882%3A142

src/components/Dialogs/ImageCropDialog/cropper.ts
```diff
@@ -53,8 +53,12 @@ export const useCropper = ({ imageEl, imageType }: UseCropperOpts) => {
...
 
   const handleZoomChange = (zoom: number) => {
-    setCurrentZoom(zoom)
-    cropper?.zoomTo(zoom)
+    const [minZoom, maxZoom] = zoomRange
+    const isInRange = zoom >= minZoom - zoomStep && zoom <= maxZoom + zoomStep
+    if (isInRange) {
+      setCurrentZoom(zoom)
+      cropper?.zoomTo(zoom)
+    }
```
kdembler: I think what could be helpful here is that the value exceeds the range, just make it min/max from that range, the we can avoid the situations where the user is less than `step` from min/max but cannot move to min/max itself. Does that make sense?

src/shared/theme/colors.ts
```diff
@@ -25,6 +25,22 @@ export default {
...
+    300: 'rgba(255,255,255, 0.32)',
+  },
+  transparentBlack: {
+    100: 'rgba(0,0,0, 0.24)',
+    200: 'rgba(0,0,0, 0.54)',
+    300: 'rgba(0,0,0, 0.66)',
+  },
+  transparentPrimary: {
+    100: 'rgba(180, 187, 255, 0.06)',
+    200: 'rgba(180, 187, 255, 0.12)',
```
kdembler: I don't think `100` `200` and `300` are correct names here, that's what you call range of colors. Here let's maybe use the percentage of opacity as the key? Figma uses names like this:
Screenshot 2021-03-17 at 16 21 35

So it would be the most straight-forward to use `colors.transparentBlack[54]` in the code

kdembler (commented):
Just last tweaks

kdembler (approved):

Issue #275, opened 2021-01-28
Create publishing sidenav


kdembler:
#### Description

We will need a separate navigation sidenav for the publishing part of Atlas.

#### Requirements

- `PublishingSidenav` added to `src/components` (with items defined inside the component)
  - same done for `ViewerSidenav`
- New sidenav integrated
  - Should be displayed instead of the regular one when the path starts with `/studio`

0xVolodya:
Hi, can I try to do this?

kdembler:
Hey @nadalfederer, thanks for your interest in this. I'd be happy if you worked on this, however, I think this may be slightly blocked by one other task currently in progress - #276. I think it'd make most sense to work on this once the other is finished.

PR #390, opened 2021-03-19
Publishing sidenav

mikkio-j:
Issue: #275 
Figma: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4469%3A177](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4469%3A177)


src/components/ViewerSideNav/ViewerSideNav.style.ts
```diff
@@ -84,13 +96,13 @@ export const SidebarNavLink = styled(Link, { shouldForwardProp: isPropValid })<S
...
+    background-color: ${({ isStudio }) => (isStudio ? colors.transparentWhite[4] : 'rgba(0,0,0,0.12)')};
   }
   &:focus {
-    background-color: rgba(0, 0, 0, 0.24);
+    background-color: ${({ isStudio }) => (isStudio ? colors.transparentWhite[6] : 'rgba(0,0,0,0.24)')};
   }
   &:active {
-    background-color: rgba(0, 0, 0, 0.4);
+    background-color: ${({ isStudio }) => (isStudio ? colors.transparentWhite[32] : 'rgba(0,0,0,0.4)')};
   }
```
eldiegod: I think we should use some of the colors from `transparentBlack` but not sure if possible for all of them

src/components/ViewerSideNav/ViewerSideNav.style.ts
```diff
@@ -106,7 +118,7 @@ export const SidebarNavLink = styled(Link, { shouldForwardProp: isPropValid })<S
     line-height: 1;
   }
   &[data-active='true'] {
-    background-color: rgba(0, 0, 0, 0.24);
+    background-color: ${({ isStudio }) => (isStudio ? colors.transparentWhite[6] : 'rgba(0,0,0,0.24)')};
```
eldiegod: same as above

eldiegod (approved):
Solid PR added and added a small comment 👍

src/components/ViewerSideNav/ViewerSideNav.tsx
```diff
@@ -24,27 +25,50 @@ import routes from '@/config/routes'
...
+    to: routes.videos(),
+  },
+  {
+    icon: 'channels',
+    name: 'Channels',
+    to: routes.channels(),
+  },
+]
 
+const ViewerSideNavbar: React.FC<SidenavProps> = ({ items, isStudio }) => {
```
kdembler: This component is used for both viewer and studio so why is it called `ViewerSideNav`? We should have some generic `SideNav` component and then `StudioSideNav` and `ViewerSideNav` based on that. Or potentially we could just have the generic one and define the items in respective layouts, that could also work 

As a sidenote, maybe we also should use `Sidenav`, without the second capitalisation? Would be more in line with `Topbar`

src/components/ViewerSideNav/ViewerSideNav.tsx
```diff
@@ -56,45 +80,51 @@ const SideNavbar: React.FC<SidenavProps> = ({ items }) => {
...
               to={item.to}
               expanded={expanded}
               subitems={item.subitems}
               itemName={item.name}
               onClick={closeSideNav}
+              isStudio={isStudio}
             >
               <Icon name={item.icon} />
-              <span>{item.name}</span>
+              <span>{expanded && item.expandedName ? item.expandedName : item.name}</span>
```
kdembler: This doesn't work very well because you can see the change when collapsing. Do we need to change those at all? I think this should work fine if we displayed `expandedName` always?

src/components/Sidenav/Sidenav.tsx
```diff
@@ -31,20 +32,23 @@ type NavItemType = {
...
-type SidenavProps = {
+export type SidenavProps = {
   items: NavItemType[]
+  isStudio?: boolean
 }
 
-const SideNavbar: React.FC<SidenavProps> = ({ items }) => {
-  const isStudio = useMatch(routes.studio())
-
+const ViewerSideNavbar: React.FC<SidenavProps> = ({ items, isStudio }) => {
```
kdembler:
```suggestion
const Sidenav: React.FC<SidenavProps> = ({ items, isStudio }) => {
```

src/components/Sidenav/Sidenav.tsx
```diff
@@ -137,5 +149,5 @@ const NavItem: React.FC<NavItemProps> = ({ expanded = false, subitems, children,
   )
 }
 
-export { SideNavbar as default, NavItem }
+export { ViewerSideNavbar as default, NavItem }
```
kdembler:
```suggestion
export { Sidenav as default, NavItem }
```

src/components/ViewerSideNav/ViewerSideNav.tsx
kdembler: @mikkio-j I think you misunderstood what I said here. I didn't mean to remove `expandedName` altogether, this clearly looks bad:
Screenshot 2021-03-23 at 17 35 27
What I meant was that we can always display the `expandedName` in that `span` the comment was targeting, we should still pass the shorter name to the `NavItem`

kdembler (approved):

PR #392, opened 2021-03-19
fix infinite playback bug

kdembler:


drillprop (approved):
I can't reproduce this bug anymore, so I think it should be good now. 🎉

Issue #262, opened 2021-01-28
Add membership related queries


kdembler:
#### Description

We will need to query some membership related info from the query node to display in publishing flows. Let's add those

#### Requirements

- Queries & hooks for membership oriented queries added (look at https://github.com/Joystream/atlas/issues/269)
- Mocking code added to support this functionality
- Example added to the playground

PR #394, opened 2021-03-22
Membership queries

mikkio-j:
Issue: #262


drillprop (approved):
didn't find anything. Looks good to me! 👍🏻

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -22,7 +27,7 @@ type Channel {
 
   avatarPhotoUrl: String
 
-  owner: Member!
+  member: Membership!
```
drillprop: Don't know why, but in the gist, there is no `member` field in `Channel`, and also there are no `channels` on `Membership`. 
Not sure what we should do with that. If the schema is wrong I think you should also add `channels` to `Membership`

mikkio-j: it's a temporary solution/assumption, currently we are waiting for the query node guys to update that. @kdembler do we have more info about that?

kdembler: As far as I can tell this is currently blocked by some missing functionality in Hydra so we have to wait. Let's leave it out for now and update when the new query node gets deployed

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -4,9 +4,14 @@ type Language {
...
 
-type Member {
+type Membership {
   id: ID!
+
   handle: String!
+
+  avatarUri: String
+
+  controllerAccount: ID!
```
kdembler: This is `String` in the schema

eldiegod (approved):
Looks good to me codewise, I don't have much deep context on the matter tho

src/api/hooks/membership.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type MembershipOpts = QueryHookOptions<GetMembershipQuery>
+
+export const useMembership = (variables: GetMembershipQueryVariables, opts?: MembershipOpts) => {
+  const { data, ...rest } = useGetMembershipQuery({
+    ...opts,
+    variables,
+  })
+  return {
+    membership: data?.membership as BasicMembershipFieldsFragment | undefined,
```
kdembler: Why do we need this type assertion?

src/api/queries/membership.graphql
```diff
@@ -0,0 +1,28 @@
+fragment BasicChannelFields on Channel {
```
kdembler: This is already defined in `channels.graphql`

kdembler: Let's call the file `memberships.graphql` to stay consistent with other files in the same directory

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -22,8 +29,6 @@ type Channel {
 
   avatarPhotoUrl: String
 
-  owner: Member!
```
kdembler, line 25: We still should have member reference in the `Channel` type. It will most likely be `ownerMember: Membership`

src/api/schemas/extendedQueryNode.graphql
```diff
@@ -171,6 +176,7 @@ type PageInfo {
 input ChannelWhereInput {
   isCurated_eq: Boolean
   isPublic_eq: Boolean
+  memberId_eq: ID
```
kdembler: Let's make this `ownerMemberId_eq` for now

src/components/PublishingTopbar/PublishingTopbar.tsx
```diff
@@ -18,53 +20,53 @@ import {
...
-type Channel = {
-  name: string
-  avatar: string
-}
 type ChannelInfoProps = {
   active?: boolean
-  member: string
-  channel: Channel
+  memberName?: string
+  channel: BasicChannelFieldsFragment | undefined
```
kdembler: If something can be `undefined` we usually use `?` notation: `channel?: BasicChannelFieldsFragment`

kdembler: More places like this in this file

src/components/PublishingTopbar/PublishingTopbar.tsx
```diff
@@ -18,53 +20,53 @@ import {
...
+} & MemberInfoProps
 
 const PublishingTopbar: React.FC = () => {
   const { activeUser, setActiveChannel } = useActiveUser()
-  // TODO Add member fetching
-  const [currentChannel, setCurrentChannel] = useState(member.channels[0])
-
+  const { memberships, loading, error } = useMemberships(
+    {
+      where: { controllerAccount_eq: activeUser?.accountId as string },
```
kdembler: why this string assertion?

src/components/PublishingTopbar/PublishingTopbar.tsx
```diff
@@ -18,53 +20,53 @@ import {
...
+    },
+    {
+      skip: activeUser?.accountId === undefined,
```
kdembler:
```suggestion
      skip: !activeUser?.accountId,
```

src/components/PublishingTopbar/PublishingTopbar.tsx
```diff
@@ -18,53 +20,53 @@ import {
...
+      onCompleted: () => {
+        const member = memberships?.find((member) => member.id === activeUser?.memberId)
+        const channel = member?.channels.find((channel) => channel.id === activeUser?.channelId)
+        setCurrentMember(member)
+        setCurrentChannel(channel)
+      },
+    }
+  )
+  const [currentMember, setCurrentMember] = useState<BasicMembershipFieldsFragment>()
+  const [currentChannel, setCurrentChannel] = useState<BasicChannelFieldsFragment>()
```
kdembler: Let's move those above the hook in which they're used

src/components/PublishingTopbar/PublishingTopbar.tsx
```diff
@@ -18,53 +20,53 @@ import {
...
+      skip: activeUser?.accountId === undefined,
+      onCompleted: () => {
+        const member = memberships?.find((member) => member.id === activeUser?.memberId)
+        const channel = member?.channels.find((channel) => channel.id === activeUser?.channelId)
+        setCurrentMember(member)
+        setCurrentChannel(channel)
+      },
+    }
+  )
+  const [currentMember, setCurrentMember] = useState<BasicMembershipFieldsFragment>()
```
kdembler: Do we need this state? We already have `activeUser.channelId`, right?

src/mocking/accessors/single.ts
```diff
@@ -1,5 +1,5 @@
-import { VideoWhereUniqueInput } from '@/api/queries'
-import { FilteringArgs, HasId } from '../types'
+import { VideoWhereUniqueInput, MembershipWhereUniqueInput } from '@/api/queries'
+import { FilteringArgs, HasId, HasControllerAccount } from '../types'
```
kdembler: Those new types don't seem to be needed here

src/mocking/handlers.ts
```diff
@@ -113,6 +120,18 @@ const queryNodeHandlers = [
...
+  // memberships
+  createQueryHandler<GetMembershipQuery, GetMembershipQueryVariables>(
+    queryNode,
+    GetMembershipDocument,
+    createSingleItemAccessor(mockMemberships)
+  ),
+  createQueryHandler<GetMembershipsQuery, GetMembershipsQueryVariables>(
+    queryNode,
+    GetMembershipsDocument,
+    createMembershipsAccessor(mockMemberships)
```
kdembler: Do we need this custom accessor? I'd think the regular `createOffsetLimitPaginationAccessor` would work here?

kdembler (changes requested):

src/components/PublishingTopbar/PublishingTopbar.tsx
```diff
@@ -18,53 +20,53 @@ import {
...
-  ],
-}
+} & MemberInfoProps
 
 const PublishingTopbar: React.FC = () => {
   const { activeUser, setActiveChannel } = useActiveUser()
-  // TODO Add member fetching
-  const [currentChannel, setCurrentChannel] = useState(member.channels[0])
-
+  const { memberships, loading, error } = useMemberships(
```
kdembler: I think here we actually can use `useMembership`. We can assume the membership will be set somewhere else (sign in view). So here we should just fetch one specific membership with ID specified in `activeUser`

kdembler (approved):

PR #400, opened 2021-03-24
clean up layout and routing

kdembler:


src/components/Sidenav/Sidenav.tsx
```diff
@@ -97,7 +97,7 @@ const Sidenav: React.FC<SidenavProps> = ({ items, isStudio }) => {
               variant="secondary"
               onClick={closeSideNav}
               icon="external"
-              to={isStudio ? routes.index() : routes.studio()}
+              to={isStudio ? routes.viewer.index() : routes.studio.index()}
```
mikkio-j: Can't go back to viewer view, this will need something like `../` or maybe separate route for going back from studio to viewer instead of `routes.viewer.index()`

drillprop: @mikkio-j are you sure? Works for me

drillprop: My bad, sorry. You were right, can't go back

src/config/routes.ts
```diff
@@ -1,34 +1,33 @@
 export default {
-  index: () => '/',
-  searchOverlay: () => `/search`,
-  search: ({ query }: { query?: string } = {}) => {
-    const basePath = '/search'
+  viewer: {
+    index: () => '',
```
drillprop:
```suggestion
    index: () => '/',
```
I think it should fix the issue mentioned by @mikkio-j

src/config/routes.ts
```diff
@@ -1,34 +1,33 @@
...
+  },
+  studio: {
+    index: () => 'studio',
+    newChannel: () => 'channel/new',
+    editChannel: () => 'channel',
+    videos: () => 'videos',
+    uploads: () => 'uploads',
+  },
+  playground: {
+    index: () => 'playground',
```
drillprop: We could list playground paths here like in the studio to keep consistency

src/views/playground/PlaygroundLayout.tsx
```diff
@@ -19,7 +26,13 @@ export const PlaygroundView = () => {
...
               <Link to="./member-active-channel">Active user/member/channel</Link>
             </LinksContainer>
-            <PlaygroundRouter />
+            <Routes>
+              <Route path="/validation-form" element={<PlaygroundValidationForm />} />
+              <Route path="/drafts" element={<PlaygroundDrafts />} />
+              <Route path="/video-metadata" element={<VideoMetaData />} />
+              <Route path="/uploading-files-data" element={<PlaygroundUploadingFilesData />} />
+              <Route path="/member-active-channel" element={<PlaygroundMemberChannel />} />
+            </Routes>
```
drillprop, line 42: By analogy with StudioLayout we could replace this with `playgroundRoutes`

drillprop (commented):
Looks really solid 👍🏻 , just a couple of things

src/config/routes.ts
kdembler: I'd like playground to be "self-contained" so I wouldn't want to leak its routes externally

src/components/Sidenav/Sidenav.tsx
kdembler: Fixed now. @drillprop your suggestion would indeed solve it but it would create another problem here:
```ts
if (
  cachedLocation.state?.overlaidLocation?.pathname === location.pathname ||
  location.pathname === [routes.viewer.index(), routes.viewer.search()].join('/')
)
```
Sometimes we will need to operate on those routes so ended up with a different approach

src/views/playground/PlaygroundLayout.tsx
kdembler: :+1: cleaned this up

drillprop (approved):

Issue #393, opened 2021-03-20
Better removal visuals from Continue Watching


bedeho:
When I remove a single video, all videos temporarily disappear, a full row of loading items is displayed for a brief moment, and then the actual videos snap back in without the video I removed. This does not look very good.

1) Removal of a given video should not result in hiding others, even temporarily.
2) The single video should gracefully disappear, and the adjacent videos should slide in to take its place, or something like this. I know the animations will get redone later by the designers, but this looks quite rough as is.

kdembler:
Yeah, I really didn't like this behaviour but this was a technical constraint for the current release related to how we fetch the data (it was always batched, so we needed to refetch the full batch). Things changed and we should be able to achieve this much more easily after `sumer` gets merged

PR #401, opened 2021-03-25
Fix continue watching carousel bug

### LINKED ISSUE #398: Continue watching carousel not updating with new videos
Continue watching carousel not updating with new videos as one watches more videos it doesn't add them to the Continue watching carousel.

Related to #282

drillprop:
Fix #398

### DIFF
--- a/src/api/client/cache.ts
+++ b/src/api/client/cache.ts
@@ -10,13 +10,14 @@ const getVideoKeyArgs = (args: Record<string, GetVideosQueryVariables['where']> 
   const isPublic = args?.where?.isPublic_eq
   const channelIdIn = args?.where?.channelId_in ? JSON.stringify(args.where.channelId_in) : ''
   const createdAtGte = args?.where?.createdAt_gte ? JSON.stringify(args.where.createdAt_gte) : ''
+  const idIn = args?.where?.id_in ? JSON.stringify(args.where.id_in) : ''
 
   // only for counting videos in HomeView
   if (args?.where?.channelId_in && !args?.first) {
     return `${createdAtGte}:${channelIdIn}`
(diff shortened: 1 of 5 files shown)


eldiegod (approved):
Perfectly balanced, as all things should be

drillprop:
I'll set it as a draft for now. Tried several approaches to fix  #393 also, but still did not manage to do this. Will look at it once more later with a fresh mind.

src/components/Sidenav/Sidenav.tsx
```diff
@@ -69,7 +69,7 @@ const Sidenav: React.FC<SidenavProps> = ({ items, isStudio }) => {
         <SidebarNavList>
           {items.map((item) => (
             <NavItem
-              key={item.name}
+              key={item.expandedName}
```
kdembler: Why is this change here?

src/api/client/cache.ts
```diff
@@ -10,13 +10,14 @@ const getVideoKeyArgs = (args: Record<string, GetVideosQueryVariables['where']>
   const isPublic = args?.where?.isPublic_eq
   const channelIdIn = args?.where?.channelId_in ? JSON.stringify(args.where.channelId_in) : ''
   const createdAtGte = args?.where?.createdAt_gte ? JSON.stringify(args.where.createdAt_gte) : ''
+  const idIn = args?.where?.id_in ? JSON.stringify(args.where.id_in) : ''
```
kdembler: Why do we need that?

drillprop: We don't need that anymore, forgot to revert this

kdembler (approved):

PR #402, opened 2021-03-26
initial polkadot setup

### LINKED ISSUE #259: Create basic web3 setup
#### Description

For publishing, we will require a lot of web3 interaction with the query nodes. Let's create a basic setup that makes a web3 client available to the app

#### Requirements

- Basic web 3 setup done
  - Should handle basic external signer (extension) integration

kdembler:
Resolves #259


src/hooks/useJoystream/hook.ts
```diff
@@ -0,0 +1,10 @@
+import { useContext } from 'react'
+import { JoystreamContext } from './provider'
+
+export const useJoystream = () => {
+  const ctx = useContext(JoystreamContext)
+  if (!ctx) {
+    throw new Error('useJoystream must be used within JoystreamProvider')
+  }
+  return ctx
+}
```
drillprop: I don't like the filename.  In my opinion, it should be named just `useJoystream` as we did in other hooks - `usePersonalData`, `useDrafts` etc.

src/joystream-lib/api.ts
```diff
@@ -0,0 +1,282 @@
...
+    this.log('Destroyed')
+  }
+
+  /* Private utilities */
+  private log(msg: string) {
+    console.log(`[JoystreamJS] ${msg}`)
+  }
+
+  private logWarn(msg: string) {
+    console.log(`[JoystreamJS] ${msg}`)
```
drillprop: If it's a warning, maybe it should be `console.warn`

src/joystream-lib/api.ts
```diff
@@ -0,0 +1,282 @@
...
+      })
+
+      this.onExtensionConnectedUpdate?.(true)
+      return true
+    } catch (e) {
+      throw new ExtensionUnknownError(e)
+    }
+  }
+
+  static async build(appName: string, endpoint: string): Promise<JoystreamJs> {
```
mikkio-j, line 89: not sure what `appName` does here. Is it needed?

src/joystream-lib/api.ts
kdembler: it certainly should

src/joystream-lib/api.ts
kdembler: yep, it's passed to `web3Enable`: https://github.com/Joystream/atlas/blob/85803b3d9814b02282a15f60d7ea8816b386a9b6/src/joystream-lib/api.ts#L65

it will be shown in the popup of the extension, asking the user if they allow connection with the app

src/hooks/useJoystream/hook.ts
kdembler: Hmm, I don't like that filename in return, but you have a better argument than preference here and that's consistency. So I'll rename it but possibly we should rename all of them at some point

kdembler: Changed

src/joystream-lib/api.ts
kdembler: Fixed

mikkio-j (approved):

Issue #286, opened 2021-01-29
Create uploads view layout


kdembler:
**TBD**, once design is ready

PR #403, opened 2021-03-26
Uploads view layout

mikkio-j:
Issue: #286 
Figma: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5799%3A23304](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5799%3A23304)


kdembler (changes requested):
Visual review for now:

1. It seems there's a lot of misalignment (all of those should be in line with each other):

(icons)
Screenshot 2021-03-26 at 11 58 56
(labels)
Screenshot 2021-03-26 at 11 59 26

2. The icons for status don't seem to be matching. Both in sizes (20px vs 24px) and in the icon shape:

(figma)
Screenshot 2021-03-26 at 12 01 44

(app)
Screenshot 2021-03-26 at 12 01 35

3. It seems we're also using a different icon for `image`:

(figma)
Screenshot 2021-03-26 at 12 02 32

(app)
Screenshot 2021-03-26 at 12 02 23

4. The progress bar seems to cover everything else, e.g. I cannot select text

5. Maybe the full area of the group should be clickable instead of just the button on the right to make it easier to see the details? What do you think?

6. I think we should improve the expand transitions. Right now only the details pane will slide down but the height of the group changes instantaneously so there's a jump in the layout. Let me know if you need help on that

7. Spacing between groups should be `24px`, it's bigger now.

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -0,0 +1,76 @@
...
+      fileStatus = <Icon name="success" />
+      break
+    case 'reconnecting':
+      fileStatus = <Icon name="error-second" />
+      fileStatusMessage = 'Reconnecting...'
+      break
+    default:
+      fileStatus = <CircularProgressbar value={asset.progress} />
+  }
+
```
drillprop, line 53: Don't like these `let`s here. I would do something like this 
```js

  const getStatus = (asset: Asset) => {
    switch (asset.status) {
      case 'uploading':
        return (
          <ProgressbarContainer>
            <CircularProgressbar value={asset.progress} />
          </ProgressbarContainer>
        )
      case 'pending':
        return (
          <ProgressbarContainer>
            <CircularProgressbar value={0} />
          </ProgressbarContainer>
        )
      case 'failed':
        return <Icon name="error-second" />
      case 'completed':
        return <Icon name="success" />
      case 'reconnecting':
        return <Icon name="error-second" />
      default:
        return <CircularProgressbar value={asset.progress} />
    }
  }
// ....

   <FileStatusContainer>{getStatus(asset)}</FileStatusContainer>
```
 and then, where the StatusMessage is 
```js
      {asset.status === 'reconnecting' && <StatusMessage variant="subtitle2">reconnecting</StatusMessage>}
```

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.style.tsx
```diff
@@ -0,0 +1,108 @@
...
+  cursor: pointer;
+  transition: all ${transitions.timings.regular} ${transitions.easing};
+`
+export const ProgressBar = styled.div<ProgressbarProps>`
+  --progress-bar-color: #b4bbff33;
+  position: absolute;
+  top: 0;
+  left: 0;
+  height: 100%;
+  width: ${({ progress }) => progress && `${progress}%`};
```
drillprop: is it possible to use `transform: scaleX()` here? Would be better for performance.

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -0,0 +1,76 @@
...
+        </FileInfoType>
+        <Text variant="body2">{resolution}</Text>
+        <Text>{size}</Text>
+      </FileInfoContainer>
+      <StatusMessage variant="subtitle2">{fileStatusMessage}</StatusMessage>
+    </FileLineContainer>
+  )
+}
+
+export default AssetLine
```
drillprop, line 71: I think you could add this also to the storybook

mikkio-j: Not sure about that. I believe `AssetLine` will only live in the context of `AssetsGroupUploadBar`, so I think it would make more sense to add the whole group instead. What do you think?

drillprop: yeah, I agree

drillprop (commented):
Found one more thing. The color of the AssetLine is a little bit different when you expand it.
Figma: 
image
App: 
image

drillprop (approved):

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.style.tsx
```diff
@@ -0,0 +1,116 @@
...
+  height: ${sizes(20)};
+  background-color: ${({ isActive }) => (isActive ? colors.gray[900] : colors.black)};
+  cursor: pointer;
+  transition: background-color ${transitions.timings.sharp} ${transitions.easing};
+  &:hover {
+    background-color: ${colors.gray[900]};
+  }
+`
+export const ProgressBar = styled.div<ProgressbarProps>`
+  --progress-bar-color: #b4bbff33;
```
kdembler: 1. Why is this defined as a CSS variable? It don't think it's reused anywhere?
2. This is a color from our theme, let's use it via `colors`

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+  AssetsGroupUploadBarContainer,
+  ProgressBar,
+  Thumbnail,
+  AssetsInfoContainer,
+  UploadInfoContainer,
+  AssetsDrawerContainer,
+} from './AssetsGroupUploadBar.style'
+import { AssetLine } from '../AssetLine'
+import { Icon, Text } from '@/shared/components'
+import { DrawerButton } from '@/components/Topbar/StudioTopbar/StudioTopbar.style'
```
kdembler: Which button is that? It's the `expand` one? If we need it here, let's elevate it to `shared`, we shouldn't rely on another view

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const [isAssetsDrawerActive, setAssetsDrawerActive] = useState(false)
+
+  const drawer = useRef<HTMLDivElement>(null)
+
+  const isChannelType = type === 'channel'
+  const isPending = files.every((file) => file.status === 'pending')
+  const hasErrorNumber = files.filter(({ status }) => status === 'failed' || status === 'reconnecting').length
+
+  const allAssetsSize = files.reduce((acc, file) => (acc = acc + file.size), 0)
+  const alreadyUploadedSize = files.reduce((acc, file) => (acc = acc + (file.size * file.progress) / 100), 0)
```
kdembler: You don't need to assign to `acc` here, only thing that matters is the return value
```suggestion
  const allAssetsSize = files.reduce((acc, file) => acc + file.size, 0)
  const alreadyUploadedSize = files.reduce((acc, file) => acc + (file.size * file.progress) / 100, 0)
```

kdembler: The brackets here are a bit misleading, if we need them, I think it's more clear if we say `acc + (file.progress / 100) * file.size`

kdembler (changes requested):

src/shared/components/Button/Button.style.ts
```diff
@@ -10,6 +10,7 @@ export type ButtonStyleProps = {
   disabled?: boolean
   full?: boolean
   hasText?: boolean
+  rotateIcon?: boolean
```
kdembler: I don't think this should be a prop on the `Button` itself, since it won't be reused in other places than one specific button we have in mind. What we can do instead, is to just create the `ExpandButton` component that will take care of this

src/components/Topbar/StudioTopbar/StudioTopbar.style.tsx
```diff
@@ -17,12 +17,10 @@ export const StyledTopbarBase = styled(TopbarBase)`
   }
 `
 
-export const DrawerButton = styled(Button)<CommonStudioTopbarProps>`
+export const DrawerButton = styled(Button)`
   svg {
     width: auto;
     height: auto;
```
kdembler: Btw, why is this needed? The `Button` component supports icons, right? So my guess would be we don't need to size them ourselves and they just work out of the box

kdembler (changes requested):

src/components/Topbar/StudioTopbar/StudioTopbar.style.tsx
mikkio-j: It was misaligned without it, but you are right - it should work out of the box. Had to replace chevron icon with the one from the Mystery Box 😃 Fixed now

kdembler (approved):

PR #404, opened 2021-03-26
make routes absolute

kdembler:


drillprop (commented):
Unfortunately, studio routes don't work right now. You cannot go to `videos` or `new channel`. 
I think we should update also the`StudioLayout` but with... relative paths 😅  
like:
```
const studioRoutes = [
  { path: 'channel/new', element: <CreateEditChannelView newChannel /> },
  { path: 'channel', element: <CreateEditChannelView /> },
  { path: 'videos', element: <MyVideosView /> },
]
```
This will be still a little bit messy. Don't know what's the best solution for this.

eldiegod:
Hmm alternatively we could keep all the routes in one file to avoid the relative paths thing (instead of nested routers I mean)

src/config/routes.ts
```diff
@@ -19,17 +25,33 @@ export default {
...
-    index: (absolute?: boolean) => `${absolute ? '/' : ''}playground`,
+    index: () => '',
   },
 }
 
+export const absoluteRoutes = Object.keys(BASE_PATHS).reduce((absoluteRoutesAcc, basePathKeyStr) => {
+  const basePathKey = basePathKeyStr as keyof typeof BASE_PATHS
+  const basePath = BASE_PATHS[basePathKey]
+
+  const routes = relativeRoutes[basePathKey]
```
drillprop:
```suggestion
export const absoluteRoutes = Object.entries(BASE_PATHS).reduce((absoluteRoutesAcc, basePathKeyStr) => {
  const [basePathKey, basePath] = basePathKeyStr

  const routes = relativeRoutes[basePathKey as keyof typeof BASE_PATHS]
```
:P

drillprop (approved):
Works well now!

Issue #406, opened 2021-03-26
Indicate new video drafts


kdembler:
As seen in the designs (https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5863%3A977), when new drafts are saved, we should show a badge indicating that "there are new ones", both in the sidenav and the drafts tab. Let's do that (should be probably persistent across sessions)

PR #408, opened 2021-03-30
Video drafts badge, Sidenav update

mikkio-j:
Issue: #406 
Figma: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4475%3A30](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4475%3A30)


eldiegod (changes requested):
Styles and component work pretty well, however, I'm not sure if we should keep the seen state on the drafts themselves, my feeling is that this should be saved in something like a useState that also saves to localstorage.
Mostly to not add things to the draft/hook that don't have relation to what is going to be uploaded, thoughts?

mikkio-j:
> Styles and component work pretty well, however, I'm not sure if we should keep the seen state on the drafts themselves, my feeling is that this should be saved in something like a useState that also saves to localstorage.
> Mostly to not add things to the draft/hook that don't have relation to what is going to be uploaded, thoughts?

Hmm I understand, I didn't like the idea of adding new field to localstorage just to mark it seen/unseen, besides it's the video draft not video data, I assume  the part of it will be passed to form and then uploaded. I would rather nest it, something like that:
```
type Draft = {
    type: 'video'
    data: {...}
    seen: boolean
}
```

What do you think?

eldiegod:
I see what you mean; what I want to avoid also is adding unneeded complexity to the draft methods like `setUpdatedDate`; I think I would rather keep it in the useDraft hook but separate from the draft themselves as they don't need to know about this as this is just a general flag that can be set apart.

`I didn't like the idea of adding new field to localstorage just to mark it seen/unseen` why is that the case?

mikkio-j:
> I see what you mean; what I want to avoid also is adding unneeded complexity to the draft methods like `setUpdatedDate`; I think I would rather keep it in the useDraft hook but separate from the draft themselves as they don't need to know about this as this is just a general flag that can be set apart.
> 
> `I didn't like the idea of adding new field to localstorage just to mark it seen/unseen` why is that the case?

Okay, that makes sense 😀 Might be a good idea

src/components/Sidenav/Sidenav.style.ts
```diff
@@ -4,6 +4,7 @@ import { breakpoints, colors, sizes, transitions, typography, zIndex } from '@/s
 import isPropValid from '@emotion/is-prop-valid'
 import styled from '@emotion/styled'
 import { Link, LinkProps } from 'react-router-dom'
+import { badgeStyles } from '@/shared/components/Tabs/Tabs.styles'
```
kdembler: If the `badgeStyles` are something we need to use outside of `Tabs`, I think they should be defined somewhere else. If we use styles from some other component, it may not be obvious that by changing styling of `Tabs`, `Sidenav` is also affected

src/components/Sidenav/Sidenav.tsx
```diff
@@ -22,6 +22,7 @@ import FollowedChannels from './FollowedChannels'
 import { usePersonalData } from '@/hooks'
 import HamburgerButton from '@/shared/components/HamburgerButton'
 import { absoluteRoutes } from '@/config/routes'
+import { BadgeType } from '@/shared/components/Tabs/Tabs'
```
kdembler: Same as with styles, let's just move badge stuff to a separate component, even if it isn't a component per se but only styles and types

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -27,9 +38,13 @@ type DraftState = {
   videos: VideoDraft[]
 }
 
+export type DraftsSeenStatusState = { id: string; seen: boolean }[]
```
kdembler: Do we need to keep state of all drafts we've ever seen? I see that list growing indefinitely? Couldn't we just remember the drafts we haven't seen and remove them as soon as we look at them?

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -4,15 +4,20 @@ import React, { useEffect, useRef, useState } from 'react'
...
 
+export type BadgeType = {
+  name: string
+  number: number
+}
 export type TabsProps = {
   tabs: string[]
   initialIndex?: number
   onSelectTab: (idx: number) => void
+  badges?: BadgeType[]
```
kdembler: Does this have to be a separate prop? Because of that, we need to later do finds to correlate badge items with tabs items. Do you see any downsides to doing the following instead?
```ts
type TabItem = {
  name: string
  badge?: number
}

type TabsProps = {
  tabs: TabItem[]
  ...
}
```

The usage would be much simpler I guess

kdembler: Current implementation also allows us to have mismatch - we can update the tab name but forget to update the badge name and it will stop working. We should always try to avoid situations like that

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -64,11 +73,24 @@ export const MyVideosView = () => {
...
   const hasNoVideos = currentTabName === 'All Videos' && totalCount === 0 && drafts.length === 0
+  const unseenDraftsLength = draftsSeenStatusState.filter((draft) => draft.seen === false).length
 
   const handleChangePage = (page: number) => {
     setCurrentPage(page)
   }
 
+  const handleSetCurrentTab = async (tab: number) => {
+    if (TABS[tab] === 'Drafts') {
+      setCurrentTab(tab)
```
kdembler: If it's not `Drafts`, we don't want to update current tab?

kdembler (changes requested):
Also, there currently isn't a way to test it out because of the staging not working atm. Can we update the `testChannelId` in `MyVideosView.tsx` so we can use it with mocked backend?

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -19,6 +29,7 @@ export type VideoDraft = {
   publishedBeforeJoystream?: string
   hasMarketing?: boolean
   isExplicit?: boolean
+  seen?: boolean
```
eldiegod: We don't need this anymore right?

src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const handleNewVideoOpen = () => {
+    // TODO add logic for opening new video view
+  }
+
+  return (
+    <SidenavBase
+      isStudio
+      items={studioNavbarItems}
+      onNewVideoOpenClick={handleNewVideoOpen}
+      unseenDraftsNumber={channelUnseenDrafts.length}
```
kdembler: Makes sense?

```suggestion
  const channelId = activeUser.channelId ?? undefined
  const { unseenDrafts } = useDrafts('video', channelId)

  const handleNewVideoOpen = () => {
    // TODO add logic for opening new video view
  }

  return (
    <SidenavBase
      isStudio
      items={studioNavbarItems}
      onNewVideoOpenClick={handleNewVideoOpen}
      unseenDraftsNumber={channelId ? unseenDrafts.length : 0}
```

src/hooks/useDrafts/useDrafts.tsx
```diff
@@ -27,9 +37,13 @@ type DraftState = {
   videos: VideoDraft[]
 }
 
+export type UnseenDrafts = { draftId: string; channelId: string }[]
```
kdembler: Let's use this as a singular type -`UnseenDraft`. We can do `UnseenDraft[]` when needed

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -97,6 +117,8 @@ export const MyVideosView = () => {
   if (error) {
     throw error
   }
+  const channelUnseenDrafts = unseenDrafts.filter((draft) => draft.channelId === channelId)
```
kdembler: Let's use the same approach as I proposed above, no need to filter twice

kdembler (changes requested):

kdembler (approved):

PR #421, opened 2021-03-31
Connectivity status

### LINKED ISSUE #405: Manage connectivity status
We will need to handle those error states:
- No internet connection
- No connection to the node (but with internet)

First we need to be able to detect those (latter should be rather simple), then we should represent those somehow in the UI. We should probably also block taking any actions (doing any transaction won't work without internet/node).

We will need to get designs for that but the logic part could be potentially done right away.

drillprop:
Fix #405 
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6744%3A354


src/components/NoConnectionIndicator/NoConnectionIndicator.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { transitions } from '@/shared/theme'
+import React from 'react'
+import { CSSTransition } from 'react-transition-group'
+import { IndicatorWrapper } from './NoConnectionIndicator.style'
+
+type NoConnectionIndicatorProps = {
+  connectionStatus: ConnectionStatus
+}
+
+const NoConnectionIndicator: React.FC<NoConnectionIndicatorProps> = ({ connectionStatus }) => {
```
mikkio-j: so the designs are ready:
[https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6744%3A354](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6744%3A354)

Might be a good idea to update the PR with it

src/components/NoConnectionIndicator/NoConnectionIndicator.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <CSSTransition
+      in={connectionStatus === 'no-internet' || connectionStatus === 'disconnected'}
+      timeout={200}
+      classNames={transitions.names.fade}
+      mountOnEnter
+      unmountOnExit
+    >
+      <IndicatorWrapper>
+        <Spinner size="small" />
```
mikkio-j, line 26: I believe Spinner will need some refactor to cover different color variations. The grey part in designs is lighter

mikkio-j (approved):
Just one thing regarding the spinner 👍

src/hooks/useConnectionStatus.tsx
```diff
@@ -0,0 +1,58 @@
...
+  setNodeConnection: (connection: ConnectionStatus) => void
+  connectionStatus: ConnectionStatus
+}
+
+export const ConnectionStatusProvider: React.FC = ({ children }) => {
+  const [connectionStatus, setNodeConnection] = useState<ConnectionStatus>('connecting')
+
+  useEffect(() => {
+    const connectionHandler = () => {
+      setNodeConnection(window.navigator.onLine ? 'connecting' : 'no-internet')
```
kdembler: Do you think `navgiator.onLine` is a good fit for our use case? From the docs:
> In Chrome and Safari, if the browser is not able to connect to a local area network (LAN) or a router, it is offline; all other conditions return true. So while you can assume that the browser is offline when it returns a false value, you cannot assume that a true value necessarily means that the browser can access the internet. You could be getting false positives, such as in cases where the computer is running a virtualization software that has virtual ethernet adapters that are always "connected." Therefore, if you really want to determine the online status of the browser, you should develop additional means for checking.

src/hooks/useConnectionStatus.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react'
+
+export type ConnectionStatus = 'connected' | 'disconnected' | 'connecting' | 'no-internet'
+
+const ConnectionStatusContext = React.createContext<undefined | ConnectionStatusValue>(undefined)
+ConnectionStatusContext.displayName = 'ConnectionStatusContext'
+
+type ConnectionStatusValue = {
+  setNodeConnection: (connection: ConnectionStatus) => void
+  connectionStatus: ConnectionStatus
```
kdembler: I'd expect a separate status here for the internet/node connection. Could you walk me through how the status is determined here?

drillprop: 1. First, the hook will check(and listen) if a user is connected to the internet. 
2. If a user is connected. it will set `connectionStatus` to `connecting`- which means basically `connecting to node`. If everything is fine with the node, it will set the state to `connected`, if not -`disconnected`
3. If a user doesn't have an internet connection it will set `connectionStatus` to `no-internet`. Now it will wait for reconnecting. If the user reconnects it will go to point `2` - set status to connecting and then it will check status of node

I thought it makes sense to keep this that way. If you cannot connect to the internet, it's impossible to determine if everything works on the node side, right?

kdembler: Yeah, I guess that mostly makes sense. Although I'm a bit worried on some details. For example, I guess it may be possible for the internet connection to go away for some short amount, like 1sec. So `offline`, `online` would get triggered. But I guess it's also possible for the node connection to not timeout in that time. Resulting in the end status of `connecting` and node not reporting getting connected since it think it always was. I guess we may have a better mental picture on this once we figure out the point above about `navigator.onLine`

drillprop: okay, I changed the approach. Now there are two separate statuses for the internet and for the node connection. 
If `navigator.online` will be true, the hook will be pinging google every three seconds to check if the user has a stable internet connection. 
If `navigator.online` will change to false, it will wait for three seconds, and then it will set the user as disconnected from the internet.

src/hooks/useConnectionStatus.tsx
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export const ConnectionStatusProvider: React.FC = ({ children }) => {
+  const [nodeConnectionStatus, setNodeConnection] = useState<ConnectionStatus>('connecting')
+  const [navigatorOnline, setNavigatorOnline] = useState(true)
+  const [isUserConnectedToInternet, setIsUserConnectedToInternet] = useState(true)
+
+  useEffect(() => {
+    let timeout: NodeJS.Timeout
+    let interval: NodeJS.Timeout
```
kdembler: Are those used inside this hook?

src/hooks/useConnectionStatus.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (navigatorOnline) {
+      return
+    }
+    const timeout = setTimeout(() => {
+      setIsUserConnectedToInternet(false)
+    }, 3000)
+    return () => {
+      clearInterval(timeout)
+    }
+  }, [navigatorOnline])
```
kdembler: Could you explain this approach? Do we need any additional logic apart from adding `setIsUserConnectedToInternet(false)` at line 39?

src/hooks/useJoystream/provider.tsx
```diff
@@ -39,10 +41,12 @@ export const JoystreamProvider: React.FC = ({ children }) => {
         joystream = await JoystreamJs.build(APP_NAME, NODE_URL)
         setJoystream(joystream)
         setAccounts(joystream.accounts)
+        setNodeConnection('connected')
 
         joystream.onAccountsUpdate = handleAccountsUpdate
         joystream.onExtensionConnectedUpdate = handleExtensionConnectedUpdate
       } catch (e) {
+        setNodeConnection('disconnected')
```
kdembler, line 50:
```suggestion
        // TODO: listen to connection events to handle disconnect/reconnect
        setNodeConnection('disconnected')
```

src/shared/components/Spinner/Spinner.style.ts
```diff
@@ -33,12 +33,11 @@ export const SpinnerWrapper = styled.div<SpinnerWrapperProps>`
...
-  border: 0.09em solid ${colors.gray[500]};
-  border-left: 0.09em solid ${colors.blue[500]};
-  margin-bottom: ${sizes(4)};
-  animation: load 1s infinite linear;
+  border: 0.09em solid ${colors.blue[500]};
+  border-left: 0.09em solid ${colors.gray[200]};
+  animation: spinnerAnimation 1s infinite linear;
   ${({ size }) => spinnerSizes(size)};
-  @keyframes load {
+  @keyframes spinnerAnimation {
```
kdembler: This likely needs rebase

kdembler (changes requested):

src/hooks/useConnectionStatus.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  useEffect(() => {
+    // ping google every three seconds to check if user is connected to internet
+    const interval = setInterval(() => {
+      checkConnection()
+    }, 3000)
+    return () => {
+      clearInterval(interval)
+    }
+  }, [])
```
kdembler, line 31: It's amazing that ESLint doesn't complain about `checkConnection` not missing from here. I think it can tell that `checkConnection` doesn't contain any dynamic references so it's safe to just use the stale reference. Magic

kdembler (approved):

PR #423, opened 2021-03-31
fix typechecking

### LINKED ISSUE #409: TS doesn't show errors inmediatly
i.e. scenario, changing the props/API of a component that's used in different places through Atlas won't cause TS to show errors on the console for these.

Expected behavior:
TS shows in what file and what the type errors are upon changing a components API and breaking it somewhere else where it's being used.

kdembler:
Fixes #409 

Turns our `tsc` was failing because of version mismatch - we were using `@3` and some libraries required `>4`. I have no idea why it failed silently though as if everything was all right instead of reporting some error. Really weird

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "react-scripts": "4.0.1",
     "react-transition-group": "^4.4.1",
     "ts-loader": "^6.2.1",
-    "typescript": "^3.8.3",
+    "typescript": "^4.2.3",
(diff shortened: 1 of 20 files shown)


drillprop (approved):
Looks good, much-needed fix 🙏🏻

eldiegod (approved):
Let's get this one merged boys 🙏

PR #424, opened 2021-04-01
Update textarea and textfield

### LINKED ISSUE #412: Update InputBase and TextArea
Inputs got an update and some consistency improvement. We should update our components to reflect this - Figma -> Components -> Text Field, Select, Text Area

Currently, as I understand it, both `TextField` and `Select` use `InputBase` for common styling. If I'm not wrong, only update to the `InputBase` should be necessary

#### Requirements:
- `TextField` and `Select` updated with latest designs
  - Those also have variants with icons in the design - I think we can skip that for the time being since it's not used in any way in the current views
  - `Select` has variants with typing inside in the design - we should ignore that (that would be something like `Autocomplete`) and just focus on `Select` itself
- `TextArea` brought in line with the other inputs - I suppose it should use `InputBase` after update
  - There's an icon included for resizing in the design. I'm not sure if that's possible to do, the drag to resize is a native browser functionality after all. I suggest to look into it and do if it's easy and ignore otherwise 
- Let's also update `HelperText` (possibly worth to elevate to its own component) to use `Text` component with `caption` styling

drillprop:
Fix #412
Figma: 
https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6604%3A530
https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6565%3A75


drillprop:
image
From the top: firefox, safari, chrome
I've added a custom resizer, but it's not ideal. There is no space around him and it doesn't work on firefox.

src/shared/components/InputBase/InputBase.styles.ts
```diff
@@ -1,101 +1,67 @@
...
+  input:-webkit-autofill:hover,
+  input:-webkit-autofill:focus,
+  textarea:-webkit-autofill,
+  textarea:-webkit-autofill:hover,
+  textarea:-webkit-autofill:focus {
+    // This is madness, but basically, it will make autofill transparent for a long time.
+    // Changing background-color and color won't work
+    transition: color 99999s, background-color 99999s;
+    transition-delay: 99999s;
+  }
```
kdembler: This looks error-prone. Do we really need this, are the autofill colors that bad?

src/shared/components/HelperText/HelperText.tsx
```diff
@@ -0,0 +1,61 @@
...
+    }
+    if (charactersCount > maxLength) {
+      setCharactersWarning('error')
+    }
+  }, [charactersCount, maxLength])
+
+  return (
+    <HelperTextsWrapper>
+      {/* caption cannot be child of <div> */}
+      <StyledHelperText variant="caption" as="span" helperTextVariant={getVariant(warning, error)}>
```
kdembler, line 49: I think the correct solution here is to not use `<caption>` as an HTML element when displaying the `caption` style. It should only be used within `<table>` so choosing it was probably a mistake on my part. Let's just make `<Text variant="caption">` render as `span` be default

src/shared/components/InputBase/InputBase.styles.ts
```diff
@@ -1,101 +1,67 @@
...
 import styled from '@emotion/styled'
 
 type FormGroupProps = {
   error?: boolean
   disabled?: boolean
 }
 
-export const FormGroup = styled.div<FormGroupProps>`
-  position: relative;
+export const FormGroup = styled.label<FormGroupProps>`
```
kdembler, line 9: What's the rationale for using `label` as a container vs just for the label text itself?

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -1,67 +1,14 @@
-import { colors, sizes, transitions, typography } from '@/shared/theme'
 import styled from '@emotion/styled'
-
-export const TextAreaWrapper = styled.div`
-  width: 100%;
-`
+import resizer from '@/shared/icons/resizer.svg'
```
kdembler: This should be imported from `@/shared/icons`

src/shared/components/InputBase/InputBase.styles.ts
```diff
@@ -1,101 +1,67 @@
...
+  input:-webkit-autofill::first-line,
+  textarea:-webkit-autofill::first-line {
+    font-size: 1rem;
+  }
+
   input,
-  button {
+  button,
+  textarea {
     transition: all ${transitions.timings.regular} ${transitions.easing};
```
kdembler: Because of this, resizing the textarea feels lagged. Let's explicitly specify properties that should transition here

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -1,67 +1,14 @@
...
-  ::placeholder {
-    color: ${colors.gray[400]};
-    transition: color ${transitions.timings.loading} ${transitions.easing};
-  }
-  :focus {
-    border-bottom: 1px solid ${colors.blue[500]};
+  display: block;
+  position: relative;
+  :disabled {
+    cursor: not-allowed;
```
kdembler: I guess this could be moved to the `InputBase`?

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -1,67 +1,14 @@
...
   }
-  :placeholder-shown {
-    border-bottom: 1px solid ${colors.gray[200]};
+  ::-webkit-resizer {
+    background: url(${resizer}) no-repeat;
```
kdembler, line 8: Wasn't aware you can do that, looks cool!

However, on Firefox it looks bad. It looked bad even before so it's not on your changes but I think we could improve that.

Our resize:
Screenshot 2021-04-06 at 13 49 12

Default Firefox resize:
Screenshot 2021-04-06 at 13 49 20

Could you look into why that happens and if there's something we could do to improve?

src/shared/components/TextArea/TextArea.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { forwardRef, useState } from 'react'
-import { getVariant } from '../InputBase'
-import { HelperText, HelperTextCount, HelperTextsWrapper, StyledTextArea, TextAreaWrapper } from './TextArea.style'
+import InputBase, { InputBaseProps } from '../InputBase'
+import { StyledTextArea } from './TextArea.style'
 
 export type TextAreaProps = {
```
kdembler, line 5: Some of props specified here are already defined in `InputBaseProps`, let's try not to duplicate those here

src/shared/components/TextArea/TextArea.tsx
```diff
@@ -11,74 +11,58 @@ export type TextAreaProps = {
...
   return (
-    <TextAreaWrapper className={className}>
+    <InputBase
+      label={label}
+      warning={warning}
+      disabled={disabled}
+      error={error}
+      helperText={helperText}
+      charactersCount={charactersCount}
+      maxLength={maxLength}
```
kdembler: If a prop is not needed inside this component, let's just destructure it with `...inputBaseProps` and pass them here in bulk, similar to `TextField`

kdembler (changes requested):

src/shared/components/InputBase/InputBase.styles.ts
drillprop: If you use the `label` as a wrapper for `input`, you don't have to add `htmlFor/for` attribute to `label` and `id` to `input`.

kdembler: Oh, cool, I wasn't aware :+1:

src/shared/components/InputBase/InputBase.styles.ts
drillprop: Maybe it's not that bad after all. 
image

I'll remove this.

src/shared/components/TextArea/TextArea.style.ts
drillprop: I cannot import SVG like that. Also, maybe I shouldn't put this in the `icons` directory - I don't use it as a `ReactComponent` in the first place? I'll put it in the `assets` folder. What do you think?

src/shared/components/TextArea/TextArea.style.ts
drillprop: Not sure if I follow here. On firefox, right now we are using a default resizer. You cannot do anything with it. It's impossible to change it and it's impossible to hide it. 
To do something with it I would have to disable resizing and implement resizing from the scratch, I think.
On other browsers, I've added a custom resizer using `::-webkit-resizer`,  but as you see, it's not ideal. I cannot change its position and size.

kdembler: Yes, we're using a default one, but for some reason it doesn't look as good as resize I've posted which is also a default one, just from a clean up. My guess is that some of our CSS may be causing it to look the way it does, be it margin, padding or something else. I'm asking if you can see what is the reason

kdembler: Not much we can do it seems, that's on Firefox

src/shared/components/TextArea/TextArea.style.ts
kdembler: ah I see, let's leave it out then

src/shared/components/TextArea/TextArea.style.ts
```diff
@@ -1,67 +1,11 @@
...
-  }
-  :hover {
-    :not(:focus) {
-      border-bottom: 1px solid ${colors.gray[500]};
-      ::placeholder {
-        color: ${colors.gray[500]};
-      }
-    }
+  display: block;
+  position: relative;
```
kdembler:
```suggestion
  position: relative;
  resize: vertical;
```

src/shared/components/TextArea/TextArea.tsx
```diff
@@ -1,84 +1,42 @@
...
-    }
+    setCharactersCount(e.target.value.length)
+    onChange?.(e)
   }
 
   return (
-    <TextAreaWrapper className={className}>
+    <InputBase disabled={disabled} charactersCount={charactersCount} {...inputBaseProps}>
       <StyledTextArea
+        className={className}
```
kdembler: Here we pass the `className` to the text area, but in the `TextField`, we pass it to the `InputBase`, we should be consistent with that. I think `InputBase` makes more sense but I'm open to discussion

kdembler (changes requested):

kdembler (approved):

Issue #410, opened 2021-03-30
update snackbars


kdembler:
New designs provide some new snackbar experience so we'll need to refresh what we have.

#### Requirements:
- API should stay the same or similar to what we have currently
- Multiple snackbars can be on screen at the same time
  - We probably should specify some max amount
- Support for `primary` and `secondary` variants
- Support for secondary text
- Support for action button
- Sliding animation as seen in codepen in figma (sliding up from bottom of the screen, on dismiss sliding left, under the sidenav)
- Preferably using CSS transitions, if hard, we can pull in some animation library

PR #426, opened 2021-04-02
Snackbar update

mikkio-j:
Issue: #410 
Figma: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6401%3A14](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6401%3A14)


src/shared/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -19,23 +29,15 @@ export default {
   ],
 } as Meta
 
-const Template: Story<SnackbarProps> = (args) => <Snackbar {...args} />
-
-const ClickableTemplate: Story<DisplaySnackbarArgs> = ({ time, message, variant, buttonText }) => {
+const ClickableTemplate: Story<DisplaySnackbarArgs> = ({ ...args }) => {
   const { displaySnackbar } = useSnackbar()
   return (
     <>
```
drillprop: you can remove this fragment

src/shared/components/Snackbar/Snackbar.tsx
```diff
@@ -1,33 +1,67 @@
...
+  const [height, setHeight] = useState(0)
+
+  useEffect(() => {
+    if (ref.current) {
+      setHeight(ref.current.clientHeight)
+    }
+  }, [])
+
   return (
     <>
```
drillprop: and this one too 😄

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -1,72 +1,128 @@
...
+export const Submessage = styled(Text)`
+  margin-top: 10px;
+  margin-bottom: 16px;
+`
+
+export const StyledIcon = styled(Icon)`
+  margin-left: 2px;
+  margin-right: ${sizes(2)};
+  width: ${sizes(6)};
+  height: ${sizes(6)};
```
drillprop:
```suggestion
  width: ${sizes(5)};
  height: ${sizes(5)};
```
Icon should be smaller

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -1,72 +1,128 @@
...
+    transform: translateY(0) translateX(0);
+    height: ${({ snackbarHeight }) => snackbarHeight && snackbarHeight}px;
+    margin-bottom: ${sizes(3)};
   }
-  .snackbar-exit-active {
-    transform: translateX(-200%);
+  &.snackbar-exit {
+    transform: translateY(0) translateX(0);
+    height: ${({ snackbarHeight }) => snackbarHeight && snackbarHeight}px;
+    margin-bottom: ${sizes(3)};
```
drillprop, line 41: lot of repeated code - you could just do: 
```
&.snackbar-enter-done, 
&.snackbar-enter-active,
....
```

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -1,72 +1,128 @@
...
+  &.snackbar-enter-active {
+    transition: height ${transitions.timings.regular} ${transitions.easing},
+      margin-bottom ${transitions.timings.regular} ${transitions.easing},
+      transform ${transitions.timings.regular} ${transitions.easing} ${transitions.timings.regular};
+  }
+  &.snackbar-exit-active {
+    transition: transform ${transitions.timings.regular} ${transitions.easing},
+      height ${transitions.timings.regular} ${transitions.easing} ${transitions.timings.regular},
+      margin-bottom ${transitions.timings.regular} ${transitions.easing} ${transitions.timings.regular};
+  }
```
drillprop:
```suggestion
  &.snackbar-enter-active,
  &.snackbar-exit-active {
    transition: height ${transitions.timings.regular} ${transitions.easing},
      margin-bottom ${transitions.timings.regular} ${transitions.easing},
      transform ${transitions.timings.regular} ${transitions.easing} ${transitions.timings.regular};
  }
```

mikkio-j: that cannot be really applied. These transitions are not the same. In `enter` there's delay on `transform` and in `exit` there's delay on `margin-bottom` and `height`

drillprop: My mistake, didn't notice

drillprop (approved):
👍🏻

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,98 @@
-import { snackbarTransitions } from '@/shared/components/Snackbar'
+import React, { createContext, useContext, useState, useRef, useEffect } from 'react'
+import { CSSTransition, TransitionGroup } from 'react-transition-group'
+import styled from '@emotion/styled'
+import faker from 'faker'
 import Snackbar from '@/shared/components/Snackbar/Snackbar'
```
kdembler: This should be imported from `@/shared/components`

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,98 @@
...
-import { transitions } from '@/shared/theme'
-import { Global } from '@emotion/react'
-import React, { createContext, useContext, useEffect, useState } from 'react'
-import { CSSTransition } from 'react-transition-group'
+import { transitions, sizes } from '@/shared/theme'
 
 export type DisplaySnackbarArgs = {
   time?: number
```
kdembler: Let's call this `timeout` so it's more clear

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,98 @@
...
-  const [isVisible, setIsVisible] = useState(false)
-  const [snackbarOpts, setSnackbarOpts] = useState<DisplaySnackbarArgs | null>(null)
+  const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
+
+  // Snackbars state reference is needed for setTimeout to read
+  const snackbarsRef = useRef(snackbars)
+  snackbarsRef.current = snackbars
+
+  const displaySnackbar = ({ time, ...args }: DisplaySnackbarArgs) => {
+    const id = faker.random.uuid()
```
kdembler: `faker` has a _huge_ bundle size, we shouldn't use it in our runtime. Let's use some alternate method to generate an ID, like number of milliseconds or something. Matter of fact, it'd be best if we put that in `utils` since we already generate IDs in couple places so we can reuse that logic. `hooks/useDrafts/utils` already uses dome ID generation, we can use that as a start

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,98 @@
...
+const SNACKBARS_LIMIT = 3
+
 export const SnackbarProvider: React.FC = ({ children }) => {
-  const [isVisible, setIsVisible] = useState(false)
-  const [snackbarOpts, setSnackbarOpts] = useState<DisplaySnackbarArgs | null>(null)
+  const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
+
+  // Snackbars state reference is needed for setTimeout to read
+  const snackbarsRef = useRef(snackbars)
+  snackbarsRef.current = snackbars
```
kdembler: Could you explain rationale behind it?

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,98 @@
...
+                message={item.message}
+                subMessage={item.subMessage}
+                variant={item.variant}
+                actionText={item.actionText}
+                onActionClick={item.onActionClick}
+                icon={item.icon}
+                onClick={() => closeSnackbar(item.id)}
+              />
+            </CSSTransition>
+          ))}
```
kdembler:
```suggestion
          {snackbars.map(({ id, ...snackbarProps }) => (
            <CSSTransition key={id} timeout={2 * parseInt(transitions.timings.regular)} classNames={'snackbar'}>
              <Snackbar
                {...snackbarProps}
                onClick={() => closeSnackbar(id)}
              />
            </CSSTransition>
          ))}
```

yarn.lock
```diff
@@ -19967,4 +19967,4 @@ zen-observable@^0.8.0, zen-observable@^0.8.14:
 zwitch@^1.0.0:
   version "1.0.5"
   resolved "https://registry.yarnpkg.com/zwitch/-/zwitch-1.0.5.tgz#d11d7381ffed16b742f6af7b3f223d5cd9fe9920"
-  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
+  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
```
kdembler: Why did this newline got deleted?

kdembler (changes requested):

src/hooks/useSnackbar/useSnackbar.tsx
mikkio-j: @kdembler I had an issue with using the state directly in setTimeout, so had to put it in container - that solution is described in details here:
[https://github.com/facebook/react/issues/14010#issuecomment-433788147](https://github.com/facebook/react/issues/14010#issuecomment-433788147)

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,89 @@
...
-  const [isVisible, setIsVisible] = useState(false)
-  const [snackbarOpts, setSnackbarOpts] = useState<DisplaySnackbarArgs | null>(null)
+  const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
+
+  // Snackbars state reference is needed for setTimeout to read
+  const snackbarsRef = useRef(snackbars)
+  snackbarsRef.current = snackbars
+
+  const displaySnackbar = ({ timeout, ...args }: DisplaySnackbarArgs) => {
+    const id = Math.random().toString(36).substr(2, 11)
```
kdembler: Like I suggested in the comment, let's elevate this into `@/utils`

src/hooks/useSnackbar/useSnackbar.tsx
kdembler: Okay, so I may have been more explicit. I actually do understand the problem and I also think there's a better solution to it than this one. I was hoping you could try to explain the problem so we may arrive to the conclusion. Do you understand why this doesn't work with just using `snackbars` directly and why the ref helps us with this problem? I want to make sure you properly understand what happens before we get to the solution

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,89 @@
...
 
 export type DisplaySnackbarArgs = {
-  time?: number
-  variant?: 'success' | 'error' | 'info'
-  buttonText?: string
+  timeout?: number
+  variant?: 'primary' | 'secondary'
+  icon?: 'success' | 'error' | 'info'
   message: string
+  subMessage?: string
```
kdembler: Sorry for not bringing this up earlier, I think we can improve the naming of those for more clarity. How about `title` for the main one and `message`/`description`/`caption`?

src/shared/components/Snackbar/Snackbar.tsx
```diff
@@ -1,33 +1,65 @@
...
+          {actionText && (
+            <SnackbarAction variant="tertiary" onClick={onActionClick}>
+              {actionText}
+            </SnackbarAction>
+          )}
         </SnackbarParagraph>
-        <SnackbarButton onClick={onClick}>{buttonText}</SnackbarButton>
-      </SnackbarWrapper>
-    </>
+        <SnackbarButton onClick={onClick} icon="close" variant="tertiary" size="small" />
```
kdembler: This doesn't seem to properly follow layout in Figma. Is there some good reason to position this button absolutely vs just including it in the layout?

Consider what's happening now:

- The header has fixed height of 32px
- The message has a margin of 10px which isn't multiple of our spacing units
- The message can overlap with the button

If you stick the close button and the title into the container, all of those problems go away. This is also how this is structured in Figma. Design should always use multiples of `4` as margins and paddings. If we use `10px` as margin then this most likely points to a mistake in our implementation

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -1,72 +1,124 @@
...
+
+export const StyledIcon = styled(Icon)`
+  margin-left: 2px;
+  margin-right: ${sizes(2)};
+  width: ${sizes(5)};
+  height: ${sizes(5)};
+`
+
+export const StyledInnerWrapper = styled.div<InnerWrapperProps>`
+  padding: ${sizes(3)} ${sizes(5)};
```
kdembler: This is constant here but in Figma it's different for snackbars without the message and those with it.

kdembler (changes requested):

src/hooks/useSnackbar/useSnackbar.tsx
mikkio-j: As I understand it, the problem is that setTimeout uses `snackbars` from the moment setTimeout is triggered, so the setState function inside operates on `snackbars` state that comes from the moment of triggering and not current state (after timeout). It is not a problem as long as nothing's changed after trigger (or state will change inside its scope) - but in our case it will happen. New snackbars will appear in between. Is that correct?

kdembler: Yep, sounds right! When we create a function that we pass as an argument to `setTimeout`, we pass `snackbars` reference to it. Since all of our data is immutable (when we do modification, we don't modify an existing array but rather replace the reference to it with a new one) this will not get updated if there's a change before the timeout executes.

Okay, so now that the problem is clear, let's get to the solution. When you have some setter from `useState`, you can pass a value to it (a new array), but you can also pass a function. This function will get executed with the current value of that state, and I guess it's available for purposes just like this one, when you need to update the state based on current value. We can use this and get rid of the ref:
```ts
setTimeout(() => {
  setSnackbars((currentSnackbars) => currentSnackbars.filter((snackbar) => snackbar.id !== id))
}, timeout)
```

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -1,76 +1,86 @@
...
 }
 
+type SnackbarsState = {
+  id: string
+} & Omit<DisplaySnackbarArgs, 'time'>
+
 type SnackbarContextValue = {
   displaySnackbar: (args: DisplaySnackbarArgs) => void
-  closeSnackbar: () => void
+  closeSnackbar: (id: string) => void
```
kdembler, line 24: There actually isn't any way for the consumer to know this ID, right?

kdembler: Let's return it from `displaySnackbar`

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -1,72 +1,111 @@
...
+    margin-top: calc(-1 * ${sizes(1)});
+  }
+  &:hover {
+    background-color: ${colors.transparent};
   }
 `
+
+export const SnackbarDescription = styled(Text)`
+  margin-top: ${sizes(1)};
+  margin-bottom: ${sizes(4)};
```
kdembler: This needs to be conditional on presence of the button

src/shared/components/Snackbar/Snackbar.style.ts
```diff
@@ -1,72 +1,111 @@
...
+  span {
+    margin-top: calc(-1 * ${sizes(1)});
+  }
+  &:hover {
+    background-color: ${colors.transparent};
   }
 `
+
+export const SnackbarDescription = styled(Text)`
+  margin-top: ${sizes(1)};
```
kdembler: Seem there's some weird style applied to this in Figma. Let's use the following here:
`padding: ${sizes(1)} 0;` instead of `margin-top`. This should bring us close to height specified in Figma. Then if button is present as well we can use `margin-bottom: ${sizes(3)};`

src/hooks/useSnackbar/useSnackbar.tsx
mikkio-j: Done

src/shared/components/Snackbar/Snackbar.style.ts
mikkio-j: have to add color here after rebase

kdembler (approved):

PR #427, opened 2021-04-02
channel integration base, general improvements

kdembler:
#### Summary of changes:

- (**Schema**) updated the language to use `iso` field
- (**Active user context**) Changed the value so it's never null (individual fields can be null)
- (**ActionBar**) Updated with the design
- (**ChannelCover**) Updated with the design, changed to use the button component
- (**Select**) Changed to generic component so any type can be used for `value`
- (**Select**) Changed so it uses (`value`/`onChange`) just the selected value, not the full item. From consumer perspective, it's not needed to keep the label in state, only the value is important
- (**Button**) Changed the default button type to `button` so any button won't submit form by default
- (**EditChannelView**) Various improvements to state handling, ability to reset the form


src/shared/components/Select/Select.tsx
```diff
@@ -1,42 +1,57 @@
-import React, { forwardRef } from 'react'
+import React from 'react'
```
kdembler, line 1: @drillprop Please take a look to make sure I didn't break anything

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -46,220 +53,240 @@ type CreateEditChannelViewProps = {
...
+      language,
+    } = channel
+
+    const avatarAssetUrl = createUrlFromAsset(avatarPhotoAvailability, avatarPhotoUrl, avatarPhotoDataObject)
+    const coverAssetUrl = createUrlFromAsset(coverPhotoAvailability, coverPhotoUrl, coverPhotoDataObject)
+
+    const foundLanguage = languages.find(({ value }) => value === language?.iso)
+
     reset({
-      ...getValues(),
```
kdembler, line 87: @mikkio-j Was this needed for some specific case? It seems to work fine without it

mikkio-j: @kdembler shouldn't be needed anymore, I added that before `description`, `isPublic`, `selectedLanguage` fields were introduced (its purpose was to include all fields in reset - if the field was not provided then it would get that from `defaultValues`)

kdembler: I think it should work the same way, since missing values should get populated with defaults anyway. Let's see then, if it breaks something we'll reintroduce it

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -46,220 +53,240 @@ type CreateEditChannelViewProps = {
...
   })
-  const handleAvatarConfirm = (blob: Blob, url: string) => {
-    setValue('avatar', { url, blob })
-    setAvatarImageUrl(url)
-    setActionBarActive(true)
-  }
-  const handleCoverConfirm = (blob: Blob, url: string) => {
-    setValue('cover', { url, blob })
-    setCoverImageUrl(url)
-    setActionBarActive(true)
```
kdembler, line 114: @mikkio-j FYI: I managed to get rid of this by sticking the dialog inside the controller, but if we went this route you can also do `setValue('cover', { url, blob }, { shouldDirty: true })` and this will properly mark the form as dirty so there wouldn't be a need for separate state like `actionBarActive`

src/shared/components/Select/Select.tsx
drillprop: Tested on playgrounds with changed values and on the storybook. Looks good for me. I like this change a lot 👍🏻

src/components/Dialogs/TransactionDialog/TransactionDialog.tsx
```diff
@@ -2,12 +2,25 @@ import React from 'react'
...
+import { ExtrinsicStatus } from '@/joystream-lib'
 
-export type TransactionDialogProps = ActionDialogProps
+export type TransactionDialogProps = Omit<
+  ActionDialogProps,
+  'additionalActionsNode' | 'primaryButtonText' | 'onPrimaryButtonClick' | 'secondaryButtonText' | 'showDialog'
+> & {
+  onCancel?: () => void
+  status: ExtrinsicStatus | null
+}
```
drillprop:
```suggestion
export type TransactionDialogProps = Pick<ActionDialogProps, 'warning' | 'error'> & {
  onCancel?: () => void
  status: ExtrinsicStatus | null
}
```
I think it should be the same, and is much more readable

src/shared/components/Button/Button.tsx
```diff
@@ -13,6 +13,7 @@ export type ButtonProps = {
   className?: string
   onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void
   to?: To
+  type?: 'button' | 'submit'
```
drillprop, line 16: it would be nice to add `cursor: pointer` to all buttons with `type=submit`(because we will not use `onClick` in forms)

drillprop (approved):
Two small things

src/components/Dialogs/TransactionDialog/TransactionDialog.tsx
kdembler: `Pick` may be a better choice, but we may need some other props. Will take a look

src/shared/components/Button/Button.tsx
kdembler: Good point

src/components/Dialogs/TransactionDialog/TransactionDialog.tsx
kdembler: Done

src/shared/components/Button/Button.tsx
kdembler: Done

mikkio-j (approved):
Great improvements right there 🎉

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -46,220 +53,240 @@ type CreateEditChannelViewProps = {
...
+        <Header>
+          <Controller
+            name="cover"
+            control={control}
+            render={({ value, onChange }) => (
+              <>
+                <ChannelCover
+                  coverPhotoUrl={loading ? null : value.url}
+                  onCoverEditClick={(e) => {
+                    e.preventDefault()
```
mikkio-j: maybe we don't need that anymore since you've changed button's type?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -46,220 +53,240 @@ type CreateEditChannelViewProps = {
...
-                  onChange={(e) => onChange(e.selectedItem)}
-                  error={!!errors.selectedLanguage && !value}
-                  helperText={(errors.selectedLanguage as FieldError)?.message}
-                />
+                <>
+                  <StyledAvatar
+                    imageUrl={value.url}
+                    size="fill"
+                    onEditClick={(e) => {
+                      e.preventDefault()
```
mikkio-j: same as above

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
kdembler: yes, those weren't needed, removed

PR #429, opened 2021-04-02
Pagination update

### LINKED ISSUE #411: update the pagination component
Some changes have been made to the pagination component as seen in Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6530%3A2477

Should be pretty self-explanatory

drillprop:
Fix #411 
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6530%3A2477

### DIFF
--- a/src/shared/components/Pagination/Pagination.style.tsx
+++ b/src/shared/components/Pagination/Pagination.style.tsx
@@ -24,7 +24,6 @@ const buttonActiveState = css`
 `
 
 export const ChevronButton = styled(Button)`
-  color: ${colors.gray[300]};
   width: ${sizes(12)};
   height: ${sizes(12)};
   @media screen and (min-width: ${breakpoints.small}) {
@@ -38,7 +37,7 @@ export const ChevronButton = styled(Button)`
 `
 
 export const PaginationButton = styled.button<PaginationButtonProps>`
-  /* show only one button on mobile */
+  /* show only one active button on mobile */
   display: ${({ isActive }) => (isActive ? 'flex' : 'none')};
   border: none;
   cursor: pointer;
@@ -58,7 +57,7 @@ export const PaginationButton = styled.button<PaginationButtonProps>`
   :hover,
   :focus,
   :active {
-    color: ${colors.white};
+    color: ${colors.gray[50]};
     background-color: ${colors.gray[600]};
   }
   ${({ isActive }) => isActive && buttonActiveState}
--- a/src/shared/components/Pagination/Pagination.style.tsx
+++ b/src/shared/components/Pagination/Pagination.style.tsx
@@ -18,14 +18,7 @@ export const PaginationWrapper = styled.div`
   }
 `
 
-const buttonActiveState = css`
-  color: ${colors.white};
-  background-color: ${colors.gray[800]};
-`
-
 export const ChevronButton = styled(Button)`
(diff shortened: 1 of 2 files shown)


src/shared/components/Pagination/Pagination.tsx
```diff
@@ -27,33 +26,39 @@ const Pagination: React.FC<PaginationProps> = ({
...
+      onChangePage(idx === 1 ? internalPage - 2 : internalPage)
+    }
+  }
+
   return (
     <PaginationWrapper>
-      <PaginationButton
-        isHidden={internalPage <= 1}
-        isChevron={true}
+      <ChevronButton
```
mikkio-j, line 37: in designs icon has `gray[50]` color when `:active`. Or maybe it has that styles only on the right side at all times. Not sure about that.

drillprop: set default colors of the `<Button />`. Should be better now

mikkio-j (approved):

src/shared/components/Pagination/Pagination.style.tsx
```diff
@@ -1,50 +1,67 @@
...
 `
 
 const buttonActiveState = css`
   color: ${colors.white};
   background-color: ${colors.gray[800]};
 `
 
+export const ChevronButton = styled(Button)`
+  width: ${sizes(12)};
+  height: ${sizes(12)};
```
kdembler: I believe those should already be specified by the button component?

src/shared/components/Pagination/Pagination.style.tsx
```diff
@@ -1,50 +1,67 @@
...
+    &:first-of-type {
+      margin-right: ${sizes(8)};
+    }
+    &:last-of-type {
+      margin-left: ${sizes(8)};
+    }
+  }
+`
+
 export const PaginationButton = styled.button<PaginationButtonProps>`
```
kdembler, line 51: The background-color we have don't seem to be matching what's in Figma. We should use those: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6530%3A2651

First for inactive item, second for hover, third for active state (so both current page and pressed button)

kdembler, line 51: The text color seems to be white when item is active, should be `gray[50]`

Screenshot 2021-04-08 at 11 05 03

kdembler (changes requested):
After testing it out, I don't think the dots should be clickable. If I'm on page 1 and I click dots, I wouldn't expect it to have the same function as "next" button. Let's make them just visual and non-interactive

src/shared/components/Pagination/Pagination.style.tsx
drillprop: Everything looks fine on Figma where we have a grayish background, but if we set these transparent colors on the black background - active buttons will be barely visible at least for my taste. I'll update this, but want to just raise the issue.
image

kdembler: Good point, I'll raise this with the designers. For now, let's use what's in the design

kdembler (approved):

Issue #413, opened 2021-03-30
Update Checkbox component


kdembler:
There's been some slight changes to the checkbox design, let's just make sure our component is up to date. Figma -> Components -> Checkbox

#### Requirements:
- Make sure all the details are up to date
- I think there may have been some update to how the checkbox is sized (padding, margin), make sure it's matching

PR #430, opened 2021-04-06
Checkbox update

drillprop:
Fix: #413 
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6596%3A1533

### DIFF
--- a/src/shared/components/Checkbox/Checkbox.styles.tsx
+++ b/src/shared/components/Checkbox/Checkbox.styles.tsx
@@ -1,5 +1,4 @@
-import { shade } from 'polished'
-import { sizes, colors, transitions } from '@/shared/theme'
+import { sizes, colors, transitions, typography } from '@/shared/theme'
 import { css } from '@emotion/react'
 import styled from '@emotion/styled/'
 import Text from '../Text'
@@ -16,11 +15,12 @@ export const CheckboxLabel = styled.label<CheckboxLabelProps>`
 `
 
 export const LabelText = styled(Text)`
-  margin-left: 10px;
-  color: ${colors.gray[400]};
+  color: ${colors.gray[50]};
+  line-height: ${sizes(7)};
 `
 
 export const Container = styled.div<CheckboxStateProps>`
+  --hoverColor: ${colors.transparentPrimary[12]};
   position: relative;
   width: max-content;
   padding: ${sizes(2)};
@@ -29,13 +29,8 @@ export const Container = styled.div<CheckboxStateProps>`
   border-radius: 100%;
   color: ${colors.gray[300]};
   transition: background ${transitions.timings.loading} ${transitions.easing};
-  ${({ isFocused, selected, disabled }) =>
-    isFocused && !disabled && `background: ${selected ? colors.blue[900] : colors.gray[800]};`}
   &:hover {
-    ${({ selected, disabled, error }) => [
-      !disabled && `background: ${selected ? colors.blue[900] : colors.gray[800]};`,
-      error && `background: ${shade(0.7, colors.error)};`,
-    ]}
+    ${({ disabled }) => [!disabled && `background: var(--hoverColor)`]}
   }
 `
 
@@ -51,10 +46,13 @@ const disabledStyles = (props: CheckboxStateProps) =>
     ? [
         css`
           cursor: not-allowed;
+          opacity: 0.5;
+          border: 1px solid ${colors.gray[200]};
+          background-color: ${colors.gray[400]};
         `,
         props.selected &&
           css`
-            background-color: ${props.disabled ? colors.gray[700] : 'transparent'};
+            background-color: ${colors.gray[700]};
             border: 1px solid ${colors.gray[700]};
             color: ${colors.gray[400]};
           `,
@@ -108,4 +106,20 @@ export const Checkmark = styled.div`
   width: 18px;
   height: 18px;
   border-radius: 1px;
+`
+
+export const TextContainer = styled.div`
+  margin-left: ${sizes(4)};
+  position: relative;
+`
+
+type CaptionProps = {
+  error?: boolean
+}
+
+export const Caption = styled(Text)<CaptionProps>`
+  display: block;
+  font-size: ${typography.sizes.caption};
+  color: ${({ error }) => (error ? colors.error : colors.gray[300])};
+  position: absolute;
 `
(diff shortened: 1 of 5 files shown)


eldiegod (commented):
Looks pretty good, would be cool to have a story with a caption also I think figma has a border radius of 1px

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -29,13 +29,8 @@ export const Container = styled.div<CheckboxStateProps>`
...
   color: ${colors.gray[300]};
   transition: background ${transitions.timings.loading} ${transitions.easing};
-  ${({ isFocused, selected, disabled }) =>
-    isFocused && !disabled && `background: ${selected ? colors.blue[900] : colors.gray[800]};`}
   &:hover {
-    ${({ selected, disabled, error }) => [
-      !disabled && `background: ${selected ? colors.blue[900] : colors.gray[800]};`,
-      error && `background: ${shade(0.7, colors.error)};`,
-    ]}
+    ${({ disabled }) => [!disabled && `background: var(--hoverColor)`]}
```
mikkio-j: I don't think we need to declare that variable if it's used only here

drillprop:
> I think figma has a border radius of 1px

Didn't notice, you were right. Fixed

mikkio-j (approved):

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -69,15 +81,22 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
...
+              onFocus={handleFocus}
+              onBlur={handleBlur}
               {...props}
             />
             <Checkmark>{isSelected ? <Icon name={isIndeterminate ? 'dash' : 'check'} /> : null}</Checkmark>
           </InnerContainer>
         </Container>
-        {label && <LabelText variant="subtitle2">{label}</LabelText>}
+        <TextContainer>
+          {label && <LabelText variant="subtitle2">{label}</LabelText>}
```
kdembler: I think this should be `vartiant="body1"` as in Figma, then you don't need the additional `line-height` style

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -69,15 +81,22 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
...
+          {helperText && (
+            <Caption error={error} variant="caption">
```
kdembler: Any reason to not use the `HelperText` component?

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -16,8 +15,8 @@ export const CheckboxLabel = styled.label<CheckboxLabelProps>`
 `
 
 export const LabelText = styled(Text)`
-  margin-left: 10px;
-  color: ${colors.gray[400]};
+  color: ${colors.gray[50]};
+  line-height: ${sizes(7)};
 `
 
 export const Container = styled.div<CheckboxStateProps>`
```
kdembler, line 19: Cannot comment on the line I want to, but the negative margin to the left should be now only `sizes(1)`: 
Screenshot 2021-04-08 at 11 26 19

kdembler (changes requested):

src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -37,48 +47,52 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
...
-              onFocus={onFocusHandler}
-              onBlur={onBlurHandler}
-              {...props}
-            />
-            <Checkmark>{isSelected ? <Icon name={isIndeterminate ? 'dash' : 'check'} /> : null}</Checkmark>
-          </InnerContainer>
-        </Container>
-        {label && <LabelText variant="subtitle2">{label}</LabelText>}
-      </CheckboxLabel>
+      <>
```
kdembler: I don't think we should display 2 separate elements here, this may have unexpected effects on the layout. In other components I think we just return the helper text as part of the label? There's also some unexpected spacing between the checkbox and the helper because of this I think.

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -108,4 +107,13 @@ export const Checkmark = styled.div`
...
   height: 18px;
   border-radius: 1px;
+  position: relative;
+  svg {
+    position: absolute;
+  }
+`
+
+export const StyledHelperText = styled(HelperText)`
+  margin-top: 0;
```
kdembler:
```suggestion
```

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -108,4 +107,13 @@ export const Checkmark = styled.div`
...
+  margin-left: ${sizes(10)};
```
kdembler:
```suggestion
  margin-left: ${sizes(9)};
```

src/shared/components/Checkbox/Checkbox.styles.tsx
```diff
@@ -108,4 +107,13 @@ export const Checkmark = styled.div`
   width: 18px;
   height: 18px;
   border-radius: 1px;
+  position: relative;
+  svg {
+    position: absolute;
+  }
```
kdembler: Why do we need this?

drillprop: check
I had an odd layout shift before, but I guess it's fixed now.

src/shared/components/Checkbox/Checkbox.styles.tsx
drillprop: Are you sure? 
9:
image
10: 
image

src/shared/components/Checkbox/Checkbox.styles.tsx
drillprop: I set it to `sizes(1)` eventually, in the design margin is a little bit different for TextField and Checkbox

src/shared/components/Checkbox/Checkbox.styles.tsx
drillprop: oh, maybe I shouldn't use`sizes` here. I'll look into this again

kdembler (approved):

Issue #414, opened 2021-03-30
Update RadioButton component


kdembler:
There's been some slight changes to the radio button design, let's just make sure our component is up to date. Figma -> Components -> Radio buttons

#### Requirements:
- Make sure all the details are up to date

PR #431, opened 2021-04-06
Radio button update

mikkio-j:
Issue: #414 
Figma: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6599%3A977](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6599%3A977)


src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -1,25 +1,41 @@
...
+  StyledLabelText,
+  StyledCaptionText,
+  RadioButtonStyleProps,
+  RadioButtonWrapper,
+} from './RadioButton.style'
 
 type RadioButtonProps = Partial<{
   selectedValue: string | number
   label: string
+  caption?: string
```
drillprop: I think we should name this `helperText` like in other inputs

src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -1,25 +1,41 @@
...
-      </Label>
+      <RadioButtonWrapper>
+        <Label disabled={disabled}>
+          <StyledInput checked={isSelected} error={error} disabled={disabled}>
+            <Input ref={ref} value={value} type="radio" disabled={disabled} {...props} checked={isSelected} />
+          </StyledInput>
+          {label && <StyledLabelText variant="body1">{label}</StyledLabelText>}
+        </Label>
+        {caption && (
+          <StyledCaptionText variant="caption" error={error}>
```
drillprop: This is wrong. If you set `variant` to `caption` the element will be set as `<caption>` and you shouldn't use the `caption` element inside `div`. Captions, if I'm not mistaken, are related only to tables.
If you use this, you'll see in the console:
image

We should update `Text` component maybe in the future. This `caption` variant is very misleading

drillprop (commented):
Found two small things

src/shared/components/RadioButton/RadioButton.tsx
drillprop: Okay, leave it as it is. I will update the `Text` component in my PR #424

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -89,15 +97,26 @@ export const StyledInput = styled.div<RadioButtonStyleProps>`
   }
```
drillprop, line 116: It's not your change, but I'm almost sure that we don't need this `&::before` above. Can you double-check this and remove it if it's unnecessary?

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -89,15 +97,26 @@ export const StyledInput = styled.div<RadioButtonStyleProps>`
...
 `
 export const StyledLabelText = styled(LabelText)`
-  font-size: ${typography.sizes.subtitle2};
+  color: ${colors.gray[50]};
+`
+
+export const StyledCaptionText = styled(Text)<RadioButtonCaptionProps>`
+  color: ${({ error }) => (error ? colors.error : colors.gray[300])};
+  margin: ${sizes(1)} 0 0 ${sizes(7)};
+  text-align: left;
```
drillprop:
```suggestion
```
It's redundant

drillprop (approved):

src/shared/components/RadioButton/RadioButton.style.ts
mikkio-j: That's correct! Fixed

src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -1,25 +1,41 @@
...
-      </Label>
+      <RadioButtonWrapper>
+        <Label disabled={disabled}>
+          <StyledInput checked={isSelected} error={error} disabled={disabled}>
+            <Input ref={ref} value={value} type="radio" disabled={disabled} {...props} checked={isSelected} />
+          </StyledInput>
+          {label && <StyledLabelText variant="body1">{label}</StyledLabelText>}
+        </Label>
+        {helperText && (
+          <StyledCaptionText variant="caption" error={error}>
```
kdembler: Let's use `HelperText` component here

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -18,44 +23,47 @@ export const Input = styled.input`
 const colorFromProps = ({ error, checked, disabled }: RadioButtonStyleProps) => {
```
kdembler, line 23: Seems we're missing the hover state? Both for unchecked and checked states (or if we have it, it's not what's in Figma)

kdembler, line 23: The active state seems to be missing the background

Figma:
Screenshot 2021-04-08 at 11 33 59

App:
Screenshot 2021-04-08 at 11 34 38

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -69,35 +77,36 @@ export const StyledInput = styled.div<RadioButtonStyleProps>`
...
   & + span {
     color: ${(props) => (props.checked ? colors.white : '')};
   }
+  transition: background-color 0.25s ease, border-color 0.25s ease;
+`
+
+export const RadioButtonWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  margin-bottom: ${sizes(4)};
```
kdembler: I don't think the component should have a margin by itself? Or at least if it's `:not(:last-of-type)`

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -69,35 +77,36 @@ export const StyledInput = styled.div<RadioButtonStyleProps>`
...
-    z-index: -1;
-  }
   ${colorFromProps};
   & + span {
     color: ${(props) => (props.checked ? colors.white : '')};
   }
+  transition: background-color 0.25s ease, border-color 0.25s ease;
+`
+
+export const RadioButtonWrapper = styled.div`
```
kdembler: Do we need that wrapper? I think for other places we just include the helper as part of the label. Not sure if that's the best approach but let's be consistent

kdembler (changes requested):

src/shared/components/RadioButton/RadioButton.style.ts
mikkio-j: Reverting that, missed that it handles hover state

src/shared/components/RadioButton/RadioButton.style.ts
mikkio-j: Removed pseudo-element responsible for that. It's back now. Fixed

src/shared/components/RadioButton/RadioButton.stories.tsx
```diff
@@ -1,12 +1,14 @@
 import React, { useState } from 'react'
 import { Meta, Story } from '@storybook/react'
 import RadioButton from './RadioButton'
+import { template } from 'lodash'
```
kdembler: This doesn't seem needed here?
```suggestion
```

src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -1,24 +1,33 @@
...
   Omit<React.InputHTMLAttributes<HTMLInputElement>, 'onClick'>
 
 const RadioButton = React.forwardRef<HTMLInputElement, RadioButtonProps>(
-  ({ value, selectedValue, label, disabled, error, ...props }, ref) => {
+  ({ value, selectedValue, label, helperText, disabled, error, ...props }, ref) => {
     const isSelected = value === selectedValue
 
     return (
       <Label disabled={disabled}>
         <StyledInput checked={isSelected} error={error} disabled={disabled}>
```
kdembler: Spacing here needs to be improved, like mentioned in the chat:

Screenshot 2021-04-09 at 11 08 43
Screenshot 2021-04-09 at 11 12 01

drillprop:
One more thing - could you also add `className` prop to `RadioButton` and pass it to the `Label`?

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -18,61 +23,86 @@ export const Input = styled.input`
...
+  --radio-button-container-size: ${sizes(6)};
+  display: inline-grid;
+  grid-template-columns: auto 1fr;
+  align-items: center;
+  grid-column-gap: ${sizes(3)};
+  grid-row-gap: ${sizes(1)};
+  cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
+  & > span:nth-of-type(1) {
+    margin: ${sizes(2)};
+  }
```
kdembler: Is this needed?

kdembler (approved):

src/shared/components/RadioButton/RadioButton.style.ts
mikkio-j: Good catch. No longer needed. Fixed

PR #437, opened 2021-04-08
channel create/update integration

kdembler:


src/components/Dialogs/TransactionDialog/TransactionDialog.tsx
```diff
@@ -3,29 +3,85 @@ import ActionDialog, { ActionDialogProps } from '../ActionDialog/ActionDialog'
...
 import { StyledTitleText, StyledDescriptionText } from '../MessageDialog/MessageDialog.style'
 import { ExtrinsicStatus } from '@/joystream-lib'
+import MessageDialog from '../MessageDialog'
 
 export type TransactionDialogProps = Pick<ActionDialogProps, 'onEnter' | 'onExit' | 'className'> & {
-  onCancel?: () => void
   status: ExtrinsicStatus | null
+  successTitle: string
+  successDescription: string
+  onClose: () => void
```
drillprop, line 12: I'm a bit confused. Could we just use `onExit`, here?

src/joystream-lib/api.ts
```diff
@@ -1,11 +1,14 @@
+/* eslint-disable new-cap */
```
drillprop: I guess you could also remove this `eslint-disable` and do: 
```ts
import { Bytes, GenericEvent, Raw, Option, Vec, TypeRegistry, u64 as U64 } from '@polkadot/types'
```
but not sure what is better here. I personally feel uncomfortable if I see `ts-ignore` or `eslint-disable` :P

src/components/Dialogs/TransactionDialog/TransactionDialog.tsx
kdembler: Probably because `onEnter` and `onExit` may have been confusing. They're not super useful so just removed them. The way they worked, `onExit` would get triggered only when `showDialog` changes from `true` to `false`. But something has to trigger that change, that's why we had `onClose` callback so consumer can react to user closing the dialog

src/joystream-lib/api.ts
kdembler: Renamed :+1:

mikkio-j (approved):
have nothing to add 🎉

PR #435, opened 2021-04-08
update typography colors and button typography

kdembler:

PR #439, opened 2021-04-09
Fix button icon color

drillprop:
image
after merging #435 button icons have default black color. I believe this PR should fix the issue

### DIFF
--- a/src/shared/components/Button/Button.style.ts
+++ b/src/shared/components/Button/Button.style.ts
@@ -167,7 +167,6 @@ export const StyledButton = styled('button', { shouldForwardProp: isPropValid })
   --horizontal-padding-large: 20px;
 
   border-width: 1px;
-  color: ${colors.gray[50]};
   text-decoration: none;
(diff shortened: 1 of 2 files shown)


kdembler:
I run into the same issue with #435 actually, that's why I updated the cheveron icon. I think we're still not entirely settled on the approach with icon colors, but I think for now we'll just use colors defined inside the icon. If you update the `close` icon with the one from the MysteryBox™️, it should work:

```svg
<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
    <path d="M3 13L8 8L13 13" stroke="#F4F6F8" stroke-width="2"/>
    <path d="M13 3L8 8L3 3" stroke="#F4F6F8" stroke-width="2"/>
</svg>

```

drillprop:
So basically, for now, we won't be using svg's `currentColor`, right?

kdembler (approved):
Yep!

PR #441, opened 2021-04-09
refactor Button, create IconButton

kdembler:


src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
```diff
@@ -83,15 +83,15 @@ const ImageCropDialogComponent: React.ForwardRefRenderFunction<
 
   const zoomControlNode = (
     <ZoomControl>
-      <Button icon="zoom-out" variant="tertiary" onClick={() => handleZoomChange(currentZoom - zoomStep)} />
+      <IconButton icon="zoom-out" variant="tertiary" onClick={() => handleZoomChange(currentZoom - zoomStep)} />
```
kosorz, line 86: It's tempting and I also catch myself doing it quite often in my pet projects but please note that because you've passed anonymous function to onClick, React will always re-render this IconButton since it receives a new anonymous function as a prop which React is unable to compare to the previous anonymous function (since they are both anonymous). You could pass in a reference to the method like onClick={handleClick} and it lets React know when nothing has changed so it does not unnecessarily re-render.

kdembler: Hey!
That's not true actually. And that's for two reasons:
1) What you said would be mostly true if this was a class component. Then we could define the click handler as a class method and its reference would be stable. However, this is a function component, so its body is only the render function. Even if I would define a named handle, like `handleClick`, the function would be created anew on every render, meaning every time the component renders, the handler reference would be different, just like with the anonymous function we use currently. To have a stable reference, we would need to use something like [`React.useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback) or [`React.useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo)
2) Even if the handler reference was stable (using `useCallback` for example), `IconButton` would actually still re-render. This is because by default, React components will re-render every time their parents re-render. There isn't automatic mechanism that will compare the props and opt-out of re-render. To avoid re-render on stable props you need to use something like [`React.PureComponent`](https://reactjs.org/docs/hooks-reference.html#usememo) or [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo). This will do a shallow comparison on props and not re-render if they're stable.

You can test out the above in this codesanbox I've created: https://codesandbox.io/s/loving-darkness-v3hm3 If you open the console, you will see that on parent render, also all of the children render, except the memoized component.

---

We _could_ do all of the above and avoid any unnecessary re-renders, however, usually optimizing React app prematurely isn't the best idea. The philosophy, also expressed in the official docs, is that optimizations shouldn't be attempted unless you actually see there's a performance problem. Because of virtual DOM, the cost of re-render isn't that high. Funnily enough, a lot of times, optimizing will actually [result in a less performant code](https://kentcdodds.com/blog/usememo-and-usecallback) that you've started with.

Thanks for your concern though!
Let me know if you have any questions :)

kosorz: Well, thanks for your answer!

You're absolutely right I have been assigning references subconsciously since pre-hooks react. 
It seems like it's time to delve deeper into the differences in mechanisms of pre hooks and current react to avoid such bloopers in the future.

kdembler: Personally I found that once you go full hooks, you won't want to go back. IMO, they're pretty awesome. But like you pointed out, they do require some mental shift :) Lmk if you wander around the codebase and have any questions 💪

src/shared/components/Button/Button.style.ts
```diff
@@ -1,183 +1,57 @@
...
-  display: inline-flex;
-  flex-shrink: 0;
-  justify-content: center;
-  align-items: center;
-  &:hover {
-    cursor: ${(props) => (!props.disabled && props.clickable ? 'pointer' : '')};
-  }
+export const StyledText = styled(Text)<ButtonSizeProps>`
+  // compensate for line-height being 1
+  ${textPaddingStyles};
```
drillprop:
```suggestion
  ${textPaddingStyles}
```

src/shared/components/Button/Button.style.ts
```diff
@@ -1,183 +1,57 @@
...
-  flex-shrink: 0;
-  width: ${typography.sizes.icon.small};
-  height: ${typography.sizes.icon.small};
-
-  margin: calc(-1 * var(--vertical-padding-small)) 0;
-  margin-right: ${({ hasText }) => hasText && sizes(2)};
-
-  filter: ${(props) => (props.disabled ? 'brightness(0.7)' : null)};
+export const StyledButtonBase = styled(ButtonBase)<ButtonSizeProps>`
+  ${sizeOverwriteStyles};
```
drillprop:
```suggestion
  ${sizeOverwriteStyles}
```

src/shared/components/Button/Button.style.ts
drillprop: For some reason, my vscode is angry about this.

drillprop (approved):
Looks good 👍🏻

src/shared/components/Button/Button.style.ts
kdembler: Did some research and we probably shouldn't include that extra semicolon. Removed those

mikkio-j (approved):
Looks good 👌

PR #446, opened 2021-04-13
icons update

kdembler:
Updates all the places with icons to use the new ones from the design.

Notable changes:

1) We manually generate React components out of SVGs now. This should let us keep better track of what is used and what is not and also give us greater control over those components. That's done with `yarn codegen:icons`

2) Icons are separated into the following categories:
- `Glyph` - 16x16 icon for informative purposes and for use inside buttons
- `Alert` - 24x24 icon for alerts (success, info, etc)
- `Player` - 24x24 icon meant to be used as a control in the video player
- `Outline` - 32x32 icon for informative purposes
- `Nav` - 32x32 icon for navigation
- `Large` - 54x54 icon for presentation

3) All icon component names consist of `Svg${IconType}${IconName}`. This way all the icons can be easily imported with code suggestions

4) The `Button` component will now accept a `icon: ReactNode` prop and can be used like this:
```tsx
<Button icon={<SvgGlyphAddVideo />}>Add video</Button>
```

5) The `Button` component will now only accept `string` as its `children`

6) The `IconButton` component will now accept icon as its children:
```tsx
<IconButton><SvgGlyphAddVideo /></IconButton>
```


src/components/CoverVideo/CoverVideo.style.ts
```diff
@@ -217,19 +217,10 @@ export const ButtonsContainer = styled.div`
...
-  svg {
-    width: auto;
-    height: auto;
-  }
 `
 
 export const SoundButton = styled(IconButton)`
   margin-left: ${sizes(4)};
   height: ${BUTTONS_HEIGHT_PX};
   width: ${BUTTONS_HEIGHT_PX};
```
drillprop, line 225: Just as a side note - I think you could also remove `high` and `width` from `SoundButton`, `high` from `PlayButton` and delete `BUTTONS_HEIGHT` above .  The difference is really subtle.

drillprop (approved):
Didn't find anything, awesome work 🎉

src/components/CoverVideo/CoverVideo.style.ts
kdembler: Let's leave this out for the viewer update

Issue #368, opened 2021-03-09
Integrate the sign in view


kdembler:
Once https://github.com/Joystream/atlas/issues/280 is done, we should add all the integrations to that view:

1) Getting accounts from polkadot extension
2) Interaction with the faucet
3) Creating a membership (make sure the handle isn't used already)

PR #447, opened 2021-04-14
Signin update

### LINKED ISSUE #433: Update the sign in view with new design
The sign in view got a face lift, let's update our app. There are some iffy parts though, contact me once you start working on this so we can go through the details.

drillprop:
Fixes #433 partly #368 (point one at least)
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=3119%3A0


kdembler (changes requested):
Looking good, can't wait to see this flow in action!

Some things I've noticed in the app:
1) In the membership hub, the `New member` button redirects directly to create membership form. However, for new membership, user will need to select the account they want to use for it as well. So we need to redirect them to the dialog and the `Select account` step
2) Seems the `Please reload after installing` message reloads the page on click, let's remove that
3) In the membership create form, let's use the avatar on top to try to display the image user provided the URL for. We should debounce on input and use that as `src`. We can also listen to `onError` on the `img` and set error in the URL input 
4) There's a bunch of UI inconsistencies. However, to not block the progress on further integration I've created a separate issue with those so we can tackle them independent of this PR: https://github.com/Joystream/atlas/issues/448

Will take a look at the code later

src/hooks/useSignInDialog/Steps/AccountStep.style.ts
```diff
@@ -0,0 +1,116 @@
...
+  align-self: center;
+  margin: 0;
+  margin-right: 20px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`
+
+export const StyledButton = styled(Button)`
+  display: block;
```
kdembler: Why is this needed?

src/hooks/useSignInDialog/Steps/AccountStep.style.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export const IconGroup = styled.div`
+  display: flex;
+  align-items: center;
+  img,
+  svg {
+    display: block;
+    margin: 0 ${sizes(2)};
+  }
```
kdembler: Couldn't this be just selected with `> *`? I'd like to avoid selecting by explicit element name, I wouldn't want our styles in the whole app to break if we stop inlining SVGs for example

src/config/routes.ts
```diff
@@ -30,9 +30,19 @@ export const relativeRoutes = {
     editChannel: () => 'channel',
     videos: () => 'videos',
     uploads: () => 'uploads',
-    signIn: () => 'signIn',
+    signIn: () => 'signin',
+    join: ({ step }: { step?: string } = {}) => {
```
mikkio-j: since this function does the same thing as the one in search route maybe you could extract these two into one that is generic?

src/hooks/useSignInDialog/Steps/AccountStep.tsx
```diff
@@ -0,0 +1,148 @@
...
+const AccountStep: React.FC = () => {
+  const navigate = useNavigate()
+  const { setActiveUser } = useActiveUser()
+  const { accounts } = useJoystream()
+  const [imageLoaded, setImageLoaded] = useState(false)
+  const [selectedAccountAddress, setSelectedAccountAddress] = useState<undefined | string>()
+
+  const handleSubmitSelectedAccount = async (e: FormEvent) => {
+    e.preventDefault()
+    if (selectedAccountAddress) {
```
kdembler: No need to change here but the opposite generally creates less nested code and is easier to read:
```ts
if (!selectedAccountAddress) {
  return
}
...
```

src/hooks/useSignInDialog/Steps/AccountStep.tsx
```diff
@@ -0,0 +1,148 @@
...
+      await setActiveUser({
+        accountId: selectedAccountAddress,
+        memberId: null,
+        channelId: null,
+      })
+      navigate('?step=2')
```
kdembler: We shouldn't use hardcoded route here. We defined that in `routes` so let's rely on that.

src/hooks/useSignInDialog/Steps/AccountStep.tsx
```diff
@@ -0,0 +1,148 @@
...
+        key={!accounts?.length ? 'no-accounts' : 'accounts'}
+        classNames={transitions.names.fadeAndSlide}
+        timeout={parseInt(transitions.timings.routing)}
+      >
+        {!accounts?.length ? (
+          <StepWrapper centered withBottomBar>
+            {!imageLoaded && <Placeholder height="180px" width="320px" />}
+            <CSSTransition in={imageLoaded} classNames="fade" timeout={100}>
+              <AccountStepImg src={accountCreation} onLoad={() => setImageLoaded(true)} />
+            </CSSTransition>
```
kdembler: Why do we need to handle image loading here?

src/hooks/useSignInDialog/Steps/AccountStep.style.ts
```diff
@@ -0,0 +1,116 @@
...
+  margin: 10px 30px;
+  counter-increment: ordered-list-counter;
+  display: flex;
+  margin-bottom: 20px;
+  align-items: center;
+  max-width: 170px;
+  flex-direction: column;
+  ::before {
+    margin-bottom: ${sizes(2)};
+    content: '0' counter(ordered-list-counter);
```
kdembler: Very elegant, I wasn't aware you can do something like this with just CSS 👌

src/hooks/useSignInDialog/Steps/AccountStep.tsx
```diff
@@ -0,0 +1,148 @@
...
+              </OrderedStep>
+              <OrderedStep secondary variant="caption" as="li">
+                Click the plus icon
+              </OrderedStep>
+              <OrderedStep secondary variant="caption" as="li">
+                Continue with instructions presented on the screen
+              </OrderedStep>
+            </OrderedSteps>
+            <StepFooter>
+              <BottomBarIcon></BottomBarIcon>
```
kdembler:
```suggestion
              <BottomBarIcon />
```

src/hooks/useSignInDialog/Steps/AccountStep.tsx
```diff
@@ -0,0 +1,148 @@
...
+        memberId: null,
+        channelId: null,
+      })
+      navigate('?step=2')
+    }
+  }
+
+  const handleSelect: (e: React.MouseEvent<HTMLElement>) => void = (e) => {
+    const element = e.currentTarget as HTMLInputElement
+    setSelectedAccountAddress(element.value)
```
kdembler: Do we need to read that value from the element? Couldn't we just do `onSelect={() => handleSelect(id)}` in line 104?

src/hooks/useSignInDialog/Steps/ExtensionStep.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { absoluteRoutes } from '@/config/routes'
+import { SvgGlyphExternal } from '@/shared/icons'
+
+const ExtensionStep: React.FC = () => {
+  const navigate = useNavigate()
+  const { extensionConnected } = useJoystream()
+
+  useEffect(() => {
+    if (extensionConnected) {
+      navigate('?step=1')
```
kdembler: Same as above with route

src/hooks/useSignInDialog/Steps/Steps.style.ts
```diff
@@ -1,28 +1,41 @@
...
+  margin-top: 24px;
+  background-color: ${colors.gray[800]};
+  width: 100%;
+  outline: 24px solid ${colors.gray[800]};
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`
+export const BottomBarIcon = styled(SvgGlyphWarning)`
+  margin-right: 10px;
```
kdembler: This isn't the correct value from the design. It should be 8/sizes(2). In general, most of the time when you cannot use `sizes`, you may have the wrong value. Take a look at the selected node on the left:

Screenshot 2021-04-16 at 11 30 52

Screenshot 2021-04-16 at 11 31 00

src/hooks/useSignInDialog/Steps/Steps.style.ts
```diff
@@ -1,28 +1,41 @@
...
   margin-top: ${sizes(2)};
-  color: ${colors.gray[200]};
   max-width: 400px;
 `
+
+export const StepFooter = styled.div`
+  margin-top: 24px;
+  background-color: ${colors.gray[800]};
+  width: 100%;
+  outline: 24px solid ${colors.gray[800]};
```
kdembler: I suggest we don't do that. This may make the layout really hard to understand, it took me some time. How about this instead:
```css
margin: 0 calc(-1 * var(--dialog-padding));
padding: 0 var(--dialog-padding);
```
wdyt?

src/hooks/useSignInDialog/Steps/TermsStep.tsx
```diff
@@ -98,27 +98,31 @@ const TermsStep: React.FC<TermsStepProps> = ({ onStepChange }) => {
...
-          <Checkbox
-            value={isAccepted}
-            onClick={() => setIsAccepted(!isAccepted)}
-            label="I’ve read and accept Terms And Conditions"
-          />
-        </CSSTransition>
-        <ContinueButton disabled={!isAccepted}>Continue</ContinueButton>
-      </TermsForm>
+      <StepFooter>
+        <TermsForm onSubmit={handleSubmit}>
```
kdembler: Is there some advantage to this being a form instead of just `div` or something?

src/hooks/useSignInDialog/useSignInDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { readFromLocalStorage, writeToLocalStorage } from '@/utils/localStorage'
+import React, { useState, useCallback, useEffect } from 'react'
+import { useRouterQuery } from '../useRouterQuery'
+import { ExtensionStep, AccountStep, TermsStep } from './Steps'
+
+type DialogStep = {
+  isActive?: boolean
+}
+
+const getDialogState = promisify(() => readFromLocalStorage<DialogStep>('dialogStep') || { isActive: false })
```
kdembler: Why do we need this in local storage?

src/hooks/useSignInDialog/useSignInDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+    },
+    {
+      title: 'Create or select a polkadot account',
+      element: <AccountStep />,
+    },
+    {
+      title: 'Accept terms and conditions',
+      element: <TermsStep onAcceptTerms={closeDialog} />,
+    },
+  ]
```
kdembler: I don't like those being here. We're really fragmenting the code here. This is very view-specific but is defined on global level

kdembler: On a second thought, why do we need this hook at all? Couldn't this be handled internally by the sign in view based on the query param?

src/joystream-lib/api.ts
```diff
@@ -314,9 +314,13 @@ export class JoystreamJs {
...
-    const accountInjector = await web3FromAddress(accountId)
-    this._selectedAccountId = accountId
-    this.api.setSigner(accountInjector.signer)
+    try {
+      const accountInjector = await web3FromAddress(accountId)
+      this._selectedAccountId = accountId
+      this.api.setSigner(accountInjector.signer)
+    } catch (error) {
+      throw AccountNotFoundError
+    }
```
kdembler: Did you run into some case where this was needed? I'm curious why the previous check (`find`) wouldn't catch this

src/hooks/useSignInDialog/Steps/AccountStep.style.ts
```diff
@@ -0,0 +1,116 @@
+import { Spinner, Button, Text, RadioButton } from '@/shared/components'
```
mikkio-j:
```suggestion
import { Button, Text, RadioButton } from '@/shared/components'
```

kdembler (changes requested):

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -35,57 +34,66 @@ const CreateMemberView = () => {
...
-        <SubTitle variant="body2">
-          Customize your membership details to easily spot the member account you would like to use when signing in.
+        <Hero variant="h2">Create Joystream membership</Hero>
+        <SubTitle variant="body1" secondary>
+          Start your journey as a Video Publisher. Create, manage and modify your channel and video content.
         </SubTitle>
       </Header>
-      <Form onSubmit={onSubmit}>
+      <Form onSubmit={handleCreateMember}>
+        <StyledAvatar size="view" imageUrl={avatarImageUrl} />
```
kdembler: From my original review:

> We can also listen to onError on the img and set error in the URL input

src/hooks/useSignInDialog/Steps/Steps.style.ts
drillprop: yeah, that's definitely a better solution, but to make this work I also had to change `width` a bit. Fixed

src/hooks/useSignInDialog/Steps/AccountStep.tsx
drillprop: I was experiencing a small layout shift here, but I will get rid of that. I'll just set a fixed height to the image.

src/joystream-lib/api.ts
drillprop: The app was breaking when the `accountId` saved in `localStorage` was invalid.
image
Not super sure if it's still needed. The previous check, for some reason, is not catching this.

drillprop: oh, I see. The previous check doesn't do anything 😅  
```js
   if (!this._accounts.find((a) => a.id)) {
      throw AccountNotFoundError
    }
```
Should be: 
```js
    if (!this._accounts.find((a) => a.id === accountId)) {
      throw AccountNotFoundError
    }
```

kdembler (approved):

Issue #432, opened 2021-04-07
Implement "auth" logic to properly redirect logged in users


kdembler:
We will need to handle users in different stages of the login process in the studio. Full flow here:

![signin-flow](https://user-images.githubusercontent.com/12646744/113906936-1325ba80-97d5-11eb-99a8-6f6cc3e2f088.png)

For any questions, DM me

PR #449, opened 2021-04-14
Redirecting  logged in users

mikkio-j:
Issue: #432


src/components/LoadingStudio.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import styled from '@emotion/styled'
+import { Spinner, Text } from '@/shared/components'
+import { TOP_NAVBAR_HEIGHT } from '@/components'
+
+export const LoadingStudio = () => {
```
kdembler:
```suggestion
export const LoadingStudio: React.FC = () => {
```

src/hooks/useJoystream/provider.tsx
```diff
@@ -11,6 +11,7 @@ type JoystreamContextValue = {
   joystream: JoystreamJs | null
   accounts: Account[]
   extensionConnected: boolean
+  extensionConnectionLoading: boolean
```
kdembler: Could we possibly reuse `extensionConnected` for that? It could have 3rd intermediary step, for example `null`. What do you think about that?

src/views/studio/StudioLayout.tsx
```diff
@@ -10,61 +10,171 @@ import {
...
+  )
+}
+
+const MainContainer = styled.main`
+  position: relative;
+  padding: ${TOP_NAVBAR_HEIGHT}px var(--global-horizontal-padding) 0;
+  margin-left: var(--sidenav-collapsed-width);
+`
+
+const StudioLayoutWrapper = () => {
```
kdembler:
```suggestion
const StudioLayoutWrapper: React.FC = () => {
```

src/views/studio/StudioLayout.tsx
```diff
@@ -10,61 +10,171 @@ import {
...
+  return (
+    <>
+      <NoConnectionIndicator
+        nodeConnectionStatus={nodeConnectionStatus}
+        isConnectedToInternet={isUserConnectedToInternet}
+      />
+      <StudioTopbar hideChannelInfo={!authenticated} />
+      {authenticated && <StudioSidenav />}
+      {extensionConnectionLoading ? (
+        <LoadingStudio />
```
kdembler: This is only displayed when `extensionConnectionLoading`, but I think per flow it should also be displayed if `accountId && memberId` are set but `channelId` is not and we're still fetching the member right? This is `wait till member details are fetched` in the diagram. Or we don't need it for some reason?

kdembler (approved):

PR #450, opened 2021-04-14
add config for staging network

kdembler:
Most of the changes are because of the differences in the most recent query node schema


.env
```diff
@@ -1,10 +1,17 @@
 # This file is commited. Do not store secrets here
```
kdembler, line 1: Let's clean this up later

drillprop (commented):
Hard to test it, because we don't have any channels and videos yet, but the code looks good.
Also when I go to `/videos` app will break.  I think we could fix that by renaming `categories` to `videoCategories` and `category` to `videoCategory` in `extendedQueryNode.graphql`(line 257 and 260), then changing `category` to `videoCategories` in `categories.graphql` and finally  in `categories`  hook.
```ts
  const { data, ...rest } = useGetCategoriesQuery({ ...opts, variables })
  return {
    categories: data?.videoCategories,
    ...rest,
  }
```
 

I guess this is something which I forgot to do when I was updating `queryNode`

kdembler:
There are still some problems with the query node getting resolved so we will not be able to see it fully working yet. That should be fixed today, if something doesn't work we can fix it further. Fixed the problem with the categories

drillprop (approved):

Issue #418, opened 2021-03-31
Add filtering to displayed videos and channels


kdembler:
We should filter the videos displayed on the viewer app:
- `isCensored: false`
- `isPublic: true`
- have all the assets so both `thumbnailAvailability` and `mediaAvailability` are `ACCEPTED`

Analogous filters for channels

PR #452, opened 2021-04-15
Filtering displayed videos and channels on viewer

mikkio-j:
Issue: #418


kdembler (approved):
Looks good, let's test this further when we have more assets in the staging network

PR #480, opened 2021-04-19
Signin integration

### LINKED ISSUE #368: Integrate the sign in view

drillprop:
Fix #368
Based on another PR #447

### DIFF
--- a/src/components/SignInSteps/AccountStep.style.ts
+++ b/src/components/SignInSteps/AccountStep.style.ts
@@ -1,10 +1,15 @@
 import { Button, Text, RadioButton } from '@/shared/components'
+import Spinner from '@/shared/components/Spinner'
 import { sizes, colors, typography, transitions } from '@/shared/theme'
 import styled from '@emotion/styled'
 
 type AccountWrapperProps = {
   isSelected?: boolean
 }
+
+export const StyledSpinner = styled(Spinner)`
+  margin: ${sizes(20)} auto;
+`
 
 export const IconGroup = styled.div`
   display: flex;
--- a/src/components/SignInSteps/AccountStep.tsx
+++ b/src/components/SignInSteps/AccountStep.tsx
@@ -5,6 +5,7 @@ import { transitions } from '@/shared/theme'
 import React, { FormEvent, useState } from 'react'
 import { CSSTransition, SwitchTransition } from 'react-transition-group'
 import {
+  StyledSpinner,
   AccountStepImg,
   AccountsWrapper,
   AccountWrapper,
@@ -22,6 +23,7 @@ import joystreamIcon from '@/assets/logo.png'
 import { StepFooter, BottomBarIcon, StepSubTitle, StepTitle, StepWrapper } from './SignInSteps.style'
 import { useNavigate } from 'react-router'
 import { SvgGlyphNewChannel, SvgOutlineConnect } from '@/shared/icons'
+import { useMemberships } from '@/api/hooks'
 
 type AccountStepProps = {
   nextStepPath: string
@@ -32,6 +34,16 @@ const AccountStep: React.FC<AccountStepProps> = ({ nextStepPath }) => {
   const { setActiveUser } = useActiveUser()
   const { accounts } = useJoystream()
   const [selectedAccountAddress, setSelectedAccountAddress] = useState<undefined | string>()
+
+  const { memberships: accountsWithMemberships, loading } = useMemberships({
+    where: {
+      controllerAccount_in: accounts.map((a) => a.id),
+    },
+  })
+
+  const accountsWithNoMembership = accounts.filter(
+    (el) => !accountsWithMemberships?.map((a) => a.controllerAccount).includes(el.id)
+  )
 
   const handleSubmitSelectedAccount = async (e: FormEvent) => {
     e.preventDefault()
@@ -49,15 +61,17 @@ const AccountStep: React.FC<AccountStepProps> = ({ nextStepPath }) => {
   const handleSelect = (id: string) => {
     setSelectedAccountAddress(id)
   }
-
+  if (loading) {
+    return <StyledSpinner />
+  }
   return (
     <SwitchTransition>
       <CSSTransition
-        key={!accounts?.length ? 'no-accounts' : 'accounts'}
+        key={!accountsWithNoMembership?.length ? 'no-accounts' : 'accounts'}
         classNames={transitions.names.fadeAndSlide}
         timeout={parseInt(transitions.timings.routing)}
       >
-        {!accounts?.length ? (
+        {!accountsWithNoMembership?.length ? (
           <StepWrapper centered withBottomBar>
             <AccountStepImg src={accountCreation} />
             <Spinner size="small" />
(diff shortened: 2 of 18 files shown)


src/components/SignInSteps/AccountStep.style.ts
```diff
@@ -1,11 +1,16 @@
 import { Button, Text, RadioButton } from '@/shared/components'
+import Spinner from '@/shared/components/Spinner'
```
kdembler, line 2: Should use `@/shared/components` I think

src/components/SignInSteps/AccountStep.tsx
```diff
@@ -33,6 +35,16 @@ const AccountStep: React.FC<AccountStepProps> = ({ nextStepPath }) => {
   const { accounts } = useJoystream()
   const [selectedAccountAddress, setSelectedAccountAddress] = useState<undefined | string>()
 
+  const { memberships: accountsWithMemberships, loading } = useMemberships({
```
kdembler: This will be an array of memberships, not accounts, the name is misleading

src/components/SignInSteps/AccountStep.tsx
```diff
@@ -33,6 +35,16 @@ const AccountStep: React.FC<AccountStepProps> = ({ nextStepPath }) => {
...
+    where: {
+      controllerAccount_in: accounts.map((a) => a.id),
+    },
+  })
+
+  const accountsWithNoMembership = accounts.filter(
+    (el) => !accountsWithMemberships?.map((a) => a.controllerAccount).includes(el.id)
```
kdembler: This can be done just once instead of every iteration of filter `!accountsWithMemberships?.map((a) => a.controllerAccount)`

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -1,62 +1,98 @@
...
 
-  const handleCreateMember = handleSubmit((data) => {
-    setLoading(true)
-    // create member here
+  useEffect(() => {
+    if (!isLoadingDialogVisible) {
+      return
+    }
+    if (status === 'error' && error) {
+      error.startsWith('Member name') && setInputError('handle', { message: error })
```
kdembler: We get an error code, but you translate it into a message and then do string comparison on that message :P It's always a lot safer to operate on a code instead of a message string

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -111,4 +150,58 @@ const CreateMemberView = () => {
...
+      },
+      body,
+    })
+    const json = await response.json()
+    if (json.error) {
+      throw Error(json.error)
+    }
+    return json
+  } catch (error) {
+    const errorMessage = handleFaucetErrors(error.message)
```
kdembler: Let's skip that here - this is logic, it's better to return computer-parseable code in case you need to use it again. The "translation" part can be handled by the render logic

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -97,9 +134,11 @@ const CreateMemberView = () => {
           error={!!errors.about}
           helperText={errors.about?.message}
         />
-        <StyledButton type="submit">Create membership</StyledButton>
+        <StyledButton disabled={nodeConnectionStatus !== 'connected'} type="submit">
+          Create membership
+        </StyledButton>
       </Form>
-      <ActionDialog showDialog={loading} exitButton={false}>
+      <ActionDialog showDialog={isLoadingDialogVisible} exitButton={false}>
```
kdembler, line 151: Let's do this for errors: Let's skip displaying them in the form altogether. And we can probably remove the friendly messages as well. In theory, our UI logic should make sure none of the errors ever shows up because we do all the checks on our own. So if there's an error from the faucet let's just show an error dialog with something like this. `Some unexpected error occurred. Error code: HandleTooLong`

src/components/SignInSteps/AccountStep.style.ts
drillprop: I can't do this. I'm getting: 
image

drillprop:
Not working now after redirect PR merge - will look at it and try to fix it

drillprop:
One thing is still not working - If you create a member on `membership/new` it won't redirect you to `/signin`. You need to refresh the page to see the membership hub. 
I couldn't figure out how to make it right.

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -1,62 +1,109 @@
...
 import TextArea from '@/shared/components/TextArea'
 import { textFieldValidation, urlValidation } from '@/utils/formValidationOptions'
 import { debounce } from 'lodash'
 import React, { useEffect, useState } from 'react'
 import { useForm } from 'react-hook-form'
 import { useNavigate } from 'react-router'
+import { FAUCET_URL } from '@/config/urls'
 import { Form, StyledButton, Wrapper, StyledText, Header, Hero, SubTitle, StyledAvatar } from './CreateMemberView.style'
+import { useMemberships } from '@/api/hooks'
+import MessageDialog from '@/components/Dialogs/MessageDialog'
```
kdembler: Should be added to import above `@/components/Dialogs`

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -1,62 +1,109 @@
...
-      clearTimeout(timeout)
+  const [shouldFetchMemberships, setShouldFetchMemberships] = useState(false)
+  const [error, setError] = useState<string | undefined>()
+  const [avatarImageUrl, setAvatarImageUrl] = useState('')
+  const [isLoadingDialogVisible, setIsLoadingDialogVisible] = useState(false)
+
+  const { memberships, startPolling } = useMemberships(
+    {
+      where: {
+        controllerAccount_in: [activeUser.accountId || ''],
```
kdembler, line 44: Why not `controllerAcount_eq`?

src/views/studio/StudioLayout.tsx
```diff
@@ -91,25 +98,25 @@ const StudioLayout = () => {
                 <PrivateRoute
                   path={relativeRoutes.studio.signIn()}
                   element={<SignInView />}
-                  isAuth={!memberSet && accountSet}
+                  isAuth={!!memberships?.length && !memberSet}
```
kdembler: If we need this membership condition in each of those, let's put that in a variable

src/views/studio/StudioLayout.tsx
```diff
@@ -140,6 +147,12 @@ const StudioLayout = () => {
   )
 }
 
+const fullWidthLayout = css`
+  :root {
+    --sidenav-collapsed-width: 0;
```
kdembler: Are you sure this is a good idea to change this for the full studio layout? Some positioning depends on it (like snackbars). If we just set it to 0, some styling will likely break when the sidenav is visible

src/views/studio/CreateMemberView/CreateMemberView.tsx
drillprop: if I change that it just won't redirect the user to the `/signin` after creating a member. It will go back to `/sign/join`

kdembler: Right, because then this would a different query than the one used for route auth so it wouldn't be refetched automatically. This is tricky, but let's keep it then. We should add a comment here to explain what's happening though

kdembler (approved):
Awesome job. Still some rough edges but it's a great improvement

Issue #457, opened 2021-04-16
Improve Tooltip component


kdembler:
I think our Tooltip component could be improved, both in terms of styling and how it's coded.

Styling changes are minor but there are some differences with the design. The most obvious one is the font weight, we should make sure all the other things are matching as well: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5976%3A7902

Regarding the coding side, the problem with the current implementation is that the Tooltip component creates two wrapper elements on top of the wrapped element. This creates issues with the layout and styling - it's hard to get to the desired layout and you may have to manually override styles of inner wrapper of the Tooltip, for example as is done in the `ActionBar`: https://github.com/Joystream/atlas/blob/beea8eba4671bd637f29d76784da1fa3431bf127/src/shared/components/ActionBar/ActionBar.style.ts#L86-L93

To get rid of that problem, we could create a Tooltip that doesn't create any wrapper element but displays the wrapped element directly. This can be achieved by displaying the tooltip in some separate overlay layer. It will also require attaching some event handlers to the displayed children. This technique can be used for it: https://frontarm.com/james-k-nelson/passing-data-props-children/#cloning-children. Also for reference take a look at the Material-UI solution for this: https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tooltip/Tooltip.js

Let me know if you have any questions or need some help with that

Also seems the story is broken a bit

PR #481, opened 2021-04-20
Tooltip improvement

mikkio-j:
Issue: #457 
Figma: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5976%3A7921](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5976%3A7921)

### DIFF
--- a/src/shared/components/Tooltip/Tooltip.style.ts
+++ b/src/shared/components/Tooltip/Tooltip.style.ts
@@ -1,95 +1,59 @@
 import styled from '@emotion/styled'
-import { css } from '@emotion/react'
-import { colors, sizes, typography, zIndex, transitions } from '@/shared/theme'
+import { colors, sizes, typography } from '@/shared/theme'
 
-type TooltipProps = {
-  isActive?: boolean
-  arrowDisabled?: boolean
-  above?: boolean
-  right?: boolean
-  offsetY?: number
-}
+export const IconWrapper = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: ${sizes(5)};
+  height: ${sizes(5)};
+  margin-right: ${sizes(1)};
+`
 
-type ChildrenContainerProps = {
-  darkenContent?: boolean
-}
-
-const positionFromProps = ({ above, right: rightProp, isActive }: TooltipProps) => {
-  const top = above ? '0' : 'auto'
-  const bottom = above ? 'auto' : '0'
-  const right = rightProp ? '0' : 'auto'
-  const left = rightProp ? 'auto' : '0'
-
-  return css`
-    top: ${top};
-    bottom: ${bottom};
-    left: ${left};
-    right: ${right};
-    opacity: ${isActive ? 1 : 0};
-  `
-}
-
-const transformTooltipFromProps = ({ above, offsetY = 0, isActive }: TooltipProps) => {
-  const translate = above ? `calc(-100% + ${offsetY}px)` : `calc(100% + 10px + ${offsetY}px)`
-  return css`
-    transform: translateY(${translate}) scale(${isActive ? 1 : 1.1});
-  `
-}
-
-const transformArrowFromProps = ({ above, offsetY = 0, isActive }: TooltipProps) => {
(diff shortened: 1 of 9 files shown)


src/shared/components/Tooltip/Tooltip.tsx
```diff
@@ -1,49 +1,56 @@
...
-        onMouseLeave={() => setActive(false)}
-        onFocus={() => setActive(false)}
-        onBlur={() => setActive(false)}
-        darkenContent={darkenContent}
-        tabIndex={0}
-      >
-        {children}
-      </ChildrenContainer>
-    </StyledTooltip>
+      <span tabIndex={0}>{children}</span>
```
kdembler: Why do we need this `span` wrapper?

src/shared/components/Tooltip/Tooltip.style.ts
```diff
@@ -1,95 +1,59 @@
...
-  display: inline-block;
-  width: 100%;
-  transition: filter 200ms ${transitions.easing};
-  &:hover {
-    filter: ${({ darkenContent }) => darkenContent && 'brightness(90%)'};
+export const StyledTooltip = styled.div`
+  display: inline-flex;
+  padding: ${sizes(2)};
+  background-color: ${colors.gray[400]};
+  span {
```
kdembler: Let's create a separate styled component for the text, we shouldn't rely on the component using some specific element for rendering. If later we'd want to change the element rendered for this text variant, this style would break

src/shared/components/Tooltip/Tooltip.tsx
mikkio-j: @kdembler  This is a workaround for children that are not forwarding the ref. Besides that, Tippy doesn't accept `ReactNode` as a child so had to do `children as React.ReactElement` assertion as well. What do you think, should we get rid of the span (this will bring back constraint regarding ref)?

kdembler: Ah right, I see. That's an interesting question - what is more painful to work with, needing refs on all components or the wrapper that we need to style and take into account in the layout

kdembler: We can try it with the wrapper, should be much less painful now that there is only one wrapper

mikkio-j: Not sure if span will need additional styling, Tippy might take care of that:
[https://github.com/atomiks/tippyjs-react#component-children](https://github.com/atomiks/tippyjs-react#component-children)

kdembler: Well, it's an additional element that's getting rendered so you may need some different layouts in some places but that's fine. Let's just add `className` to the span

src/shared/components/Tooltip/Tooltip.tsx
```diff
@@ -1,49 +1,59 @@
...
+      placement={placement}
+      offset={[offsetX, offsetY]}
+      render={(attrs) => (
+        <CSSTransition
+          in={isVisible}
+          timeout={parseInt(transitions.timings.sharp)}
+          classNames={transitions.names.fade}
+          unmountOnExit
+        >
+          <StyledTooltip {...attrs} className={className}>
```
kdembler:
```suggestion
          <StyledTooltip {...attrs}>
```

src/shared/components/Tooltip/Tooltip.tsx
```diff
@@ -1,49 +1,59 @@
...
-        onMouseLeave={() => setActive(false)}
-        onFocus={() => setActive(false)}
-        onBlur={() => setActive(false)}
-        darkenContent={darkenContent}
-        tabIndex={0}
-      >
-        {children}
-      </ChildrenContainer>
-    </StyledTooltip>
+      <span tabIndex={0}>{children}</span>
```
kdembler:
```suggestion
      <span tabIndex={0} className={className}>{children}</span>
```

kdembler (approved):

Issue #460, opened 2021-04-19
General integration for video edit sheet


kdembler:
The video sheet currently has mostly only UI integration. We need to:
- integrate it with the draft system so editing and handling of drafts is fully integrated.
- make sure the validation works properly and handles all cases
- add delete button for existing videos
- handle closing the video sheet when there are open drafts (as seen here https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5863%3A977)
- handle draft interactions from the my videos view - opening in a new tab and deleting

As a result of this task, the video upload sheet should be ready for blockchain integration.

PR #493, opened 2021-04-22
Video sheet draft integration

mikkio-j:
Issue: #460

### DIFF
--- a/src/hooks/useDrafts/utils.ts
+++ b/src/hooks/useDrafts/utils.ts
@@ -22,7 +22,7 @@ export const addDraft = async (draftProps: Omit<Draft, 'updatedAt' | 'id'>) => {
   return newDraft
 }
 
-export const updateDraft = async (draftId: string, draftProps: Omit<RawDraft, 'channelId'>) => {
+export const updateDraft = async (draftId: string, draftProps: RawDraft) => {
   const currentDrafts = await getDrafts()
   const updatedAt = new Date().toISOString()
   const newDrafts = currentDrafts.map((draft) =>
--- a/src/hooks/useDrafts/utils.ts
+++ b/src/hooks/useDrafts/utils.ts
@@ -22,7 +22,7 @@ export const addDraft = async (draftProps: Omit<Draft, 'updatedAt' | 'id'>) => {
   return newDraft
 }
 
-export const updateDraft = async (draftId: string, draftProps: RawDraft) => {
+export const updateDraft = async (draftId: string, draftProps: Omit<RawDraft, 'channelId'>) => {
   const currentDrafts = await getDrafts()
   const updatedAt = new Date().toISOString()
   const newDrafts = currentDrafts.map((draft) =>
--- a/src/hooks/useEditVideoSheet.tsx
+++ b/src/hooks/useEditVideoSheet.tsx
@@ -137,7 +137,10 @@ export const useVideoEditSheetRouting = (): Location => {
     }
     setCachedSheetState(sheetState)
 
-    if (sheetState === 'minimized' && cachedSheetState === 'open') {
+    if (
+      (sheetState === 'minimized' && cachedSheetState === 'open') ||
+      (sheetState === 'closed' && cachedSheetState !== 'minimized')
+    ) {
       // restore the old location when sheet was minimized/closed
       const oldLocation = location.state?.overlaidLocation ?? absoluteRoutes.studio.index()
       navigate(oldLocation)
--- a/src/hooks/useEditVideoSheet.tsx
+++ b/src/hooks/useEditVideoSheet.tsx
@@ -137,7 +137,7 @@ export const useVideoEditSheetRouting = (): Location => {
     }
     setCachedSheetState(sheetState)
 
-    if (sheetState === 'minimized' || (sheetState === 'closed' && cachedSheetState !== 'minimized')) {
+    if (sheetState === 'minimized' && cachedSheetState === 'open') {
       // restore the old location when sheet was minimized/closed
       const oldLocation = location.state?.overlaidLocation ?? absoluteRoutes.studio.index()
       navigate(oldLocation)
--- a/src/shared/components/Button/Button.tsx
+++ b/src/shared/components/Button/Button.tsx
@@ -3,8 +3,10 @@ import { ButtonIconWrapper, StyledButtonBase, StyledText } from './Button.style'
 import { TextVariant } from '../Text'
 import { ButtonBaseProps, ButtonSize } from '../ButtonBase'
 
+export type TextColorVariant = 'default' | 'error'
 export type ButtonProps = {
   icon?: React.ReactNode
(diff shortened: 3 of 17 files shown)


kdembler (commented):
Looking good overall! Will do the code review later in the day, notes for now:
1) Assuming I have 2 drafts open and I close the currently selected one, the other one should get selected. Currently the closed draft's details are still visible in the form
1) If I close the draft that is not currently selected, it gets selected on close 😅 
1) The open in new tab button doesn't have a `pointer`
1) Opening draft in new tab shouldn't mark it as selected, just add it as a tab
1) If the sheet is closed, the open in new tab should make it open (as minimized)
1) The `Delete video` should be visible only when editing an existing video (not a draft)
1) Seems that according to the design the delete video shouldn't span the full width - https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=7389%3A6887
1) When I select a video file, it seems the tab name properly updates to that file's name but the draft name in the form stays `New Draft` and same in drafts tab? I would expect all of them to update
1) I think we don't have this ⬇️ Do you think we should add it?
Screenshot 2021-04-23 at 12 45 51

drillprop (commented):
Didn't find anything besides things that were mentioned before. Awesome job! 👍🏻  I Will look at it again after you fix the issues

mikkio-j:
@kdembler Added fixes. Some things I want to mention regarding the notes:

~~5. Sheet is now minimized on open tab click, but it also redirects to "All videos" tab. Still looking for reason for that.~~
7. Design seems to be inconsistent, Delete button takes full width here: [https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=7426%3A19746](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=7426%3A19746) 
9. As I understand it, it should show up when user's trying to make a new transaction? Maybe it's better to implement that in blockchain integration task?

src/hooks/useEditVideoSheet.tsx
```diff
@@ -137,7 +137,7 @@ export const useVideoEditSheetRouting = (): Location => {
     }
     setCachedSheetState(sheetState)
 
-    if (sheetState === 'minimized' || (sheetState === 'closed' && cachedSheetState !== 'minimized')) {
+    if (sheetState === 'minimized' && cachedSheetState === 'open') {
```
kdembler: Why did you remove the other condition? Do you understand why it was there?

src/hooks/useDrafts/utils.ts
```diff
@@ -22,7 +22,7 @@ export const addDraft = async (draftProps: Omit<Draft, 'updatedAt' | 'id'>) => {
   return newDraft
 }
 
-export const updateDraft = async (draftId: string, draftProps: RawDraft) => {
+export const updateDraft = async (draftId: string, draftProps: Omit<RawDraft, 'channelId'>) => {
```
kdembler: Why does draft not have the channel information now? Is this captured in some other way? We wouldn't want the drafts to be global across channels right?

src/shared/components/Button/Button.tsx
```diff
@@ -3,8 +3,10 @@ import { ButtonIconWrapper, StyledButtonBase, StyledText } from './Button.style'
 import { TextVariant } from '../Text'
 import { ButtonBaseProps, ButtonSize } from '../ButtonBase'
 
+export type TextColorVariant = 'default' | 'error'
```
kdembler: Let's move this to `Button.style` to not create a two-way dependencies

src/shared/components/VideoPreviewBase/PullUp.tsx
```diff
@@ -12,8 +12,8 @@ type PullUpProps = StateProps & {
 
 export const PullUp: React.FC<PullUpProps> = ({ disabled, onClick }) => (
   <Tooltip text={disabled ? 'Video already added to edit list' : 'Open in new tab'} placement="top-start">
-    <Container disabled={disabled} onClick={onClick}>
-      <IconButton disabled={disabled} variant="tertiary">
+    <Container disabled={disabled}>
```
kdembler: Do we even need this?

src/utils/formValidationOptions.ts
```diff
@@ -43,3 +44,18 @@ export const urlValidation: (name: string) => RegisterOptions = (name) => ({
     message: `${name} must be a valid url`,
   },
 })
+
+// Validates DD/MM/YYYY formatted dates
+export const dateValidation = (date: string, required = false) => {
+  if (!date) return !required
+
+  const reversedDate = date.split('/').reverse().join('-')
+  const parsedDate = parseISO(reversedDate)
```
kdembler: Is this really needed? I think `date-fns` should be able to parse an arbitrary date format? https://date-fns.org/v2.21.1/docs/parse

src/utils/formValidationOptions.ts
```diff
@@ -43,3 +44,18 @@ export const urlValidation: (name: string) => RegisterOptions = (name) => ({
     message: `${name} must be a valid url`,
   },
 })
+
+// Validates DD/MM/YYYY formatted dates
+export const dateValidation = (date: string, required = false) => {
```
kdembler: Let's call this `pastDateValidation` since that's what it's doing

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -1,11 +1,17 @@
...
+import {
+  FormContainer,
+  StyledHeaderTextField,
+  StyledRadioContainer,
+  DeleteVideoContainer,
+  DeleteVideoButton,
+} from './EditVideoForm.style'
 
 const visibilityOptions: SelectItem[] = [
   { name: 'Public (Anyone can see this video)', value: 'public' },
```
kdembler: This could use `true/false` as values, the same as we do in channel edit. Then we don't need to do comparison when saving/loading the draft

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -58,25 +73,31 @@ export const EditVideoSheet: React.FC = () => {
...
+  })
 
   // Tabs
-  const { addDraft } = useDrafts('video', channelId)
+  const { addDraft, drafts, getDraft, updateDraft } = useDrafts('video', channelId)
 
   const selectTab = useCallback(
-    (tab: EditVideoSheetTab) => {
+    async (tab: EditVideoSheetTab) => {
+      const currentDraft = await getDraft(tab.id)
```
kdembler: I don't think we need to fetch the draft directly from the local storage here. The hook exposes `drafts` which should be a list of drafts , we can use that.

kdembler (changes requested):

kdembler (approved):

PR #494, opened 2021-04-22
Test image downsizing

### LINKED ISSUE #468: Test image downsizing
The cropping library we use offers some options for specifying desired size (in pixels) of the cropped image. I specified some values for it. We should do a test in which we check if the images are properly downsized. That is, if user for example selects super detailed image that's 500MB in size, cropping should scale that down so the assets' sizes make sense. We should create a playground to test that

drillprop:
Fix #468

Some images do not work. For example: [~100mb](https://upload.wikimedia.org/wikipedia/en/f/f3/Franti%C5%A1ek_Kupka_-_Katedr%C3%A1la_-_Google_Art_Project.jpg) , [~300mb](https://upload.wikimedia.org/wikipedia/commons/7/76/Vincent_van_Gogh_-_De_slaapkamer_-_Google_Art_Project.jpg)
On firefox, if you try using image cropping after you click confirm, you will see nothing. On chrome, the cropper dialog will not display the image at all and it will break if you touch zoom controls.
Also worth noticing - for some reason, if you open these images on chrome, you will see a small square 16px on 16px(or something). 
Not sure if I should open an issue for that. I think these images are just broken somehow.

### DIFF
--- a/src/components/Dialogs/ImageCropDialog/cropper.ts
+++ b/src/components/Dialogs/ImageCropDialog/cropper.ts
@@ -145,14 +145,18 @@ export const useCropper = ({ imageEl, imageType }: UseCropperOpts) => {
         return
       }
       const imageCropData = cropper.getCropBoxData()
-      cropper.getCroppedCanvas(CANVAS_OPTS_PER_TYPE[imageType]).toBlob((blob) => {
-        if (!blob) {
-          console.error('Empty blob from cropped canvas', { blob })
-          return
-        }
-        const url = URL.createObjectURL(blob)
-        resolve([blob, url, imageCropData])
-      })
+      cropper.getCroppedCanvas(CANVAS_OPTS_PER_TYPE[imageType]).toBlob(
+        (blob) => {
+          if (!blob) {
+            console.error('Empty blob from cropped canvas', { blob })
+            return
+          }
+          const url = URL.createObjectURL(blob)
+          resolve([blob, url, imageCropData])
+        },
+        'image/jpeg',
+        0.7
+      )
     })
   }
 
--- a/src/components/Dialogs/ImageCropDialog/cropper.ts
+++ b/src/components/Dialogs/ImageCropDialog/cropper.ts
@@ -145,18 +145,14 @@ export const useCropper = ({ imageEl, imageType }: UseCropperOpts) => {
         return
       }
       const imageCropData = cropper.getCropBoxData()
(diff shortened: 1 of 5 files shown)


kdembler:
Do you think we could add some info to the playground? Like original file size and cropped file size?

drillprop:
> Do you think we could add some info to the playground? Like original file size and cropped file size?

Added

src/views/playground/Playgrounds/ImageDownsizing.tsx
```diff
@@ -147,9 +147,8 @@ const ImageDownsizing = () => {
             Original image size: {originalAvatarSizes?.width} X {originalAvatarSizes?.height}
           </p>
           <p>Original image filesize: {formatBytes(originalAvatarSizes?.fileSize || 0)}</p>
-          <p>
-            Cropped image size: {avatarSizes.width} X {avatarSizes.height}
-          </p>
+          {/* constant, they are always the same - defined by CANVAS_OPTS_PER_TYPE in cropper.ts  */}
+          <p>Cropped image size: 256 X 256</p>
```
kdembler: I think the point here is to make sure (test) that they actually are of the specified size. Can we just read that from the image?

src/components/Dialogs/ImageCropDialog/cropper.ts
```diff
@@ -145,14 +145,18 @@ export const useCropper = ({ imageEl, imageType }: UseCropperOpts) => {
...
+        (blob) => {
+          if (!blob) {
+            console.error('Empty blob from cropped canvas', { blob })
+            return
+          }
+          const url = URL.createObjectURL(blob)
+          resolve([blob, url, imageCropData])
+        },
+        'image/jpeg',
+        0.7
```
kdembler: Do we need to go that low? I wouldn't want the images to get too noisy, especially for the cover images. `toBlob` specifies some default here, doesn't that work well?

src/views/playground/Playgrounds/ImageDownsizing.tsx
kdembler: I think we can get that easily with `naturalHeight` & `naturalWidth`?

src/components/Dialogs/ImageCropDialog/cropper.ts
drillprop: work well, fixed

kdembler (approved):

PR #499, opened 2021-04-22
add uploads manager

### LINKED ISSUE #371: Create uploads manager
Once the user goes through some of the flows (like creating video or channel), we will need to upload some resources to the storage providers. We should create some logic that will handle that. This needs to be some kind of a context that exposes a hook to start a new upload and offers info on progress of the existing uploads. Let's start this simple and improve along the way

This possibly should integrate with work that has been done here: https://github.com/Joystream/atlas/issues/267 It's not a strong requirement though, it's possible it will be easier without doing that integration and that works as well.

Let's create a playground for this so we can test it out with minimum interaction with existing views at first

kdembler:
Merged with the older `useUploadingFilesData`

Resolves #371


public/mockServiceWorker.js
```diff
@@ -194,7 +194,10 @@ async function getResponse(event, client, requestId) {
 
   switch (clientMessage.type) {
     case 'MOCK_SUCCESS': {
-      return delayPromise(() => respondWithMock(clientMessage), clientMessage.payload.delay)
+      return delayPromise(
```
kdembler, line 197: This file is included in `.prettierignore` but my IDE keeps formatting it and I cannot make it stop. Let's just format this for my sake 🙏

src/hooks/useUploadsManager/store.ts
```diff
@@ -0,0 +1,65 @@
...
+      return [...state, action.asset]
+    case 'UPDATE_ASSET':
+      return state.map((asset) => {
+        if (asset.contentId !== action.contentId) {
+          return asset
+        }
+
+        const assetUpdates = {
+          ...(action.lastStatus ? { lastStatus: action.lastStatus } : {}),
+        }
```
drillprop: `const assetUpdates = action.lastStatus ? { lastStatus: action.lastStatus } : {}` isn't enough?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -179,33 +186,40 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
     setTransactionStatus(ExtrinsicStatus.ProcessingAssets)
 
     const assets: ChannelAssets = {}
+    let avatarContentId = ''
+    let coverContentId = ''
+
     if (dirtyFields.avatar) {
       if (data.avatar.blob && avatarHashPromise) {
-        assets.avatar = {
+        ;[assets.avatar, avatarContentId] = joystream.createFileAsset({
```
drillprop: This semicolon is making me feel uncomfortable 😅   At first I thought that it's a mistake,  but I guess it works.

drillprop (approved):
Found one small thing. Besides that works well.  
Feels great to see uploaded avatars in the top bar 🎉 
image

mikkio-j (approved):
Took me a while to get through this but it looks good. Did some testing and it also went well  🎉 🎉

src/hooks/useUploadsManager/store.ts
kdembler: It is, I also updated other fields here previously, that's why it was this way

kdembler: Changed

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
kdembler: It's not only not a mistake, it's actually required here. This is one of the exceptions in which you need to explicitly add a semicolon even though we are not usually using them. This is because the line is starting with `[`. I've changed it now so we don't need that

kdembler: Actually, thinking about it now, in this exact setup the semicolon may not be required because it's a first line after an opening brace, so would likely be parsed properly but Prettier will insert a semicolon at the start of any line that starts with a `[` or `(`

drillprop (approved):

Issue #469, opened 2021-04-19
Handle video edit sheet covering different view's action bar 


kdembler:
Consider this scenario:
1. I go to channel view and makes some edits in the form
2. The action bar shows up to allow me to save changes to my channel
3. I open the video edit sheet with "new video" button
4. The action sheet slides up
5. I minimize action bar, so it leaves out the draft tabs. Now both action sheet and action bar try to occupy the same space in the layout and will cover each other.

We need this to be handled smoothly - the underlying action bar probably needs to fade out when the action sheet opens and fade back in when it closes/minimizes. Given it has higher `z-index` the action bar will be on top and cover the minimized tabs until its closed - this is desired behaviour

PR #501, opened 2021-04-23
add Actionbar fade out on video sheet enter

mikkio-j:
Issue: #469

### DIFF
--- a/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
+++ b/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
@@ -402,7 +402,7 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
             <CSSTransition
               in={sheetState !== 'open'}
               timeout={2 * parseInt(transitions.timings.loading)}
-              classNames={'actionbar'}
+              classNames={transitions.names.fade}
               unmountOnExit
             >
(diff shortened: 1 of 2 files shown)


src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -396,15 +399,22 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
...
-              fullWidth={!channelId}
-              primaryButtonText={newChannel ? 'Create channel' : 'Publish changes'}
-              secondaryButtonText="Cancel"
-              onCancelClick={() => reset()}
-              onConfirmClick={handleSubmit}
-            />
+            <CSSTransition
+              in={sheetState !== 'open'}
+              timeout={2 * parseInt(transitions.timings.loading)}
+              classNames={'actionbar'}
```
kdembler: Is this different than the regular `fade`?

mikkio-j: It has `transitions.timings.regular` delay on enter so it doesn't fade in immediately when the sheet begins to slide down. Tried to overwrite `fade` with delay property, CSSTransition didn't seem to work with that approach

update: made it work 😄

kdembler (approved):

PR #504, opened 2021-04-23
Signin tweaks

### LINKED ISSUE #448: sign in view UI tweaks
Followups from #447:

1) The multistepper design wasn't updated (step headers)
Screenshot 2021-04-14 at 11 25 38

2) In the create account step:
a) the `add account` image needs to be exported as SVG, it's pixelated
b) we shouldn't show the spinner
c) we're lacking spacing between the header and `follow instructions`
d) the step number circles don't seem to be properly sized and the text also doesn't seem fully centred

3) In the select account step:
a) both the polkadot and joystream icons need to be svgs, they're pixelated
b) spacing between account select items seems excessive
c) seems we're unnecessarily capping the `max-width` of the address, why not display it fully if there's space?
d) let's use the `glyph-channel` icon for the account item

4) There seems to be strange border next to the checkbox (Firefox). Possibly related to the `Checkbox` itself, but let's get rid of it:
Screenshot 2021-04-14 at 11 45 22

5) In the member create form we could use some more spacing between the inputs:
Screenshot 2021-04-14 at 11 48 01

6) Spinner in the membership create dialog seems too big

7) The sign in views don't look very good on bigger screens. We should use the same styling as for the rest of studio to restrict max width

drillprop:
Fix #448 
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6826%3A165271

regarding: 
> a) the add account image needs to be exported as SVG, it's pixelated

I had problems getting the working SVG there. When I add this I see image which is incomplete:  
image
And the original SVG looks like this: 
image
Was experimenting with values inside the SVG, but didn't manage to fix it. Also, it's worth mentioning that if you export this image from Figma it won't be a pure SVG, it will contain some raster images inside - like for example polkadot logo. 

> There seems to be strange border next to the checkbox (Firefox). Possibly related to the Checkbox itself, but let's get rid of it:

Didn't find this. Is it still actual? Also maybe I should get rid of the checkbox entirely there like in the design? 

image


kdembler:
Regarding the checkbox, I cannot longer see the problem I've brought up so maybe it was fixed along the way.

Regarding the polkadot account image, we have to do something about it, this doesn't look good:
Screenshot 2021-04-23 at 16 07 23

I exported the SVG and it seem to work for me?
Screenshot 2021-04-23 at 16 19 25

src/views/studio/SignInJoinView/SignInProccessView/SignInProccessView.tsx
```diff
@@ -28,11 +28,11 @@ const SignInProccessView = () => {
       element: <ExtensionStep nextStepPath={absoluteRoutes.studio.signInJoin({ step: '2' })} />,
     },
     {
-      title: 'Create or select a polkadot account',
+      title: 'Connect accounts',
```
kdembler: Wait, why are those specified in two places?

kdembler (commented):
Except the account image and one comment, everything looking good!

kdembler:
Ah, just one more thing, when the membership doesn't have an avatar we should just use the regular avatar placeholder instead of what we have:
Screenshot 2021-04-23 at 16 58 29

drillprop:
Regarding this SVG - I think it was mac m1 related issue. Sometimes some SVG or images are not displaying for me correctly in chrome. Next time I'll check on the other browsers first.

src/components/SignInSteps/SignInStepsStepper.tsx
```diff
@@ -0,0 +1,41 @@
+import { useRouterQuery } from '@/hooks'
+import React from 'react'
+import { useNavigate } from 'react-router'
+import { Multistepper } from '../Dialogs'
+import AccountStep from './AccountStep'
+import ExtensionStep from './ExtensionStep'
+import TermsStep from './TermsStep'
+
+type SignInStepsStepperProps = {
+  path: (query: { step: string }) => string
```
kdembler: Does this component need this path? I think it could just use the route at which it is displayed and only change the query param?

kdembler (approved):

Issue #464, opened 2021-04-19
Integrate the uploads view


kdembler:
Once #371 is done, we should be able to use that information in the uploads view to integrate it. This view should consume two sources of information:
- data from the query nodes (e.g. query all the assets that have `pending` status - that mean they haven't been provided yet)
- data from #371 - to enrich the data from above with details like upload status, some additional file metadata etc

We also should add the view for no videos as in the design

PR #512, opened 2021-04-27
Uploads view integration

mikkio-j:
Issue: #464


src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -67,6 +68,9 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
   const [avatarHashPromise, setAvatarHashPromise] = useState<Promise<string> | null>(null)
   const [coverHashPromise, setCoverHashPromise] = useState<Promise<string> | null>(null)
 
+  const [avatarCropData, setAvatarCropData] = useState<ImageCropData>()
+  const [coverCropData, setCoverCropData] = useState<ImageCropData>()
```
kdembler: Couldn't those be part of the form state, same as `blob` and `url`?

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -10,45 +11,46 @@ import {
...
-            <CircularProgressbar value={asset.progress} />
-          </ProgressbarContainer>
-        )
+  const renderStatusIndicator = (asset: UploadData) => {
+    if (asset.liaisonJudgement === LiaisonJudgement.Pending && asset.progress === 0) {
+      return (
+        <ProgressbarContainer>
+          <CircularProgressbar value={0} />
+        </ProgressbarContainer>
+      )
```
kdembler: Do we need this? Seems the return on line 49 would return the same thing

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.tsx
```diff
@@ -13,48 +15,39 @@ import { AssetLine } from '../AssetLine'
...
-  const isChannelType = type === 'channel'
-  const isPending = files.every((file) => file.status === 'pending')
-  const hasErrorNumber = files.filter(({ status }) => status === 'failed' || status === 'reconnecting').length
+  const isChannelType = uploadData[0].parentObject.type === 'channel'
 
-  const allAssetsSize = files.reduce((acc, file) => acc + file.size, 0)
-  const alreadyUploadedSize = files.reduce((acc, file) => acc + (file.progress / 100) * file.size, 0)
+  const isPending = uploadData.every(
+    (file) => file.liaisonJudgement === LiaisonJudgement.Pending && file.progress === 0
+  )
```
kdembler: I'm a bit confused here - `Pending` in the context of judgement means that the upload was not yet provided - it can be being uploaded atm though. Here we say that the group is pending only if every file is at 0 progress. I guess this is supposed to represent something like `isWaiting`?

kdembler: I guess it's just a difference in language the data and the design use

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.tsx
```diff
@@ -13,48 +15,39 @@ import { AssetLine } from '../AssetLine'
...
+  const hasErrorNumber = uploadData.filter(({ lastStatus }) => lastStatus === 'error').length
```
kdembler: Can we use some other name? Like `errorsCount`?

src/views/studio/MyUploadsView/EmptyUploadsView.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <ContainerView>
+      <StyledEmptyUploadsIllustration />
+      <InnerContainerView>
+        <MessageView>
+          <Text variant="body2" secondary>
+            There are no uploads on your list
+          </Text>
+        </MessageView>
+        <div>
```
kdembler: Is this div needed?

src/views/studio/MyUploadsView/EmptyUploadsView.tsx
```diff
@@ -0,0 +1,55 @@
...
+    </ContainerView>
+  )
+}
+
+const ContainerView = styled.div`
+  position: relative;
+`
+
+const InnerContainerView = styled.div`
+  position: absolute;
```
kdembler: Why absolute? We could do this easier with flex I think. It slightly breaks on mobile screens

src/views/studio/MyUploadsView/EmptyUploadsView.tsx
```diff
@@ -0,0 +1,55 @@
...
+  left: 50%;
+  top: ${sizes(78)};
```
kdembler: Let's just use value directly in cases like this :P

src/views/studio/MyUploadsView/MyUploadsView.tsx
```diff
@@ -1,128 +1,65 @@
...
+  const { channel, loading: channelLoading } = useChannel(channelId)
+  const { videos, loading: videosLoading } = useVideos({
+    where: {
+      channelId_eq: channelId,
+    },
+  })
+
+  const channelDataObjects = [channel?.avatarPhotoDataObject, channel?.coverPhotoDataObject]
+  const videosDataObjects =
+    videos?.map((video) => [{ title: video.title, ...video.mediaDataObject }, video.thumbnailPhotoDataObject]).flat() ||
```
kdembler, line 28: btw, you can do the same with `flatMap` (no need to change)

src/views/studio/MyUploadsView/MyUploadsView.tsx
```diff
@@ -1,128 +1,65 @@
...
+    []
+  const allDataObjects = [...channelDataObjects, ...videosDataObjects]
+
+  // Enriching data with pending/accepted/rejected status
+  const uploadsStateWithLiaisonJudgement = uploadsState.map((asset) => {
+    const dataObject = allDataObjects.find((dataObject) => dataObject?.joystreamContentId === asset.contentId)
+    if (dataObject) {
+      return { ...asset, liaisonJudgement: dataObject.liaisonJudgement, progress: asset.progress * 100 }
```
kdembler: Maybe let's be consistent on representing the progress in data. I used 0..1, here we use 0..100, I'm fine with both as long as we use the same

kdembler (commented):
Looking good, some small tweaks.

One thing we could possibly improve is the sync with the query node. Imagine a case where you use some other computer/browser to publish a video and you don't finish an upload (or you do that in incognito). Then without access to that browser and it's data, you won't be able to ever finish the upload. We could add logic that adds missing files to the uploads manager. Doesn't have to happen in this PR though, just sharing my observation, lmk what you think

mikkio-j:
> One thing we could possibly improve is the sync with the query node. Imagine a case where you use some other computer/browser to publish a video and you don't finish an upload (or you do that in incognito). Then without access to that browser and it's data, you won't be able to ever finish the upload. We could add logic that adds missing files to the uploads manager. Doesn't have to happen in this PR though, just sharing my observation, lmk what you think

Sounds good, but I believe it can be really tricky to implement. Maybe it could be something to work on during post-sumer?

kdembler (approved):

Issue #509, opened 2021-04-26
Properly handle updates to files


kdembler:
When updating a channel and modifying an avatar/cover, it'd be best if that change is reflected in the local state of the app. After the update flow finishes, the form shouldn't anymore be marked as dirty but at the same time we cannot really use state from the query node right away, because it may not have yet processed the file upload (see #465 for more info). Instead we should do some local update, most likely to Apollo cache, to use a local URL for this session.

PR #516, opened 2021-04-28
Display updated assets

mikkio-j:
Issue: #509

### DIFF
--- a/src/components/Topbar/StudioTopbar/StudioTopbar.tsx
+++ b/src/components/Topbar/StudioTopbar/StudioTopbar.tsx
@@ -157,12 +157,7 @@ const StudioTopbar: React.FC<StudioTopbarProps> = ({ hideChannelInfo, fullWidth 
             {loading ? (
               <ChannelInfoPlaceholder />
             ) : membership?.channels.length ? (
-              <ChannelInfo
-                channel={currentChannel}
-                memberName={membership.handle}
-                onClick={handleDrawerToggle}
-                // cachedAvatarUrl={cachedAvatarUrl}
-              />
+              <ChannelInfo channel={currentChannel} memberName={membership.handle} onClick={handleDrawerToggle} />
             ) : (
               <ChannelInfoContainer onClick={handleDrawerToggle}>
                 <NewChannelAvatar newChannel size="small" />
--- a/src/components/Topbar/StudioTopbar/StudioTopbar.tsx
+++ b/src/components/Topbar/StudioTopbar/StudioTopbar.tsx
@@ -78,23 +78,19 @@ const StudioTopbar: React.FC<StudioTopbarProps> = ({ hideChannelInfo, fullWidth 
 
   const { sheetState } = useEditVideoSheet()
 
-  const cachedAvatarUrl = readUrlFromCache({
-    fileType: 'avatar',
-    channelId: activeUser?.channelId ?? '',
-    client,
-  })
-
-  const currentChannel = {
-    ...membership?.channels.find((channel) => channel.id === activeUser?.channelId),
-    cachedAvatarUrl,
-  } as ChannelFieldsWithCachedAvatar
-
   const channels = membership?.channels.map((channel) => {
-    if (channel.id === currentChannel.id) {
+    const cachedAvatarUrl = readUrlFromCache({
+      fileType: 'avatar',
+      channelId: channel.id,
+      client,
+    })
+    if (cachedAvatarUrl) {
       return { ...channel, cachedAvatarUrl }
     }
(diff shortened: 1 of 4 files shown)


src/components/Topbar/StudioTopbar/StudioTopbar.tsx
```diff
@@ -193,7 +213,7 @@ const ChannelInfo = React.forwardRef<HTMLDivElement, ChannelInfoProps>(
 
     return (
       <ChannelInfoContainer onClick={onClick} isActive={active} ref={ref}>
-        <StyledAvatar size="small" imageUrl={avatarPhotoUrl} />
+        <StyledAvatar size="small" imageUrl={avatarPhotoUrl || channel?.cachedAvatarUrl} />
```
drillprop: I found that It will not update the avatar in the top bar immediately. It will update only if you click the top bar.
Nagranie z ekranu 2021-04-29 o 10 54 30
 Can we do something about it?

src/utils/cachingAssets.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+`
+
+const cachedAvatarUrlFragment = gql`
+  fragment AvatarUrlField on Channel {
+    cachedAvatarUrl
+  }
+`
+
+export const writeUrlInCache: (arg: WriteUrlInCacheArg) => void = ({ url, fileType, channelId, client }) => {
```
drillprop:
```suggestion
export const writeUrlInCache = ({ url, fileType, channelId, client }: WriteUrlInCacheArg) => {
```
I think we don't need to specify returned type here. Typescript will figure it out

src/utils/cachingAssets.ts
```diff
@@ -0,0 +1,43 @@
...
+  client.writeFragment({
+    id: `Channel:${channelId}`,
+    fragment: fileType === 'avatar' ? cachedAvatarUrlFragment : cachedCoverUrlFragment,
+    data: {
+      [field]: url,
+    },
+  })
+}
+
+export const readUrlFromCache: (arg: ReadUrlFromCacheArg) => string = ({ fileType, channelId, client }) => {
```
drillprop, line 36: The same
```suggestion
export const readUrlFromCache = ({ fileType, channelId, client }: ReadUrlFromCacheArg) => {
```

drillprop: Actually, maybe I'm wrong. Let's leave it as it is.

drillprop (commented):
The code generally looks good. Find one minor thing which could be improved

src/components/Topbar/StudioTopbar/StudioTopbar.tsx
```diff
@@ -67,10 +72,25 @@ const StudioTopbar: React.FC<StudioTopbarProps> = ({ hideChannelInfo, fullWidth
       skip: !activeUser?.memberId,
     }
   )
+  const { client } = useChannel(activeUser?.channelId ?? '', {
+    skip: !activeUser?.channelId,
+  })
```
kdembler: How about just this instead? 😅 
```ts
```

src/components/Topbar/StudioTopbar/StudioTopbar.tsx
mikkio-j: Hmmm this doesn't happen when we're updating both avatar and cover

src/utils/cachingAssets.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+`
+
+const cachedAvatarUrlFragment = gql`
+  fragment AvatarUrlField on Channel {
+    cachedAvatarUrl
+  }
+`
+
+export const writeUrlInCache = ({ url, fileType, channelId, client }: WriteUrlInCacheArg) => {
```
kdembler, line 25: Let's use the existing fields here

src/utils/cachingAssets.ts
```diff
@@ -0,0 +1,43 @@
...
+  client.writeFragment({
+    id: `Channel:${channelId}`,
+    fragment: fileType === 'avatar' ? cachedAvatarUrlFragment : cachedCoverUrlFragment,
+    data: {
+      [field]: url,
+    },
+  })
+}
+
+export const readUrlFromCache: (arg: ReadUrlFromCacheArg) => string = ({ fileType, channelId, client }) => {
```
kdembler: And once we move away from the custom fields, I don't think we need the read function at all right?

mikkio-j: that's correct

kdembler (approved):

PR #520, opened 2021-04-28
Add checkout component to create channel view

### LINKED ISSUE #496: Add Checkout component to Create Channel View
We need to add Checkout component to CreateEditChannelView:
192 168 0 220_6006__path=_story_general-noconnectionindicator--regular

Make sure it's:
- visible only when new channel is being created
- following form and updating progress

drillprop:
Fix #496 
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6691%3A2575

### DIFF
--- a/src/shared/components/Checkout/Checkout.styles.tsx
+++ b/src/shared/components/Checkout/Checkout.styles.tsx
@@ -76,13 +76,14 @@ export const StepState = styled.div<StepStateProps>`
   ]};
 `
 
-export const Step = styled.div`
-  cursor: pointer;
+export const Step = styled.div<StepStateProps>`
+  cursor: ${({ completed }) => (completed ? 'initial' : 'pointer')};
   display: grid;
   grid-auto-flow: column;
   justify-content: space-between;
   align-items: center;
   padding: ${sizes(2)};
+  margin-bottom: ${sizes(2)};
   :last-of-type {
     position: relative;
     ::after {
@@ -94,8 +95,9 @@ export const Step = styled.div`
       background-color: ${colors.gray[300]};
     }
   }
+
   :hover {
-    background-color: ${colors.transparentPrimary[18]};
+    background-color: ${({ completed }) => !completed && colors.transparentPrimary[18]};
   }
 `
 
--- a/src/shared/components/Checkout/Checkout.styles.tsx
+++ b/src/shared/components/Checkout/Checkout.styles.tsx
@@ -86,7 +86,7 @@ export const Step = styled.div`
   :last-of-type {
     position: relative;
     ::after {
-      bottom: -12px;
+      bottom: -${sizes(4)};
       position: absolute;
       content: '';
       width: 100%;
--- a/src/shared/components/Checkout/Checkout.styles.tsx
+++ b/src/shared/components/Checkout/Checkout.styles.tsx
@@ -34,7 +34,7 @@ export const BottomRowContainer = styled.div`
   justify-content: space-between;
   align-items: center;
   position: relative;
-  z-index: 6;
+  z-index: 1;
 `
 
 type StepsContainerProps = {
@@ -106,6 +106,6 @@ export const StepsCompletedText = styled(Text)`
 export const StepInnerContainer = styled.div`
   display: grid;
   grid-auto-flow: column;
-  grid-gap: 12px;
+  grid-gap: ${sizes(3)};
   align-items: center;
 `
--- a/src/shared/components/Checkout/Checkout.styles.tsx
+++ b/src/shared/components/Checkout/Checkout.styles.tsx
@@ -1,46 +1,68 @@
-import React from 'react'
 import styled from '@emotion/styled'
-import { colors } from '@/shared/theme'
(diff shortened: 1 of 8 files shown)


src/shared/components/Checkout/Checkout.tsx
```diff
@@ -1,49 +1,55 @@
...
       </StepsContainer>
+      <BottomRowContainer>
+        <StepsProgressContainer>
+          <CircularProgresaBarContainer>
+            <StyledCircularProgressbar value={stepsCompletedNumber} maxValue={steps.length} />
+            <StepsCompletedText variant="body2">
+              {stepsCompletedNumber}/{steps.length}
+            </StepsCompletedText>
+          </CircularProgresaBarContainer>
+          <IconButton variant="secondary" onClick={() => setIsHidden(!isHidden)}>
```
mikkio-j:
```suggestion
          <IconButton variant="secondary" onClick={() => setIsHidden(!isHidden)} size="small">
```

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -58,9 +80,23 @@ export const Step = styled.div`
   cursor: pointer;
   display: grid;
   grid-auto-flow: column;
-  grid-gap: 12px;
   justify-content: space-between;
   align-items: center;
+  padding: ${sizes(2)};
+  :last-of-type {
+    position: relative;
+    ::after {
```
mikkio-j, line 89: It seems like this horizontal line is vertically misaligned  a little bit. In design there is 16px of space above and below the line

mikkio-j (commented):
just 2 minor things 👌

src/shared/components/Checkout/Checkout.tsx
```diff
@@ -1,49 +1,55 @@
...
+      <StepsContainer isHidden={isHidden}>
         {steps.map((step, idx) => (
           <Step key={step.title + idx} onClick={step.onClick}>
             <StepInnerContainer>
-              <StepState completed={step.completed}>{step.completed && <SvgGlyphCheck />}</StepState>{' '}
+              <StepState completed={step.completed}>{step.completed && <SvgGlyphCheck />}</StepState>
               <Text variant="body2">{step.title}</Text>
             </StepInnerContainer>
-            <SvgGlyphChevronDown />
+            <SvgGlyphChevronRight />
```
kdembler: The design seems to hide this when the step is completed already:
Screenshot 2021-04-30 at 08 51 46

Maybe we should also not make it clickable when already completed? Wdyt?

src/shared/components/Checkout/Checkout.styles.tsx
```diff
@@ -1,46 +1,68 @@
-import React from 'react'
 import styled from '@emotion/styled'
-import { colors } from '@/shared/theme'
+import { colors, sizes, transitions } from '@/shared/theme'
 import Text from '../Text'
+import CircularProgressbar from '../CircularProgressbar'
+import { SvgGlyphChevronDown } from '@/shared/icons'
 
 export const Container = styled.div`
```
kdembler, line 7: Not sure to which component this exactly refers to, but seems we're lacking some spacing between steps:
Screenshot 2021-04-30 at 08 53 01

src/views/studio/CreateEditChannelView/CreateEditChannelView.style.tsx
```diff
@@ -43,3 +45,13 @@ export const StyledAvatar = styled(Avatar)`
     height: 136px;
   }
 `
+
+export const StyledCheckout = styled(Checkout)`
+  display: none;
+  position: fixed;
+  bottom: ${TOP_NAVBAR_HEIGHT}px;
```
kdembler: Is this really a given that the bottom action bar will be the same height as topbar? I think this may be just a coincidence. Looking at the action bar styles, it doesn't actually specify explicit height at all, it relies on its content. So not sure, we could plug in the resize observer or something

kdembler: Or, on a second thought, maybe we could just put checkout and action bar in one shared fixed container and position them in that container with flex? Then we don't need action bar height at all? Just an idea

kdembler (commented):
Looking good, just some minor things

src/shared/components/Checkout/Checkout.tsx
drillprop: Good idea. I'll do that.

src/views/studio/CreateEditChannelView/CreateEditChannelView.style.tsx
drillprop: To accomplish that I had to do some changes to ActionBar and ActionBarTransaction. Not sure if it's the best approach, but wrapping one fixed element with another fixed element gave me some unexpected results and I couldn't get it right.

kdembler (approved):
:+1: I like this approach, it makes sense for the checkout to be paired with the action bar

PR #525, opened 2021-04-29
Handle storage providers

### LINKED ISSUE #466: Handle dynamic storage providers
Currently we use a single hard-coded (via env vars) storage provider. We need to remove that constraint. After https://github.com/Joystream/joystream/issues/2252 is done, the query node should offer the following information:
- list of all storage providers with URLs
- details of the storage provider that accepted a given asset

Given that info, we should be able to remove constraint on pre-defined storage provider and use info from the query node.

drillprop:
Fix #466


src/api/hooks/storageProviders.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+export const useRandomStorageProviderUrl = () => {
+  const { totalCount } = useStorageProvidersCount({ where: { metadata_contains: 'http' } })
+  const randomStorageIdx = totalCount && getRandomIntInclusive(1, totalCount)
+
+  const { storageProviders } = useStorageProviders(
+    {
+      where: { metadata_contains: 'http' },
```
kdembler: Very good idea

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -13,16 +12,15 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
...
       const setAssetUploadProgress = (progress: number) => {
         setUploadsProgress((prevState) => ({ ...prevState, [asset.contentId]: progress }))
       }
 
       try {
         addAsset({ ...asset, lastStatus: 'inProgress', size: file.size })
         setAssetUploadProgress(0)
-
-        const assetUrl = new URL(asset.contentId, STORAGE_NODE_URL)
+        const assetUrl = new URL(asset.contentId, storageNodeUrl + '/asset/v0/')
```
kdembler: Let's put that suffix somewhere in the `@/config`

kdembler: I think maybe we should extract this full line somewhere because we use it in two places. Let's just make it a separate function in `@/utils/assets`

src/api/hooks/storageProviders.ts
kdembler: We should also check for `isActive: true`

src/api/hooks/storageProviders.ts
```diff
@@ -0,0 +1,58 @@
...
+export const useStorageProvidersCount = (variables: GetStorageProvidersQueryVariables, opts?: StorageProvidersOpts) => {
+  const { data: connectionData, loading, ...rest } = useGetStorageProvidersCountQuery({ ...opts, variables })
+  return {
+    totalCount: connectionData?.storageProvidersConnection.totalCount,
+    loading,
+    ...rest,
+  }
+}
+
+export const useRandomStorageProviderUrl = () => {
```
kdembler, line 33: Maybe we just fetch storage providers list with some higher limit (like 100 or whatever) and just skip fetching the count? Should be simpler

kdembler (commented):
Great job, some minor comments

kdembler (approved):

PR #526, opened 2021-04-29
video edit

kdembler:
summary of changes:
- ability to edit existing videos
- improved handling of assets
- saving only drafts that had some edits
- debounced saving of drafts on values change
- a lot of small tweaks

happy to discuss any changes


src/hooks/useEditVideoSheet.tsx
```diff
@@ -1,19 +1,34 @@
...
+import { useDrafts } from '@/hooks/useDrafts'
+import { useActiveUser } from '@/hooks/useActiveUser'
+import { useVideo } from '@/api/hooks'
+import { InputFilesState } from '@/shared/components/MultiFileSelect/MultiFileSelect'
+import { createUrlFromAsset } from '@/utils/asset'
+
+export type EditVideoSheetTab = {
+  id: string
+  isDraft?: boolean
+  isFresh?: boolean
```
drillprop: 🐟  
`isFresh` means `new` here, right? Maybe `isNew` will be a more appropriate name here :P?

src/hooks/useEditVideoSheet.tsx
```diff
@@ -74,27 +159,74 @@ const useEditVideoSheetContext = () => {
...
+    title: tab.isDraft ? draft?.title ?? 'New Draft' : video?.title ?? '',
+    description: (tab.isDraft ? draft?.description : video?.description) ?? '',
+    category: (tab.isDraft ? draft?.category : video?.category?.id) ?? null,
+    language: (tab.isDraft ? draft?.language : video?.language?.iso) ?? 'en',
+    isPublic: (tab.isDraft ? draft?.isPublic : video?.isPublic) ?? true,
+    isExplicit: (tab.isDraft ? draft?.isExplicit : video?.isExplicit) ?? null,
+    hasMarketing: (tab.isDraft ? draft?.hasMarketing : video?.hasMarketing) ?? false,
+    publishedBeforeJoystream:
+      (tab.isDraft ? draft?.publishedBeforeJoystream : video?.publishedBeforeJoystream?.toISOString()) ?? null,
+    assets,
```
drillprop, line 227: got the error, after clicking one of the videos.
image
 Not sure if you can reproduce. Maybe this particular video was broken somehow, or this is because the processor is down right now

drillprop: Oh, I think it was the video. The id of the video is  "72". It has set up `publishedBeforeJoystream` field

drillprop (commented):
The code looks good, but I can't really update the video for now. 
If I'll try to do some changes, I'll get this error:
image

src/hooks/useEditVideoSheet.tsx
kdembler: Sure, that works as well

src/hooks/useEditVideoSheet.tsx
kdembler: Fixed

kdembler:
All fixed now, also improved some behaviours

drillprop (approved):
Found one more problem but I think I'll create a separate issue for that. 
If you update the video thumbnail and then click again on this video, you'll see the previous thumbnail. 
video

besides that, looks good to me!

drillprop:
Didn't notice that we already have the issue for that #528, so all good.

Issue #474, opened 2021-04-19
Handle errors in uploads


kdembler:
We should automatically handle errors in uploaded assets. For example, if the storage provider becomes unreachable, we should pick another one and start over. Full flow can be seen here: https://github.com/Joystream/atlas/issues/381#issuecomment-802704377

This will also include some additional handling in the uploads view in case the app doesn't have access to the file anymore. When the app loads and there are some pending uploads, we should display a toast, like so: 
<img width="456" alt="Screenshot 2021-04-26 at 10 47 49" src="https://user-images.githubusercontent.com/12646744/116055308-dbe05600-a67c-11eb-846e-7237645e8347.png">

The uploads manager should probably also use the query node to do some initial sync on assets

eldiegod:
Should this be fixed here?:

After clicking "start publishing" without having added a thumbnail the file drop shows this error but the upload starts and its successful 

<img width="1674" alt="Screenshot 2021-05-04 at 10 40 03" src="https://user-images.githubusercontent.com/10615263/116979459-520c3a80-acc5-11eb-86ee-55d96df351c6.png">

@kdembler

kdembler:
huh, that shouldn't happen, but no, that's a separate issue

kdembler:
fixed in #543 (issue reported by @eldiegod in a comment)

PR #530, opened 2021-04-30
Handle uploads errors

mikkio-j:
Issue: #474

### DIFF
--- a/src/hooks/useUploadsManager/types.ts
+++ b/src/hooks/useUploadsManager/types.ts
@@ -1,5 +1,6 @@
 import { ChannelId, VideoId } from '@/joystream-lib'
 import { AssetDimensions, ImageCropData } from '@/types/cropper'
+import { LiaisonJudgement } from '@/api/queries'
 
 type AssetType = 'video' | 'thumbnail' | 'cover' | 'avatar'
 type AssetParent = 'video' | 'channel'
@@ -15,11 +16,14 @@ export type AssetUpload = {
   owner: ChannelId
   type: AssetType
   lastStatus: AssetUploadStatus
+  liaisonJudgement?: LiaisonJudgement
+  ipfsContentId?: string
   // size in bytes
   size: number
   dimensions?: AssetDimensions
   imageCropData?: ImageCropData
   metadata?: string
+  title?: string | null
 }
 export type AssetUploadWithProgress = AssetUpload & {
   // progress of upload - 0...1
--- a/src/hooks/useUploadsManager/types.ts
+++ b/src/hooks/useUploadsManager/types.ts
@@ -29,10 +29,19 @@ export type InputAssetUpload = Omit<AssetUpload, 'lastStatus' | 'size'>
 
 export type UploadsManagerState = AssetUpload[]
 
+export type StartFileUploadOptions = {
+  isReUpload?: boolean
+  changeHost?: boolean
+}
+
 export type UploadManagerValue = {
   uploadsState: AssetUploadWithProgress[][]
-  removeAsset: (contentId: string) => void
-  startFileUpload: (file: File | Blob, asset: InputAssetUpload, storageMetadata: string) => void
+  startFileUpload: (
+    file: File | Blob | null,
+    asset: InputAssetUpload,
+    storageMetadata: string,
+    opts?: StartFileUploadOptions
+  ) => void
   isLoading: boolean
 }
 export type UploadsProgressRecord = Record<string, number>
--- a/src/hooks/useUploadsManager/types.ts
+++ b/src/hooks/useUploadsManager/types.ts
@@ -30,8 +30,10 @@ export type InputAssetUpload = Omit<AssetUpload, 'lastStatus' | 'size'>
 export type UploadsManagerState = AssetUpload[]
 
 export type UploadManagerValue = {
-  uploadsState: AssetUploadWithProgress[]
+  uploadsState: AssetUploadWithProgress[][]
+  removeAsset: (contentId: string) => void
   startFileUpload: (file: File | Blob, asset: InputAssetUpload, storageMetadata: string) => void
+  isLoading: boolean
 }
 export type UploadsProgressRecord = Record<string, number>
 
--- a/src/hooks/useUploadsManager/types.ts
+++ b/src/hooks/useUploadsManager/types.ts
@@ -4,7 +4,7 @@ import { AssetDimensions, ImageCropData } from '@/types/cropper'
 type AssetType = 'video' | 'thumbnail' | 'cover' | 'avatar'
 type AssetParent = 'video' | 'channel'
 
-export type AssetUploadStatus = 'completed' | 'inProgress' | 'error'
+export type AssetUploadStatus = 'completed' | 'inProgress' | 'error' | 'reconnecting' | 'reconnectionError'
 
 export type AssetUpload = {
   contentId: string
--- a/src/hooks/useUploadsManager/useUploadsManager.tsx
+++ b/src/hooks/useUploadsManager/useUploadsManager.tsx
@@ -1,26 +1,96 @@
-import React, { useCallback, useContext, useState } from 'react'
+import React, { useCallback, useContext, useState, useEffect } from 'react'
 import axios from 'axios'
 import * as rax from 'retry-axios'
 import { useNavigate } from 'react-router'
-import { useSnackbar } from '@/hooks/useSnackbar/useSnackbar'
+import { useSnackbar, useActiveUser } from '@/hooks'
+import { useChannel, useVideos } from '@/api/hooks'
 import { absoluteRoutes } from '@/config/routes'
 import { ChannelId } from '@/joystream-lib'
 import { useUploadsManagerStore } from './store'
 import { InputAssetUpload, AssetUploadWithProgress, UploadManagerValue, UploadsProgressRecord } from './types'
 import { createStorageNodeUrl } from '@/utils/asset'
-import { useSnackbar } from '@/hooks/useSnackbar'
+import { LiaisonJudgement } from '@/api/queries'
 
 const RETRIES_COUNT = 5
+const RECONNECTION_ERROR_MESSAGE = 'Reconnection failed'
+
+type GroupByParentObjectIdAcc = {
+  [key: string]: AssetUploadWithProgress[]
+}
 
 const UploadManagerContext = React.createContext<UploadManagerValue | undefined>(undefined)
 UploadManagerContext.displayName = 'UploadManagerContext'
 
 export const UploadManagerProvider: React.FC = ({ children }) => {
   const navigate = useNavigate()
-  const { uploadsState, addAsset, updateAsset } = useUploadsManagerStore()
+  const { uploadsState, addAsset, updateAsset, removeAsset } = useUploadsManagerStore()
   const { displaySnackbar } = useSnackbar()
   const [uploadsProgress, setUploadsProgress] = useState<UploadsProgressRecord>({})
-  const { displaySnackbar } = useSnackbar()
+  const { activeUser } = useActiveUser()
+  const channelId = activeUser.channelId ?? ''
+  const { channel, loading: channelLoading } = useChannel(channelId)
+  const { videos, loading: videosLoading } = useVideos(
+    {
+      where: {
+        id_in: uploadsState.filter((item) => item.parentObject.type === 'video').map((item) => item.parentObject.id),
+      },
+    },
+    { skip: !uploadsState.length }
+  )
+
+  const uploadsStateWithProgress: AssetUploadWithProgress[] = uploadsState.map((asset) => ({
+    ...asset,
+    progress: uploadsProgress[asset.contentId] ?? 0,
+  }))
+
+  const channelDataObjects = [channel?.avatarPhotoDataObject, channel?.coverPhotoDataObject]
+  const videosDataObjects = videos?.flatMap((video) => [video.mediaDataObject, video.thumbnailPhotoDataObject]) || []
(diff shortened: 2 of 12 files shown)


src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,187 @@
...
+    displaySnackbar({
+      title: `(${lostConnectionAssets.length}) Asset${
+        lostConnectionAssets.length > 1 ? 's' : ''
+      } waiting to resume upload`,
+      description: 'Reconnect files to fix the issue',
+      actionText: 'See assets',
+      onActionClick: () => navigate(absoluteRoutes.studio.uploads()),
+      iconType: 'warning',
+    })
+  }, [lostConnectionAssets.length])
```
drillprop: If we cannot do anything with it  let's just add above ` // eslint-disable-next-line react-hooks/exhaustive-deps`

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,187 @@
...
+    ) => {
       const setAssetUploadProgress = (progress: number) => {
         setUploadsProgress((prevState) => ({ ...prevState, [asset.contentId]: progress }))
       }
-
+      const isFileInState = assetsFiles?.find((file) => file.contentId === asset.contentId)
+      if (!isFileInState && file) {
+        setAssetsFiles([...assetsFiles, { contentId: asset.contentId, blob: file }])
+      }
+      const savedFile = assetsFiles.find((file) => file.contentId === asset.contentId)?.blob
```
drillprop: I don't think, it's necessary. You used `find` on line 116, right?

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -12,26 +17,116 @@ import {
...
+          {
+            contentId: asset.contentId,
+            owner: asset.owner,
+            parentObject: {
+              type: asset.parentObject.type,
+              id: asset.parentObject.id,
+            },
+            type: asset.type,
+          },
+          randomStorageProviderUrl ?? '',
```
drillprop: not sure, but I think we shouldn't run `startFileUpload` if the `randomStorageProviderUrl` is `falsy`.

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -12,26 +17,116 @@ import {
...
+      {
+        contentId: asset.contentId,
+        owner: asset.owner,
+        parentObject: {
+          type: asset.parentObject.type,
+          id: asset.parentObject.id,
+        },
+        type: asset.type,
+      },
+      randomStorageProviderUrl ?? '',
```
drillprop: Same as above

src/hooks/useUploadsManager/types.ts
```diff
@@ -28,9 +28,20 @@ export type InputAssetUpload = Omit<AssetUpload, 'lastStatus' | 'size'>
...
 
+export type StartFileUploadOptions = {
+  isReUpload?: boolean
+  changeHost?: boolean
+}
+
 export type UploadManagerValue = {
-  uploadsState: AssetUploadWithProgress[]
-  startFileUpload: (file: File | Blob, asset: InputAssetUpload, storageMetadata: string) => void
+  uploadsState: AssetUploadWithProgress[][]
```
kdembler, line 42: Why is this a nested array?

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,188 @@
...
+    progress: uploadsProgress[asset.contentId] ?? 0,
+  }))
+
+  const channelDataObjects = [channel?.avatarPhotoDataObject, channel?.coverPhotoDataObject]
+  const videosDataObjects = videos?.flatMap((video) => [video.mediaDataObject, video.thumbnailPhotoDataObject]) || []
+  const allDataObjects = [...channelDataObjects, ...videosDataObjects]
+
+  // Enriching data with pending/accepted/rejected status
+  const uploadsStateWithLiaisonJudgement = uploadsStateWithProgress.map((asset) => {
+    const dataObject = allDataObjects.find((dataObject) => dataObject?.joystreamContentId === asset.contentId)
```
kdembler: Maybe we could add a warning if `dataObject` is not found? That would point to some local state corruption, might be useful for development

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,188 @@
...
+    displaySnackbar({
+      title: `(${lostConnectionAssets.length}) Asset${
+        lostConnectionAssets.length > 1 ? 's' : ''
+      } waiting to resume upload`,
+      description: 'Reconnect files to fix the issue',
+      actionText: 'See assets',
+      onActionClick: () => navigate(absoluteRoutes.studio.uploads()),
+      iconType: 'warning',
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kdembler: Why do we need that, what dependency do we want to ignore?

src/hooks/useUploadsManager/useUploadsManager.tsx
kdembler: I guess we would display it only if we're not loading data anymore

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,188 @@
...
+  }, [lostConnectionAssets.length])
+
+  // Enriching video type assets with video title
+  const uploadsStateWithVideoTitles = uploadsStateWithLiaisonJudgement.map((asset) => {
+    if (asset.type === 'video') {
+      const video = videos?.find((video) => video.mediaDataObject?.joystreamContentId === asset.contentId)
```
kdembler, line 99: Likewise, we could show a warning here

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,188 @@
...
+      const key = asset.parentObject.id
+      !acc[key] ? (acc[key] = [{ ...asset }]) : acc[key].push(asset)
+      return acc
+    }, {})
+  )
 
   const startFileUpload = useCallback(
-    async (file: File | Blob, asset: InputAssetUpload, storageMetadata: string) => {
+    async (
+      file: File | Blob | null,
```
kdembler, line 130: We need to do one of the following:
1) Handle case where both `file` and `fileInState` are null
2) Create a separate fn for resuming/retrying a download

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -12,26 +17,122 @@ import {
...
+          {
+            isReUpload: true,
+          }
+        )
+      }
+    },
+    [asset, randomStorageProviderUrl, startFileUpload]
+  )
+
+  const { getRootProps, getInputProps, open } = useDropzone({
```
kdembler: Let's rename `open` on destructure so it's more descriptive

src/hooks/useUploadsManager/types.ts
kdembler: Is this about grouping the assets together? So something like?
```
[
  [video-1-thumbnail, video-1-media],
  [video-2-thumbnail, video-2-media],
]
```

kdembler (commented):
Just the code for now, I'll test it out later because the query node is down

src/hooks/useUploadsManager/types.ts
mikkio-j: > Is this about grouping the assets together? So something like?
> 
> ```
> [
>   [video-1-thumbnail, video-1-media],
>   [video-2-thumbnail, video-2-media],
> ]
> ```

Yes, that's exactly it

src/hooks/useUploadsManager/useUploadsManager.tsx
mikkio-j: `displaySnackbar` is causing updates. Wasn't sure how should I handle that

kdembler: Almost always when we disable `react-hooks/exhaustive-deps` something will break. Not including `displaySnackbar` in the dependency array will cause the hook to always use the stale reference to the function. `displaySnackbar` in turn uses `...snackbars` so using old `displaySnackbar` reference may cause some of the snackbars to disappear.

Disabling this rule is almost always not the right solution. In this case, we should just make the `displaySnackbar` reference stable

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,201 @@
...
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [lostConnectionAssets.length, navigate])
+
+  // Enriching video type assets with video title
+  const uploadsStateWithVideoTitles = uploadsStateWithLiaisonJudgement.map((asset) => {
+    if (asset.type === 'video') {
+      const video = videos?.find((video) => video.mediaDataObject?.joystreamContentId === asset.contentId)
+      if (!video && !videosLoading) {
+        console.warn(`Video not found. ContentId: ${asset.contentId}`)
```
kdembler: When doing upload I get a lot (like 30 or something) warning like this in the console. Could you look into that?

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -47,19 +148,43 @@ const AssetLine: React.FC<AssetLineProps> = ({ isLast = false, asset }) => {
...
+        showDialog={showDialog}
+        variant="warning"
+        onSecondaryButtonClick={() => {
+          setShowDialog(false)
+          if (asset.parentObject.type === 'video') {
+            navigate(absoluteRoutes.studio.editVideo())
+          }
+          if (asset.parentObject.type === 'channel') {
+            navigate(absoluteRoutes.studio.editChannel())
+          }
```
kdembler, line 164: I think we got a misunderstanding on what's supposed to happen here. You redirect the user to "new video" screen when they click `Upload new`. But what should actually happen is that we create an update transaction with the new file and try to submit it, right from this screen and then start the upload. That's also what's in the diagram linked in the issue. However, maybe let's try to elevate that to separate task so we don't drag this PR for too long

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,62 +1,212 @@
...
+    if (asset?.type === 'video') {
+      const video = videos?.find((video) => video.mediaDataObject?.joystreamContentId === asset.contentId)
+      const title = video?.title ?? null
+      return { ...asset, title }
+    }
+    return asset
+  })
+
+  // Check if liaison data and video title is available
+  uploadsStateWithVideoTitles.map((asset) => {
```
kdembler, line 107: In general, in cases like this where you don't care about return value but just want to iterate over the values, `.forEach` may be a better choice. It's a bit more performant, although negligibly, but mostly just for readability improvement

No need to change here

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -11,28 +16,122 @@ import {
...
 
-  const renderStatusIndicator = (asset: UploadData) => {
+  const renderStatusMessage = (asset: AssetUploadWithProgress) => {
+    if (asset.lastStatus === 'reconnecting') {
+      return 'Trying to reconnect...'
+    }
+    if (asset.lastStatus === 'reconnectionError') {
+      return (
+        <Button size="small" variant="secondary" icon={<SvgGlyphUpload />} onClick={handleChangeHost}>
+          Change host (0.2 JOY)
```
kdembler:
```suggestion
          Change host
```

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -47,19 +146,43 @@ const AssetLine: React.FC<AssetLineProps> = ({ isLast = false, asset }) => {
...
-        </FileInfoType>
-        <Text variant="body2">{dimension}</Text>
-        <Text>{size}</Text>
-      </FileInfoContainer>
-      <StatusMessage variant="subtitle2">{fileStatusMessage}</StatusMessage>
-    </FileLineContainer>
+    <>
+      <MessageDialog
+        title="Different file was selected!"
+        description="We detected that you selected a different file than the one you uploaded previously. Select the same file to continue the upload or make a new transaction (0.2JOY) and upload a new version of file."
```
kdembler: I'd like the copy here to be `We detected that you selected a different file than the one you uploaded previously. Select the same file to continue the upload or edit "your channel"/"the video" to use the new file.`

We should show the correct variant depending on whether it's a video or a channel

src/views/studio/MyUploadsView/AssetLine/AssetLine.tsx
```diff
@@ -47,19 +146,43 @@ const AssetLine: React.FC<AssetLineProps> = ({ isLast = false, asset }) => {
...
+          if (asset.parentObject.type === 'channel') {
+            navigate(absoluteRoutes.studio.editChannel())
+          }
+        }}
+        primaryButtonText="Reselect file"
+        onPrimaryButtonClick={() => {
+          setShowDialog(false)
+          openFileSelect()
+        }}
+        secondaryButtonText="Upload new (0.2 JOY)"
```
kdembler: `Edit channel` / `Edit video` same as above

kdembler (approved):

Issue #467, opened 2021-04-19
Pimp up the transaction dialog


kdembler:
Transaction dialog should get a very slight refresh - we should do a better job to represent current step and progress. Designs currently include grayish step indicator at the top of the dialog. The designs are also supposed to get some illustration for each step. Once that's ready, we should update our component

PR #531, opened 2021-04-30
Update transaction dialog

drillprop:
Fix: #467
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=4754%3A770

### DIFF
--- a/src/components/Dialogs/TransactionDialog/TransactionDialog.style.ts
+++ b/src/components/Dialogs/TransactionDialog/TransactionDialog.style.ts
@@ -1,16 +1,45 @@
 import styled from '@emotion/styled'
-import { sizes, media } from '@/shared/theme'
+import { sizes, media, colors, transitions } from '@/shared/theme'
 import { ReactComponent as TransactionIllustration } from '@/assets/transaction-illustration.svg'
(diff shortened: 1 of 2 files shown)


src/components/Dialogs/TransactionDialog/TransactionDialog.style.ts
```diff
@@ -1,16 +1,45 @@
...
+
+  display: grid;
+  grid-template-columns: repeat(auto-fit, minmax(10px, 1fr));
+  grid-gap: ${sizes(1)};
+  width: 100%;
+  height: ${sizes(2)};
+`
+
+export const Step = styled.div<StepProps>`
+  background-color: ${({ isActive }) => (isActive ? colors.gray[400] : colors.gray[600])};
```
kdembler, line 23: I think the design specifies that not only the active step should be highlighted but also the previous ones, take a look: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6545%3A157596

kdembler (approved):

PR #544, opened 2021-05-04
Allow deleting videos

### LINKED ISSUE #529: Allow deleting videos
Both from the video sheet and my videos preview. Should show a confirmation dialog, as in the design

drillprop:
Fix #529


src/joystream-lib/api.ts
```diff
@@ -424,6 +424,19 @@ export class JoystreamJs {
     }
   }
 
+  private async _deleteVideo(videoId: VideoId, memberId: MemberId, cb?: ExtrinsicStatusCallbackFn) {
```
kdembler: I think in this case we can just put this in the public function body, since there are no separate variants (e.g. create/update)

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -205,3 +235,73 @@ const getPublicness = (currentTabName: typeof TABS[number]) => {
       return undefined
   }
 }
+
+export const useDeleteVideo = (memberId: string | null) => {
+  const { joystream } = useJoystream()
+  const [videoIdToDelete, setVideoIdToDelete] = useState<undefined | string>()
```
mikkio-j: I think it would be better to go for `null` here instead, so we don't need to pass those `undefined` below. For videoId we could go with `videoId ?? null`

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -205,3 +235,73 @@ const getPublicness = (currentTabName: typeof TABS[number]) => {
       return undefined
   }
 }
+
+export const useDeleteVideo = (memberId: string | null) => {
```
mikkio-j: not exactly sure if we should keep it here. Maybe we should keep the consistency and move it to hooks folder

src/hooks/useDeleteVideo.tsx
```diff
@@ -0,0 +1,77 @@
...
+    setDeleteTransactionStatus(null)
+  }
+
+  return {
+    handleConfirmDeleteVideo,
+    handleDeleteTransactionClose,
+    handleCancel,
+    handleDeleteVideoClick,
+    videoIdToDelete,
+    deleteTransactionStatus,
```
kdembler: Let's not use `handle*` names for functions returned from hooks. Let's reserve that name for functions that are supposed to handle some kinds of events. Those should be named like actions - `confirmDeleteVideo`, `cancelDeleteVideo`, etc.

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -259,6 +259,13 @@ export const EditVideoSheet: React.FC = () => {
     setTransactionStatus(null)
   }
 
+  const handleDeleteVideo = async (videoId: string) => {
+    const videoTabIdx = videoTabs.findIndex((vt) => vt.id === videoId)
+    await refetchVideo()
```
kdembler: Do we need that?

src/hooks/useDeleteVideo.tsx
```diff
@@ -0,0 +1,77 @@
+import { useQueryNodeStateSubscription } from '@/api/hooks'
+import { ExtensionSignCancelledError, ExtrinsicStatus } from '@/joystream-lib'
+import { useState, useEffect } from 'react'
+import { useSnackbar, useJoystream } from '@/hooks'
+
+export const useDeleteVideo = (memberId: string | null) => {
+  const { joystream } = useJoystream()
+  const [videoIdToDelete, setVideoIdToDelete] = useState<string | null>(null)
```
kdembler: I guess we don't need that?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -130,11 +139,30 @@ export const MyVideosView = () => {
...
+        primaryButtonText="Delete video"
+        secondaryButtonText="Cancel"
+      />
+      <TransactionDialog
+        status={deleteTransactionStatus}
+        successTitle={'Video successfully deleted!'}
+        successDescription={'Your video was deleted from the blockchain.'}
+        onClose={() => {
+          handleDeleteTransactionClose()
+          refetch({})
```
kdembler: Let's leave it out here

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -414,13 +423,41 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
+                }}
+              >
                 Delete video
               </DeleteVideoButton>
             </DeleteVideoContainer>
           )}
         </InputsContainer>
       </FormWrapper>
+      <MessageDialog
+        title="Do you want to remove this video from your videos?"
```
kdembler:
```suggestion
        title="Delete this video?"
```

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -130,11 +139,30 @@ export const MyVideosView = () => {
...
               }}
               onEditVideoClick={() => handleVideoClick(video.id)}
-              onDeleteVideoClick={() => {
-                // TODO: handle delete
-              }}
+              onDeleteVideoClick={() => handleDeleteVideoClick(video.id)}
             />
           ))}
+      <MessageDialog
+        title="Do you want to remove this video from your videos?"
```
eldiegod:
```suggestion
        title="Do you want to remove this video from your channel?"
```
I think makes more sense

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -414,13 +423,41 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
+        exitButton={false}
+        description="Video will be removed permanently and all its data will be lost. Joystream studio do not keep any of your data after you remove your video."
```
kdembler:
```suggestion
        description="You will not be able to undo this. Deletion requires a blockchain transaction to complete. Currently there is no way to remove uploaded video assets."
```

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -414,13 +423,41 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
+        onPrimaryButtonClick={handleConfirmDeleteVideo}
+        error
+        variant="warning"
+        primaryButtonText="Delete video"
+        secondaryButtonText="Cancel"
+      />
+      <TransactionDialog
+        status={deleteTransactionStatus}
+        successTitle={'Video successfully deleted!'}
+        successDescription={'Your video was deleted from the blockchain.'}
```
kdembler:
```suggestion
        successDescription="Your video was marked as deleted and it will no longer show up on Joystream."
```

FYI, the reason for this copy change is that once something is saved on the blockchain, it cannot really be deleted ever, so I guess it's good to not give false understanding to the user

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -130,11 +139,30 @@ export const MyVideosView = () => {
                 handleVideoClick(video.id, { minimized: true })
               }}
               onEditVideoClick={() => handleVideoClick(video.id)}
-              onDeleteVideoClick={() => {
-                // TODO: handle delete
-              }}
+              onDeleteVideoClick={() => handleDeleteVideoClick(video.id)}
             />
           ))}
+      <MessageDialog
```
kdembler, line 149: Same copy changes here

eldiegod (commented):
Functionality-wise working good 👍

src/views/studio/MyVideosView/MyVideosView.tsx
drillprop: will apply Klaudiusz suggestion here https://github.com/Joystream/atlas/pull/544#discussion_r626435019

kdembler (approved):

Issue #470, opened 2021-04-19
Clean up notifications


kdembler:
The design includes displaying success/failure snackbars/dialogs after specific flows succeed/fail. We should add those kind of announcements to the app. General idea is this:

- if something major happened that may change the users flow, display the dialog with message, for example:
    - new channel created and after accepting the user will be redirected
    - new video published
    - error in transaction happened
- if something more minor happened, we should use a snackbar:
    - channel updates saved
    - asset successfully uploaded

We should first create a list of all kinds of interactions like this with proposed notification and then implement that

mikkio-j:
Here's the list of notifications yet to be implemented:

###### Snackbars:

- Draft/video deleted
- Assets uploaded
- Assets uploading
- Draft saved
- Draft opened in tab
- internet connection restored

###### Dialogs:

- remove draft and possibly remove video (if not covered in #529)

kdembler:
- What about channel/video edit?
- I don't think we should have a snackbar on draft saved because that will be happening continuously
- Probably also asset upload error, unless that's part of uploads error handling
- What about "Internet connection lost"? Does that exist already?

mikkio-j:
> * What about channel/video edit?
> * I don't think we should have a snackbar on draft saved because that will be happening continuously
> * Probably also asset upload error, unless that's part of uploads error handling
> * What about "Internet connection lost"? Does that exist already?

1. Ah yes, that I've already implemented and forgot to put it here
2. Yeah, that would be an issue. Understood.
3. That's the part of asset upload error
4. Does not exist, forgot to put it here

PR #546, opened 2021-05-04
Notifications

mikkio-j:
Issue: #470

### DIFF
--- a/src/hooks/useConnectionStatus.tsx
+++ b/src/hooks/useConnectionStatus.tsx
@@ -47,7 +47,7 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
   }, [displaySnackbar])
 
   useEffect(() => {
-    // ping google every three seconds to check if user is connected to internet
+    // ping google every five seconds to check if user is connected to internet
     const interval = setInterval(() => {
       checkConnection()
     }, 5000)
--- a/src/hooks/useConnectionStatus.tsx
+++ b/src/hooks/useConnectionStatus.tsx
@@ -1,4 +1,5 @@
 import React, { useContext, useEffect, useState } from 'react'
+import { useSnackbar } from '@/hooks/useSnackbar'
 
 export type ConnectionStatus = 'connected' | 'disconnected' | 'connecting'
 
@@ -19,6 +20,8 @@ const withTimeout = async <T,>(promise: Promise<T>, timeout: number) => {
 export const ConnectionStatusProvider: React.FC = ({ children }) => {
   const [nodeConnectionStatus, setNodeConnection] = useState<ConnectionStatus>('connecting')
   const [isUserConnectedToInternet, setIsUserConnectedToInternet] = useState(true)
+  const [showSnackbar, setShowSnackbar] = useState(false)
+  const { displaySnackbar } = useSnackbar()
 
   useEffect(() => {
     // ping google every three seconds to check if user is connected to internet
@@ -30,6 +33,13 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
     }
   }, [])
 
+  useEffect(() => {
+    if (showSnackbar) {
+      displaySnackbar({ title: 'Network connection restored', iconType: 'success' })
+      setShowSnackbar(false)
+    }
+  }, [displaySnackbar, showSnackbar])
+
   const checkConnection = async () => {
     try {
       const res = await withTimeout(
@@ -40,10 +50,16 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
         4000
       )
       if (res) {
-        setIsUserConnectedToInternet(true)
+        setIsUserConnectedToInternet((previousState) => {
+          if (previousState === false) {
+            setShowSnackbar(true)
+          }
+          return true
+        })
       }
     } catch (error) {
       setIsUserConnectedToInternet(false)
+      displaySnackbar({ title: 'No network connection', iconType: 'error' })
     }
   }
 
--- a/src/hooks/useConnectionStatus.tsx
+++ b/src/hooks/useConnectionStatus.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState, useCallback } from 'react'
+import React, { useContext, useEffect, useState, useCallback, useRef } from 'react'
 import { useSnackbar } from '@/hooks/useSnackbar'
 
 export type ConnectionStatus = 'connected' | 'disconnected' | 'connecting'
@@ -20,7 +20,6 @@ const withTimeout = async <T,>(promise: Promise<T>, timeout: number) => {
 export const ConnectionStatusProvider: React.FC = ({ children }) => {
   const [nodeConnectionStatus, setNodeConnection] = useState<ConnectionStatus>('connecting')
   const [isUserConnectedToInternet, setIsUserConnectedToInternet] = useState(true)
-  const [showSnackbar, setShowSnackbar] = useState(false)
   const { displaySnackbar } = useSnackbar()
 
   const checkConnection = useCallback(async () => {
@@ -33,18 +32,12 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
         4000
       )
       if (res) {
-        setIsUserConnectedToInternet((previousState) => {
-          if (previousState === false) {
-            setShowSnackbar(true)
-          }
-          return true
-        })
+        setIsUserConnectedToInternet(true)
       }
     } catch (error) {
       setIsUserConnectedToInternet(false)
-      displaySnackbar({ title: 'No network connection', iconType: 'error' })
     }
-  }, [displaySnackbar])
+  }, [])
 
   useEffect(() => {
     // ping google every five seconds to check if user is connected to internet
@@ -56,12 +49,19 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
     }
   }, [checkConnection])
 
+  const isInitialMount = useRef(true)
+
   useEffect(() => {
-    if (showSnackbar) {
-      displaySnackbar({ title: 'Network connection restored', iconType: 'success' })
-      setShowSnackbar(false)
+    if (isInitialMount.current) {
+      isInitialMount.current = false
+    } else {
+      if (isUserConnectedToInternet) {
+        displaySnackbar({ title: 'Network connection restored', iconType: 'success' })
+      } else {
+        displaySnackbar({ title: 'No network connection', iconType: 'error' })
+      }
     }
-  }, [displaySnackbar, showSnackbar])
+  }, [displaySnackbar, isUserConnectedToInternet])
 
   return (
     <ConnectionStatusContext.Provider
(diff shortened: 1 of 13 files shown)


src/hooks/useConnectionStatus.tsx
```diff
@@ -30,6 +33,13 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
     }
   }, [])
```
drillprop: I think, we can get rid of this warning easily by adding `checkConnection` to the dependency array and wrapping `checkConnection` in useCallback

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -36,15 +38,15 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
         // TODO: remove assets from the same parent if all finished
         updateAsset(asset.contentId, 'completed')
         setAssetUploadProgress(100)
-
+        displaySnackbar({ title: 'Asset uploaded', iconType: 'success' })
         // TODO: add snackbar?
```
drillprop:
```suggestion
```

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -332,6 +332,22 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
...
+      if (
+        data.cover.blob &&
+        coverContentId &&
+        storageProviderUrl &&
+        data.avatar.blob &&
+        avatarContentId &&
+        storageProviderUrl
+      ) {
+        displaySnackbar({ title: '(2) Assets being uploaded', iconType: 'info' })
+      }
```
drillprop, line 344: I think you could refactor this. For example, you could just return if there is no `storageProviderUrl`. You also used `storageProviderUrl` twice 😅  above

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -242,6 +242,22 @@ export const EditVideoSheet: React.FC = () => {
...
+        displaySnackbar({ title: 'Asset being uploaded', iconType: 'info' })
+      }
+      if (
+        videoInputFile?.blob &&
+        videoContentId &&
+        randomStorageProviderUrl &&
+        thumbnailInputFile?.blob &&
+        thumbnailContentId &&
+        randomStorageProviderUrl
+      ) {
```
drillprop: same as above

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -290,7 +306,17 @@ export const EditVideoSheet: React.FC = () => {
...
+          handleTransactionClose()
+          if (isEdit && transactionStatus === ExtrinsicStatus.Completed) {
+            closeSheet()
+            displaySnackbar({ title: 'Video successfully updated', iconType: 'success' })
+          }
+          if (!isEdit && transactionStatus === ExtrinsicStatus.Completed) {
+            closeSheet()
+            displaySnackbar({ title: 'Video successfully created', iconType: 'success' })
+          }
+        }}
```
drillprop:
```suggestion
        onClose={() => {
          handleTransactionClose()
          if (transactionStatus === ExtrinsicStatus.Completed) {
            closeSheet()
            displaySnackbar({ title: `Video successfully ${isEdit ? 'updated' : 'created'}`, iconType: 'success' })
          }
        }}
```

src/hooks/useConnectionStatus.tsx
```diff
@@ -40,12 +33,35 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
...
       }
     } catch (error) {
       setIsUserConnectedToInternet(false)
+      displaySnackbar({ title: 'No network connection', iconType: 'error' })
     }
-  }
+  }, [displaySnackbar])
+
+  useEffect(() => {
+    // ping google every three seconds to check if user is connected to internet
```
drillprop:
```suggestion
    // ping google every five seconds to check if user is connected to internet
```
:P but I guess we could just remove this comment

drillprop (commented):
Found visual bug
image
Also, it's worth checking how the snack bars looks with the loading screen(skeleton) on `/uploads`

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -290,7 +295,13 @@ export const EditVideoSheet: React.FC = () => {
             ? 'Your video was created and saved on the blockchain. Upload of video assets may still be in progress.'
             : 'Changes to your video were saved on the blockchain.'
         }
-        onClose={handleTransactionClose}
+        onClose={() => {
+          handleTransactionClose()
+          if (transactionStatus === ExtrinsicStatus.Completed) {
+            closeSheet()
+            displaySnackbar({ title: `Video successfully ${isEdit ? 'updated' : 'created'}`, iconType: 'success' })
+          }
```
drillprop: Last thing :P - I think, you could just move this to `handleTransactionClose`

kdembler: Definitely!

src/hooks/useConnectionStatus.tsx
```diff
@@ -40,12 +33,35 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
...
+          }
+          return true
+        })
       }
     } catch (error) {
       setIsUserConnectedToInternet(false)
+      displaySnackbar({ title: 'No network connection', iconType: 'error' })
     }
-  }
+  }, [displaySnackbar])
```
kdembler: Won't this cause problems? I think our current `displaySnackbar` reference isn't stable? We probably should just slap `useCallback` on top of it so we're safe wherever we use it

src/hooks/useConnectionStatus.tsx
```diff
@@ -40,12 +33,35 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
...
+      clearInterval(interval)
+    }
+  }, [checkConnection])
+
+  useEffect(() => {
+    if (showSnackbar) {
+      displaySnackbar({ title: 'Network connection restored', iconType: 'success' })
+      setShowSnackbar(false)
+    }
+  }, [displaySnackbar, showSnackbar])
```
kdembler: I'm a bit lost here, why do we use this instead of just calling `displaySnackbar` where needed?

I think we could simplify all this code and just add `useEffect` that fires snackbar when `isUserConnectedToInternet` changes

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -135,9 +147,16 @@ export const MyVideosView = () => {
                 onPullupClick={(e) => {
                   e.stopPropagation()
                   handleVideoClick(draft.id, { draft: true, minimized: true })
+                  const index = videoTabs.findIndex((item) => item.id === draft.id)
+                  displaySnackbar({
+                    title: 'Video opened in new tab',
+                    iconType: 'success',
+                    actionText: 'Remove',
+                    onActionClick: () => removeVideoTab(index),
+                  })
```
kdembler: Index here will always be `-1` and this will not work. Try to figure out why, let me know if you struggle

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -176,6 +195,17 @@ export const MyVideosView = () => {
         successDescription="Your video was marked as deleted and it will no longer show up on Joystream."
         onClose={handleVideoDeleted}
       />
+      <MessageDialog
+        title="Do you want to remove this draft from your drafts?"
+        description="Draft will be removed permanently and all its data will be lost. Drafts are stored locally - that’s why you don’t need to make a transaction to confirm it."
```
```suggestion
        title="Delete this draft?"
        description="You will not be able to undo this."
```

kdembler (changes requested):
Some additional comments:

- This doesn't fully follow the requirements set in the issue. We have double notifications for some places and it could be improved. Consider channel update flow. You go with the transaction and at the end of it you get a success dialog that says "Channel successfully updated". You can look at it for some time and then you hit close. Once you click close, the snackbar will show up that says "Channel successfully updated". I don't think it makes much sense to show both of those in the same flow. It probably should be one or the other, as described in the issue. Lmk if we need to clear that up more

- The button will need some padding, doesn't look great atm:
Screenshot 2021-05-10 at 11 07 57

-

src/views/studio/MyVideosView/MyVideosView.tsx
mikkio-j: this was resolved by returning index from added tab, but it still has to use `removeVideoTab`, which did not operate on current state. I believe it's because of the how current snackbar works. Did a workaround with adding useEffect. Please take a look if that solution is satisfying

src/hooks/useConnectionStatus.tsx
```diff
@@ -45,7 +37,31 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
...
+    // ping google every five seconds to check if user is connected to internet
+    const interval = setInterval(() => {
+      checkConnection()
+    }, 5000)
+    return () => {
+      clearInterval(interval)
+    }
+  }, [checkConnection])
+
+  const isInitialMount = useRef(true)
```
kdembler, line 53: Why do we need that? Usually when we have to add some non-obvious workaround, it's a good idea to leave a comment right away

src/hooks/useEditVideoSheet.tsx
```diff
@@ -54,13 +54,15 @@ export const EditVideoSheetProvider: React.FC = ({ children }) => {
...
-      setVideoTabs([...videoTabs, tabToAdd])
+      const newVideoTabs = [...videoTabs, tabToAdd]
+      setVideoTabs(newVideoTabs)
 
       if (shouldSelect) {
         const newTabIdx = videoTabs.length
         setSelectedVideoTabIdx(newTabIdx)
       }
+      // return index of added tab
+      return newVideoTabs.length - 1
```
kdembler:
```suggestion

      setVideoTabs([...videoTabs, tabToAdd])
      const newTabIdx = videoTabs.length

      if (shouldSelect) {
        setSelectedVideoTabIdx(newTabIdx)
      }
      return newTabIdx
```

kdembler: Probably will be different after we return a function to remove the tab

src/hooks/useConnectionStatus.tsx
mikkio-j: Comment added

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -98,6 +102,14 @@ export const MyVideosView = () => {
       return
     }
     addVideoTab({ id, isDraft: opts.draft })
+
+    displaySnackbar({
+      title: 'Video opened in new tab',
+      iconType: 'success',
+      actionText: 'Remove',
```
```suggestion
      title: 'Video opened in a new tab',
      iconType: 'success',
      actionText: 'Undo',
```

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -116,6 +128,28 @@ export const MyVideosView = () => {
     closeDeleteTransactionDialog()
   }
 
+  const confirmRemoveDraft = (id: string) => {
+    removeDraft(id)
+    displaySnackbar({
+      title: 'Draft successfully deleted',
```
```suggestion
      title: 'Draft deleted',
```

kdembler (approved):

PR #551, opened 2021-05-04
update to the new testnet

kdembler:


drillprop (approved):
Did initial testing and I cannot upload any assets right now, but I don't think it's on our side(getting CORS errors). 

Besides that, looks good

mikkio-j:
> Did initial testing and I cannot upload any assets right now, but I don't think it's on our side(getting CORS errors).
> 
> Besides that, looks good

I do not get these errors. Interesting

update: Yup, I'm getting CORS on localhost

update 2: Works now

kdembler:
Yeah, probably should have mentioned here, the preview will be still running the old testnet, I will update the env vars before merge, for now please test on localhost

drillprop (approved):

Issue #528, opened 2021-04-30
Preserve form state on video edit


kdembler:
Currently (#526) the form state will reset for edited videos when switching to a different video tab. That's because we don't save updates as drafts. We should cache that state somewhere and restore it when user comes to the tab again

PR #554, opened 2021-05-05
Preserve form state on video edit

eldiegod:
#528


src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -171,11 +173,12 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
 
   // with react-hook-form v7 it's possible to call watch((data) => update()), we should use that instead when we upgrade
   const handleFormChange = () => {
+    const data = getValues()
     if (!selectedVideoTab?.isDraft) {
-      return
+      updateTabData(data)
+    } else {
+      debouncedDraftSave.current(selectedVideoTab, data, addDraft, updateDraft, updateSelectedVideoTab)
```
kdembler, line 248: We probably should update the debounced fn name, because now it doesn't save drafts only

src/hooks/useEditVideoSheet.tsx
```diff
@@ -180,19 +180,30 @@ export const useEditVideoSheetTabData = (tab?: EditVideoSheetTab) => {
     activeUser: { channelId },
   } = useActiveUser()
   const { drafts } = useDrafts('video', channelId ?? '')
-
   const { selectedVideoTabCachedAssets } = useEditVideoSheet()
 
   const { video, loading, error } = useVideo(tab?.id ?? '', { skip: tab?.isDraft })
+  const [cachedDirtyVideoTabsData, setcachedDirtyVideoTabs] = useState<Array<EditVideoFormFields & EditVideoSheetTab>>(
```
kdembler: I'm a bit worried about this getting out of sync since each hook invocation will have its own state (we don't back this by any context, so will be a separate state for each component that uses this hook). Maybe the data should be persisted by the `EditVideoSheetProvider`? What do you think?

kdembler: Also I'm wondering whether this data should come from this hook versus the general `useEditVideoSheet`, we already return stuff like `setSelectedVideoTabCachedAssets` from it so maybe form state should also be modified from it?

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
eldiegod: what would you suggest here?

src/hooks/useEditVideoSheet.tsx
eldiegod: Yeah I was a bit confused about where to put it at first; so added it to the provider and moved to `useEditVideoSheet`

src/hooks/useEditVideoSheet.tsx
```diff
@@ -180,19 +203,27 @@ export const useEditVideoSheetTabData = (tab?: EditVideoSheetTab) => {
...
-
-  const { selectedVideoTabCachedAssets } = useEditVideoSheet()
+  const { selectedVideoTabCachedAssets, selectedVideoTabCachedDirtyFormData } = useEditVideoSheet()
 
   const { video, loading, error } = useVideo(tab?.id ?? '', { skip: tab?.isDraft })
 
   if (!tab) {
     return {
-      data: null,
+      updateTabData: () => ({}),
```
drillprop: I'm a little confused here, why we need this? Could you explain?

drillprop (commented):
I would need some clarification on one thing, but it works really well. Solid work 👍🏻

src/hooks/useEditVideoSheet.tsx
eldiegod: oops, that was not meant to stay there, good catch 👌

src/hooks/useEditVideoSheet.tsx
```diff
@@ -42,7 +44,9 @@ export const EditVideoSheetProvider: React.FC = ({ children }) => {
   const [sheetState, setSheetState] = useState<EditVideoSheetState>('closed')
   const [cachedSheetState, setCachedSheetState] = useState<EditVideoSheetState>('closed')
   const [assetsCache, setAssetsCache] = useState<EditVideoAssetsCache>({})
-
+  const [selectedVideoTabCachedDirtyFormData, _setSelectedVideoTabCachedDirtyFormData] = useState<
+    Array<EditVideoFormFields & EditVideoSheetTab>
+  >([])
```
kdembler: Maybe we should make it an object instead, like we do for the assets? Then later we won't need the `findIndex`, we will be able to just do `selectedVideoTabCachedDirtyFormData[selectedVideoTab.id]`

kdembler: The updates would be easier as well

src/hooks/useEditVideoSheet.tsx
```diff
@@ -42,7 +44,9 @@ export const EditVideoSheetProvider: React.FC = ({ children }) => {
   const [sheetState, setSheetState] = useState<EditVideoSheetState>('closed')
   const [cachedSheetState, setCachedSheetState] = useState<EditVideoSheetState>('closed')
   const [assetsCache, setAssetsCache] = useState<EditVideoAssetsCache>({})
-
+  const [selectedVideoTabCachedDirtyFormData, _setSelectedVideoTabCachedDirtyFormData] = useState<
```
kdembler: We call this `selectedVideoTabCachedDirtyFormData` so it should be the data for the currently selected tab, but this keeps track of _all_ the tabs, there's some confusion here. The name of the state in this function needs to be different, we can keep the `selectedVideoTabCachedDirtyFormData` as the context value but it should return data only for the selected tab

kdembler (commented):
Looks good but this doesn't work fully though, we need to also manage the form dirty state somehow so the form knows whether there actually were any updates. For example, if I modify video title, the transaction bar will show up with options to either reset to original state or publish changes. Now, if I switch tab and go back to my edit, the edited title will be there but the form will keep thinking that this is original data and will not show the transaction bar. This may also break some of our update logic which depends on information about which fields are dirty. Right now once you go back to edit, none of the fields will be dirty. We need to solve that somehow. Let me know if something isn't clear

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
kdembler: I'm actually not sure what I meant here, the fn still only saves drafts.. Let's ignore this comment :P

eldiegod:
Came out with a solution to get around the form isDirty state when editing a video and switching tabs.

src/hooks/useEditVideoSheet.tsx
```diff
@@ -42,7 +45,9 @@ export const EditVideoSheetProvider: React.FC = ({ children }) => {
   const [sheetState, setSheetState] = useState<EditVideoSheetState>('closed')
   const [cachedSheetState, setCachedSheetState] = useState<EditVideoSheetState>('closed')
   const [assetsCache, setAssetsCache] = useState<EditVideoAssetsCache>({})
-
+  const [selectedVideoTabsCachedDirtyFormData, _setSelectedVideoTabsCachedDirtyFormData] = useState<
```
kdembler: This is `videoTabsCached...`, not `selectedVideoTabs` right? It's a record of all of them

src/hooks/useEditVideoSheet.tsx
```diff
@@ -20,6 +20,7 @@ export type EditVideoSheetTab = {
 }
 
 type EditVideoAssetsCache = Record<string, InputFilesState>
+type EditVideoTabCachedDirtyFormData = Record<string, Partial<EditVideoFormFields> & EditVideoSheetTab>
```
kdembler: Why do we need `EditVideoSheetTab` fields here? Aren't form fields enough?

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -149,7 +154,28 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
     setFileSelectError(null)
     reset(tabData)
-  }, [selectedVideoTab, cachedSelectedVideoTabId, forceReset, reset, tabDataLoading, tabData, updateSelectedVideoTab])
+
+    if (selectedVideoTabCachedDirtyFormData) {
+      // allow a render for the form to reset first and then set fields dirty
+      setTimeout(() => {
+        const keys = Object.keys(selectedVideoTabCachedDirtyFormData) as Array<keyof EditVideoFormFields>
+        keys.forEach((key) => {
+          console.log({ key, value: selectedVideoTabCachedDirtyFormData[key] })
```
kdembler:
```suggestion
```

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -149,7 +154,28 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
 
     setFileSelectError(null)
     reset(tabData)
-  }, [selectedVideoTab, cachedSelectedVideoTabId, forceReset, reset, tabDataLoading, tabData, updateSelectedVideoTab])
+
+    if (selectedVideoTabCachedDirtyFormData) {
+      // allow a render for the form to reset first and then set fields dirty
+      setTimeout(() => {
```
kdembler, line 200: It'd be best if we can avoid the timeout to not have the action bar sliding up and down for one render. Let's continue discussion on rocket chat

kdembler (commented):
Seems to work well! One thing we probably will need to also coordinate is to when to clean this cached state. In @drillprop PR, we clear the cache on sheet close if I'm not mistaken. We will likely need some similar behaviour here. Please coordinate on this @eldiegod @drillprop, I'm fine with this being done in either of your PRs

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
kdembler: Let's leave it for now, but let's create an issue to capture this behaviour

src/hooks/useEditVideoSheet.tsx
eldiegod: yup

src/hooks/useEditVideoSheet.tsx
eldiegod: Yes I think after changing the `videoTabsCachedDirtyFormData` from an array to an object map EditVideoSheetTab should be enough

eldiegod:
> Seems to work well! One thing we probably will need to also coordinate is to when to clean this cached state. In @drillprop PR, we clear the cache on sheet close if I'm not mistaken. We will likely need some similar behaviour here. Please coordinate on this @eldiegod @drillprop, I'm fine with this being done in either of your PRs

@drillprop let me know where/what PR you think would make more sense to add the cache reset

drillprop:
@eldiegod  in my PR I did clearing the asset cache when we close the sheet https://github.com/Joystream/atlas/pull/577/files#diff-01e1c39885af3a05a0018a8d94117d21e2f5e56e54daf06ec3719de02f694a70R135
I think it makes sense to do it that way.

eldiegod:
>  I think it makes sense to do it that way.

Yup that was my thought, did the same here 👍

kdembler:
I think it's ready to go but needs a rebase. Also @drillprop I'd appreciate if you approved here concerning the cache reset

drillprop (approved):
Looks good to me! 👍🏻

PR #564, opened 2021-05-05
query node workers update

kdembler:


src/api/hooks/storageProviders.ts
```diff
@@ -15,7 +15,7 @@ export const useStorageProvider = (id: string, opts?: StorageProviderOpts) => {
     variables: { where: { id } },
   })
   return {
-    storageProvider: data?.storageProviderByUniqueInput,
+    storageProvider: data?.workerByUniqueInput,
```
drillprop: I'm not entirely sure about this. It can be confusing in the future. Maybe, we should also change the name of`storageProviders`(and corresponding hooks) to just `workers`?

kdembler: It makes sense this way, because storage providers are only one type of worker. And here we are interested only in storage providers. Will add additional query filter by type

src/api/hooks/workers.ts
```diff
@@ -0,0 +1,51 @@
...
+    variables: { where: { id } },
+  })
+  return {
+    storageProvider: data?.workerByUniqueInput,
+    ...queryRest,
+  }
+}
+
+type WorkersOpts = QueryHookOptions<GetWorkersQuery>
+export const useStorageProviders = (variables: GetWorkersQueryVariables, opts?: WorkersOpts) => {
```
drillprop, line 25:
```suggestion
export const useWorkers = (variables: GetWorkersQueryVariables, opts?: WorkersOpts) => {
```

drillprop (approved):

Issue #553, opened 2021-05-05
Don't filter channels without assets 


kdembler:
Currently we use the following query params for displaying channels:
```
where: {
  isPublic_eq: true,
  isCensored_eq: false,
  avatarPhotoAvailability_eq: AssetAvailability.Accepted,
  coverPhotoAvailability_eq: AssetAvailability.Accepted,
},
```
Channel cover/avatar aren't required though so we should also display channels that don't have them. We could also allow `invalid` state or maybe we should just drop those filters for channels

PR #565, opened 2021-05-05
Removed PhotoAvailability filters

eldiegod:
Fix: #553

Removed only from the query variables as the filters themselves might come handy later


mikkio-j (approved):
That should be good. We could also clear that enum out from import since we don't need that here, for now

Issue #560, opened 2021-05-05
Indicate form error better in video sheet


kdembler:
If your screen isn't high enough, the content rating input may be not visible but it's required. So if you missed it, submitting the form will fail but because you don't see it, you may not know what the problem is. We need to somehow elevate that error better

PR #569, opened 2021-05-05
Scroll to errors

eldiegod:
Fix: #560 

Came up with this solution to scroll into the errored required fields as react-use-form wasn't doing it by default for some reason.

### DIFF
--- a/src/shared/components/Select/Select.tsx
+++ b/src/shared/components/Select/Select.tsx
@@ -14,6 +14,7 @@ export type SelectProps<T = string> = {
   value?: T | null
   items: SelectItem<T>[]
   placeholder?: string
+  containerRef?: Ref<HTMLDivElement>
 } & InputBaseProps
 
 // don't use React.FC so we can use a generic type on a component
@@ -25,9 +26,9 @@ const Select = <T,>({
   value,
   disabled,
   onChange,
-  customRef,
+  containerRef,
   ...inputBaseProps
-}: SelectProps<T> & { customRef?: Ref<HTMLDivElement> }) => {
+}: SelectProps<T>) => {
   const itemsValues = items.map((item) => item.value)
 
   const handleItemSelect = (changes: UseSelectStateChange<T>) => {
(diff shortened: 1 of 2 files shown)


mikkio-j (approved):
Looks good 👍

drillprop (approved):
Looks good to me 👍🏻

src/shared/components/Select/Select.tsx
```diff
@@ -25,8 +25,9 @@ const Select = <T,>({
   value,
   disabled,
   onChange,
+  customRef,
```
kdembler: I guess `custom` here comes from the fact that we're not using `forwardRef` but not sure if that should leak to the consumer. Maybe let's just call this `containerRef`?

src/shared/components/Select/Select.tsx
```diff
@@ -25,8 +25,9 @@ const Select = <T,>({
...
   ...inputBaseProps
-}: SelectProps<T>) => {
+}: SelectProps<T> & { customRef?: Ref<HTMLDivElement> }) => {
```
kdembler: Let's move that into `SelectProps`

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -329,8 +331,15 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
               name="category"
               control={control}
               rules={requiredValidation('Video category')}
+              onFocus={() => {
+                categorySelectRef.current?.scrollIntoView({
+                  behavior: 'smooth',
+                  block: 'start',
+                })
```
kdembler: Let's move that block into some function like `handleFieldFocus` that can accept ref as an argument

src/shared/components/Select/Select.tsx
eldiegod: thats right, added the changes 👍

kdembler (approved):

Issue #479, opened 2021-04-19
Channel cover responsive tweaks


kdembler:
- remove cover button not visible on medium sized screens with a bright cover image
- long handle name not allowing to click on the cover image on medium screens

PR #570, opened 2021-05-06
Channel cover tweaks

mikkio-j:
Issue: #479

### DIFF
--- a/src/shared/components/ChannelCover/ChannelCover.style.ts
+++ b/src/shared/components/ChannelCover/ChannelCover.style.ts
@@ -142,7 +142,7 @@ export const EditableControls = styled.div`
   }
 
   ${media.xlarge} {
-    height: 70%;
+    height: 80%;
   }
 `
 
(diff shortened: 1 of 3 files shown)


drillprop (commented):
I still see some issues. Maybe it's a good idea to tackle these problems as a part of this PR?
1. If you hover on the edit button on medium sizes is barely visible.
image
2. IMHO remove button not looking great on medium sizes: 
image 
3. Also remove button looks weird together with the edit button. 
image
4. There is a small issue with an overlay on large screens 
image You could fix this by setting ` height: 80%;` on the line 145 in ChannelCover.style

drillprop (approved):
I just noticed that the design changed a lot, and looks much different than the view we currently have, but we can ignore it for now. I guess It's not in the scope of this task.

Looks good to me  👍🏻

mikkio-j:
> I just noticed that the design changed a lot, and looks much different than the view we currently have, but we can ignore it for now. I guess It's not in the scope of this task.
> 
> Looks good to me 👍🏻

Yup, channel cover got the rework

kdembler (approved):

PR #571, opened 2021-05-06
sentry console integration

kdembler:


mikkio-j (approved):
don't have possibility to test that integration 😃  but read sentry docs and the code looks good so 👍  The question would be if we want to capture just error level calls

kdembler:
Yeah I think just the errors are good enough for now :+1:

Issue #535, opened 2021-04-30
Use correct dimensions for uploaded images


kdembler:
Currently the uploaded image dimensions aren't correct values. My guess is we use height and width of the picker area instead of the dimensions of the cropped image which are different

PR #575, opened 2021-05-06
Images dimensions

mikkio-j:
Issue: #535 
Also added video's thumbnail and video's dimensions to uploads

### DIFF
--- a/src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
+++ b/src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
@@ -1,5 +1,4 @@
 import { IconButton } from '@/shared/components'
-import { ImageCropData } from '@/types/cropper'
 import React, { forwardRef, useCallback, useImperativeHandle, useRef, useState } from 'react'
 import { ActionDialogProps } from '../ActionDialog'
 import { CropperImageType, useCropper } from './cropper'
@@ -17,14 +16,14 @@ import {
 } from './ImageCropDialog.style'
 import { SvgGlyphPan, SvgGlyphZoomIn, SvgGlyphZoomOut } from '@/shared/icons'
 
-export type Dimensions = {
+export type AssetDimensions = {
   width: number
   height: number
 }
 
 export type ImageCropDialogProps = {
   imageType: CropperImageType
-  onConfirm: (croppedBlob: Blob, croppedUrl: string, imageCropData: Dimensions) => void
+  onConfirm: (croppedBlob: Blob, croppedUrl: string, imageCropData: AssetDimensions) => void
 } & Pick<ActionDialogProps, 'onExitClick'>
 
 export type ImageCropDialogImperativeHandle = {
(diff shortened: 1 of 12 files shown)


src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
```diff
@@ -17,9 +17,14 @@ import {
...
 
+export type Dimensions = {
+  width: number
+  height: number
+}
+
 export type ImageCropDialogProps = {
   imageType: CropperImageType
-  onConfirm: (croppedBlob: Blob, croppedUrl: string, imageCropData: ImageCropData) => void
+  onConfirm: (croppedBlob: Blob, croppedUrl: string, imageCropData: Dimensions) => void
```
drillprop: I think we could actually get rid of `cropper.ts`  in `src/types` with the `ImageCropDialog` type.
We should also remove `import { ImageCropData } from '@/types/cropper'` at the top of this file

drillprop (commented):
Find one thing to improve

src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
drillprop: On the second thought, maybe it's a better idea to move `Dimensions` to `@/types/cropper`? wdyt? Also don't like the name Dimensions - it seems too broad for my taste.

mikkio-j: You're right, `AssetDimensions` should be better, can't get more specific because this type goes for all assets

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -413,7 +411,9 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
                 />
                 <ImageCropDialog
                   imageType="cover"
-                  onConfirm={(blob, url, imageCropData) => onChange({ blob, url, imageCropData })}
+                  onConfirm={(blob, url, assetDimensions, imageCropData) =>
```
drillprop, line 414: We could simplify this:
```ts
<ImageCropDialog imageType="cover" onConfirm={onChange} ref={coverDialogRef} />
```

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -435,7 +435,9 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
                   />
                   <ImageCropDialog
                     imageType="avatar"
-                    onConfirm={(blob, url, imageCropData) => onChange({ blob, url, imageCropData })}
+                    onConfirm={(blob, url, assetDimensions, imageCropData) =>
```
drillprop, line 438: Same as above 
```ts
<ImageCropDialog imageType="avatar" onConfirm={onChange} ref={avatarDialogRef} />
```

drillprop: Erase that. I was wrong.  Leave it as it is

kdembler (approved):

Issue #519, opened 2021-04-28
No max number of videos in a grid row


kdembler:
On `master` we're capping the number of videos per row to 6 in a grid. Seems we broke that on `sumer`

<img width="1770" alt="Screenshot 2021-04-28 at 14 44 37" src="https://user-images.githubusercontent.com/12646744/116405884-70d78080-a830-11eb-8645-eb6b63649ac6.png">

PR #576, opened 2021-05-06
Limit grid columns to 6

eldiegod:
issue: #519


drillprop (approved):
Yep, I can confirm, this should fix the issue. 👍🏻  Good catch

Issue #559, opened 2021-05-05
Improve dialog handling


kdembler:
I think we could use some better dialog management, possibly with its own management hook. It'd be useful to be able to stack dialogs on top of each other (for example, transaction status covering video delete confirmation dialog), maybe we could use some kind of imperative API, as we do with the snackbars. Also #549 could be tackled here. We need to find some good approach for this, let's discuss. Let's create a proposal API first and only then start working on it

drillprop:
I started doing some tests already in the playground and I think doing something like we do in the `useSnackbar` is a great idea.

1. First, we could create some context. It will look something like this
```ts
type DialogState = {
  id: string
} & SomeDialogProps

export const DialogProvider: React.FC = ({ children }) => {
	const [dialogs, setDialogs] = useState<DialogState[]>([])
	
	const openDialog = (customId: string, dialogConfig?: SomeDialogProps) => {
           setDialogs([...dialogs, { id: customId, showDialog: true, ...dialogConfig }])
         }

	const closeDialog = (id: string) => {
              // filter dialogs here 
  }

return (
    <DialogContext.Provider value={{ openDialog, closeDialog }}>
      {dialogs.map(({ id, onExitClick, onPrimaryClick ...dialogProps }, idx) => (
        <SomeDialog
          key={idx}
          {...dialogProps}
            onExitClick={(e) => {
            closeDialog(id)
             onExitClick?.(e)
           }}
	   onPrimaryButtonClick={(e) => {
            closeDialog(id)
            onPrimaryButtonClick?.(e)
           }}
					
        />
      ))}
      {children}
    </DialogContext.Provider>
  )
}
```
As you see, it's very similar to the `useSnackbar` context hook. We will keep there an array of dialogs, and just filter or add more of them. Also, as you see I think we should close the dialog every time the user clicks every dialog button(exit button, primary button, or secondary button).  I just think it makes sense to do this as a default behavior. What do you think?

2. Worth mentioning, that we have a minor problem with dialogs. If we have two open dialogs at once and we close one, both of them will disappear: 
 
![dialog bug](https://user-images.githubusercontent.com/51168865/117929894-9f0f9280-b2fd-11eb-8e4d-2507c6dc30c7.gif)

This is caused by `BaseDialog` `useEffect`.To make this work, we will need to move `Portal` here(from `BaseDialog`) and call similar `useEffect` as we call in BaseDialog (but with different condition). Like this:
```ts
useEffect(() => {
[...]

PR #577, opened 2021-05-06
Indicate losing changes in upload video sheet

### LINKED ISSUE #527: Indicate losing changes in upload video sheet
Because the file data cannot be saved in a draft, we should do a better job of indicating that to the user. We already do the browser "Are you sure you want to quit" dialog on browser tab close, but we also should show something when the user closes the video sheet in-app. The design specifies this: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5863%3A977 I think that would work fine. We should make sure it only happens when there's actually some loss of data. For example, if the user has 10 drafts open, but none of them contains files, we shouldn't show any warning.

To do:
- Add a warning dialog when closing the action sheet and there would be loss of data
- Make sure the browser warning shows up only when it's needed

drillprop:
Fix #527 
Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=5863%3A977


drillprop:
I really wanted to not repeat myself here(with dialogs and methods), but I had some hard times accomplishing that. I'll try to look at it again and try to refactor it in the evening.  My mind is just too foggy right now.

Feel free to test it out and give suggestions, if you have any

src/hooks/useDisplayDataLostWarning.tsx
```diff
@@ -0,0 +1,47 @@
+import { MessageDialog } from '@/components'
+import React, { useState } from 'react'
+
+type OpenWarningDialogArgs = {
+  cancelCallback?: () => void
+  confirmCallback?: () => void
+}
+
+export const useDisplayDataLostWarning = () => {
```
mikkio-j, line 9: Not so sure about this hook, it's too specific imho. We have issue for handling dialogs (#559), so it will probably need to be removed after merging new dialog handling. Or It can be used as a base for that.

mikkio-j: so I guess we can stick with that for now

mikkio-j (commented):
Do you think we should add warning before creating new channel? Currently we can create new channel during the flow and there's no warning before transaction (after creation active channel's changed so we lose changes)

drillprop:
> Do you think we should add warning before creating new channel? Currently we can create new channel during the flow and there's no warning before transaction (after creation active channel's changed so we lose changes)

@mikkio-j added that

mikkio-j (approved):
Looks like everything's good 👍

src/components/Sidenav/SidenavBase.style.ts
```diff
@@ -79,7 +79,8 @@ export const ButtonGroup = styled.div`
   display: flex;
   flex-direction: column;
   justify-content: flex-end;
-  a:not(:first-of-type) {
+  a:not(:first-of-type),
+  button:first-of-type {
```
kdembler:
```suggestion
  * + * {
```

I think this will solve it?

src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
```diff
@@ -46,39 +50,52 @@ export const StudioSidenav: React.FC = () => {
...
     return item
   })
+  const { haveVideoTabsAssetCache } = useEditVideoSheet()
+
+  const handleClick = () => {
+    if (haveVideoTabsAssetCache) {
+      openWarningDialog({
+        confirmCallback: () => {
+          setExpanded(false)
+          navigate(absoluteRoutes.viewer.index())
```
kdembler, line 60: Shouldn't we reset the cache here?

src/hooks/useDisplayDataLostWarning.tsx
```diff
@@ -0,0 +1,47 @@
+import { MessageDialog } from '@/components'
+import React, { useState } from 'react'
+
+type OpenWarningDialogArgs = {
+  cancelCallback?: () => void
+  confirmCallback?: () => void
```
kdembler: how about `onCancel` and `onConfirm`?

src/hooks/useDisplayDataLostWarning.tsx
kdembler: Let's keep it as an exploration for now

src/hooks/useDisplayDataLostWarning.tsx
```diff
@@ -0,0 +1,47 @@
...
+      showDialog={lostDataDialogVisible}
+      onPrimaryButtonClick={confirmCloseSheet}
+      onSecondaryButtonClick={cancelCloseSheet}
+      onExitClick={cancelCloseSheet}
+      secondaryButtonText="Cancel"
+      variant="warning"
+    />
+  )
+  return {
+    DataLostWarningDialog: renderWarning,
```
kdembler, line 44: I don't like this pattern much, but I'm fine with using it for now since it will get reworked

src/hooks/useDisplayDataLostWarning.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const cancelCloseSheet = () => {
+    setLostDatadDialogVisible(false)
+    cancelWarningCallback?.()
+  }
+
+  const renderWarning = () => (
+    <MessageDialog
+      title="Video & image data will be lost"
+      description="Drafts are stored locally and dont contain metadata for video and image file - if you abandon the proccess those files will have to be uploaded again."
```
kdembler:
```suggestion
      title="Drafts' video & image data will be lost"
      description="Drafts' assets aren't stored permanently. If you proceed, you will need to reselect the files again."
```

src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
kdembler: I guess this will happen automatically because the video sheet will get closed?

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -66,6 +69,22 @@ export const EditVideoSheet: React.FC = () => {
...
+  const { DataLostWarningDialog, openWarningDialog } = useDisplayDataLostWarning()
+
+  useEffect(() => {
+    if (sheetState === 'closed' || !haveVideoTabsAssetCache) {
+      return
+    }
+
+    const beforeUnload = (e: BeforeUnloadEvent) => {
+      e.preventDefault()
+      e.returnValue = 'Do you want to leave this page? Changes that you made may not be saved.'
```
kdembler: Is this actually respected in any way? Seems all the dialog have default values?

src/hooks/useEditVideoSheet.tsx
```diff
@@ -30,6 +30,8 @@ type ContextValue = {
   setSelectedVideoTabCachedAssets: (files: InputFilesState) => void
   sheetState: EditVideoSheetState
   setSheetState: (state: EditVideoSheetState) => void
+  haveVideoTabsAssetCache: boolean
+  hasVideoTabAssetCache: (tabIdx: number) => boolean
```
kdembler: I think we need to improve the naming. It's hard to tell which is a boolean and which a function. How about `anyVideoTabsCachedAssets' for the boolean and 'hasVideoTabAnyCachedAssets'? A bit explicit but should be more readable

kdembler (commented):
Looking very good, just some minor stuff regarding the code

src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
drillprop: I don't think we need to do more work here. If we go to the viewer, the sheet will unmount and all the asset cache should be removed.

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
drillprop: You're right, I think it's deprecated

kdembler (approved):

Issue #517, opened 2021-04-28
No margin in viewer channel details


kdembler:
<img width="546" alt="Screenshot 2021-04-28 at 14 42 19" src="https://user-images.githubusercontent.com/12646744/116405478-fdce0a00-a82f-11eb-9db0-f9fa14d73073.png">

PR #578, opened 2021-05-06
Fix media query typos

eldiegod:
Fix: #517


kdembler (approved):
oh wow, good job for catching that

PR #579, opened 2021-05-06
Update member validation

### LINKED ISSUE #572: Validate member handle
when creating the new membership, we should validate the handle, it should match this regex: https://github.com/Joystream/joystream/blob/c5a8bbaf4e4aa19de435278c22fd3530bacedfc9/pioneer/packages/joy-members/src/EditForm.tsx#L22

also, we should change the placeholder value because the handle cannot contain spaces, we can make it `johnsmith` instead

drillprop:
Fix #572

### DIFF
--- a/src/utils/formValidationOptions.ts
+++ b/src/utils/formValidationOptions.ts
@@ -1,5 +1,6 @@
 import { ValidationRule, Message, Validate } from 'react-hook-form'
 import { isValid } from 'date-fns'
+import { URL_PATTERN } from '@/config/regex'
 
 type RegisterOptions = Partial<{
   required: Message | ValidationRule<boolean>
@@ -17,6 +18,7 @@ type TextValidationArgs = {
   minLength?: number
   required?: boolean
   pattern?: RegExp
+  message?: string
 }
 
 export const textFieldValidation = ({
@@ -25,6 +27,7 @@ export const textFieldValidation = ({
   maxLength,
   required = false,
   pattern,
+  message,
 }: TextValidationArgs): RegisterOptions => ({
   required: {
     value: required,
@@ -42,7 +45,7 @@ export const textFieldValidation = ({
     ? {
         pattern: {
           value: pattern,
-          message: `${name} must be a valid`,
+          message: message ? `${name} ${message}` : `${name} must be a valid`,
         },
       }
     : {}),
@@ -57,7 +60,7 @@ export const requiredValidation: (name: string) => RegisterOptions = (name) => (
 
 export const urlValidation: (name: string) => RegisterOptions = (name) => ({
   pattern: {
-    value: /[-a-zA-Z0-9@:%._\\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\\+.~#?&//=]*)?/gi,
+    value: URL_PATTERN,
     message: `${name} must be a valid url`,
   },
 })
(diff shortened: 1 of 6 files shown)


src/utils/formValidationOptions.ts
```diff
@@ -11,12 +11,21 @@ type RegisterOptions = Partial<{
...
+  pattern?: RegExp
+}
+
+export const textFieldValidation = ({
+  name,
+  minLength = 0,
+  maxLength,
+  required = false,
+  pattern,
+}: TextValidationArgs): RegisterOptions => ({
```
mikkio-j: that's what we needed 👍

mikkio-j (approved):

src/utils/formValidationOptions.ts
```diff
@@ -29,6 +38,14 @@ export const textFieldValidation: (
     value: maxLength,
     message: `${name} cannot be longer than ${maxLength} characters.`,
   },
+  ...(pattern
+    ? {
+        pattern: {
+          value: pattern,
+          message: `${name} must be a valid`,
```
kdembler: I think we will need to do a slightly better job of indicating to the user what it means for the handle to be `valid`. We don't want them to have to guess what works and what doesn't. So I guess it'd be best if we allowed passing a custom message to display in this case. (I guess we still can use the field name to stay consistent)

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -137,9 +137,17 @@ const CreateMemberView = () => {
...
+          placeholder="johnysmith"
           label="Member Name"
-          ref={register(textFieldValidation('Member name', 4, 40, true))}
+          ref={register(
+            textFieldValidation({
+              name: 'Member name',
+              maxLength: 40,
+              minLength: 4,
+              required: true,
+              pattern: /^[a-z0-9_]+$/,
```
kdembler: Let's put this in the `@/config` somewhere 🙏

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -148,7 +156,7 @@ const CreateMemberView = () => {
           label="About"
           placeholder="Anything you'd like to share about yourself with the Joystream community"
           maxLength={100}
-          ref={register(textFieldValidation('About', 0, 100))}
+          ref={register(textFieldValidation({ name: 'About', maxLength: 100 }))}
```
kdembler: Let's make that more like 1000

src/utils/formValidationOptions.ts
kdembler: I'd like the message to be something like this: `Member name may contain only lowercase letters, numbers and underscores`

src/utils/formValidationOptions.ts
```diff
@@ -11,12 +12,23 @@ type RegisterOptions = Partial<{
...
-  minLength: number,
-  maxLength: number,
+type TextValidationArgs = {
+  name: string
+  maxLength: number
+  minLength?: number
   required?: boolean
-) => RegisterOptions = (name, minLength, maxLength, required = false) => ({
+  pattern?: RegExp
+  message?: string
```
kdembler: Maybe `patternMessage` or something alike? This could hint that this is a generic message

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -137,9 +138,18 @@ const CreateMemberView = () => {
         />
         <StyledTextField
           name="handle"
-          placeholder="Johnny Smith"
+          placeholder="johnysmith"
```
kdembler: I'm sorry 🙃 

```suggestion
          placeholder="johnnysmith"
```

kdembler (approved):

PR #580, opened 2021-05-06
Fix transaction-illustration

### LINKED ISSUE #561: SVG bug when uploading first video
It seems the SVGs are colliding when uploading the first video. The SVG in transaction dialog seems to be masked by the SVG in "no videos yet". We should fix that

drillprop:
Fix #561


mikkio-j:
I believe we should also update `empty-videos-illustration.svg` ids cause it might happen again with different illustration

kdembler:
Good idea @mikkio-j 

In general maybe we should just do the same preprocessing on illustrations as we do for icons, then we would have certainty that the IDs are unique (not in scope here)

drillprop:
Ok, updated. There are still a couple of SVGs in assets that have ids mask0, mask1, etc. I think in the future we should do what @kdembler suggests. I'll create an issue for that

mikkio-j (approved):

Issue #573, opened 2021-05-06
Add license input to video upload


kdembler:
We've missed specifying the license in the initial design, it's been added since: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=8118%3A0

We should use license from `knownLicenses.json` for that

PR #582, opened 2021-05-07
add license select input

mikkio-j:
Issue: #573

### DIFF
--- a/src/joystream-lib/api.ts
+++ b/src/joystream-lib/api.ts
@@ -26,6 +26,7 @@ import {
 } from '@joystream/types/content'
 import {
   ChannelMetadata,
+  License,
   MediaType,
   PublishedBeforeJoystream,
   VideoMetadata,
@@ -284,6 +285,20 @@ export class JoystreamJs {
     }
     if (inputMetadata.hasMarketing != null) {
       protoMeta.setHasMarketing(inputMetadata.hasMarketing)
+    }
+
+    const protoLicenseType = new License()
+    if (inputMetadata.license?.code != null) {
+      protoLicenseType.setCode(inputMetadata.license.code)
+    }
+    if (inputMetadata.license?.attribution != null) {
+      protoLicenseType.setAttribution(inputMetadata.license.attribution)
+    }
+    if (inputMetadata.license?.customText != null) {
+      protoLicenseType.setCustomText(inputMetadata.license.customText)
+    }
+    if (protoLicenseType != null) {
+      protoMeta.setLicense(protoLicenseType)
     }
 
     const protoMediaType = new MediaType()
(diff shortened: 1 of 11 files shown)


src/joystream-lib/api.ts
```diff
@@ -286,6 +287,20 @@ export class JoystreamJs {
...
+    if (inputMetadata.license?.code != null) {
+      protoLicenseType.setCode(inputMetadata.license.code)
+    }
+    if (inputMetadata.license?.attribution != null) {
+      protoLicenseType.setAttribution(inputMetadata.license.attribution)
+    }
+    if (inputMetadata.license?.customText != null) {
+      protoLicenseType.setCustomText(inputMetadata.license.customText)
+    }
+    if (protoLicenseType != null) {
```
kdembler: I don't think we need that check, we create `protoLicenseType` ourselves so I don't think it will ever be nullish?

kdembler (commented):
Just a very quick look at how it works, will review the code next week.

There seems to be some weird delay on changing the license type. The attribution/custom text input shows up after some time. Maybe you need to rely on `watch` rather than `getValues` for conditional render?

https://user-images.githubusercontent.com/12646744/117437477-8df70800-af30-11eb-9613-d23cc6b8f278.mov

src/shared/components/Select/Select.tsx
```diff
@@ -15,6 +23,9 @@ export type SelectProps<T = string> = {
   items: SelectItem<T>[]
   placeholder?: string
   containerRef?: Ref<HTMLDivElement>
+  icon?: boolean
+  tooltipTexts?: string[]
+  headerTooltipTexts?: string[]
```
kdembler: I think those could be just part of the `SelectItem` type?

src/shared/components/Select/Select.tsx
```diff
@@ -15,6 +23,9 @@ export type SelectProps<T = string> = {
   items: SelectItem<T>[]
   placeholder?: string
   containerRef?: Ref<HTMLDivElement>
+  icon?: boolean
```
kdembler: I think the icon could also be done on by-item basis.

kdembler (commented):
Seem to be working fine, just some minor comments

src/shared/components/Select/Select.tsx
```diff
@@ -87,10 +83,10 @@ const Select = <T,>({
                 key={`${item.name}-${index}`}
                 {...getItemProps({ item: item.value, index })}
               >
-                {icon && tooltipTexts?.length && (
+                {item.tooltipText && (
```
kdembler, line 86: 👍

kdembler (approved):

PR #586, opened 2021-05-07
auth tweaks

kdembler:
summary of changes:
- `useActiveUser` -> `useUser`, new hook `useAuthorizedUser` for all the user fields should be set (no need to check if `channelId` is null for example)
- moved extensions accounts integration into `useUser`, this way joystream lib is decoupled from the browser env
- moved all membership queries usages into `useUser`, now all of the views don't need to reuse the exact same query, they will just get data from `useUser`
- minor tweaks in auth

Resolves #449 and other undocumented issues


src/components/StudioEntrypoint.tsx
```diff
@@ -15,36 +14,24 @@ type StudioEntrypointProps = {
...
-    },
-    {
-      skip: !accounts.length,
-    }
-  )
+    activeAccountId,
+    activeMemberId,
+    activeChannelId,
+    setActiveUser,
+    extensionConnected: extensionStatus,
```
drillprop:
```suggestion
    extensionConnected
```
and we could just remove line 28.

src/components/Topbar/StudioTopbar/StudioTopbar.tsx
```diff
@@ -51,43 +50,35 @@ type NavDrawerProps = {
...
       return
     }
-    setActiveChannel(channelId)
+    setActiveUser({ channelId })
     setDrawerActive(false)
   }
 
   const handleLogout = async () => {
-    await removeActiveUser()
+    resetActiveUser()
```
drillprop: We could get rid of `async` above

src/hooks/useUser/useUser.tsx
```diff
@@ -0,0 +1,177 @@
...
+    const account = accounts.find((a) => a.id === activeUserState.accountId)
+
+    if (!account) {
+      console.warn('Selected accountId not found in extension accounts, resetting user')
+      resetActiveUser()
+    }
+  }, [accounts, activeUserState.accountId, extensionConnected, resetActiveUser])
+
+  const userInitialized =
+    (extensionConnected === true && (!!memberships || !accounts?.length)) || extensionConnected === false
```
drillprop, line 119:
```suggestion
const userInitialized = !!(extensionConnected && (!!memberships || !accounts?.length))
```
Should work, right? Not 100% sure

src/views/studio/StudioLayout.tsx
```diff
@@ -47,25 +45,24 @@ const StudioLayout = () => {
...
-  } = useActiveUser()
-
-  const { memberships, loading: membershipsLoading } = useMemberships({
-    where: { controllerAccount_in: accounts.map((a) => a.id) },
-  })
+    activeAccountId,
+    activeMemberId,
+    activeChannelId,
+    extensionConnected: extensionStatus,
+    accounts,
```
drillprop:
```suggestion
```

src/views/studio/StudioLayout.tsx
```diff
@@ -47,25 +45,24 @@ const StudioLayout = () => {
...
-    loading: activeUserLoading,
-  } = useActiveUser()
-
-  const { memberships, loading: membershipsLoading } = useMemberships({
-    where: { controllerAccount_in: accounts.map((a) => a.id) },
-  })
+    activeAccountId,
+    activeMemberId,
+    activeChannelId,
+    extensionConnected: extensionStatus,
```
drillprop:
```suggestion
    extensionConnected
```
Analogous to StudioEntrypoint. Remove line 58 and then on the line 63: 
`const accountSet = !!activeAccountId && !!extensionConnected`
I think it's better

drillprop (commented):
I think I've tested all the sign-in scenarios and it works really great 👍🏻  The code also looks solid and it's definitely easier to handle.

src/hooks/useUser/useUser.tsx
kdembler: I don't think it does work because `false` is a valid state for `extensionConnected` (user doesn't have the extension yet)

kdembler:
@drillprop Thanks for catching those, should be good now

drillprop (approved):
Much-needed changes. Looks good to me 🎉

src/views/studio/StudioLayout.tsx
```diff
@@ -47,25 +45,24 @@ const StudioLayout = () => {
...
 
   const [enterLocation] = useState(location.pathname)
-  const extensionConnectionLoading = extensionStatus === null
-  const extensionConnected = extensionStatus === true
   const hasMembership = !!memberships?.length
 
-  const accountSet = !!accountId && extensionConnected
-  const memberSet = accountSet && !!memberId && hasMembership
-  const channelSet = memberSet && !!channelId && hasMembership
+  const accountSet = !!activeAccountId && extensionConnected
```
drillprop:
```suggestion
  const accountSet = !!activeAccountId && !!extensionConnected
```
sorry, last thing. Should work now

kdembler: Yep, already pushed that :+1:

PR #596, opened 2021-05-11
Legal DMCA

eldiegod:
Issue: #498

### DIFF
--- a/src/views/LegalView/LegalView.style.tsx
+++ b/src/views/LegalView/LegalView.style.tsx
@@ -1,0 +1,60 @@
+import React from 'react'
+import { Text } from '@/shared/components'
+import { colors, sizes } from '@/shared/theme'
+import styled from '@emotion/styled'
+
+export const Box = styled.div`
+  scroll-behavior: smooth;
+  text-align: left;
+  margin-bottom: ${sizes(10)};
+  position: relative;
+  height: 85vh;
+  width: 100%;
+  background-color: ${colors.gray[800]};
+  overflow: auto;
+`
+export const TextWrapper = styled.div`
+  margin: ${sizes(9)} ${sizes(17)} ${sizes(9)} ${sizes(8)};
+  a {
+    text-decoration: none;
+    color: ${colors.gray[50]};
+  }
+`
+const _Paragraph = styled(Text)`
+  margin-top: ${sizes(6)};
+  ol + & {
+    margin-top: 0;
+  }
+`
+
+export const Paragraph: React.FC<{ header?: boolean }> = ({ header, ...props }) => (
+  <_Paragraph
+    variant={header ? 'h6' : 'body2'}
+    // secondary={!header}
+    {...props}
+  />
+)
+// wrapper so we can use a specific element to render
+const _StyledText = styled(Text)``
+export const ListItem: React.FC = (props) => (
+  <li>
+    <_StyledText
+      variant="body2"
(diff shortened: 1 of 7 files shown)


src/views/LegalView/LegalView.style.tsx
```diff
@@ -0,0 +1,60 @@
...
+  />
+)
+// wrapper so we can use a specific element to render
+const _StyledText = styled(Text)``
+export const ListItem: React.FC = (props) => (
+  <li>
+    <_StyledText
+      variant="body2"
+      // secondary
+      // as="li"
```
drillprop, line 56:
```suggestion
```

src/views/LegalView/LegalView.style.tsx
```diff
@@ -0,0 +1,60 @@
...
+  margin-top: ${sizes(6)};
+  ol + & {
+    margin-top: 0;
+  }
+`
+
+export const Paragraph: React.FC<{ header?: boolean }> = ({ header, ...props }) => (
+  <_Paragraph
+    variant={header ? 'h6' : 'body2'}
+    // secondary={!header}
```
drillprop, line 36:
```suggestion
```

src/views/LegalView/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+            <br />
+            Designated DMCA Agent Jsgenesis AS
+            <br />
+            CO UMA Workspace Stenersgata 8
+            <br />
+            Oslo, 0184 Norway
+            <br />
+            <br />
+            Or by telephone: +44789553019
+          </Paragraph>
```
drillprop, line 83: A small thing, but you could use the `<address />` tag here, and add links to the phone number and email address: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

src/views/LegalView/LegalView.style.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const Box = styled.div`
+  scroll-behavior: smooth;
+  text-align: left;
+  margin-bottom: ${sizes(10)};
+  position: relative;
+  height: 85vh;
+  width: 100%;
+  background-color: ${colors.gray[800]};
+  overflow: auto;
```
kdembler, line 14: There are 2 scroll containers in the view, one is this Box, second is the `body` itself. Once you stop scrolling the box, the body will scroll. Maybe we should just make the Box grow and let the user scroll the body?

src/views/LegalView/LegalView.style.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+`
+const _Paragraph = styled(Text)`
+  margin-top: ${sizes(6)};
+  ol + & {
+    margin-top: 0;
+  }
+`
+
+export const Paragraph: React.FC<{ header?: boolean }> = ({ header, ...props }) => (
```
kdembler, line 33: We need to adjust colors for the text. The body text is too dark for the background, the contrast is too low. In terms step we use `secondary` for everything except the header

src/views/LegalView/LegalView.style.tsx
eldiegod: Left this as comments so when we merge to summer we can uncomment those as they don't work on master, should have mentioned in the PR 😜

kdembler (approved):

Issue #491, opened 2021-04-22
Update my videos empty states


kdembler:
The design has recently added empty state to specific tabs in my videos view. Let's update what we have

PR #598, opened 2021-05-11
Empty videos tab

eldiegod:
issue: #491


src/assets/theater-mask.svg
```diff
@@ -0,0 +1,126 @@
...
+<path fill-rule="evenodd" clip-rule="evenodd" d="M286.458 220.575C286.72 216.828 284.004 215.806 283.296 214.922C285.524 215.991 287.587 216.716 288.347 218.698C289.152 220.802 288.996 223.764 284.697 224.217C285.249 223.502 286.301 222.823 286.458 220.575Z" fill="white"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M306.906 198.933C302.717 198.277 300.949 199.887 299.686 201.165C302.817 200.721 303.806 200.555 306.561 201.212C309.316 201.868 310.082 202.815 311.714 204.032C311.825 202.189 311.095 199.59 306.906 198.933Z" fill="black"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M286.465 194.208C283.492 192.472 281.423 193.415 279.891 194.201C282.516 194.669 283.356 194.794 285.214 196.135C287.072 197.476 287.259 198.528 288.003 200.057C288.884 198.452 289.437 195.944 286.465 194.208Z" fill="black"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M315.464 228.808C312.823 231.209 308.133 234.614 303.928 236.301C307.52 232.147 307.317 231.156 312.709 226.674C318.102 222.192 315.276 217.425 318.454 216.393C320.498 221.466 319.881 224.79 315.464 228.808Z" fill="white"/>
+</g>
+<defs>
+<clipPath id="clip0">
```
mikkio-j: let's make it more specific like `theater-mask-clip0`. You know why 😄  Other thing is that `clip-path` may cause some conflicts. Might be worth double-checking that.

Also, this illustration is almost the same as `empty-uploads-illustration` (this one doesn't have this circular shadow). Not sure if we should use the same in both places or not

src/views/studio/MyVideosView/EmptyVideosView.tsx
```diff
@@ -87,38 +92,39 @@ const MessageView = styled.div`
...
-export const EmptyVideos = () => {
+export const EmptyVideos: React.FC<{ isDraftTab: boolean }> = ({ isDraftTab }) => {
   return (
     <Container>
-      <WellIllustration />
+      <TheaterMaskIllustration />
       <Message>
-        <Text variant="h5">No videos found...</Text>
+        <Text secondary variant="body2">
+          There are no {isDraftTab ? 'drafts' : 'videos'} currently
```
mikkio-j: In designs texts are specified to each tab, like `Currently there are no unlisted videos` for unlisted etc.
Maybe we should do that here as well

src/assets/theater-mask.svg
eldiegod: removed the clip-path at the end it was not needed and used only theater-mask.svg as they are indeed the same

src/views/studio/MyVideosView/EmptyVideosView.tsx
```diff
@@ -87,38 +92,39 @@ const MessageView = styled.div`
...
   return (
     <Container>
-      <WellIllustration />
+      <TheaterMaskIllustration />
       <Message>
-        <Text variant="h5">No videos found...</Text>
+        <Text secondary variant="body2">
+          {text}
+        </Text>
+        <Button icon={<SvgGlyphAddVideo />} to={absoluteRoutes.studio.editVideo()} size="large">
```
kdembler, line 103: This button will grow with the container, it should be sized independently:
Screenshot 2021-05-17 at 11 52 06

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -230,7 +230,19 @@ export const MyVideosView = () => {
...
+            {((isDraftTab && drafts.length === 0) || (!isDraftTab && totalCount === 0 && !loading)) && (
+              <EmptyVideos
+                text={
+                  currentTabName === 'All Videos'
+                    ? 'There are no videos currently'
+                    : currentTabName === 'Published'
+                    ? 'There are no published videos currently'
+                    : currentTabName === 'Drafts'
+                    ? 'There are no drafts currently'
+                    : 'There are no unlisted videos currently'
```
```suggestion
                    ? "You don't have any published videos at the moment"
                    : currentTabName === 'Published'
                    ? "You don't have any public videos at the moment"
                    : currentTabName === 'Drafts'
                    ? "You don't have any drafts at the moment"
                    : "You don't have any unlisted videos at the moment"
```

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -230,7 +230,19 @@ export const MyVideosView = () => {
             <Grid maxColumns={null} onResize={handleOnResizeGrid}>
               {gridContent}
             </Grid>
-            {((isDraftTab && drafts.length === 0) || (!isDraftTab && totalCount === 0 && !loading)) && <EmptyVideos />}
+            {((isDraftTab && drafts.length === 0) || (!isDraftTab && totalCount === 0 && !loading)) && (
+              <EmptyVideos
+                text={
+                  currentTabName === 'All Videos'
+                    ? 'There are no videos currently'
+                    : currentTabName === 'Published'
```
kdembler: Let's change that tab name to `Public`, all of the videos are "Published" tbh

src/views/studio/MyVideosView/EmptyVideosView.tsx
```diff
@@ -87,38 +92,39 @@ const MessageView = styled.div`
...
   }
 `
 
 const Message = styled.div`
-  display: flex;
-  flex-direction: column;
+  position: relative;
+  top: -256px;
+  display: grid;
+  gap: ${sizes(4)};
```
drillprop, line 126:
```suggestion
  gap: ${sizes(4)};
  justify-items: center;
```
Maybe it's just a matter of preference, but I think the button should have always the same width

drillprop (commented):
Found one thing, but overall looks good 👍🏻

src/views/studio/MyVideosView/EmptyVideosView.tsx
drillprop: This line should fix what @kdembler brought here https://github.com/Joystream/atlas/pull/598#discussion_r633392594

drillprop (approved):

kdembler (approved):

Issue #588, opened 2021-05-10
Ensure handle for new member isn't used already


kdembler:
We should do some client-side validation to ensure the handle for new member isn't already in use. This will potentially save the user an error modal. We can use `membershipByUniqueInput(where:{handle:"klaudiusz"})` query for that and see if it returns anything

PR #600, opened 2021-05-11
validation for already used member's handle

mikkio-j:
Issue: #588

### DIFF
--- a/src/views/studio/CreateMemberView/CreateMemberView.tsx
+++ b/src/views/studio/CreateMemberView/CreateMemberView.tsx
@@ -6,6 +6,7 @@ import TextArea from '@/shared/components/TextArea'
 import { textFieldValidation, urlValidation } from '@/utils/formValidationOptions'
 import { debounce } from 'lodash'
 import React, { useEffect, useState } from 'react'
+import { useLazyQuery } from '@apollo/client'
 import { useForm } from 'react-hook-form'
 import { useNavigate } from 'react-router'
 import { FAUCET_URL } from '@/config/urls'
@@ -20,6 +21,7 @@ import {
   StyledTextField,
 } from './CreateMemberView.style'
 import { useQueryNodeStateSubscription } from '@/api/hooks'
+import { GetMembershipDocument, GetMembershipQuery, GetMembershipQueryVariables } from '@/api/queries'
 import axios, { AxiosError } from 'axios'
 import { MemberId } from '@/joystream-lib'
 import { MEMBERSHIP_NAME_PATTERN } from '@/config/regex'
@@ -35,7 +37,7 @@ const CreateMemberView = () => {
   const { nodeConnectionStatus } = useConnectionStatus()
 
   const navigate = useNavigate()
-  const { register, handleSubmit, errors, trigger, setError: setInputError } = useForm<Inputs>({
+  const { register, handleSubmit, errors, trigger, setError: setInputError, getValues } = useForm<Inputs>({
     shouldFocusError: false,
     defaultValues: {
       handle: '',
@@ -54,6 +56,13 @@ const CreateMemberView = () => {
     throw queryNodeStateError
   }
 
+  const [getMember, { data: memberData }] = useLazyQuery<GetMembershipQuery, GetMembershipQueryVariables>(
+    GetMembershipDocument,
+    {
+      onCompleted: () => createMemberOnCompleted(),
+    }
+  )
+
   // success
   useEffect(() => {
     if (!isSubmitting || !membershipBlock || !queryNodeState || !activeAccountId) {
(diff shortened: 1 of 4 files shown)


src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -54,6 +56,13 @@ const CreateMemberView = () => {
     throw queryNodeStateError
   }
 
+  const [getMember, { data: memberData }] = useLazyQuery<GetMembershipQuery, GetMembershipQueryVariables>(
```
drillprop: We could create a custom hook for this query analogous to `useMembership`.

drillprop (commented):
It's doesn't work as it supposed to. If you create a user with the taken handle, and then when you click  `create membership` a second time with the same handle input, the error will disappear and nothing will happen.

src/views/studio/CreateMemberView/CreateMemberView.tsx
mikkio-j: Had some struggles with introducing custom lazy hook. It would require some additional work. Since it will be used only here I decided to stick with it. If we will need more lazy hooks (or we will use that one frequently) in the future I believe we can add them then.

drillprop: No problem, I'm okay with that 👍🏻

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -97,6 +104,16 @@ const CreateMemberView = () => {
     debounceAvatarChange(value)
   }
 
+  const debounceHandleChange = debounce(async (value) => {
+    clearErrors('handle')
+    getMember({ variables: { where: { handle: value } } })
+  }, 1000)
+
+  const handleHandleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
drillprop: handleHandleChange ❤️  
handleMemberName should also work well I think 😅 , but not a big deal for me, you can leave it as it is now

drillprop (approved):
Works well now 👍🏻 . Found some issues regarding avatar validation, but maybe I'll create a separate issue for that.

src/views/studio/CreateMemberView/CreateMemberView.tsx
mikkio-j: Yup, chuckled a little as well when writing this 😅   On the other hand, wanted to be consistent here

kdembler: I mean, it makes sense... 😅

src/views/studio/CreateMemberView/CreateMemberView.tsx
kdembler: Our codegen actually generates the lazy hooks as well. If we don't want to go through `@/api/hooks`, let's just use `useGetMembershipLazyQuery` that's already generated

kdembler (changes requested):
I think we will need to tweak it up a bit. With the current solution, it's possible to write up an already used handle and quickly submit before the request finishes. The submit won't wait for validation to complete and will proceed.

react-hook-form should actually allow you to do both async and debounced validation, just with the `validate` option, check out this snippet:
```
validate: debounce(async value => {
  await sleep(300);
  return value === "bill";
})
```

In our case, we could call the `getMember` inside the `validate`. Let me know if you run into trouble with that

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -133,10 +138,25 @@ const CreateMemberView = () => {
...
+                    const {
+                      data: { membershipByUniqueInput },
+                    } = await refetchMember()
+                    resolve(membershipByUniqueInput?.handle !== handle)
+                  }, 500)(value)
+                })
+              },
             })
           )}
+          onChange={async () => await trigger('handle')}
```
kdembler: Do we need that line? I think `react-hook-form` will automatically trigger validation on submit? The side effect of this is that user will get error messages when typing (for example "handle too short" when I just started typing). It's a bad UX to display errors too early. If they submit once, the error checking mode should change to `onChange` and the user should get updated right away

mikkio-j: Good point, removed that `onChange`. It's not needed for reason other than showing error before submit

kdembler (approved):

src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -133,10 +138,24 @@ const CreateMemberView = () => {
...
               patternMessage: 'may contain only lowercase letters, numbers and underscores',
+              validate: async (value) => {
+                // Wrapping it up with promise and resolving comparison inside debounce
+                // debounce() will not automatically do the return, which is needed for validation
+                return new Promise((resolve) => {
+                  debounce(async (handle) => {
+                    const {
+                      data: { membershipByUniqueInput },
+                    } = await refetchMember()
+                    resolve(membershipByUniqueInput?.handle !== handle)
```
kdembler: Was about to merge but just noticed this, I'm not sure if this check makes sense. We want to make sure that the membership with that handle doesn't exist right? So I'd expect
```suggestion
                    resolve(!membershipByUniqueInput)
```

PR #601, opened 2021-05-12
Extend video sheet backdrop over the topnav

### LINKED ISSUE #597: Extend video sheet backdrop over the topnav
Design indicates that the backdrop the video sheet has should extend over the topnav. It makes sense since expanding the member drawer will not work anyway, so let's just cover it:

Screenshot 2021-05-11 at 14 10 11

drillprop:
Fix #597

### DIFF
--- a/src/shared/components/HamburgerButton/HamburgerButton.style.ts
+++ b/src/shared/components/HamburgerButton/HamburgerButton.style.ts
@@ -14,7 +14,7 @@ export const Hamburger = styled.button`
   top: 18px;
   // FIXME: this shouldn't be set to 'fixed', button should work on its own, consumer may set it to fixed
   position: fixed;
-  z-index: ${zIndex.header + 1};
+  z-index: ${zIndex.sideNav + 1};
   cursor: pointer;
   background-color: ${colors.gray[700]};
   border: none;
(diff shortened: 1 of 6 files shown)


src/shared/components/HamburgerButton/HamburgerButton.style.ts
```diff
@@ -14,7 +14,7 @@ export const Hamburger = styled.button`
   top: 18px;
   // FIXME: this shouldn't be set to 'fixed', button should work on its own, consumer may set it to fixed
   position: fixed;
-  z-index: ${zIndex.header + 1};
+  z-index: ${zIndex.sideNav + 1};
```
mikkio-j: I know that this incrementation was here before, but maybe since we're adding `sideNav` index, we could also do `nearSideNav` with value like 210 or so.

mikkio-j (approved):
Looks good, one thing worth noticing 👍

src/shared/components/HamburgerButton/HamburgerButton.style.ts
drillprop: Should work now, didn't have to add `nearSideNav`

kdembler (approved):

PR #603, opened 2021-05-12
Fix issue with missing assets when editing video

drillprop:
There was a bug with the missing thumbnail and video when a user was trying to edit the video. It looked like this: 
thumbnail bug

This PR should fix the problem

### DIFF
--- a/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
+++ b/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
@@ -133,6 +133,9 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
   })
 
   useEffect(() => {
+    if (isEdit) {
+      return
+    }
     // reset multifileselect when sheetState is closed
     if (sheetState === 'closed') {
       setValue('assets', {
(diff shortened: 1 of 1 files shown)


kdembler:
I don't really understand how that change fixes the problem. That `useEffect` makes changes only if the sheet is closed. How does that break anything?

drillprop:
So, in the first place, I used that useEffect to fix the issue with not resetting MulitfileSelect. We had a problem which you bring up to me when I was working on warnings:
> 1. create new draft
> 2. set video
> 3. close the action sheet
> 4. confirm losing changes
> 
> and then I open that draft for edit again, the form looks like it has the video file selected

The problem is when we close the video sheet, we don't get rid of the edit video form, we just hide it(maybe we should change this?).  And then when we open again, it will have the previous state, which was updated by this `useEffect`.

When we editing videos, every time when we open the sheet, we are not fetching again the assets. So, to fix that, I just add this condition. 
I hope it makes sense

kdembler:
I see. I think the better solution would be to make sure that when the form loads it always loads the correct values, regardless of the current state. Nevertheless, this fixes the bug so let's get it merged!

kdembler (approved):

Issue #492, opened 2021-04-22
Allow changing sort direction in my videos


kdembler:
We should allow user to sort the videos in ascending/descending order by the creation date.

kdembler:
We need to make sure this doesn't conflict with logic for refetching videos after we create a new one

PR #604, opened 2021-05-12
Sorting Videos

eldiegod:
Issue: #492


src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -2,15 +2,25 @@ import { useVideos } from '@/api/hooks'
...
+  StyledDismissibleMessage,
+  TabsContainer,
+  ViewContainer,
+} from './MyVideos.styles'
 import { removeVideoFromCache } from '@/utils/cachingAssets'
+import { VideoOrderByInput } from '@/api/queries'
 
 const TABS = ['All Videos', 'Published', 'Drafts', 'Unlisted'] as const
+const SORT_OPTIONS = [
+  { name: 'ascending', value: VideoOrderByInput.CreatedAtAsc },
```
mikkio-j: I think we should change naming here to "Newest/Oldest"

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -238,6 +265,20 @@ export const MyVideosView = () => {
           <>
             <TabsContainer>
               <Tabs initialIndex={0} tabs={mappedTabs} onSelectTab={handleSetCurrentTab} />
+              <SortContainer>
+                <Text variant="body2">Sort by</Text>
+                <Select
```
mikkio-j, line 247: We should be able to change height of that element. I know it has 50px everywhere in forms, but in this case it has 40px, according to design

mikkio-j (commented):
Logic looks good to me, two minor things I noticed

src/views/studio/MyVideosView/MyVideosView.tsx
kdembler: It seems in the design selects now have two sizing options. Would it be a lot of work to add that?

src/views/studio/MyVideosView/MyVideosView.tsx
kdembler: let's make it `Newest first`/`Oldest first`

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -238,6 +265,20 @@ export const MyVideosView = () => {
...
+                  helperText={false}
+                  value={sortVideosBy}
+                  items={[...SORT_OPTIONS]}
```
kdembler, line 250: Why create new copy on every render instead of just passing the constant array?

src/api/hooks/video.ts
```diff
@@ -32,8 +33,13 @@ export const useVideos = (variables?: GetVideosQueryVariables, opts?: VideosOpts
       where: variables?.where,
     },
   })
+  const videos =
+    variables?.orderBy === VideoOrderByInput.CreatedAtAsc
+      ? data?.videos?.slice()?.sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime())
+      : data?.videos?.slice()?.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime())
```
kdembler: I'm not sure if this actually achieves the goal of this issue. It will work fine when you have just one page of videos, but when your videos will span more than one page and you select `Oldest first`, this will still make the pages ordered in the newest first order, but just sort the items on each page. This will need further improvement

kdembler (changes requested):

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: TS doesnt like it

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: That should be in a different PR no?

src/api/hooks/video.ts
eldiegod: True, will do it from the read

src/api/client/cache.ts
```diff
@@ -44,7 +44,10 @@ const cache = new InMemoryCache({
             // if offset and limit are not provided.
             const offset = args?.offset ?? 0
             const limit = args?.limit ?? existing?.length
-            return existing?.slice(offset, offset + limit)
+            const sortingASC = args?.orderBy === VideoOrderByInput.CreatedAtAsc
+            return sortingASC
+              ? existing?.slice(offset, offset + limit)
+              : existing?.slice(offset, offset + limit).reverse()
```
mikkio-j: I'm not sure if that's fully working, seems it's not covering the case when we have 2 pages of videos

eldiegod: True, I think the reverse should come before the slice here

mikkio-j (approved):

src/api/client/cache.ts
```diff
@@ -44,7 +44,13 @@ const cache = new InMemoryCache({
...
             const offset = args?.offset ?? 0
             const limit = args?.limit ?? existing?.length
-            return existing?.slice(offset, offset + limit)
+            const sortingASC = args?.orderBy === VideoOrderByInput.CreatedAtAsc
+            return sortingASC
+              ? existing?.slice(offset, offset + limit)
+              : existing
+                  ?.slice()
+                  .reverse()
+                  ?.slice(offset, offset + limit)
```
kdembler: Let's split slicing and sorting to improve readability here. Also I'm not sure if `reverse()` call is enough. There is no guarantee that the data in the cache is already sorted. Maybe let's do a proper sort here

src/shared/components/InputBase/InputBase.tsx
```diff
@@ -5,7 +5,7 @@ import { FormGroup, LabelText } from './InputBase.styles'
 export type InputBaseProps = {
   error?: boolean
   warning?: boolean
-  helperText?: string
+  helperText?: string | boolean
```
kdembler: Why would that be a `boolean`?

src/views/studio/MyVideosView/MyVideosView.tsx
kdembler: If we remove `as const` it shouldn't complain?

kdembler (commented):
Needs rebase as well

src/api/client/cache.ts
eldiegod: Fair added full sorting just in case

src/shared/components/InputBase/InputBase.tsx
eldiegod: It's was a work around for removing the helper text node that even if empty was adding the container for it and occpuying space, other option cold be to check for empty string arg for not rendering the HelperText node

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: That would make TS show the type of the array as `{
    name: string;
    value: VideoOrderByInput;
}[]` which would not give type safety for things we do like `typeof SORT_OPTIONS[number]` as any string/VideoOrderByInput combination would be valid instead we want that only the literals defined in the array to be valid + TS would allow the array to be mutable

kdembler:
@eldiegod needs rebase again, instructions in the chat

Issue #477, opened 2021-04-19
Improve dialog responsiveness 


kdembler:
Let's adjust dialog behaviour on mobile when there's a lot of vertical content. We should probably make it scrollable so it doesn't overflow the screen. It needs `max-height` or something like this

PR #605, opened 2021-05-13
add max-height and scroll to base dialog

mikkio-j:
Issue: #477

### DIFF
--- a/src/components/Dialogs/BaseDialog/BaseDialog.style.ts
+++ b/src/components/Dialogs/BaseDialog/BaseDialog.style.ts
@@ -12,7 +12,7 @@ export const StyledContainer = styled.div`
   width: 90%;
   max-width: 440px;
   min-height: 150px;
-  max-height: 70vh;
+  max-height: 80vh;
   overflow: auto;
   margin: ${sizes(16)} auto;
   color: ${colors.white};
--- a/src/components/Dialogs/BaseDialog/BaseDialog.style.ts
+++ b/src/components/Dialogs/BaseDialog/BaseDialog.style.ts
@@ -12,17 +12,14 @@ export const StyledContainer = styled.div`
   width: 90%;
   max-width: 440px;
   min-height: 150px;
-  max-height: 550px;
-  overflow-y: scroll;
+  max-height: 70vh;
+  overflow: auto;
   margin: ${sizes(16)} auto;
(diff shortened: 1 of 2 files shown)


src/components/Dialogs/BaseDialog/BaseDialog.style.ts
```diff
@@ -12,13 +12,17 @@ export const StyledContainer = styled.div`
   width: 90%;
   max-width: 440px;
   min-height: 150px;
-  overflow: hidden;
+  max-height: 550px;
+  overflow-y: scroll;
```
drillprop: I think `overflow: auto;` should be a better choice for this. We don't want to always show scroll, we want to show the scroll only when it's needed.

src/components/Dialogs/BaseDialog/BaseDialog.style.ts
```diff
@@ -12,13 +12,17 @@ export const StyledContainer = styled.div`
...
   margin: ${sizes(16)} auto;
   color: ${colors.white};
   background-color: ${colors.gray[700]};
   padding: var(--dialog-padding);
   box-shadow: 0 8px 8px rgba(0, 0, 0, 0.12), 0 24px 40px rgba(0, 0, 0, 0.16);
-
+  ${media.small} {
+    max-height: 650px;
+    overflow: hidden;
```
drillprop: Not 100% sure, but with `overflow: auto` we could get rid of the `overflow: hidden` entirely, and just adjust `max-height`. Hidden overflow could be bad for laptop screens. But it needs some testing.

Also maybe we should consider setting these values based on the height of the viewport, not width.

src/components/SignInSteps/AccountStep.style.ts
```diff
@@ -36,7 +36,7 @@ export const SubTitle = styled(Text)`
 
 export const AccountsWrapper = styled.div`
   width: 100%;
-  max-height: 300px;
+  max-height: 200px;
```
drillprop: ok, so we have max-height and min-height with the same value. We could just replace them with just `height`, right?

drillprop (approved):
Last thing 😅

drillprop (approved):
👍🏻

kdembler (approved):

Issue #465, opened 2021-04-19
Indicate asset being unavailable


kdembler:
When viewing channel/video details in edit views, we need to be able to handle some assets still being uploaded. For example, if you update your channel with a new avatar, its upload may take some time. In the meantime, we need to be able to present that in the UI in some way. We may not have access to the file.

There's design for this now in the channel view in Figma

PR #611, opened 2021-05-13
add image asset upload failed indicator

mikkio-j:
Resolves: #465


src/components/VideoPreview.tsx
```diff
@@ -78,6 +85,7 @@ export const VideoPreviewPublisher: React.FC<VideoPreviewWPublisherProps> = ({ i
       duration={video?.duration}
       views={video?.views}
       thumbnailUrl={thumbnailPhotoUrl}
+      hasThumbnailUploadFailed={hasThumbnailUploadFailed}
```
drillprop, line 85: what if the video upload failed, which is more common?
Generally,  it's a bigger problem, because the design is not clear about it. After the transaction is done we are adding the video to MyVideosView, but it's not available yet, it still needs to be uploaded. 
We can even go to the URL where is located(but we can't play it).
Maybe we should add a tab - `videos that's need to be uploaded` or smth(bad name for a tab I know :P)? 

Just want to raise a subject here, I don't think, it's not required to do this in this particular PR

src/shared/components/Avatar/Avatar.tsx
```diff
@@ -66,6 +68,10 @@ const Avatar: React.FC<AvatarProps> = ({
               <StyledPlaceholder rounded />
             ) : imageUrl ? (
               <StyledImage src={imageUrl} onError={onError} />
+            ) : hasAvatarUploadFailed ? (
+              <NewChannelAvatar>
+                <SvgLargeUploadFailed />
+              </NewChannelAvatar>
```
drillprop, line 74: image
This looks bad on TopBar, but not sure if we will show this icon on the top bar

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -339,6 +340,16 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
     },
   ]
 
+  const hasAvatarUploadFailed =
```
drillprop: I think we could create some function, for all the assets(thumbnail, cover, avatar) to not repeat these lines over and over.

src/shared/components/Avatar/Avatar.tsx
mikkio-j: this won't be displayed in topbar, at least designs didn't include that. I think we should go for the silhouette here

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
mikkio-j: these are not the same lines, tried to make that a generic function, but got into too much trouble with types. I'll leave it as it is for now since I want it to keep it simple

src/components/Topbar/StudioTopbar/StudioTopbar.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useState, useEffect, useRef } from 'react'
 import { useUser, useDisplayDataLostWarning, useEditVideoSheet } from '@/hooks'
-import { BasicChannelFieldsFragment } from '@/api/queries'
+import { AssetAvailability, BasicChannelFieldsFragment } from '@/api/queries'
```
kdembler: Doesn't seem needed here?

src/components/VideoPreview.tsx
```diff
@@ -68,6 +70,11 @@ export const VideoPreviewPublisher: React.FC<VideoPreviewWPublisherProps> = ({ i
     video?.channel.avatarPhotoDataObject
   )
 
+  const hasThumbnailUploadFailed =
+    video?.thumbnailPhotoAvailability === AssetAvailability.Pending &&
+    uploadsState.flat().find((asset) => asset.ipfsContentId === video.thumbnailPhotoDataObject?.ipfsContentId)
+      ?.lastStatus !== 'completed'
```
kdembler: I think we shouldn't need to rely on `uploadsState`, I don't like this coupling much. Wouldn't just checking if the asset is pending work?

mikkio-j: At the beginning it relied only on `AssetAvailability`, but then I discovered some cases where this didn't quite work. Will go back to that and double-check that

mikkio-j: should be working now without `uploadsState`

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -88,7 +89,7 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
   const { channel, loading, error, refetch: refetchChannel, client } = useChannel(activeChannelId || '', {
     skip: newChannel || !activeChannelId,
   })
-  const { startFileUpload } = useUploadsManager(activeChannelId || '')
+  const { startFileUpload, uploadsState } = useUploadsManager(activeChannelId || '')
```
kdembler:
```suggestion
  const { startFileUpload } = useUploadsManager(activeChannelId || '')
```

kdembler (approved):

Issue #610, opened 2021-05-13
Edit video sheet will rerender on every upload progress update


kdembler:
Because the `EditVideoSheet` uses uploads manager to be able to start the download, every time the context changes (on upload progress, couple times a second), the form will rerender. This can lead to a janky experience, with the sheet animation looking really bad. We should probably store file upload progress some other way so updates to it don't make the video sheet rerender. Maybe debouncing the updates would be enough

PR #618, opened 2021-05-14
add throttle to onUploadProgress

mikkio-j:
Issue: #610 

that should do it. Please let me know if you still experience sheet's stuttering

### DIFF
--- a/src/hooks/useUploadsManager/useUploadsManager.tsx
+++ b/src/hooks/useUploadsManager/useUploadsManager.tsx
@@ -149,11 +149,17 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
         if (!opts?.isReUpload && file) {
           addAsset({ ...asset, lastStatus: 'inProgress', size: file.size })
         }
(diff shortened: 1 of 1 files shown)


src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -170,10 +174,13 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
               }
             },
           },
-          onUploadProgress: ({ loaded, total }: ProgressEvent) => {
-            updateAsset(asset.contentId, 'inProgress')
```
kdembler, line 174: Will this work without this? If we do a reconnect we will set status to `reconnecting`, I don't think there's anything that will change it back?

mikkio-j: If the `opts?.changeHost` is true it will set it again to `inProgress`, but yeah forgot about case if e.g. third try of reconnecting will succeed. Will add that

drillprop (approved):

Issue #599, opened 2021-05-11
Make the video sheet properly react to window height change


kdembler:
Currently the video sheet will use `window.innerHeight` to know how deep it should slide down. We don't react to changes to the window size and if you resize your browser the video sheet will stay in place making it invisible/cover the screen. Let's use some hook to measure the screen height there instead

PR #619, opened 2021-05-14
Video sheet properly reacts to window height change

eldiegod:
Issue: #599

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "react-scripts": "4.0.1",
     "react-spring": "^8.0.27",
     "react-transition-group": "^4.4.1",
+    "react-use-measure": "^2.0.4",
     "retry-axios": "^2.4.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -113,6 +113,7 @@
     "react-scripts": "4.0.1",
     "react-spring": "^8.0.27",
     "react-transition-group": "^4.4.1",
+    "react-use-measure": "^2.0.4",
```
mikkio-j, line 116: I believe you were using that in one of the first drafts for the sheet, that's a cool util, we might use that in other areas in the future so 👌

mikkio-j (approved):

package.json
eldiegod: Yes thats right, very useful 😎

kdembler: Yeah it's on me for removing it, I may have tried to oversimplify things 😅

kdembler (approved):

PR #620, opened 2021-05-14
Improve dialogs

### LINKED ISSUE #549: Dialogs don't properly handle being replaced
When there is one dialog visible on the screen, and we replace it with a different one, the original will still do a fade out animation before unmounting, which will cause the new one to shift in layout.

https://user-images.githubusercontent.com/12646744/117005991-2a7a9980-ace8-11eb-9b2a-36ee51a3b716.mov

### LINKED ISSUE #559: Improve dialog handling

drillprop:
Fix #559 
Fix #549 

Set is as a draft for now, because animations of dialog are kinda broken now and I couldn't find a good solution how to bring them back.  Will look at it again tomorrow. 

However, you can test it and throw in your suggestions of course!


src/views/studio/CreateMemberView/CreateMemberView.tsx
```diff
@@ -30,6 +29,9 @@ type Inputs = {
   about: string
 }
 
+const ERROR_DIALOG = 'ERROR_DIALOG'
+const CREATE_MEMBERSHIP_DIALOG = 'CREATE_MEMBERSHIP_DIALOG'
```
mikkio-j: I think we should move all of those (that includes other places where the `openDialog` is used) to `useDialog` and put them as enum or type, should be better to have it in one place. Or is it here to be sure that they will not be used twice?

mikkio-j (commented):
localhost_3000_playground_dialogs
Got it when called two openDialog on mount. Should it be this way or should we put them on top of each other?

drillprop:
> Got it when called two openDialog on mount. Should it be this way or should we put them on top of each other?

Shouldn't be this way. I think I broke this when I was working on retrieving animations. Will look at it again.

src/views/studio/CreateMemberView/CreateMemberView.tsx
drillprop: I don't like this - every time we use `openDialog` we would have to add something to `type` or `enum`, this feels weird. 

But to make sure to not create dialog with the same id twice I've updated openDialog function, and add some console warnings.

drillprop:
> Got it when called two openDialog on mount. Should it be this way or should we put them on top of each other?

Fixed

src/views/studio/CreateMemberView/CreateMemberView.tsx
eldiegod: An idea could be to make `openDialog` return a uniq id for that dialog, no need to pass anything ourselves, and we could save that on a useState locally and later use to close the dialog

eldiegod:
Looks like its working good functionality-wise and added a small optional comment 👍

eldiegod (approved):

mikkio-j (approved):

drillprop:
I did some changes. I was highly inspired by `react-modal-hook` library, so some code may look similar. 
For now, If you want to use dialog, you can use it in three ways: 
1. You can pass message dialog props to useDialog directly and then call it:
```js
const [openYourDialog, closeYourDialog] = useDialog({...messageDialogProps})
// ...
openDialog()
```
2. You can pass message dialog props to returned function like this:
```js
const [openYourDialog, closeYourDialog] = useDialog()
// ...
openDialog({...messageDialogProps})
```
3. You can also pass a callback that returns some JSX, like this:
```js
const [openYourDialog, closeYourDialog]  = useDialog(() => (
  <div>Some dialog</div>
))
```
Not sure if we actually need the latter, but I could get rid of it easily

As you can see, you don't need a hardcoded ID anymore. If you want to use two dialogs in a component, you just using `useDialog` again: 
```js
const [openFirstDialog, closeFirstDialog] = useDialog({...messageDialogProps})
const [openSecondDialog, closeSecondDialog] = useDialog({...messageDialogProps})
// ...
openFirstDialog()
openSecondDialog()
```

src/views/playground/PlaygroundLayout.tsx
```diff
@@ -32,34 +36,39 @@ const playgroundRoutes = [
...
-        </UploadManagerProvider>
+                  </NavContainer>
+                  <ContentContainer>
+                    <Routes>
+                      {playgroundRoutes.map((route) => (
+                        <Link key={route.path} to={route.path}>
+                          {route.name}
+                        </Link>
+                      ))}
+                    </Routes>
```
kdembler: I don't think that's what you wanted to do :P

src/components/Dialogs/BaseDialog/BaseDialog.tsx
```diff
@@ -1,42 +1,36 @@
...
-import { StyledContainer, StyledExitButton } from './BaseDialog.style'
 import { transitions } from '@/shared/theme'
-import { SvgGlyphClose } from '@/shared/icons'
 
 export type BaseDialogProps = {
   showDialog?: boolean
   exitButton?: boolean
   onExitClick?: (e: React.MouseEvent<HTMLButtonElement>) => void
   className?: string
+  isActionDialog?: boolean
```
kdembler: I don't think this is used?

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,121 @@
+import MessageDialog, { MessageDialogProps } from '@/components/Dialogs/MessageDialog'
+import { createId } from '@/utils/createId'
+import React, { useCallback, useEffect, useState, useContext, useMemo } from 'react'
+import { TransitionGroup } from 'react-transition-group'
+
+type DialogRendererProps = {
+  component: React.FunctionComponent
+}
+
+interface DialogsContainerProps {
```
kdembler: We use `type` for those

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+  dialogs: Record<string, React.FunctionComponent>
+  component?: React.ComponentType
+  container?: Element
```
kdembler: Is this used in any way?

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const DialogsRenderer = ({ component, ...rest }: DialogRendererProps) => component(rest)
+
+export const DialogsContainer = ({ dialogs }: DialogsContainerProps) => {
+  return (
+    <TransitionGroup>
+      {Object.keys(dialogs).map((key) => (
+        <DialogsRenderer key={key} component={dialogs[key]} />
+      ))}
```
kdembler: I think something like this would work? We can get rid of `DialogsRenderer` then

```suggestion
      {Object.keys(dialogs).map((key) => {
        const Component = dialogs[key]
        return <Component key={key} />
      })}
```

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+type DialogContextValue = {
+  openDialog: (key: string, dialog: React.FunctionComponent<{ in?: boolean }>) => void
+  closeDialog: (key: string) => void
+}
+const DialogContext = React.createContext<undefined | DialogContextValue>(undefined)
+DialogContext.displayName = 'DialogsContext'
+
+type DialogsProviderProps = {
+  container?: Element
+}
```
kdembler: Likewise, I don't think it's needed?

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const closeDialog = useCallback(
+    (key: string) =>
+      setDialogs((dialogs) => {
+        if (!dialogs[key]) {
+          return dialogs
+        }
+        const newDialogs = { ...dialogs }
+        delete newDialogs[key]
+        return newDialogs
```
kdembler: Should be simpler

```suggestion
        const { [key]: _, ...filteredDialogs } = dialogs
        return filteredDialogs
```

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+interface DialogsContainerProps {
+  dialogs: Record<string, React.FunctionComponent>
+  component?: React.ComponentType
+  container?: Element
+}
+
+const DialogsRenderer = ({ component, ...rest }: DialogRendererProps) => component(rest)
+
+export const DialogsContainer = ({ dialogs }: DialogsContainerProps) => {
```
kdembler: I guess we could just move it inside the context and get rid of the props and additional components?

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+  if (ctx === undefined) {
+    throw new Error('useDialog must be used within a DialogProvider')
+  }
+
+  const { openDialog, closeDialog } = ctx
+
+  const [isShown, setShown] = useState(false)
+  const showModal = useCallback(() => setShown(true), [])
+  const hideModal = useCallback(() => setShown(false), [])
+  const dialogId = useMemo(() => createId(), [])
```
kdembler: _In theory_, we shouldn't use `useMemo` this way, docs state there is no guarantee that this will not re-run (it won't in current version of react but still). I think we could just use this instead?
```suggestion
  const dialogId = useRef(createId()).current
```

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }, [closeDialog, dialogProps, dialogId, isShown, openDialog])
+
+  const _openDialog = useCallback(
+    (args?: MessageDialogProps) =>
+      dialogProps === undefined
+        ? openDialog(dialogId, ({ in: inAnimation }) => <MessageDialog {...args} showDialog={inAnimation} />)
+        : showModal(),
+    [dialogProps, dialogId, openDialog, showModal]
+  )
```
kdembler: I'm not sure about this bypassing `isShown`. We should use a single approach for displaying a dialog, not two different ways to call `openDialog`

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+  useEffect(() => {
+    if (isShown) {
+      openDialog(dialogId, ({ in: inAnimation }) => <MessageDialog {...dialogProps} showDialog={inAnimation} />)
+    } else {
+      closeDialog(dialogId)
+    }
+    return () => {
+      closeDialog(dialogId)
+    }
+  }, [closeDialog, dialogProps, dialogId, isShown, openDialog])
```
kdembler: Won't this keep opening and closing the dialog? If you pass some object to `useDialog` its reference will be different on every render so this will keep calling `closeDialog` and `openDialog`?

kdembler (commented):
Reviewed the most important parts, will take a look at the rest tomorrow

src/hooks/useDialog/useDialog.tsx
drillprop: I just got rid of this `useEffect` entirely. It should work well now

src/hooks/useDialog/useDialog.tsx
drillprop: nice! Forgot that I can use computed property like this

src/hooks/useDialog/useDialog.tsx
```diff
@@ -0,0 +1,76 @@
+import MessageDialog, { MessageDialogProps } from '@/components/Dialogs/MessageDialog'
+import { createId } from '@/utils/createId'
+import React, { useCallback, useState, useContext, useMemo, useRef } from 'react'
+import { TransitionGroup } from 'react-transition-group'
+
+type DialogRendererProps = {
+  component: React.FunctionComponent
+}
```
kdembler: This isn't used anymore right?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -350,25 +350,52 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
...
+        <StyledTitleSection className={transitions.names.slide}>
           <Controller
-            name="cover"
+            name="avatar"
             control={control}
             render={({ value, onChange }) => (
               <>
-                <ChannelCover
-                  coverPhotoUrl={loading ? null : value.url}
-                  hasCoverUploadFailed={hasCoverUploadFailed}
```
kdembler, line 364: This shouldn't be removed?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -378,159 +405,125 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
...
             <Controller
-              name="avatar"
+              name="language"
               control={control}
+              rules={requiredValidation('Language')}
               render={({ value, onChange }) => (
-                <>
-                  <StyledAvatar
-                    imageUrl={value.url}
-                    hasAvatarUploadFailed={hasAvatarUploadFailed}
```
kdembler, line 395: This shouldn't be removed?

src/views/studio/StudioLayout.tsx
```diff
@@ -37,6 +38,7 @@ import {
   PrivateRoute,
   StudioLoading,
 } from '@/components'
+import { TransitionGroup } from 'react-transition-group'
```
kdembler: Doesn't seem needed here

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
kdembler: The diff here is a bit broken because of whitespace changes but `hasCoverUploadFailed` line seems to be removed

kdembler (commented):
Did some testing and other than those minor comments, all looks good (needs rebase as well)! Really happy to have this, I think it will be a great improvement!

kdembler (approved):

PR #621, opened 2021-05-14
Fix issue with updating video without thumbnail

### LINKED ISSUE #581: It's possible to update video with no thumbnail
if you remove the thumbnail, but not replace it, you can actually still publish changes. 
After the transaction is done, the thumbnail will remain the same though. 
upload bug

drillprop:
Should fix #581

### DIFF
--- a/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
+++ b/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
@@ -221,7 +221,7 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
       setFileSelectError('Video file cannot be empty')
       return
     }
-    if (!data.assets.thumbnail?.url && !dirtyFields.assets?.thumbnail) {
+    if (!data.assets.thumbnail?.url) {
       setFileSelectError('Thumbnail cannot be empty')
       return
     }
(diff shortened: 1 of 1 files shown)


mikkio-j (approved):

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -221,7 +221,7 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
       setFileSelectError('Video file cannot be empty')
       return
     }
-    if ((!isEdit || dirtyFields.assets?.thumbnail) && !data.assets.thumbnail?.blob) {
+    if (!data.assets.thumbnail?.url && !dirtyFields.assets?.thumbnail) {
```
kdembler: It seems to work but I don't understand how. The condition for the error is `no URL and field not dirty`. How does that work for a new video? In that case the field shouldn't be dirty if the thumbnail wasn't provided right?

drillprop: Oh, wait. I just noticed that `dirtyFields.assets?.thumbnail` is always `undefined`, no matter what you do with it. I could just set this to `if (!data.assets.thumbnail?.url)` and I'll do that

kdembler (approved):

Issue #595, opened 2021-05-11
General responsive tweaks


mikkio-j:
During testing of responsiveness we've found some things that need to be improved, here is the list:

- ~~Signin~~
  - ~~Coins on small screen should not be visible (found in:  `join?step=0`)~~
  ~~(it should follow RWD from design:~~ (covered in: #626 )
[https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=7677%3A2556](https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=7677%3A2556))
<img src="https://user-images.githubusercontent.com/59691518/117788462-b68c4400-b247-11eb-8963-7028bd0612d9.png" width="200" />

  - ~~Dialog's content not scrollable - covered in other issue (#477) (found in: `join?step=2`)~~
  - ~~Joystream logo not aligned on smaller screens~~ (covered in: #626 )
 <img src="https://user-images.githubusercontent.com/59691518/117788600-d1f74f00-b247-11eb-86a2-ed2e095ccae1.png" width="200" />

- New channel
  - Select for language and privacy for smaller screens (some languages are not visible as well as second item in privacy)
  <img src="https://user-images.githubusercontent.com/59691518/117788644-dc194d80-b247-11eb-9f0d-b8d5ad78f502.png" width="200" />

- ~~My videos~~
  - ~~no videos svg illustration overflow (horizontal scroll)~~
- ~~Video sheet~~
  - ~~"Add new video" should not be visible on smaller screens~~ (covered in: #633  )
- Uploads view
  - Needs some general tweaks in upload bars and dropdown (not covered in designs). We've discussed that during one of the meetings and came to conclusion that it's not in the scope of sumer release

drillprop:
`Topbar` should be scrollable. If I have a lot of channels I can't click logout on smaller screens.
![image](https://user-images.githubusercontent.com/51168865/117816874-48f01000-b267-11eb-8b71-e54411f07a2c.png)

PR #626, opened 2021-05-17
Update sign in views to the latest design

### LINKED ISSUE #506: Update sign in views to the latest design
There were some changes in sign-in views recently. Mostly regarding RWD

Figma: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6737%3A0

- the coins illustration in `sign in process` has different sizes and slightly different positions on different views. We need to update this. 
- make sure that every view is centered on smaller screens
- I think we could skip changes to the membership hub(the carousel).

drillprop:
Fix #506 
Should also fix some issues addressed in #595


src/views/studio/SignInJoinView/SignInProcessView/SignInProcessView.style.ts
```diff
@@ -22,23 +32,30 @@ export const SubTitle = styled(Text)`
...
-  position: fixed;
+  position: relative;
   margin-top: 50px;
-  ${media.small} {
-    width: 300px;
+  width: 640px;
+  display: none;
+  ${media.large} {
+    left: -100px;
+    display: block;
```
mikkio-j: According to design coins illustration is visible on medium devices as well. We should include that, it will probably require some fiddling with position and size

mikkio-j (commented):
Maybe we should include logo alignment issue? Since it's only the case n sign in views

drillprop:
> Maybe we should include logo alignment issue? Since it's only the case n sign in views

Fixed

mikkio-j (approved):

kdembler (approved):

Issue #627, opened 2021-05-17
Infinite fetch loop on video delete


kdembler:
I've found one case that makes the app break:

1. Have one video
2. Open the "Published" tab in my videos
3. Open the video edit
4. Click "Delete video" in the video sheet
5. The video will get deleted, and we will keep firing requests to refetch videos on the "Published" tab
6. If I go to "All videos" tab, the requests will stop and the app will hide the tabs and show the "No videos" state instead

My guess is that this will work the same without editing the video (with delete from the context menu). Anyway, we need to fix this

PR #630, opened 2021-05-17
Fix MyVideos filters

### LINKED ISSUE #624: Filters in My videos not working
It seems the tabs in my videos view are not properly filtering videos when it comes to "unlisted" tab, it shows all the videos, not only unlisted ones

drillprop:
Fix #624
Fix the problem which was brought here: https://github.com/Joystream/atlas/pull/630#issuecomment-842889836

### DIFF
--- a/src/views/studio/MyVideosView/MyVideosView.tsx
+++ b/src/views/studio/MyVideosView/MyVideosView.tsx
@@ -258,18 +258,6 @@ export const MyVideosView = () => {
   )
 }
 
-const getPublicness = (currentTabName: typeof TABS[number]) => {
-  switch (currentTabName) {
-    case 'Published':
-      return true
-    case 'Unlisted':
-      return false
-    case 'All Videos':
-    default:
-      return undefined
-  }
-}
-
 export default MyVideosView
 
 const usePagination = (currentTab: number) => {
--- a/src/views/studio/MyVideosView/MyVideosView.tsx
+++ b/src/views/studio/MyVideosView/MyVideosView.tsx
@@ -15,23 +15,28 @@ const ROWS_AMOUNT = 4
 
 export const MyVideosView = () => {
   const navigate = useNavigate()
-  const { setSheetState, videoTabs, addVideoTab, removeVideoTab } = useEditVideoSheet()
+  const {
+    setSheetState,
+    videoTabs,
+    addVideoTab,
+    removeVideoTab,
+    currentVideosTab,
+    setCurrentVideosTab,
+    isDraftTab,
+    isPublic_eq,
+  } = useEditVideoSheet()
   const { displaySnackbar } = useSnackbar()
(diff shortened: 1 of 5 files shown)


drillprop:
~~My second commit should also fix #627~~
~~I decided to not create another PR for that. They're not related but I think it will be easier to test both of them at one go.~~

drillprop:
Just noticed, there is one more problem. Steps to reproduce 
1. Select the unlisted tab
2. Click create video
3. Create a public video


After creating a video, the video sheet should minimize and you'll see your newly created public video in the unlisted tab. An analogous situation will happen if you select the published tab and create an unlisted video.
Not sure how to tackle this, yet 

Just to clarify, tested also on `sumer`, this bug was not introduced in this PR, it was there previously.

mikkio-j (approved):

mikkio-j:
> Just noticed, there is one more problem. Steps to reproduce
> 
> 1. Select the unlisted tab
> 2. Click create video
> 3. Create a public video
> 
> After creating a video, the video sheet should minimize and you'll see your newly created public video in the unlisted tab. An analogous situation will happen if you select the published tab and create an unlisted video.
> Not sure how to tackle this, yet
> 
> Just to clarify, tested also on `sumer`, this bug was not introduced in this PR, it was there previously.

Could you make new issue for that and mark it as bug? Looked into it, but not sure how to resolve it

eldiegod:
Works good and as mentioned there is that issue, pretty sure it has to do with the fn `writeVideoDataInCache` but doesn't look easy to solve

> Just noticed, there is one more problem. Steps to reproduce
> 
>     1. Select the unlisted tab
> 
>     2. Click create video
> 
>     3. Create a public video
> 
> 
> After creating a video, the video sheet should minimize and you'll see your newly created public video in the unlisted tab. An analogous situation will happen if you select the published tab and create an unlisted video.
> Not sure how to tackle this, yet
> 
> Just to clarify, tested also on `sumer`, this bug was not introduced in this PR, it was there previously.

eldiegod (approved):

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -41,7 +46,7 @@ export const MyVideosView = () => {
         isPublic_eq,
       },
     },
-    { notifyOnNetworkStatusChange: true }
+    { notifyOnNetworkStatusChange: true, nextFetchPolicy: 'cache-and-network' }
```
drillprop: This is needed because if I, for example, create a public video when I have selected the unlisted tab, it will not refetch the video if I switch to  `All videos` tab. 
Just noticed - there is a downside to this - If I create an unlisted video when I have an unlisted tab selected we will not see a thumbnail right away. 
Don't know what to do about it.

drillprop:
Pagination is broken with these changes. I need to figure it out, what's happening

kdembler (approved):

kdembler (approved):

Issue #590, opened 2021-05-10
Clicking on edit video doesn't make it selected


kdembler:
If the tab for video edit is already open and there are other tabs and other tab is selected. Clicking on edit video (in video preview) will make the video sheet open but not select the desired video

PR #631, opened 2021-05-17
Select video on pullup/edit

eldiegod:
fix: #590

### DIFF
--- a/src/views/studio/MyVideosView/MyVideosView.tsx
+++ b/src/views/studio/MyVideosView/MyVideosView.tsx
@@ -94,8 +94,6 @@ export const MyVideosView = () => {
       return
     }
     addVideoTab({ id, isDraft: opts.draft })
-    const tabIdx = videoTabs.findIndex((t) => t.id === id)
-    if (tabIdx >= 0) setSelectedVideoTabIdx(tabIdx)
     if (opts.minimized) {
       displaySnackbar({
         title: 'Video opened in a new tab',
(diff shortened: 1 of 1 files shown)


drillprop (approved):
Looks good to me!

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -94,7 +94,8 @@ export const MyVideosView = () => {
       return
     }
     addVideoTab({ id, isDraft: opts.draft })
-
+    const tabIdx = videoTabs.findIndex((t) => t.id === id)
+    if (tabIdx >= 0) setSelectedVideoTabIdx(tabIdx)
```
kdembler: We probably should call `addVideoTab` only if the tab is not found right?

kdembler (changes requested):
I think this doesn't do exactly what we'd want to. In my head, "open in new tab" shouldn't select the tab, just open it. Only the edit should select it as well

src/views/studio/MyVideosView/MyVideosView.tsx
eldiegod: Not necessarily as the `addVideoTab` fn already checks for that so would be redundant

eldiegod:
> want to. In my head, "open in new tab" shouldn't select the tab, just open it. Only the edit should select it as well

I think it wouldn't be an unexpected behavior in this case since the bar is minimized anyway so it doesn't switch the user's context. Although if we decide for the behaviour to be only open and not select then we run into the issue of having tabs open but nothing selected which crashes the app, wdyt? @kdembler

kdembler (approved):
Ok, let's not overcomplicate this

Issue #489, opened 2021-04-22
Improve video tabs bar


kdembler:
The current button will always keep to the right, which may make it not always visible if there's a lot of tabs open. The designers have proposed a solution in which the button is getting sticky to the right. Let's implement something similar. I don't think it needs to be 1:1 with the design, let's discuss what works without too much effort into this.

The design has also removed the `Add new video` header on the left, we should get rid of it

PR #633, opened 2021-05-18
Tabs bar update

mikkio-j:
Issue: #489


src/views/studio/EditVideoSheet/EditVideoTabsBar/EditVideoTabsBar.style.ts
```diff
@@ -19,10 +23,12 @@ export const TabsContainer = styled.div`
   display: flex;
   align-items: center;
   overflow: auto hidden;
+  width: 100%;
+  &::-webkit-scrollbar {
+    height: 0;
+  }
```
drillprop, line 29: This should hide the scrollbar on chrome, right? What about firefox? After quick research I think we could add experimental `scrollbar-width: none;` I think I used it also in the regular `Tabs` component.

drillprop (approved):

PR #635, opened 2021-05-18
Fix error handling issues

### LINKED ISSUE #606: Error handling issues
1. This is not related entirely to error handling. When a user will go to the studio for the first time, he will see this dialog: 
image
Unfortunately, if he chooses to reject he will just see `Add extension` step in `/signin`. Then if the user changes his mind and he will want to allow access,  it could be not obvious to him how to do it (he will need to go polkadot extension ⇒ settings ⇒ manage website access). Maybe we could add some information in the `ExtensionStep`: 

> If you have installed the extension and still see this dialog, follow these steps to fix it.... -> and some steps to fix it 

2. Signed user still can go to `studio/membership/new` and if he tries to create a member he will see this dialog. 
image
which doesn't say much. **Solution**: we should redirect the user if the accountId is set in localStorage.
3. If a user has no internet connection he could still perform some actions and it's not ideal. He could click `new membership` on the membership hub and even go through all steps. He can also click his accounts on the membership hub. In my opinion, as a solution, we should just disable these buttons/tiles.
4. There is a bigger problem when there is no node connection. A user can still publish changes and he won't see any error(except `no node connection` notification). For example, if he uploads a video,  he'll see waiting for signature dialog just hanging infinitely. 
5. Unfortunately there is still a problem with image validation. My PR https://github.com/Joystream/atlas/pull/587 didn't solve this entirely. I didn't cover channel creation, only video creation. In create channel view, image crop dialog will show up even if you pick an image that is not supported or it's broken. We didn't handle errors there.
6. There are two issues with validation in the create member view.
Images validation not working correctly:
membership bug
You can still submit the form when you exceed the about's maximum characters limit
image

If you find more I'll edit this issue and add it to the list.

drillprop:
Fix #606
Regarding point one - I tried to add some message beneath the button, but it kinda looks bad. I will ignore it for now.  It's not something urgent anyway, I think

### DIFF
--- a/src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
+++ b/src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
@@ -83,10 +83,8 @@ const ImageCropDialogComponent: React.ForwardRefRenderFunction<
       const fileUrl = URL.createObjectURL(selectedFile)
       setEditedImageHref(fileUrl)
       setShowDialog(true)
-    } catch (error: unknown) {
-      if (error instanceof Error) {
-        onError?.(error)
-      }
+    } catch (error) {
+      onError?.(error)
       console.error(error)
     }
   }
--- a/src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
+++ b/src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
@@ -16,6 +16,7 @@ import {
   ZoomControl,
 } from './ImageCropDialog.style'
 import { SvgGlyphPan, SvgGlyphZoomIn, SvgGlyphZoomOut } from '@/shared/icons'
+import { validateImage } from '@/utils/image'
 
 export type ImageCropDialogProps = {
   imageType: CropperImageType
@@ -25,6 +26,7 @@ export type ImageCropDialogProps = {
     assetDimensions: AssetDimensions,
     imageCropData: ImageCropData
   ) => void
+  onError?: (error: Error) => void
 } & Pick<ActionDialogProps, 'onExitClick'>
 
 export type ImageCropDialogImperativeHandle = {
@@ -34,7 +36,7 @@ export type ImageCropDialogImperativeHandle = {
 const ImageCropDialogComponent: React.ForwardRefRenderFunction<
   ImageCropDialogImperativeHandle,
   ImageCropDialogProps
-> = ({ imageType, onConfirm, onExitClick }, ref) => {
+> = ({ imageType, onConfirm, onExitClick, onError }, ref) => {
   const [showDialog, setShowDialog] = useState(false)
   const inputRef = useRef<HTMLInputElement>(null)
   const [imageEl, setImageEl] = useState<HTMLImageElement | null>(null)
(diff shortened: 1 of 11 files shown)


drillprop:
Draft, I forgot about point 5

mikkio-j (commented):
Code looks and covers almost everything mentioned in the issue. ~~Everything but the first point, I mean displaying message if the access to page is denied in polkadot extension? Do you think we should add that here?~~

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -361,6 +363,12 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
                   onConfirm={(blob, url, assetDimensions, imageCropData) =>
                     onChange({ blob, url, assetDimensions, imageCropData })
                   }
+                  onError={() =>
+                    displaySnackbar({
+                      title: 'Cannot load the image. Choose another.',
```
kdembler, line 370:
```suggestion
                      title: 'Cannot load the image. Please use a different one.',
```

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -385,6 +393,12 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
                     onConfirm={(blob, url, assetDimensions, imageCropData) =>
                       onChange({ blob, url, assetDimensions, imageCropData })
                     }
+                    onError={() =>
+                      displaySnackbar({
+                        title: 'Cannot load the image. Choose another.',
```
kdembler, line 400:
```suggestion
                      title: 'Cannot load the image. Please use a different one.',
```

kdembler (approved):

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -385,6 +393,12 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
                     onConfirm={(blob, url, assetDimensions, imageCropData) =>
                       onChange({ blob, url, assetDimensions, imageCropData })
                     }
+                    onError={() =>
+                      displaySnackbar({
+                      title: 'Cannot load the image. Please use a different one.',
```
kdembler:
```suggestion
                        title: 'Cannot load the image. Please use a different one.',
```

mikkio-j (approved):

src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
```diff
@@ -69,16 +71,24 @@ const ImageCropDialogComponent: React.ForwardRefRenderFunction<
...
-    setEditedImageHref(fileUrl)
-    setShowDialog(true)
+    try {
+      const selectedFile = files[0]
+      await validateImage(selectedFile)
+      const fileUrl = URL.createObjectURL(selectedFile)
+      setEditedImageHref(fileUrl)
+      setShowDialog(true)
+    } catch (error: unknown) {
+      if (error instanceof Error) {
```
kdembler: why wouldn't we call `onError` if `error` isn't instance of `Error` (somehow)?

kdembler (commented):
looks good

Issue #584, opened 2021-05-07
Automatic crop on re-selecting image assets


mikkio-j:
In crop dialog we should add functionality that will crop the image automatically for the users when they re-pick the same file. Also we need to implement that in uploads view (on reselect, check if the same image was picked and crop it based on cropData)

PR #636, opened 2021-05-19
Automatic crop

mikkio-j:
Issue: #584 

There's one issue with the current state of this PR. Re-selected images are being cropped correctly with data from `cropData`, but the `canvas.toBlob()` sometimes renders an image that slightly differs in size (e.g. for avatars the difference is around 50 bytes). Not all the time though (according to my attempts around 10%-20% of files have different size). Difference in size causes change in file hash, so we obviously can't let that happen.

If you guys have some suggestions how we can resolve it, it would be great 🙏


drillprop (commented):
The code looks good, but I have some doubts about how its works right now. 
In my opinion when a user clicks `Reconnect file` he shouldn't see `ImageCropDialog`. He can't change the position/zoom of the image anyway, because we will reject this file. 

We should just open the file dialog and let the user only pick the right file. Users shouldn't have the possibility to edit images in this case.

drillprop (approved):
> There's one issue with the current state of this PR. Re-selected images are being cropped correctly with data from `cropData`, but the `canvas.toBlob()` sometimes renders an image that slightly differs in size (e.g. for avatars the difference is around 50 bytes). Not all the time though (according to my attempts around 10%-20% of files have different size). Difference in size causes change in file hash, so we obviously can't let that happen.
> 
> If you guys have some suggestions how we can resolve it, it would be great 🙏

Did you fix it?
Managed to reupload images several times with no problems. Works well for me. 👍🏻

mikkio-j:
> Did you fix it?
> Managed to reupload images several times with no problems. Works well for me. 👍🏻

Nope, you must have been lucky I guess 😄

src/components/Dialogs/ImageCropDialog/ImageCropDialog.tsx
```diff
@@ -97,7 +103,7 @@ const ImageCropDialogComponent: React.ForwardRefRenderFunction<
 
   const zoomControlNode = (
     <ZoomControl>
-      <IconButton variant="tertiary" onClick={() => handleZoomChange(currentZoom - zoomStep)}>
+      <IconButton variant="tertiary" onClick={() => handleZoomChange(currentZoom - zoomStep)} disabled={!!cropData}>
```
kdembler: Let's move that into some variable to avoid duplicated condition (`!!cropData`)

kdembler (commented):
If I try to reselect the file and select the correct image on the first try, everything seems to be working properly. However, if I first pick a different file and do a crop, I will get the warning that I selected the wrong file (as I should). Then if I click the `Reselect file` in the warning dialog, I don't have a chance to do a crop again and I keep seeing the dialog that I selected the wrong file

kdembler (approved):

Issue #609, opened 2021-05-13
uploads view not having enough bottom padding


kdembler:
if there are a lot of uploaded items and you have the video sheet minimized, you won't be able to fully see the last item:
<img width="1283" alt="Screenshot 2021-05-13 at 12 40 51" src="https://user-images.githubusercontent.com/12646744/118115010-a3fd4080-b3e8-11eb-9609-ec709468335a.png">

PR #637, opened 2021-05-19
add padding to uploads view, general responsive tweaks

mikkio-j:
Resolves: #609 and #595


drillprop (approved):
Checked on firefox and chrome - should fix the problem.

drillprop (approved):

kdembler (approved):

PR #638, opened 2021-05-19
Handle asset liaison being not active

### LINKED ISSUE #632: Handle asset liaison being not active
It's possible that storage provider that accepted a given asset upload is no longer active. We need to check their `isActive` and if it's false, we need to use a different storage provider, otherwise the asset download will surely fail.

drillprop:
Fix #632


src/hooks/useAsset.tsx
```diff
@@ -0,0 +1,27 @@
...
+    if (dataObject?.liaison?.isActive && dataObject?.liaison?.metadata) {
+      return createStorageNodeUrl(dataObject.joystreamContentId, dataObject?.liaison?.metadata)
+    }
+    if (!dataObject?.liaison?.isActive && randomStorageUrl) {
+      return createStorageNodeUrl(dataObject.joystreamContentId, randomStorageUrl)
+    }
+  }
+
+  return { getAssetUrl }
+}
```
drillprop, line 29: Not sure if it's enough for this task. It's kinda hard to test it because we have only one active worker with set metadata(URL).

kdembler: You can use the reckless deployment which has more providers for testing

src/hooks/useAsset.tsx
```diff
@@ -0,0 +1,27 @@
...
+    if (assetUrls?.length) {
+      return assetUrls[0]
+    }
+    if (!dataObject?.joystreamContentId) {
+      return
+    }
+    if (dataObject?.liaison?.isActive && dataObject?.liaison?.metadata) {
+      return createStorageNodeUrl(dataObject.joystreamContentId, dataObject?.liaison?.metadata)
+    }
+    if (!dataObject?.liaison?.isActive && randomStorageUrl) {
```
kdembler: I think we can just use `if (randomStorageUrl)` right? If for some reason the liaison is active but doesn't have metadata we'd want to fallback here as well

src/hooks/useAsset.tsx
```diff
@@ -0,0 +1,27 @@
...
+      return createStorageNodeUrl(dataObject.joystreamContentId, randomStorageUrl)
+    }
```
kdembler, line 25: I don't suppose we will ever hit this but let's add this 

```suggestion
    }
    console.error('No storage provider available to serve asset')
```

kdembler (approved):

Issue #591, opened 2021-05-10
Changing video tabs doesn't reset the file select step


kdembler:
As in the title, if you upload video for one draft and switch to a new draft, it will stay in the thumbnail select step but should go back to step 1 (select video)

PR #639, opened 2021-05-19
Reset the file select step component on tab change

eldiegod:
Issue: Changing video tabs doesn't reset the file select step #591

### DIFF
--- a/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
+++ b/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
@@ -319,6 +319,7 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
           control={control}
           render={({ value }) => (
             <MultiFileSelect
+              key={selectedVideoTab?.id}
               files={value}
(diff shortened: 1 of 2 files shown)


eldiegod:
Found a clean solution for this, I think this should do it

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -319,6 +319,7 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
           control={control}
           render={({ value }) => (
             <MultiFileSelect
+              key={selectedVideoTab?.id}
```
drillprop: Is this needed?

drillprop (approved):
Looks good to me! 👍🏻

mikkio-j (approved):

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
eldiegod: Actually no

kdembler (approved):
Really neat 👌

Issue #644, opened 2021-05-20
Improve asset upload notifications 


kdembler:
We should do the following:
1) Move notifications handling inside uploads manager, not make the consumer handle that
2) Group notifications from some time period. If I upload channel assets, it's better to show `2 assets uploaded` instead of two separate notifications.
3) Change the copy for starting uploads to `2 assets being uploaded`.
4) Let's also provide a button for 👆 to `See` and navigate to `My uploads`
5) Let's add some timeout (like 5s or something to those)

PR #653, opened 2021-05-21
Notifications improvements in upload manager

### LINKED ISSUE #645: Connection notifications tweaks
Let's:
1. Add some timeout to notifications about network connectivity (like 5s)
2. Change the copy: `Network connection restored`/`Network connection lost`

mikkio-j:
Issue: #644 

update: pushed changes that resolve #645 as well

### DIFF
--- a/src/components/NoConnectionIndicator/NoConnectionIndicator.tsx
+++ b/src/components/NoConnectionIndicator/NoConnectionIndicator.tsx
@@ -29,7 +29,7 @@ const NoConnectionIndicator: React.FC<NoConnectionIndicatorProps> = ({
         </IconWrapper>
         <TextWrapper>
           {!isConnectedToInternet ? (
-            <Text variant="subtitle2">No network connection</Text>
+            <Text variant="subtitle2">Network connection lost</Text>
           ) : (
             nodeConnectionStatus === 'disconnected' && <Text variant="body2">No node connection</Text>
           )}
--- a/src/components/NoConnectionIndicator/NoConnectionIndicator.tsx
+++ b/src/components/NoConnectionIndicator/NoConnectionIndicator.tsx
@@ -29,7 +29,7 @@ const NoConnectionIndicator: React.FC<NoConnectionIndicatorProps> = ({
         </IconWrapper>
         <TextWrapper>
           {!isConnectedToInternet ? (
-            <Text variant="subtitle2">Network connection lost</Text>
+            <Text variant="subtitle2">No network connection</Text>
           ) : (
             nodeConnectionStatus === 'disconnected' && <Text variant="body2">No node connection</Text>
           )}
--- a/src/hooks/useSnackbar/useSnackbar.tsx
+++ b/src/hooks/useSnackbar/useSnackbar.tsx
@@ -9,7 +9,6 @@ import { SvgAlertError, SvgAlertInfo, SvgAlertSuccess, SvgAlertWarning } from '@
 type SnackbarIconType = 'success' | 'error' | 'info' | 'warning'
 
 export type DisplaySnackbarArgs = {
-  customId?: string
   timeout?: number
   variant?: 'primary' | 'secondary'
   iconType?: SnackbarIconType
@@ -45,8 +44,8 @@ const SNACKBARS_LIMIT = 3
 export const SnackbarProvider: React.FC = ({ children }) => {
   const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
 
-  const displaySnackbar = useCallback(({ customId, timeout, ...args }: DisplaySnackbarArgs) => {
-    const id = customId || createId()
+  const displaySnackbar = useCallback(({ timeout, ...args }: DisplaySnackbarArgs) => {
+    const id = createId()
     setSnackbars((currentSnackbars) => {
       return [...currentSnackbars, { id, ...args }]
     })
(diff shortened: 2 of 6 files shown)


src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -55,9 +60,24 @@ export const SnackbarProvider: React.FC = ({ children }) => {
     return id
   }, [])
 
-  const closeSnackbar = (id: string) => {
-    setSnackbars(snackbars.filter((snackbar) => snackbar.id !== id))
-  }
+  const updateSnackbar = useCallback((id: string, opts: Omit<DisplaySnackbarArgs, 'id'>) => {
+    setSnackbars((currentSnackbars) => {
+      const newSnackbars = currentSnackbars.map((snackbar) => {
```
drillprop: you could make it much shorter. We especially don't need to create `newSnackbars` variable here.

drillprop: Maybe 
```tsx
  const updateSnackbar = useCallback((id: string, opts: Omit<DisplaySnackbarArgs, 'id'>) => {
    setSnackbars((currentSnackbars) =>
      currentSnackbars.map((snackbar) => (snackbar.id === id ? { ...snackbar, ...opts } : snackbar))
    )
  }, [])
``` 
?

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -55,9 +60,24 @@ export const SnackbarProvider: React.FC = ({ children }) => {
...
+      return newSnackbars
+    })
+  }, [])
+
+  const closeSnackbar = useCallback(
+    (id: string) => {
+      setSnackbars(snackbars.filter((snackbar) => snackbar.id !== id))
+    },
+    [snackbars]
+  )
```
drillprop, line 80:
```suggestion
  const closeSnackbar = useCallback((id: string) => {
    setSnackbars((snackbars) => snackbars.filter((snackbar) => snackbar.id !== id))
  }, [])
```

src/components/NoConnectionIndicator/NoConnectionIndicator.tsx
```diff
@@ -29,7 +29,7 @@ const NoConnectionIndicator: React.FC<NoConnectionIndicatorProps> = ({
         </IconWrapper>
         <TextWrapper>
           {!isConnectedToInternet ? (
-            <Text variant="subtitle2">No network connection</Text>
+            <Text variant="subtitle2">Network connection lost</Text>
```
kdembler: I think here we can leave the old copy. The difference is that snackbar tells you about something that happened, so "Network connection was lost". But this indicator is more permanent and it tells you about current status, so "No network connection"

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -55,9 +60,18 @@ export const SnackbarProvider: React.FC = ({ children }) => {
...
+    setSnackbars((currentSnackbars) =>
+      currentSnackbars.map((snackbar) => (snackbar.id === id ? { ...snackbar, ...opts } : snackbar))
+    )
+  }, [])
+
+  const closeSnackbar = useCallback(
+    (id: string) => {
+      setSnackbars(snackbars.filter((snackbar) => snackbar.id !== id))
+    },
+    [snackbars]
```
kdembler: Let's use value from setter here as well. So `setSnackbars((currentSnackbars) => ...`

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -161,6 +177,15 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
           { leading: true }
         )
 
+        displaySnackbar({
```
kdembler: I don't understand what's happening here. It seems we're adding a separate snackbar for each upload and then use `useEffect` to filter them back? It doesn't seem like a good pattern, `UploadsManager` shouldn't be managing the snackbars that closely. It will now also rerender every time snackbars are modified (not really caring what the reason for that was).

I think that we should try to just throttle this on uploads manager end and not manage the snackbars at all. We could follow this policy: "When new upload starts, wait up to 500ms to see if other uploads start and count all of them. After 500ms pass without added upload, display a snackbar with number of uploads". This would be analogous for when asset finished uploading

kdembler: Let me know if you need more info on that

kdembler (changes requested):

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -42,9 +45,11 @@ const SNACKBARS_LIMIT = 3
 export const SnackbarProvider: React.FC = ({ children }) => {
   const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
 
-  const displaySnackbar = useCallback(({ timeout, ...args }: DisplaySnackbarArgs) => {
-    const id = createId()
-    setSnackbars((s) => [...s, { id, ...args }])
+  const displaySnackbar = useCallback(({ customId, timeout, ...args }: DisplaySnackbarArgs) => {
+    const id = customId || createId()
```
drillprop: Do we actually need this `customId`? We never using it

drillprop (commented):
Found one thing to consider. Besides, that looks good to me!

src/hooks/useSnackbar/useSnackbar.tsx
kdembler: True, it's a left over from the previous implementation, will remove it

PR #661, opened 2021-05-24
Fix margin/padding issues with EditVideoForm

drillprop:
🙏🏻

### DIFF
--- a/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
+++ b/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
@@ -33,6 +33,7 @@ import {
   DeleteVideoContainer,
   DeleteVideoButton,
   FormWrapper,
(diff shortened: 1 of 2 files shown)


src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -519,7 +520,8 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
               )}
             />
           </FormField>
-          <FormField
+          <StyledLastFormField
```
mikkio-j: Not a big fan of that solution, but as you said in the chat - didn't find any better. Maybe you could add some comment explaining why `StyledLastFormField` exists? Also, moving that e.g. above will break that solution, right? Then we should include that in the comment

kdembler (approved):

mikkio-j (approved):

PR #665, opened 2021-05-25
Fix infinite fetch loop on video delete

### LINKED ISSUE #627: Infinite fetch loop on video delete

drillprop:
Fix #627

### DIFF
--- a/src/hooks/useDeleteVideo.tsx
+++ b/src/hooks/useDeleteVideo.tsx
@@ -1,11 +1,15 @@
 import { useState } from 'react'
 import { useJoystream, useAuthorizedUser, useTransactionManager } from '@/hooks'
+import { useApolloClient } from '@apollo/client'
+import { removeVideoFromCache } from '@/utils/cachingAssets'
 
 export const useDeleteVideo = () => {
   const { joystream } = useJoystream()
   const { handleTransaction } = useTransactionManager()
   const { activeMemberId } = useAuthorizedUser()
   const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false)
+
+  const client = useApolloClient()
 
   const confirmDeleteVideo = async (videoId: string, onTxSync?: () => void) => {
     if (!joystream) {
@@ -16,7 +20,10 @@ export const useDeleteVideo = () => {
 
     handleTransaction({
       txFactory: (updateStatus) => joystream.deleteVideo(videoId, activeMemberId, updateStatus),
-      onTxSync,
+      onTxSync: async () => {
+        removeVideoFromCache(videoId, client)
+        onTxSync?.()
+      },
       successMessage: {
         title: 'Video successfully deleted!',
         description: 'Your video was marked as deleted and it will no longer show up on Joystream.',
--- a/src/hooks/useDeleteVideo.tsx
+++ b/src/hooks/useDeleteVideo.tsx
@@ -9,7 +9,7 @@ export const useDeleteVideo = () => {
   const { activeMemberId, activeChannelId } = useAuthorizedUser()
   const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false)
 
-  const { refetchCount: refetchVideosCount, client } = useVideos({
+  const { client } = useVideos({
     where: {
       channelId_eq: activeChannelId,
     },
@@ -25,7 +25,6 @@ export const useDeleteVideo = () => {
     handleTransaction({
       txFactory: (updateStatus) => joystream.deleteVideo(videoId, activeMemberId, updateStatus),
       onTxSync: async () => {
-        await refetchVideosCount()
         removeVideoFromCache(videoId, client)
         onTxSync?.()
       },
--- a/src/hooks/useDeleteVideo.tsx
+++ b/src/hooks/useDeleteVideo.tsx
@@ -1,19 +1,11 @@
-import { useVideos } from '@/api/hooks'
 import { useState } from 'react'
 import { useJoystream, useAuthorizedUser, useTransactionManager } from '@/hooks'
-import { removeVideoFromCache } from '@/utils/cachingAssets'
 
 export const useDeleteVideo = () => {
   const { joystream } = useJoystream()
   const { handleTransaction } = useTransactionManager()
-  const { activeMemberId, activeChannelId } = useAuthorizedUser()
+  const { activeMemberId } = useAuthorizedUser()
   const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false)
-
-  const { client } = useVideos({
-    where: {
-      channelId_eq: activeChannelId,
-    },
-  })
 
   const confirmDeleteVideo = async (videoId: string, onTxSync?: () => void) => {
     if (!joystream) {
@@ -24,10 +16,7 @@ export const useDeleteVideo = () => {
 
     handleTransaction({
       txFactory: (updateStatus) => joystream.deleteVideo(videoId, activeMemberId, updateStatus),
-      onTxSync: async () => {
-        removeVideoFromCache(videoId, client)
-        onTxSync?.()
-      },
+      onTxSync,
       successMessage: {
         title: 'Video successfully deleted!',
         description: 'Your video was marked as deleted and it will no longer show up on Joystream.',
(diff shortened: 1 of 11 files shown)


src/hooks/useDeleteVideo.tsx
```diff
@@ -24,11 +16,7 @@ export const useDeleteVideo = () => {
 
     handleTransaction({
       txFactory: (updateStatus) => joystream.deleteVideo(videoId, activeMemberId, updateStatus),
-      onTxSync: async () => {
-        await refetchVideosCount()
-        removeVideoFromCache(videoId, client)
```
kdembler, line 29: Why not keep that here?

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -203,18 +201,19 @@ export const EditVideoSheet: React.FC = () => {
...
           writeVideoDataInCache({
-            data: fetchedVideo.data.videoByUniqueInput,
+            edge: fetchedVideo.data.videosConnection.edges[0],
             thumbnailUrl: data.assets.thumbnail?.url,
             client,
           })
         }
         // update videos count only after inserting video in cache to not trigger refetch in "my videos" on missing video
-        await refetchVideosCount()
+        // !! await refetchVideosCount()
```
kdembler: for removal

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -203,18 +201,19 @@ export const EditVideoSheet: React.FC = () => {
     }
 
     const refetchDataAndCacheAssets = async (videoId: VideoId) => {
-      const fetchedVideo = await refetchVideo({ where: { id: videoId } })
-
+      const fetchedVideo = await refetchVideoWithCursor({ where: { id_in: [videoId] } })
+      // const fetchVideoWithCursor = await
+      console.log(data.assets.thumbnail?.url)
```
kdembler: for removal

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -203,18 +201,19 @@ export const EditVideoSheet: React.FC = () => {
     }
 
     const refetchDataAndCacheAssets = async (videoId: VideoId) => {
-      const fetchedVideo = await refetchVideo({ where: { id: videoId } })
-
+      const fetchedVideo = await refetchVideoWithCursor({ where: { id_in: [videoId] } })
```
kdembler: I think better pattern for when we need to fetch something conditionally and only once is to do something like we currently have in  `CreateMemberView`:
```ts
await client.query<GetMembershipQuery, GetMembershipQueryVariables>({
  query: GetMembershipDocument,
  variables: { where: { handle: value } },
})
```

src/utils/cachingAssets.ts
```diff
@@ -65,21 +68,28 @@ export const writeUrlInCache = ({ url, fileType, parentId, client }: WriteUrlInC
...
+          ...existingVideos,
+          pageInfo: {
+            ...existingVideos.pageInfo,
+          },
+          totalCount: existingVideos.totalCount + 1,
+          edges: [{ ...edge, node: video }, ...(existingVideos?.edges ? existingVideos.edges : [])],
+        }
       },
     },
   })
```
kdembler, line 92: Below here, we also need to make change to `removeVideoFromCache` to modify `videosConnection` query as well

src/utils/cachingAssets.ts
```diff
@@ -65,33 +68,51 @@ export const writeUrlInCache = ({ url, fileType, parentId, client }: WriteUrlInC
...
+
+export const removeVideoFromCache = (videoId: string, client: ApolloClient<object>) => {
   client.cache.modify({
     fields: {
       videos: (existingVideos = []) => {
         client.cache.evict({ id: `Video:${videoId}` })
         return existingVideos.filter((video: VideoFieldsFragment) => video.id !== videoId)
       },
+      videosConnection: (existing = {}, opts) => {
+        client.cache.evict({ id: `Video:${videoId}` })
```
kdembler: Let's move evict out of `cache.modify`

src/utils/cachingAssets.ts
```diff
@@ -65,33 +68,51 @@ export const writeUrlInCache = ({ url, fileType, parentId, client }: WriteUrlInC
...
+        return (
+          existing && {
+            ...existing,
+            totalCount: existing.totalCount - 1,
```
kdembler: This will result in decrementing `totalCount` regardless of whether the given video was actually included in that query cache, right? E.g. I delete public video but `totalCount` is decremented for queries targeting only unlisted videos as well. Correct?

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -203,18 +198,24 @@ export const EditVideoSheet: React.FC = () => {
...
+      const fetchedVideo = await client.query<GetVideosConnectionQuery, GetVideosConnectionQueryVariables>({
+        query: GetVideosConnectionDocument,
+        variables: {
+          where: {
+            channelId_eq: activeChannelId,
+            isPublic_eq: data.isPublic,
+            id_eq: videoId,
+          },
+        },
+      })
```
kdembler: Why do we use a connection here instead of a regular `videoByUniqueInput` query? We care only about one video details, right?

drillprop: I think it works and I'll change that, but what about the `cursor` field? I cannot add a video cursor to the cache when I do the `videoByUniqueInput` query. Do we need that in this case?

src/utils/cachingAssets.ts
drillprop: Yes, you're right - should be fixed now.

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
drillprop: Noticed one more thing - when I'm using `videoByUniqueInput` after adding video it refetches the whole `videosConnection` list. When I'm using `videosConnection` it doesn't. To be honest, I'm not sure why. 

Also,  I need to do something with `totalCount: existing.totalCount + 1`, because analogous to `removeVideoFromCache`, `totalCount` is incremented in unlisted, public and all videos query.

kdembler: > Noticed one more thing - when I'm using videoByUniqueInput after adding video it refetches the whole videosConnection list.

Hmm that's weird.

Maybe let's use `videosConnection` after all, it will give us a cursor as well. However, if we want to get the same cursor as other queries, we will need to add `orderBy` here, it changes the cursor.

kdembler: Regarding totalCount I'm not sure... it's messy. I think once we get a normalized cache, we can just return `totalCount` locally instead

kdembler: Does it break things a lot right now or can we ignore it for the time being?

drillprop: >Maybe let's use videosConnection after all, it will give us a cursor as well. However, if we want to get the same cursor as other queries, we will need to add orderBy here, it changes the cursor.

So, I'll revert that.

> Does it break things a lot right now or can we ignore it for the time being?

Not really. I think we can ignore it for now.

src/views/studio/EditVideoSheet/EditVideoSheet.tsx
```diff
@@ -203,18 +198,24 @@ export const EditVideoSheet: React.FC = () => {
...
-
+      const fetchedVideo = await client.query<GetVideosConnectionQuery, GetVideosConnectionQueryVariables>({
+        query: GetVideosConnectionDocument,
+        variables: {
+          where: {
+            channelId_eq: activeChannelId,
+            isPublic_eq: data.isPublic,
+            id_eq: videoId,
+          },
+        },
```
kdembler: We will need `orderBy` here to make sure we get the same cursor. Also, do we need `channelId_eq` and `isPublic_eq` here?

kdembler (approved):

PR #675, opened 2021-05-26
Fix sign in views(RWD)

drillprop:


kdembler (commented):
For the step with instructions, I think it still could use some improvement on 1080p:
Screenshot 2021-05-27 at 10 42 02

The button is still not visible without any scroll (remember the browser will take some vertical space for URL bar, etc, so the height is actually lower than 1080px)

We probably could just lose some margin on top

---

Same for create member view:
Screenshot 2021-05-27 at 10 49 00

The margin on top if too excessive

kdembler (approved):

PR #681, opened 2021-05-27
fix cache policy for unique video and channel

kdembler:


src/api/client/cache.ts
```diff
@@ -69,63 +76,72 @@ const createCachedAvailabilityHandler = () => ({
   },
 })
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type CachePolicyFields<T extends string> = Partial<Record<T, FieldPolicy<any> | FieldReadFunction<any>>>
```
drillprop:
```suggestion
type CachePolicyFields<T extends string> = Partial<Record<T, FieldPolicy | FieldReadFunction>>
```
I think it should work, right? 
I looked in the `FieldPolicy` and `FieldReadFunction` and `any` is set as default there.

drillprop (approved):
Just one thing.
Besides that looks good

PR #698, opened 2021-05-31
Global state management

eldiegod:
Docs
https://mobx-state-tree.js.org/intro/welcome
https://mobx.js.org/README.html


src/models/UploadsManagerStore.ts
```diff
@@ -0,0 +1,202 @@
...
+  })
+  .actions((self) => {
+    function addAsset(asset: SnapshotOrInstance<typeof AssetUpload>) {
+      self.uploadingAssetsState.push(asset)
+    }
+    function updateAsset(asset: SnapshotOrInstance<typeof AssetUpload>) {
+      const index = self.uploadingAssetsState.findIndex((a) => a.contentId === asset.contentId)
+      if (index >= 0) {
+        self.uploadingAssetsState[index] = { ...self.uploadingAssetsState[index], ...cast(asset) }
+      }
```
kdembler, line 119: We do it this way because we want to mutate data now, right? So we don't want to override the full array like we would previously but just want to modify one entry inside it?

src/models/UploadsManagerStore.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+    function removeAsset(asset: SnapshotOrInstance<typeof AssetUpload>) {
+      self.uploadingAssetsState.remove(cast(asset))
```
kdembler, line 122: How does this work? `remove` is some special function for removing an element from an array?

kdembler (commented):
Disclaimer: I don't have experience with MobX or MST so it's possible some of the following come from my lack of understanding. Please indulge me for now, I will do my research soon 😅 

**What I like:**
- How you can just use `self.` from actions and don't care about setters on `useState` etc, pretty cool
- That you can put `debounced` functions into `volatile` and you don't need any `useRef` magic for it to work properly
- Automatic persistence without any additional work

**What I don't like:**
- I see that currently we build `mst` models and then extract TS interface from that. Is it possible to do it the other way around? We already have everything in Typescript, I'd really like to avoid rewriting everything to model syntax. If in the future we decide to change the solution we use, I wouldn't want to rewrite everything back to Typescript
- The pattern with storing hooks functions in the store seem like a _very_ bad idea to me, I think there's a lot of ways this can go wrong really fast. Or maybe that's just some workaround for the time being only?
- It seems to me like there's a bit too much logic moved into the store? I understand that the upload logic depends on store heavily so maybe it does make sense. However, I'm not sure if the store should be responsible for displaying snackbars? That seems more like a responsibility for the uploads manager? Dunno, maybe I just need to wrap my head around this
- I don't think components should use the store action (or maybe store in general) directly. For example, I think `startFileUpload` should still come from `useUploadsManager`. That hook can extract it from the store and pass without changes if needed. But I'd like the store to be abstracted in some way.

**What I don't understand:**
- Why are some components wrapped in `observer` and others not?
- On what basis does specific component gets rerendered? For example, what stops components using only an action from rerendering when state changes? Does mst check which fields from `uploadsManagerStore` get accessed in a component and use that to determine whether it needs to rerender?
- I see that persistence is done via IndexedDB instead of local storage. I've heard a lot about it but haven't used it before myself. Is there anything we should be aware of when it comes to differences between those?

eldiegod:
@kdembler let me go over all the points over here:

> I see that currently we build mst models and then extract TS interface from that. Is it possible to do it the other way around? We already have everything in Typescript, I'd really like to avoid rewriting everything to model syntax. If in the future we decide to change the solution we use, I wouldn't want to rewrite everything back to Typescript

Yes that's right, it is not possible to do it the other way around because it's type system offers features TS does not like identifiers, validation or late types and as it is JS it does checks in runtime as well.

On the other hand not everything has to be converted to, it has a util to scaffold the models of any graphql document so that a big part of our types done, also the typescript enum to MST model is also trivial, but yeah some manual work would have to be done tho it's fairly similar

> The pattern with storing hooks functions in the store seem like a very bad idea to me, I think there's a lot of ways this can go wrong really fast. Or maybe that's just some workaround for the time being only?

Yep, it's not great, I put it together as a workaround to have the PR stay focused on the UploadsManager as possible, it's not something that's gonna be needed later

> It seems to me like there's a bit too much logic moved into the store? I understand that the upload logic depends on store heavily so maybe it does make sense. However, I'm not sure if the store should be responsible for displaying snackbars? That seems more like a responsibility for the uploads manager? Dunno, maybe I just need to wrap my head around this

Maybe this is not clear yet but the `store` its the `rootStore` and directly inside it many other stores will live that will take care of different things, as of now only the hooks store (provisional) and the `uploadsManagerStore` exist so in that case there isn't much logic added to the store because that store **is** the uploadsManager.

but you are right tho it shouldn't display the snackbars, there should be a store caller notificationsStore or something on the line that one can access from the other stores to trigger all kind of notifications for the user and later on a client, in this case, a react component, but could be react native as well, can consume it and show the appropriate snackbar

So in a sense most logic would come from the centralized Store(s) and the idea would be for react components to be as "dumb" as they can

> I don't think components should use the store action (or maybe store in general) directly. For example, I think startFileUpload should still come from useUploadsManager. That hook can extract it from the store and pass without changes if needed. But I'd like the store to be abstracted in some way.

The store is already abstracted, it only exposes the actions and views, the volatile state is always private, and the plan is for it to have exactly the same public API as the useUploadsManager.

The actions are exactly for that so it can be consumed by the client since the state of the store can be consumed but it's protected and won't allow direct changes to it that don't go trough an action.

> Why are some components wrapped in observer and others not?

The observer wrapper should be used in any component that consumes the store, it converts a component into a reactive component, which tracks which observables are used automatically and re-renders the component when one of these values changes.
It is cheap to use so if in doubt add the observer wrapper.

> On what basis does specific component gets rerendered? For example, what stops components using only an action from rerendering when state changes? Does mst check which fields from uploadsManagerStore get accessed in a component and use that to determine whether it needs to rerender?

All changes to and uses of the data are tracked at runtime, building a dependency tree that captures all relations between state and output. This guarantees that computations depending on your state, like React components, run only when strictly needed. There is no need to manually optimize components with error-prone and sub-optimal techniques like memoization and selectors. 

> I see that persistence is done via IndexedDB instead of local storage. I've heard a lot about it but haven't used it before myself. Is there anything we should be aware of when it comes to differences between those?

indexedDB was designed to work with significantly larger amounts of data and supports asynchronous API out of the box. + it supports simple types (e.g. Date) and has the usual database features if needed. 

Although the abstraction used works almost in the exact same way as local storage except for the asynchronous part that's really nice to have and defaults to localSotrage if indexedDB is not supported

src/models/UploadsManagerStore.ts
eldiegod: Yes, modify the entry in place and merge the new data to the object that is already there, no need to worry about immutability inside the store

kdembler:
> that store is the uploadsManager

Yeah I guess that's where my confusion came from. Like I said, probably just need to wrap my head around it. One thing I'm curious about, we still have the `UploadManagerProvider`, it runs some transforms on data, but I guess that could be done in the store as well as a `view`? However, we don't have access to data from graphql in the store and I guess there isn't any way to call `useVideos` from within it?

---

> The store is already abstracted, it only exposes the actions and views, the volatile state is always private, and the plan is for it to have exactly the same public API as the useUploadsManager.
> The actions are exactly for that so it can be consumed by the client since the state of the store can be consumed but it's protected and won't allow direct changes to it that don't go trough an action.

I partly agree but I guess this also touches upon some of the other points. For example, if we want to start the upload and display a notification about it I guess we could do that in 2 ways:
1. Just use `uploadsManagerStore.startFileUpload` directly and that action itself triggers a notification
2. Have a `useUploadsManager` that looks like this for example:
```ts
const useUploadsManager = () => {
  const { uploadsManagerStore, notificationsStore } = useMST()
  const startFileUpload = (...params) => {
    uploadsManagerStore.startFileUpload(...params)
    notificationsStore.displaySnackbar('starting upload')
  }
  return { ...uploadsManagerStore, startFileUpload }
}
```

Are you able to point out some drawbacks/upsides of one of those? I'd expect 2nd approach gives a bit more control and doesn't create a coupling inside a store but then again I'm not sure if it's that bad.

Regarding abstracting it away, I meant that it'd be nice if components didn't had to know they are using `MST` under the hood. If we use a simple hook like I used above, we can always add some additional logic to action or even swap out the entire state library without changing any of the components. Basically a wrapper, like we do for hooks generated by `graphql-code-generator`.

---

Other than that I think it looks fine, I'm very excited to get this integrated, will surely be a huge improvement

src/models/UploadsManagerStore.ts
eldiegod: Since this is an observable array it has some extra utils built on top remove is one of them and it removes the element from the array
https://mobx.js.org/api.html#collection-utilities-

drillprop:
Just took a look at the code, it will require diving more into documentation. Have one quick question though, are there available dev tools for mobx/mst? Do they work well?

eldiegod:
@kdembler 
> However, we don't have access to data from graphql in the store and I guess there isn't any way to call `useVideos` from within it?

We don't have access to any data from graphql yet, that's why the UploadManagerProvider is still there to fetch that rn; one of the best options to solve that is using https://github.com/mobxjs/mst-gql that is package to do just that and glue it all together. But yes since the stores are not react components you can't use hooks from the inside otherwise that would be in a view no problem

>I partly agree but I guess this also touches upon some of the other points. For example, if we want to start the upload and display a notification about it I guess we could do that in 2 ways:

I see option one is the clear winner here unless there is a major cause that can't be done in the store, I don't see too much use in creating a separate hook to handle that part of the state outside the store, it would actually be detrimental because the store wouldn't have access to it and again what we want to avoid state to be spread around instead we want to as possible keep it centralized in the stores as between stores they can communicate really well and reliably as the system is built with that in mind

>Regarding abstracting it away, I meant that it'd be nice if components didn't had to know they are using MST under the hood. If we use a simple hook like I used above, we can always add some additional logic to action or even swap out the entire state library without changing any of the components. Basically a wrapper, like we do for hooks generated by graphql-code-generator.

For graphql-code-generator def makes sense because we need to build stuff on top off the basic hooks but for MST we wouldn't do that because it would on one hand be more boilerplate of hooks that's not needed and on the other hand would spread the logic again which would cause inconsistencies for when something is called from inside the stores and when called from the hooks causing different behaviours. There is no limit to what actions we can add to the store or how much customization to add so shouldn't be a problem to handle different use cases 

>Other than that I think it looks fine, I'm very excited to get this integrated, will surely be a huge improvement

Yep me too, the uploadManagers doesn't even show the full potential of it yet but it will come really handy as we scale

eldiegod:
@drillprop 
> Just took a look at the code, it will require diving more into documentation. Have one quick question though, are there available dev tools for mobx/mst? Do they work well?

Def needs some getting use to but docs are really good! feel free to ask anything you don't get.

And yeah it has several dev tools that can be used, it has a logger that logs every action, patch and snapshot to the console with details.
There is a store inspector as well that lets you look at any changes and explore the store
And I also like to in development do window.store = store so I expose the store to the window obj and from there I can do anything from the console like see values or do actions, which is awesome for debugging
Also the redux-devtools work with it as well
Whenever there is a runtime error its also pretty detailed on what went wrong

drillprop:
Had some questions, but was already answered in the discussion above. Other small ambiguities I confronted with documentation. I guess more questions will come when I start to work with it.  Really like how models work - it kinda reminds me `mongoose`. Looking forward to trying it out in action!

src/models/SnackbarStore.ts
```diff
@@ -0,0 +1,70 @@
...
+      self = { ...self, ...snackbar }
+    },
+  }))
+
+export const SnackbarStore = types
+  .model('SnackbarStore', {
+    snackbars: types.array(Snackbar),
+  })
+  .actions((self) => ({
+    afterAttach() {
```
drillprop, line 38: This part is a little confusing for me. Why do we need to do this? And why we use `autorun` and pass empty function below?

src/models/SnackbarStore.ts
```diff
@@ -0,0 +1,70 @@
...
+    description: types.maybe(types.string),
+    timeout: types.maybe(types.number),
+    variant: types.maybe(types.enumeration(['primary', 'secondary'])),
+    iconType: types.maybe(types.enumeration(['success', 'error', 'info', 'warning'])),
+    actionText: types.maybe(types.string),
+  })
+  .volatile((self) => ({
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    onActionClick() {},
+  }))
```
drillprop, line 22: What will happen if we don't add this?

src/models/SnackbarStore.ts
eldiegod: So, afterAttach is a lifecycle hook that's called when the model is attached to the tree and autorun is like an useEffect that handles dependencies automatically so it runs once and whenever something inside changes it will rerun again

Oh, removed the empty autorun not sure how it ended up in there.

src/models/SnackbarStore.ts
eldiegod: that adds some temp state to the Model and makes it aware of it so later on so we can set a custom function from outside with the `setOnActionClick` action

So we are just setting it in there to change it later

kdembler:
Closing in favour of #742 Thanks for your work on this!

Issue #697, opened 2021-05-31
Fix asset status on refresh in upload manager


mikkio-j:
Two bugs occur currently in upload manager:

1. If you start upload and quickly refresh, "Reconnect file" button will show up properly. But after another refresh it will disappear. Asset's status `inProgress` persists after refresh - should have `error` instead.
2. If the storage provider is not available upload manager is trying to reconnect, showing the message "Trying to reconnect...". If we refresh page during reconnection, `reconnecting` status will persists making reselecting file impossible. Again, asset should have `error` status at this point

PR #699, opened 2021-06-01
fix asset status on refresh in upload manager

mikkio-j:
Issue: #697


kdembler (commented):
I think this changes the behaviour we previously had a bit. If your upload didn't finish and you refreshed the app, you'd get a snackbar telling you that files are waiting to reconnect. With those changes I don't see that snackbar anymore

kdembler (approved):

PR #704, opened 2021-06-02
SVG maintenance

### LINKED ISSUE #583: Make sure that all the SVG in assets folder using unique ids
We encountered some issues with SVGs, which are related to the fact that they are using the same ids. For example #561 
To avoid this in the future we could do the same preprocessing on illustrations as we do for icons.

drillprop:
Fix #583


patches/svgo+1.3.2.patch
```diff
@@ -0,0 +1,12 @@
...
+--- a/node_modules/svgo/plugins/_collections.js
++++ b/node_modules/svgo/plugins/_collections.js
+@@ -1519,6 +1519,7 @@ exports.elems = {
+             'x',
+             'y',
+             'width',
++            'mask-type',
+             'height',
+             'maskUnits',
+             'maskContentUnits'
```
drillprop, line 12: Had to add this. SVGR using the old version of the SVGO package, which is removing `mask-type` from attributes https://github.com/gregberge/svgr/issues/336. Some of our SVGs have this attribute(e.g. transaction illustration, avatar silhouette) and it's needed to properly display them

svgr.config.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
   typescript: true,
   svgo: true,
   prettier: true,
+  svgoConfig: {
+    plugins: [{ removeViewBox: false }],
+  },
```
drillprop, line 12: Decided to add this. Some illustrations, for example `empty-state-illustration.svg` was not displayed correctly on smaller devices. I think we should apply this plugin to all of our SVGs. You can read why [here](https://github.com/svg/svgo/pull/1461)

package.json
```diff
@@ -29,7 +29,9 @@
     "mocking:channels": "node scripts/mocking/generateChannels.js",
     "codegen:graphql": "graphql-codegen --config codegen.config.yml",
     "codegen:graphql-watch": "yarn codegen:graphql --watch",
-    "codegen:icons": "svgr --config-file svgr-icons.config.js -d src/shared/icons src/shared/icons/svgs",
+    "codegen:icons": "svgr --config-file svgr.config.js -d src/shared/icons src/shared/icons/svgs",
+    "codegen:illustrations": "svgr --config-file svgr.config.js -d src/shared/illustrations src/shared/illustrations/svgs",
+    "codegen:svg-r": "yarn codegen:illustrations && yarn codegen:icons",
```
kdembler:
```suggestion
    "codegen:svgs": "yarn codegen:illustrations && yarn codegen:icons",
```

mikkio-j (approved):
Checked those svgs and everything seems to be looking good. I don't think I missed something. Nothing to add when it comes to code or config for svgr

kdembler (approved):

Issue #684, opened 2021-05-27
Display a message when user rejects Polkadot access


kdembler:
We should display some message, best would be a dialog, maybe snackbar, with information that user rejected Polkadot extension access. We should also include steps to re-enable access

We could also include some link to open that same dialog on the `Install extension` step, so users can see the instructions again

PR #708, opened 2021-06-04
display a message when user rejects Polkadot access

mikkio-j:
Issue: #684

### DIFF
--- a/src/components/SignInSteps/ExtensionStep.style.ts
+++ b/src/components/SignInSteps/ExtensionStep.style.ts
@@ -1,8 +1,21 @@
 import { Button } from '@/shared/components'
-import { sizes } from '@/shared/theme'
+import { sizes, colors, typography } from '@/shared/theme'
 import styled from '@emotion/styled'
+import { StepFooter } from './SignInSteps.style'
 
 export const StyledButton = styled(Button)`
-  margin-top: ${sizes(4)};
-  margin-bottom: ${sizes(12)};
+  margin: ${sizes(4)} 0;
+`
+
+export const StyledStepFooter = styled(StepFooter)`
+  margin-top: ${sizes(12)};
+`
+
+export const StyledListItem = styled.li`
+  text-align: left;
+  color: ${colors.gray[200]};
+  font-size: ${typography.sizes.caption};
+  & + & {
+    margin-top: ${sizes(2)};
+  }
 `
--- a/src/components/SignInSteps/ExtensionStep.style.ts
+++ b/src/components/SignInSteps/ExtensionStep.style.ts
@@ -19,5 +19,8 @@ export const StyledListItem = styled(Text)`
 `
 
 export const PolkadotExtensionRejectedWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
   text-align: center;
 `
--- a/src/components/SignInSteps/ExtensionStep.style.ts
+++ b/src/components/SignInSteps/ExtensionStep.style.ts
@@ -1,5 +1,5 @@
-import { Button } from '@/shared/components'
-import { sizes, colors, typography } from '@/shared/theme'
+import { Button, Text } from '@/shared/components'
+import { sizes } from '@/shared/theme'
 import styled from '@emotion/styled'
 import { StepFooter } from './SignInSteps.style'
 
@@ -11,11 +11,13 @@ export const StyledStepFooter = styled(StepFooter)`
   margin-top: ${sizes(12)};
 `
 
-export const StyledListItem = styled.li`
+export const StyledListItem = styled(Text)`
   text-align: left;
-  color: ${colors.gray[200]};
-  font-size: ${typography.sizes.caption};
   & + & {
     margin-top: ${sizes(2)};
   }
+`
+
+export const PolkadotExtensionRejectedWrapper = styled.div`
+  text-align: center;
 `
--- a/src/components/SignInSteps/ExtensionStep.tsx
+++ b/src/components/SignInSteps/ExtensionStep.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 import { StyledButton, StyledStepFooter, StyledListItem } from './ExtensionStep.style'
 import { BottomBarIcon, StepSubTitle, StepTitle, StepWrapper, StyledPolkadotLogo } from './SignInSteps.style'
 import { Text, Button } from '@/shared/components'
@@ -15,11 +15,49 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
   const step = useRouterQuery('step')
   const { extensionConnected, extensionRejected } = useUser()
 
+  const [showDisabledExtensionMessage, setShowDisabledExtensionMessage] = useState(false)
+
   useEffect(() => {
     if (extensionConnected && step === '1') {
       navigate(nextStepPath)
     }
   }, [extensionConnected, navigate, nextStepPath, step])
+
+  const PolkadotExtensionNotInstalled: React.FC = () => (
+    <>
+      <StepTitle variant="h4">Add Polkadot extension</StepTitle>
+      <StepSubTitle secondary variant="body2">
+        To manage your blockchain account, you will need a Polkadot browser extension. Please install it using the
+        following link:
+      </StepSubTitle>
+      <StyledButton icon={<SvgGlyphExternal />} to="https://polkadot.js.org/extension/">
+        Install extension
+      </StyledButton>
+      <Button variant="tertiary" size="small" onClick={() => setShowDisabledExtensionMessage(true)}>
+        Polkadot extension already installed? Click here
+      </Button>
+    </>
+  )
+
+  const PolkadotExtensionRejected: React.FC = () => (
+    <>
+      <StepTitle variant="h4">Enable Polkadot extension website access</StepTitle>
+      <StepSubTitle secondary variant="body2">
+        It seems like you have disabled Polkadot extension access to the app. Please follow the steps below to enable
+        it:
+      </StepSubTitle>
+      <ol>
+        <StyledListItem>Open Polkadot extension menu in the right upper corner of the screen</StyledListItem>
+        <StyledListItem>In the menu open settings and Manage Website Access</StyledListItem>
+        <StyledListItem>Find play.joystream.org address and switch it to allowed</StyledListItem>
+      </ol>
+      {showDisabledExtensionMessage && (
+        <Button variant="tertiary" size="small" onClick={() => setShowDisabledExtensionMessage(false)}>
+          Go back
+        </Button>
+      )}
+    </>
+  )
 
   return (
     <StepWrapper withBottomBar>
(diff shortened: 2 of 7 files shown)


src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -33,14 +35,47 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
...
+      )}
+      {showDisabledExtensionMessage && (
+        <Button variant="tertiary" size="small" onClick={() => setShowDisabledExtensionMessage(false)}>
+          Go back
+        </Button>
+      )}
+      <StyledStepFooter>
         <BottomBarIcon />
         <Text variant="body2" secondary>
           Please reload the page and allow access after installing the extension
```
drillprop, line 45: I think this text should be also updated, the user already has the extension, right?

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -33,14 +35,47 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
...
       </StyledButton>
-      <StepFooter>
+      <Button variant="tertiary" size="small" onClick={() => setShowDisabledExtensionMessage(true)}>
+        Polkadot extension already installed? Click here
+      </Button>
+    </>
+  )
+
+  return (
+    <StepWrapper withBottomBar>
```
drillprop: not your change, but I think I forgot to remove it. We can get rid of `withBottomBar` and `StepWrapperProps` in `SignInSteps.style.ts`

src/components/SignInSteps/ExtensionStep.style.ts
```diff
@@ -1,8 +1,21 @@
...
-  margin-top: ${sizes(4)};
-  margin-bottom: ${sizes(12)};
+  margin: ${sizes(4)} 0;
+`
+
+export const StyledStepFooter = styled(StepFooter)`
+  margin-top: ${sizes(12)};
+`
+
+export const StyledListItem = styled.li`
```
drillprop: Not sure, but I would use `styled(Text)` here and then in the component 
```js
    <ol>
      <StyledListItem secondary as="li" variant="caption">
        Open Polkadot extension menu in the right upper corner of the screen
      </StyledListItem>
      <StyledListItem secondary as="li" variant="caption">
        In the menu open settings and Manage Website Access
      </StyledListItem>
      <StyledListItem secondary as="li" variant="caption">
        Find play.joystream.org address and switch it to allowed
      </StyledListItem>
    </ol>
```

src/components/SignInSteps/ExtensionStep.style.ts
```diff
@@ -1,8 +1,21 @@
...
+  text-align: left;
+  color: ${colors.gray[200]};
+  font-size: ${typography.sizes.caption};
+  & + & {
```
drillprop, line 16: `line-height` will also work well here, right?

src/hooks/useUser/useUser.tsx
```diff
@@ -1,10 +1,15 @@
...
 import { web3Accounts, web3AccountsSubscribe, web3Enable } from '@polkadot/extension-dapp'
 import { AccountId } from '@/joystream-lib'
 import { WEB3_APP_NAME } from '@/config/urls'
 import { useMembership, useMemberships } from '@/api/hooks'
+import { useCheckBrowser, useDialog } from '@/hooks'
 import { InjectedAccountWithMeta } from '@polkadot/extension-inject/types'
+import { PolkadotExtensionRejected } from '@/components/SignInSteps/ExtensionStep'
+
+const POLKADOT_EXTENSION_ID = 'mopnmbcafieddcagagdcbnhejhlodfdd'
+const EXTENSION_URL = `chrome-extension://${POLKADOT_EXTENSION_ID}/page.js`
```
drillprop: Let's move it to the `@/config` maybe?

src/hooks/useUser/useUser.tsx
```diff
@@ -61,6 +70,25 @@ export const ActiveUserProvider: React.FC = ({ children }) => {
...
   }
 
+  const checkIfPolkadotExtensionInstalled = useCallback(async () => {
+    const res = await fetch(EXTENSION_URL)
+    const polkaDotExtensionInstalled = res.ok
+    if (polkaDotExtensionInstalled) {
+      console.warn('Polkadot extension disabled')
+      openDialog({
+        additionalActionsNode: (
+          <div style={{ textAlign: 'center' }}>
```
drillprop: We should add this in `PolkadotExtensionRejected`.

drillprop: and we shouldn't use inline styles - should be extracted to `.style.ts`

mikkio-j: right, wanted to check how it looks like really fast and missed it before commit

src/components/SignInSteps/ExtensionStep.style.ts
mikkio-j: could work, but not with the caption value (I will stick with the margin to not overwrite caption line-height)

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -1,29 +1,31 @@
...
   useEffect(() => {
-    if (extensionConnected && step === '1') {
+    if (extensionConnected && nextStepPath && step === '1') {
       navigate(nextStepPath)
     }
   }, [extensionConnected, navigate, nextStepPath, step])
 
-  return (
-    <StepWrapper withBottomBar>
+  const PolkadotExtensionNotInstalled: React.FC = () => (
```
kdembler: We shouldn't define a component inside another component

src/hooks/useUser/useUser.tsx
```diff
@@ -61,6 +67,21 @@ export const ActiveUserProvider: React.FC = ({ children }) => {
     throw activeMembershipError
   }
 
+  const checkIfPolkadotExtensionInstalled = useCallback(async () => {
+    const res = await fetch(EXTENSION_URL)
```
kdembler: This will throw if extension isn't found, exception needs to be handled

src/hooks/useUser/useUser.tsx
```diff
@@ -30,9 +33,12 @@ ActiveUserContext.displayName = 'ActiveUserContext'
 
 export const ActiveUserProvider: React.FC = ({ children }) => {
   const { activeUserState, setActiveUser, resetActiveUser } = useActiveUserStore()
+  const [openDialog, closeDialog] = useDialog()
+  const browser = useCheckBrowser()
```
kdembler: Let's use the new browser checks from https://github.com/Joystream/atlas/pull/710

I think we should get rid of `useCheckBrowser`

src/hooks/useUser/useUser.tsx
```diff
@@ -61,6 +67,21 @@ export const ActiveUserProvider: React.FC = ({ children }) => {
     throw activeMembershipError
   }
 
+  const checkIfPolkadotExtensionInstalled = useCallback(async () => {
+    const res = await fetch(EXTENSION_URL)
+    const polkaDotExtensionInstalled = res.ok
+    if (polkaDotExtensionInstalled) {
+      console.warn('Polkadot extension disabled')
+      openDialog({
```
kdembler: I'm not sure if this should open for user right away. If I visit studio for the first time and have the extension installed, I will get a prompt to enable access from the extension. Once I enable it, I will see this dialog even though I just allowed the access. That's a bad UX. What if instead we show it only in `Add Polkadot extension` step?

kdembler: Maybe we should just skip the detection entirely and just show a link to this in the mentioned step in sign in? I'm not sure what additional value this brings us

mikkio-j: > Maybe we should just skip the detection entirely and just show a link to this in the mentioned step in sign in? I'm not sure what additional value this brings us

yeah, that seems to make most sense at the moment. Removed detection - currently dialog with instruction will appear on click in the sign in step

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -39,14 +36,41 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
...
-      </StepFooter>
+      </StyledStepFooter>
     </StepWrapper>
   )
 }
 
+export const PolkadotExtensionRejected: React.FC = () => (
+  <PolkadotExtensionRejectedWrapper>
+    <StyledPolkadotLogo />
+    <StepTitle variant="h4">Enable Polkadot extension website access</StepTitle>
```
kdembler:
```suggestion
    <StepTitle variant="h4">Allow Polkadot extension access</StepTitle>
```

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -39,14 +36,41 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
...
+    <StepSubTitle secondary variant="body2">
+      It seems like you have disabled Polkadot extension access to the app. Please follow the steps below to enable it:
```
kdembler:
```suggestion
      If you've denied Polkadot extension access for this website, you won't be able to use Joystream studio. To allow access, you can take the following steps:
```

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -39,14 +36,41 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
...
+    </StepSubTitle>
+    <ol>
+      <StyledListItem secondary as="li" variant="caption">
+        Open Polkadot extension menu in the right upper corner of the screen
```
kdembler:
```suggestion
        Open the extension popup with the icon in your browser bar
```

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -39,14 +36,41 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
...
+      </StyledListItem>
+      <StyledListItem secondary as="li" variant="caption">
+        In the menu open settings and Manage Website Access
```
kdembler:
```suggestion
        Use cog icon in upper right corner to open settings and select "Manage Website Access"
```

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -39,14 +36,41 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
...
+      </StyledListItem>
+      <StyledListItem secondary as="li" variant="caption">
+        Find play.joystream.org address and switch it to allowed
+      </StyledListItem>
+      <StyledListItem secondary as="li" variant="caption">
+        Reload the page after enabling the extension
```
kdembler:
```suggestion
        Reload the page
```

src/components/SignInSteps/ExtensionStep.tsx
```diff
@@ -21,6 +14,10 @@ const ExtensionStep: React.FC<ExtensionStepProps> = ({ nextStepPath }) => {
   const navigate = useNavigate()
   const step = useRouterQuery('step')
   const { extensionConnected } = useUser()
+  const [openEnableExtensionDialog, closeEnableExtensionDialog] = useDialog({
+    additionalActionsNode: <PolkadotExtensionRejected />,
```
kdembler: This is a bit of a hack to use a custom body right? Maybe we should change this line https://github.com/Joystream/atlas/blob/ff0f6d1cfbc04ab338e1e112697ef5fbabaf0e7e/src/components/Dialogs/MessageDialog/MessageDialog.tsx#L11

and allow `description` to be `React.ReactNode` instead? Then we can pass the component as `description` (ideally that'd be `body` instead but not change for now)

mikkio-j: > This is a bit of a hack to use a custom body right? Maybe we should change this line
> 
> https://github.com/Joystream/atlas/blob/ff0f6d1cfbc04ab338e1e112697ef5fbabaf0e7e/src/components/Dialogs/MessageDialog/MessageDialog.tsx#L11
> 
> and allow `description` to be `React.ReactNode` instead? Then we can pass the component as `description` (ideally that'd be `body` instead but not change for now)

Yes, that's kind of a hack. 
Hmm I see, we could go for a description instead for now.

kdembler (approved):

Issue #589, opened 2021-05-10
Display alert for unsupported browser


kdembler:

kdembler:
We should probably expand the scope on this task a bit. I'm seeing some errors from users not having `ResizeObserver` in their browser for example. We could define a list of supported browser version and show a dialog to the user if they are on an unsupported version. Another approach would be to check whether the browser contains all the necessary APIs (ResizeObservers, IntersectionObserver, etc)

PR #710, opened 2021-06-04
Unsupported browser Dialog

eldiegod:
#589 

This should handle the issue where browsers can't install the Polkadot extension but we will need a different fix to check when browser doesn't support APIs like ResizeObservers, IntersectionObserver... doing some 

Another approach would be to check whether the browser contains all the necessary APIs (, etc)

### DIFF
--- a/src/utils/broswer.ts
+++ b/src/utils/broswer.ts
@@ -1,3 +1,10 @@
 export const copyToClipboard = (text: string) => {
   navigator.clipboard.writeText(text)
 }
+
+// @ts-ignore not worth typing
+export const isFirefox = (): boolean => typeof InstallTrigger !== 'undefined'
+// @ts-ignore not worth typing
+export const isChromiumBased = (): boolean => !!window.chrome
+
+export const isAllowedBrowser = () => isFirefox() || isChromiumBased()
--- a/src/utils/broswer.ts
+++ b/src/utils/broswer.ts
@@ -7,4 +7,8 @@ export const isFirefox = (): boolean => typeof InstallTrigger !== 'undefined'
 // @ts-ignore not worth typing
 export const isChromiumBased = (): boolean => !!window.chrome
 
+// Seen in https://stackoverflow.com/a/9851769
+// it will check for the global chrome object for chromium based browsers and
+// it will check Firefox's API to install add-ons which as of now all firefox versions have
+// to detect if the users browser it's one where the polkadot extension can be installed on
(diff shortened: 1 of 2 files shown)


src/utils/broswer.ts
```diff
@@ -1,3 +1,10 @@
 export const copyToClipboard = (text: string) => {
   navigator.clipboard.writeText(text)
 }
+
+// @ts-ignore not worth typing
+export const isFirefox = (): boolean => typeof InstallTrigger !== 'undefined'
+// @ts-ignore not worth typing
+export const isChromiumBased = (): boolean => !!window.chrome
```
kdembler, line 8: Let's document that a bit with why this makes sense. I see this probably came from https://stackoverflow.com/a/9851769, maybe let's include that link, it also includes some background info

src/views/studio/StudioLayout.tsx
```diff
@@ -55,17 +57,30 @@ const StudioLayout = () => {
...
   const accountSet = !!activeAccountId && !!extensionConnected
   const memberSet = accountSet && !!activeMemberId && hasMembership
   const channelSet = memberSet && !!activeChannelId && hasMembership
 
+  useEffect(() => {
+    if (!isAllowedBrowser()) {
+      openUnsupportedBrowserDialog({
+        variant: 'warning',
+        title: 'Unsupported browser',
+        description: 'Please use a recent version of Chrome, Firefox or Edge',
```
kdembler:
```suggestion
        title: 'Unsupported browser detected',
        description: 'It seems the browser you are using is not fully supported by Joystream Studio. Some of the features may not be accessible. For the best experience, please use a recent version of Chrome, Firefox or Edge.',
```

src/views/studio/StudioLayout.tsx
```diff
@@ -55,17 +57,30 @@ const StudioLayout = () => {
...
   const [enterLocation] = useState(location.pathname)
   const hasMembership = !!memberships?.length
 
   const accountSet = !!activeAccountId && !!extensionConnected
   const memberSet = accountSet && !!activeMemberId && hasMembership
   const channelSet = memberSet && !!activeChannelId && hasMembership
 
+  useEffect(() => {
+    if (!isAllowedBrowser()) {
+      openUnsupportedBrowserDialog({
```
kdembler, line 70: Let's also add an `I understand` primary button that will close the dialog as well

kdembler (commented):
What about the dialog for missing APIs?

kdembler (approved):

Issue #706, opened 2021-06-02
Sidenav broken on Safari


kdembler:
<img width="349" alt="Screenshot 2021-06-02 at 21 52 33" src="https://user-images.githubusercontent.com/12646744/120543690-e238bf00-c3ec-11eb-9f09-73f6b05d966e.png">

PR #712, opened 2021-06-04
sidenav safari fix

mikkio-j:
Issue: #706


src/components/Sidenav/SidenavBase.style.ts
```diff
@@ -29,7 +29,7 @@ type SidebarNavLinkProps = {
 export const SidebarNav = styled.nav<SidebarNavProps>`
   position: fixed;
   top: 0;
-  bottom: 0;
+  height: 100vh;
```
kdembler: I think this could also work any may be a bit more safer?

```suggestion
  bottom: 0;
  height: 100%;
```

kdembler (approved):

PR #718, opened 2021-06-05
introduce style linting

### LINKED ISSUE #656: Introduce style linting

drillprop:
Fix #656

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,0 +1,22 @@
+module.exports = {
+  extends: [
+    'stylelint-config-standard',
+    'stylelint-config-prettier',
+    'stylelint-config-recommended',
+    'stylelint-config-styled-components',
+  ],
+  defaultSeverity: 'warning',
+  rules: {
+    'function-name-case': null,
+    'custom-property-empty-line-before': null,
+    'unit-no-unknown': null,
+    'declaration-block-no-duplicate-properties': true,
+    'value-keyword-case': null,
+    'rule-empty-line-before': [
+      'always',
+      {
+        'ignore': 'first-nested',
+      },
+    ],
+  },
+}
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,14 +1,10 @@
 module.exports = {
-  extends: [
-    'stylelint-config-standard',
-    'stylelint-config-prettier',
-    'stylelint-config-recommended',
-    'stylelint-config-styled-components',
-  ],
+  extends: ['stylelint-config-recommended'],
   defaultSeverity: 'warning',
   rules: {
     'function-name-case': null,
     'custom-property-empty-line-before': null,
+    'no-empty-source': null,
     'unit-no-unknown': null,
     'declaration-block-no-duplicate-properties': true,
     'value-keyword-case': null,
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,10 +1,9 @@
 module.exports = {
-  extends: ['stylelint-config-recommended'],
+  extends: ['stylelint-config-standard', 'stylelint-config-prettier'],
   defaultSeverity: 'warning',
   rules: {
     'function-name-case': null,
     'custom-property-empty-line-before': null,
-    'no-empty-source': null,
     'unit-no-unknown': null,
     'declaration-block-no-duplicate-properties': true,
     'value-keyword-case': null,
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -1,10 +1,19 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const path = require('path')
-const { override, addBabelPreset, addBabelPlugin, addWebpackAlias, addWebpackModuleRule } = require('customize-cra')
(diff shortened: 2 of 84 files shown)


drillprop:
Need to set it as a draft, found some issues with comments

.stylelintrc
```diff
@@ -0,0 +1,34 @@
...
+    "value-keyword-case": null,
+    "function-name-case": null,
+    "custom-property-empty-line-before": null,
+    "unit-no-unknown": null,
+    "rule-empty-line-before": ["always", {
+      "ignore": "first-nested"
+    }]
+  }
+}
+
```
mikkio-j: Maybe you could add filename extension `.stylelintrc.js`, or `.stylelintrc.json` to get syntax checking and highlighting

drillprop: that's a good point, will change that

mikkio-j:
I'm getting this error when firing lint with `--fix` flag:

```
TypeError: this.input.fromOffset is not a function
at TemplateSafeParser.getPosition (/atlas-style-linting/node_modules/postcss-safe-parser/node_modules/postcss/lib/parser.js:379:26)
```

drillprop:
> I'm getting this error when firing lint with `--fix` flag:
> 
> ```
> TypeError: this.input.fromOffset is not a function
> at TemplateSafeParser.getPosition (/atlas-style-linting/node_modules/postcss-safe-parser/node_modules/postcss/lib/parser.js:379:26)
> ```
Found this https://github.com/stylelint/stylelint/issues/5096 and downgrading `react-scripts` indeed solves the problem, but I don't think we want that.
Ran of ideas about what to do here

drillprop:
Also, worth mentioning,  it worked before - all of the fixes were made with `--fix` flag, I didn't do them manually. 
Tried to remove all the configs from `extends` and wipe out all the rules and still nothing. 
Possibly I had installed some packages in node_modules(because I tried several of them) which possibly allow me to run `fix` flag without an error. Don't know.

drillprop:
Fixed by installing `stylelint-rational-order-plugin` as a workaround, which is far from ideal
@mikkio-j  Can you try again?

mikkio-j (approved):
works now 👍 

for the workaround, imo it's better than downgrading `react-scripts`

.stylelintrc.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  extends: [
+    'stylelint-config-standard',
+    'stylelint-config-prettier',
+    'stylelint-config-recommended',
```
kdembler: It seems `stylelint-config-standard` already extends `stylelint-config-recommended`, I don't think we need it here?

config-overrides.js
```diff
@@ -1,10 +1,19 @@
...
+  addBabelPlugin,
+  addWebpackAlias,
+  addWebpackModuleRule,
+  addWebpackPlugin,
+} = require('customize-cra')
 
 module.exports = {
   webpack: override(
     addBabelPlugin('@emotion/babel-plugin'),
+    addWebpackPlugin(new StylelintPlugin({ files: './src/**/*.{tsx,ts}' })),
```
kdembler, line 16: I'm wondering whether we need this. I have a feeling this slows down "rebuild" time in development although cannot say for sure, hard to measure for sure. I think editor+CI linting would be enough. What do you think?

.stylelintrc.js
```diff
@@ -0,0 +1,22 @@
...
+    'stylelint-config-styled-components',
```
kdembler: Do we need this? According to docs, this is needed when you're using `stylelint-processor-styled-components`. With `stylelint` supporting this out of the box it doesn't seem needed

package.json
```diff
@@ -134,7 +136,14 @@
     "@storybook/react": "^6.1.16",
     "@storybook/theming": "^6.1.16",
     "@svgr/cli": "^5.5.0",
-    "@trivago/prettier-plugin-sort-imports": "^2.0.2"
+    "@trivago/prettier-plugin-sort-imports": "^2.0.2",
+    "stylelint": "^13.13.1",
+    "stylelint-config-prettier": "^8.0.2",
+    "stylelint-config-recommended": "^5.0.0",
+    "stylelint-config-standard": "^22.0.0",
+    "stylelint-rational-order-plugin": "^1.0.2",
```
kdembler: Seems we could replace that with
```
  "resolutions": {
    "postcss-safe-parser": "4.0.2"
  },
```

config-overrides.js
drillprop: Editor linting doesn't work without `stylelint` extension.  This can be helpful for people who don't use it. They will see the nice output in the console what's wrong -  for example
image
As for performance - I didn't spot the difference, but I may be wrong.

kdembler: Okay, let's add that. However, if we move to some other build tool it's quite possible we'll get rid of both `eslint` and `stylelint` plugins

.stylelintrc.js
drillprop: yeah, we don't need that, we don't need also `config-prettier'.

kdembler:
@drillprop needs rebase

PR #720, opened 2021-06-06
Sorting My videos

eldiegod:
Issue: #492

"rebased" sorting PR

### DIFF
--- a/src/api/client/cache.ts
+++ b/src/api/client/cache.ts
@@ -1,9 +1,10 @@
 import { InMemoryCache } from '@apollo/client'
-import { offsetLimitPagination, Reference, relayStylePagination, StoreObject } from '@apollo/client/utilities'
+import { offsetLimitPagination, relayStylePagination } from '@apollo/client/utilities'
 import { parseISO } from 'date-fns'
 import {
   AllChannelFieldsFragment,
   AssetAvailability,
+  GetVideosConnectionQueryVariables,
   GetVideosQueryVariables,
   Query,
   VideoConnection,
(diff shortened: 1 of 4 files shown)


src/api/client/cache.ts
```diff
@@ -84,16 +86,34 @@ const queryCacheFields: CachePolicyFields<keyof Query> = {
...
+            .sort(
+              (a, b) =>
+                (readField('createdAt', b.node) as Date).getTime() - (readField('createdAt', a.node) as Date).getTime()
+            )
+        : filteredEdges
+            ?.slice()
+            .sort(
+              (a, b) =>
+                (readField('createdAt', a.node) as Date).getTime() - (readField('createdAt', b.node) as Date).getTime()
+            )
```
kdembler: I think we could simplify that, we basically repeat the same snippet. Possibly we could just do regular sorting once and `reverse` if needed?

src/api/client/cache.ts
```diff
@@ -84,16 +86,34 @@ const queryCacheFields: CachePolicyFields<keyof Query> = {
...
     ...relayStylePagination(getVideoKeyArgs),
-    read(existing: VideoConnection, opts) {
-      const isPublic = opts.args?.where?.isPublic_eq
-      const filteredEdges = existing?.edges.filter(
-        (edge) => opts.readField('isPublic', edge.node) === isPublic || isPublic === undefined
-      )
+    read(
+      existing: VideoConnection,
+
+      { args, readField }: { args: GetVideosQueryVariables | null; readField: ReadFieldFunction }
```
kdembler: I think typing should be `GetVideosConnectionQueryVariables` instead?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -199,6 +219,20 @@ export const MyVideosView = () => {
           <>
             <TabsContainer>
               <Tabs initialIndex={0} tabs={mappedTabs} onSelectTab={handleSetCurrentTab} />
+              <SortContainer>
+                <Text variant="body2">Sort by</Text>
+                <Select
+                  helperText={null}
+                  value={sortVideosBy}
+                  items={[...SORT_OPTIONS]}
```
kdembler: Let's remove that destructuring here. If needed let's remove `as const` as well, we can use the type of value directly (`VideoOrderByInput`)

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -199,6 +219,20 @@ export const MyVideosView = () => {
...
+                  onChange={(value) => {
+                    if (value) {
+                      setSortVideosBy(value)
+                      refetch({ orderBy: value })
+                    }
+                  }}
```
kdembler: Let's move that to a named handler above `return`

src/api/client/cache.ts
eldiegod: oh, yes thats right

kdembler (approved):

Issue #643, opened 2021-05-20
Richer notifications handling


kdembler:
I think we could have richer notifications in some scenarios. For example for asset uploads, I think we could show video title in the notification and say something like `2/2 assets uploaded`. I think we should also show only one notification per video, so if the previous notification said `1/2`, when the update happens, the old one should go away and be replaced with the new one. That would require some kind of id handling to be able to remove some specific notification. Let's discuss

PR #723, opened 2021-06-07
richer notifications handling

mikkio-j:
Issue: #643

### DIFF
--- a/src/hooks/useSnackbar/useSnackbar.tsx
+++ b/src/hooks/useSnackbar/useSnackbar.tsx
@@ -17,6 +17,7 @@ export type DisplaySnackbarArgs = {
   title: string
   description?: string
   actionText?: string
+  onExit?: () => void
   onActionClick?: () => void
   onExitClick?: () => void
 }
@@ -47,7 +48,7 @@ const SNACKBARS_LIMIT = 3
 export const SnackbarProvider: React.FC = ({ children }) => {
   const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
 
-  const displaySnackbar = useCallback(({ customId, timeout, ...args }: DisplaySnackbarArgs) => {
+  const displaySnackbar = useCallback(({ customId, timeout, onExit, ...args }: DisplaySnackbarArgs) => {
     const id = customId ?? createId()
     setSnackbars((currentSnackbars) => {
       return [...currentSnackbars, { id, ...args }]
@@ -55,6 +56,7 @@ export const SnackbarProvider: React.FC = ({ children }) => {
 
     if (timeout) {
       setTimeout(() => {
+        onExit?.()
         setSnackbars((currentSnackbars) => currentSnackbars.filter((snackbar) => snackbar.id !== id))
       }, timeout)
     }
--- a/src/hooks/useSnackbar/useSnackbar.tsx
+++ b/src/hooks/useSnackbar/useSnackbar.tsx
@@ -47,7 +47,7 @@ export const SnackbarProvider: React.FC = ({ children }) => {
   const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
 
   const displaySnackbar = useCallback(({ customId, timeout, ...args }: DisplaySnackbarArgs) => {
-    const id = customId || createId()
+    const id = customId ?? createId()
     setSnackbars((currentSnackbars) => {
       return [...currentSnackbars, { id, ...args }]
     })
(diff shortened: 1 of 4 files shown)


eldiegod (approved):
Looks good added a tiny suggestion

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -44,8 +45,8 @@ const SNACKBARS_LIMIT = 3
 export const SnackbarProvider: React.FC = ({ children }) => {
   const [snackbars, setSnackbars] = useState<SnackbarsState[]>([])
 
-  const displaySnackbar = useCallback(({ timeout, ...args }: DisplaySnackbarArgs) => {
-    const id = createId()
+  const displaySnackbar = useCallback(({ customId, timeout, ...args }: DisplaySnackbarArgs) => {
+    const id = customId || createId()
```
eldiegod:
```suggestion
    const id = customId ?? createId()
```

eldiegod (approved):

src/hooks/useSnackbar/useSnackbar.tsx
eldiegod: customId might carry a falsy value

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -133,6 +138,13 @@ export const MyVideosView = () => {
   // Workaround for removing drafts from video sheet tabs via snackbar
   // Snackbar will probably need a refactor to handle actions that change state
   useEffect(() => {
+    if (videoTabs.length === 0) {
+      closeSnackbar(OPEN_TAB_SNACKBAR)
+      return
+    }
+    updateSnackbar(OPEN_TAB_SNACKBAR, {
+      title: `${videoTabs.length > 1 ? `${videoTabs.length} videos` : 'Video'} opened in a new tab`,
```
kdembler: That won't produce the correct output though right? You can have 5 tabs already opened, add only 1, but you will get snackbar that you added 6

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -54,7 +54,16 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
...
+
+  const pendingUploadingNotificationsCounts = useRef(0)
+  type AssetParentObjectID = string
+  const assetsNotificationsCount = useRef<{
+    uploading: AssetParentObjectID[]
+    uploaded: AssetParentObjectID[]
+  }>({
+    uploading: [],
+    uploaded: [],
+  })
```
kdembler: Wouldn't it be easier to just use an objects here instead? So something like
```
{
  uploading: {
    '12': 2
  }
}
```

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -54,7 +54,16 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
     },
     { skip: !uploadsState.length }
   )
-  const pendingNotificationsCounts = useRef({ uploading: 0, uploaded: 0 })
+
+  const pendingUploadingNotificationsCounts = useRef(0)
+  type AssetParentObjectID = string
+  const assetsNotificationsCount = useRef<{
```
kdembler, line 59: Also, it's possible for there to be a clash bettwen channel and video ID, I think we will need to include parent type in the key

src/hooks/useUploadsManager/useUploadsManager.tsx
mikkio-j: Maybe it would, but now I added type of parent object to it so I'm not so sure about how would we store parent object type as well with that solution

src/hooks/useSnackbar/useSnackbar.tsx
```diff
@@ -10,13 +10,16 @@ import { createId } from '@/utils/createId'
...
+  customId?: string
   timeout?: number
   variant?: 'primary' | 'secondary'
   iconType?: SnackbarIconType
   title: string
   description?: string
   actionText?: string
+  onExit?: () => void
   onActionClick?: () => void
+  onExitClick?: () => void
```
kdembler: I see why we need `onExit` now, I guess one question I have is do we need separate callback for exit and exit click? Maybe both of those could be just `onExit`?

src/hooks/useUploadsManager/useUploadsManager.tsx
kdembler: Could we just use something like `${parentType}-${parentId}` as a key in that object? So for example `channel-12`

kdembler (approved):

PR #725, opened 2021-06-07
fix overflowing draft titles

### LINKED ISSUE #714: Draft titles not wrapped
This may be the case in other places where there is a video preview.

Screenshot 2021-06-04 at 14 28 12

drillprop:
Fix #714


src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -238,6 +238,9 @@ export const TitleHeader = styled(Text)<MainProps & ScalesWithCoverProps & Click
   ${({ main }) => main && fluidRange({ prop: 'fontSize', fromSize: '24px', toSize: '40px' })};
   line-height: ${({ main }) => (main ? 1 : 1.25)};
   cursor: ${(props) => (props.clickable ? 'pointer' : 'auto')};
+  max-width: 280px;
```
kdembler: This shouldn't be a constant value. Component will scale based on screen size so this should rather be expressed in percentage of the width

kdembler (changes requested):

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
drillprop: To be honest, I struggled a bit to find which properties I need to change here 😄 . Should be good now

kdembler (approved):

Issue #428, opened 2021-04-02
Update react-hook-form to v7


kdembler:
What we want to achieve:
- Currently we use `Controller` in couple places to include things like selects in our form validation. However, it seems it doesn't do proper typechecking on `value` / `onChange` and it a potential place for errors. Let's figure out how to type it. Maybe `react-hook-form` v7 which was recently released would help.
- In `EditVideoForm` we have `handleChanges` general handler to be called when there were changes made to the form. I think v7 introduced something like `watch((changes) => {}` that should allow us to get rid of this

kdembler:
It has been just fixed with 7.0.1 - https://github.com/react-hook-form/react-hook-form/issues/4599

PR #735, opened 2021-06-08
react-hook-form v7 upgrade

mikkio-j:
Issue: #428 

~~For using `watch` instead of form change handler - I decided not to do that. React-hook-form documentation warns about performance issues. I did experience stuttering in video sheet when using `watch((data) => {})`. Instead handler is now being triggered by `FormWrapper` change, no need to put it on each field~~

### DIFF
--- a/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
+++ b/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
@@ -96,10 +96,9 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
     register,
     handleSubmit: createSubmitHandler,
     control,
-    formState: { isDirty, dirtyFields },
+    formState: { isDirty, dirtyFields, errors },
     watch,
     reset,
-    errors,
   } = useForm<Inputs>({
     defaultValues: {
       avatar: { url: null, blob: null, assetDimensions: null, imageCropData: null },
@@ -356,7 +355,7 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
         <Controller
           name="cover"
           control={control}
-          render={({ value, onChange }) => (
+          render={({ field: { value, onChange } }) => (
             <>
               <ChannelCover
                 coverPhotoUrl={loading ? null : value.url}
@@ -387,7 +386,7 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
           <Controller
             name="avatar"
             control={control}
-            render={({ value, onChange }) => (
+            render={({ field: { value, onChange } }) => (
               <>
                 <StyledAvatar
                   imageUrl={value.url}
@@ -421,7 +420,7 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
                   name="title"
                   control={control}
                   rules={textFieldValidation({ name: 'Channel name', minLength: 3, maxLength: 40, required: true })}
-                  render={({ value, onChange }) => (
+                  render={({ field: { value, onChange } }) => (
                     <Tooltip text="Click to edit channel title">
                       <StyledHeaderTextField
                         ref={titleRef}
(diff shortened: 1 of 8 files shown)


eldiegod (approved):
Working good !!

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -453,19 +452,22 @@ const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ newChanne
...
+                    placeholder="Description of your channel to share with your audience"
+                    rows={8}
+                    ref={descriptionRef}
+                    value={value}
+                    onChange={onChange}
+                    maxLength={1000}
+                    error={!!errors.description}
+                    helperText={errors.description?.message}
+                  />
+                )}
```
kdembler: Why did we move to a `Controller` here? Is that because of sharing `ref` usage? I'm pretty sure react-hook-form@7 has to have some kind of solution for that without using `Controller`?

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -412,50 +405,49 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
-                  onChange={(value) => {
-                    onChange(value)
-                    handleFormChange()
-                  }}
-                  error={!!errors.category && !value}
-                  helperText={errors.category?.message}
-                />
-              )}
+              render={({ field: { value, onChange }, fieldState: { invalid } }) => {
+                if (invalid) handleFieldFocus(categorySelectRef)
```
kdembler: This won't work like it used to. Previously we would scroll to the field only when `onFocus` happened, so if the user tried to submit the form that had errors. Here you call the scroll in the `render` function so the scroll will happen every time this field renders which is obviously not what we want to happen. Take a look:

https://user-images.githubusercontent.com/12646744/121869840-bd085280-cd02-11eb-87cd-8961acbe23b4.mov

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -330,11 +330,11 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
 
   return (
     <>
-      <FormWrapper>
+      <FormWrapper onChange={handleFormChange}>
```
kdembler: This doesn't work like it used to. `handleFormChange` won't be called when a select value changes since it's not really part of a form and thus browser doesn't know about its value changing. We need to either:
a) use the previous approach
b) fix this somehow
c) use `watch` - you mentioned `React-hook-form documentation warns about performance issues`, can you link that?

kdembler (changes requested):

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
mikkio-j: Yes, it is because of the `ref` usage. And yes, it has solution for sharing refs but it would require registering description above the return (and not in the component itself) with something like:
`const { ref, ...rest } = register('description');`

 Thought it might be confusing to register a component in a different place than others

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
mikkio-j: Last point of `Rules`:
https://react-hook-form.com/api/useform/watch/#main

kdembler: Is that the case for when we pass callback to `watch` as well? My guess would be that the callback option was introduced exactly for that reason - to not require the component to rerender on change

kdembler: https://codesandbox.io/s/serene-waterfall-it1me?file=/src/App.tsx

just tried it out, it seems callback gets called without a rerender of the component

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -224,44 +240,46 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
+      ref.scrollIntoView({
+        behavior: 'smooth',
+        block: 'start',
+      })
     }
+  )
 
-    const callback = () => {
-      setForceReset(true)
+  watch((data) => {
```
kdembler: This will add a new callback on every render. You can easily get to a point at which this will be called 50 times on each key press :P

We need to put this inside some `useEffect`. Calling `watch` also returns a reference to subscription that we can use for cleanup if needed

kdembler (changes requested):

kdembler (approved):

Issue #647, opened 2021-05-20
Sometimes created videos are not displayed in uploads and they're not deleted from drafts


drillprop:
This will need some investigation. Don't know how to reproduce this, but I experienced this bug several times.

Sometimes when I creating the video after a transaction is done and the sheet is minimized, the video is not added to the uploads and it's not removed from drafts. When Im trying to close the tab I also see a warning about data lost. 
When I refresh the site, I see snackbar: `Asset waiting to resume upload.` but don't see `AssetLine` in UploadsView with this video(If I remember it correctly)
Not 100% sure, but I think it's not happening when you upload only one video.  It happens when you upload the second video in the row.

kdembler:
I think we managed to resolve that more or less. If the behaviour returns, let's open a new issue with updated details

PR #742, opened 2021-06-08
Zustand

drillprop:
snackbars first, will work on uploads tomorrow.


src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -59,9 +41,8 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
         id_in: uploadsState.filter((item) => item.parentObject.type === 'video').map((item) => item.parentObject.id),
       },
     },
-    { skip: !uploadsState.length }
+    { skip: !uploadsState.length, fetchPolicy: 'network-only' }
```
drillprop: I probably should not add this, but without this policy new videos weren't added to uploads.
Also as a sidenote - it partly fixes #647 - maybe it's worth investigating in the future

src/store/store.tsx
```diff
@@ -0,0 +1,45 @@
...
+    name: ZUSTAND_LOCAL_STORAGE_KEY,
+    whitelist: ['uploadsState'],
+  })
+)
+
+const StoreProvider: React.FC = ({ children }) => {
+  const navigate = useNavigate()
+  useEffect(() => {
+    store.setState((state) => ({ ...state, navigate }))
+  }, [navigate])
```
drillprop: I want to put navigate function inside the store. Not sure if it's a good pattern

src/store/store.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ...stateCreators: StateCreator<any>[]
+): StateCreator<TState> => (set, get, api) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let values: any = {}
+
+  stateCreators.forEach((sc) => {
+    values = Object.assign({}, values, sc(set, get, api))
+  })
+  return values
+}
```
drillprop: found here https://codesandbox.io/s/zustand-combine-state-creators-y0eev?file=/src/zustand/middleware.ts
We could also do something like this: 
```tsx
export const store = create<RootState>(persist((set, get, api) => ({
  ...useSnackbar(set, get, api),
  ...useUploadsManagerStore(set, get, api),
}), {
  name: ZUSTAND_LOCAL_STORAGE_KEY,
  whitelist: ['uploadsState'],
}))
```

src/store/snackbars.tsx
```diff
@@ -0,0 +1,61 @@
...
+} & Omit<DisplaySnackbarArgs, 'time'>
+
+export type SnackbarState = {
+  snackbars: Snackbar[]
+  displaySnackbar: (args: DisplaySnackbarArgs, onActionClick?: () => void) => string
+  updateSnackbar: (id: string, opts: Omit<DisplaySnackbarArgs, 'id'>) => void
+  closeSnackbar: (id: string) => void
+}
+
+const immer = <T extends State>(config: StateCreator<T, (fn: (draft: Draft<T>) => void) => void>): StateCreator<T> => (
```
drillprop: I decided to not use immer after all.  I don't want to overcomplicate things for now.  In the future probably we could create `middleware.ts` and move this function there and use it in the root store.

src/store/store.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const store = create<RootState>(
+  persist(combineStateCreators(useSnackbar, useUploadsManagerStore), {
+    name: ZUSTAND_LOCAL_STORAGE_KEY,
+    whitelist: ['uploadsState'],
+  })
+)
```
mikkio-j: So the state is currently spread, but we should be able to nest it, right? Current implementation force us to call store for each field separately:
```
  const uploadsProgress = useStore((state) => state.uploadsProgress)
  const updateAsset = useStore((state) => state.updateAsset)
  const uploadsState = useStore((state) => state.uploadsState)
```
Ideal api would be something similar to this:
```
  const { uploadsProgress, updateAsset, uploadsState}  = useStore((state) => state.uploads)
```

src/store/uploads.tsx
```diff
@@ -0,0 +1,240 @@
...
+      setAssetUploadProgress(0)
+
+      const setUploadProgressThrottled = throttle(
+        ({ loaded, total }: ProgressEvent) => {
+          get().updateAsset(asset.contentId, 'inProgress')
+          setAssetUploadProgress((loaded / total) * 100)
+        },
+        3000,
+        { leading: true }
+      )
```
mikkio-j: It seems like introducing state management resolves the issue of upload progress causing multiple rerenders, so that's a good thing. We will not need that throttle then

src/store/store.tsx
drillprop: Probably we should in the end split the store(for snack bars and for uploads). 

I'll just leave it to show that creating a root state is possible, and it's also a little bit awkward

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -28,29 +22,30 @@ type GroupByParentObjectIdAcc = {
...
-  const [uploadsProgress, setUploadsProgress] = useState<UploadsProgressRecord>({})
+
+  const displaySnackbar = useStore((state) => state.displaySnackbar)
+  const addAsset = useStore((state) => state.addAsset)
+  const updateAsset = useStore((state) => state.updateAsset)
+  const uploadsState = useStore((state) => state.uploadsState)
+  const uploadsProgress = useStore((state) => state.uploadsProgress)
+  const setUploadsProgress = useStore((state) => state.setUploadsProgress)
+  const assetsFiles = useStore((state) => state.assetsFiles)
+  const setAssetsFiles = useStore((state) => state.setAssetsFiles)
```
drillprop: Possibly I could do: 
```tsx
const {displaySnackbar, addAsset, updateAsset, uploadsState, uploadsProgress, setUploadsProgress, assetsFiles, setAssetsFiles} = useStore()
``` 
but I'm afraid that it may cause some unintended rerenders - not 100% sure

Ideally, as I mentioned [before](https://github.com/Joystream/atlas/pull/742/files#r649792760) - we should just split the global store into smaller stores.

src/hooks/useSnackbar/store.ts
```diff
@@ -0,0 +1,57 @@
...
+} & Omit<DisplaySnackbarArgs, 'time'>
+
+export type SnackbarState = {
+  snackbars: Snackbar[]
+  displaySnackbar: (args: DisplaySnackbarArgs, onActionClick?: () => void) => string
+  updateSnackbar: (id: string, opts: Omit<DisplaySnackbarArgs, 'id'>) => void
+  closeSnackbar: (id: string) => void
+}
+
+export const useSnackbar = create<SnackbarState>((set) => ({
```
kdembler: How about we call this `useSnackbarStore` and then create separate `useSnackbar` that would look like this:
```ts
const useSnackbar = () => {
  return useSnackbarStore((state) => state.displaySnackbar)
}
```

Then consumers don't need to bother with extracting the state. What do you think?

src/hooks/useStartFileUpload.tsx
```diff
@@ -0,0 +1,205 @@
+import axios, { AxiosError } from 'axios'
```
kdembler: This file should be in the same directory as the rest of uploads manager stuff

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -1,75 +1,48 @@
...
+
+  const displaySnackbar = useSnackbar((state) => state.displaySnackbar)
+  const updateAsset = useUploadsStore((state) => state.updateAsset)
+
+  const uploadsStateIds = useUploadsStore(
+    (state) =>
+      state.uploadsState.filter((item) => item.parentObject?.type === 'video').map((item) => item.parentObject.id),
+    shallow
+  )
+  const chanelUploadsState = useUploadsStore(
```
kdembler: should be `channel` with double `n`, needs replace in couple places

src/hooks/useUploadsManager/store.tsx
```diff
@@ -0,0 +1,71 @@
...
+  uploadsState: AssetUpload[]
+  addAsset: (asset: AssetUpload) => void
+  setUploadsProgress: (contentId: string, progress: number) => void
+  updateAsset: (contentId: string, lastStatus: AssetUploadStatus) => void
+  removeAsset: (contentId: string) => void
+  uploadsProgress: UploadsProgressRecord
+  notifications: {
+    uploading: number
+    uploaded: number
+  }
```
kdembler: What do we need those for? Seems `useStartFileUpload` already has its own notification counts?

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -293,5 +90,5 @@ export const useUploadsManager = () => {
   if (ctx === undefined) {
     throw new Error('useUploadsManager must be used within a UploadManagerProvider')
   }
-  return ctx
+  return { ...ctx }
```
kdembler: Why a copy of `ctx`?

src/views/playground/PlaygroundLayout.tsx
```diff
@@ -59,13 +60,14 @@ export const PlaygroundLayout = () => {
                       <Route key={route.path} path={route.path} element={route.element} />
                     ))}
                   </Routes>
+                  <Snackbars />
```
kdembler: Maybe let's just include that once in `<App />`?

src/hooks/useSnackbar/store.ts
drillprop: Good idea, updated

src/views/studio/MyUploadsView/MyUploadsView.tsx
```diff
@@ -1,17 +1,43 @@
...
+      const key = asset.parentObject.id
+      if (!acc[key]) {
+        acc[key] = []
+      }
+      acc[key].push(asset)
+      return acc
+    }, {})
+  )
 
+  const hasUploads = groupedUploadsState.length > 0
```
eldiegod, line 40: Any reason to move this logic into the view?

src/hooks/useUploadsManager/useStartFileUpload.tsx
```diff
@@ -0,0 +1,207 @@
...
+      assetsFiles,
+      displaySnackbar,
+      getStorageProvider,
+      markStorageProviderNotWorking,
+      navigate,
+      setAssetsFiles,
+      setUploadsProgress,
+      updateAsset,
+    ]
+  )
```
eldiegod, line 204: Tying to understand if it would it be possible to have this logic inside the store or are we limited by something here

eldiegod (commented):
Looks good overall, I think the snackbars came out pretty clean and got some questions about the uploads store.

Besides that I still think we should add localForage as our persistence solution to benefit from the async api, saw that it should be possible in zustand. Not sure if there is a way to add that config to every store by default

Also why did we remove immer seemed like the lib was a step right direction?

src/hooks/useUploadsManager/useStartFileUpload.tsx
kdembler: It actually was part of the store previously but I asked @drillprop to move it here. The reason is I'd like to separate store (state) and logic for it. I'm aware this may not be the most usual approach with libraries like Redux where you would put all your HTTP calls in the store, but I'd like our store to be lightweight, let's try it out

kdembler (approved):

Issue #733, opened 2021-06-08
Filter videos pending upload from search results


bedeho:
I am seeing videos eppear in Atlas before the primary asset has been uploaded, this is akward, as playback will not work, and the thumbnail will possibly not be visible as well. I thought we had used the expressive query node API to pick up on this and filter this out. I saw such videos both in search results and on the channel page, but a bit less sure about the latter.

kdembler:
Yep, we already do that for videos but we missed that in the search results. Shouldn't happen for videos in the channel page though, we do proper filtering there

PR #752, opened 2021-06-09
Filter pending videos search

mikkio-j:
Issue: #733

### DIFF
--- a/src/views/viewer/SearchOverlayView/SearchResults/SearchResults.tsx
+++ b/src/views/viewer/SearchOverlayView/SearchResults/SearchResults.tsx
@@ -18,20 +18,21 @@ const tabs = ['all results', 'videos', 'channels']
 
 const SearchResults: React.FC<SearchResultsProps> = ({ query }) => {
   const [selectedIndex, setSelectedIndex] = useState(0)
-  const { data, loading, error } = useSearch({ text: query })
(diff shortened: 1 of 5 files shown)


src/views/viewer/SearchOverlayView/SearchResults/SearchResults.tsx
```diff
@@ -25,7 +25,13 @@ const SearchResults: React.FC<SearchResultsProps> = ({ query }) => {
...
     }
     const results = data
-    const videos = results.flatMap((result) => (result.item.__typename === 'Video' ? [result.item] : []))
+    const videos = results.flatMap((result) =>
+      result.item.__typename === 'Video' &&
+      result.item.mediaAvailability === AssetAvailability.Accepted &&
+      result.item.thumbnailPhotoAvailability === AssetAvailability.Accepted
+        ? [result.item]
+        : []
+    )
```
kdembler: Why do we do client-side filtering instead of specifying an argument for the query?

kdembler: I see now that the `search` query we have in the schema doesn't have all the possible arguments. Check https://hydra.joystream.org/graphql (docs on the right edge of the screen)

kdembler (approved):

PR #753, opened 2021-06-09
Scrollbar styles

eldiegod:
Scrollbar styles for firefox and webkit browsers
https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=7465%3A195160

### DIFF
--- a/src/shared/components/GlobalStyle/GlobalStyle.tsx
+++ b/src/shared/components/GlobalStyle/GlobalStyle.tsx
@@ -9,6 +9,17 @@ import { media, colors, sizes, typography } from '../../theme'
 const globalStyles = css`
   ${emotionNormalize};
 
+  body::-webkit-scrollbar {
+    width: 8px;
+  }
+  body::-webkit-scrollbar-track {
+    background: var(--scrollbarBG);
+  }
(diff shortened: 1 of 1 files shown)


mikkio-j:
Looks good! In firefox it seems like there is full coverage for all scrollbars, but in Chromium-based browsers I miss styles in following places:

- followed channels list in viewer's sidenav
- channels list in studio's topbar dropdown
- select input
- video sheet form

eldiegod:
The chrome issues should be fixed in the last commit

drillprop (commented):
Tested on chrome and firefox. Works on chrome, but on firefox, I see just a regular one - is it intentional or I missing something? 
Tested it on mac, btw.

drillprop (approved):
> Tested on chrome and firefox. Works on chrome, but on firefox, I see just a regular one - is it intentional or I missing something?
> Tested it on mac, btw.

I think we can ignore this. Turns out that it works well, but you have to enable the option `show scrollbar: always` on mac preferences.

mikkio-j (approved):

kdembler (approved):

PR #754, opened 2021-06-09
Add open in tab option in publisher video context menu

### LINKED ISSUE #745: Launch video from My Videos
I often want to launch a video from `My Videos`, however I end up having to use the `Copy video URL` action + opening a new tab + pasting in the URL + hitting enter. There really should be a `Open in Tab` option above  :

Screenshot 2021-06-08 at 20 39 58

rafalpawlow:
Closes #745

### DIFF
--- a/src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
+++ b/src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
@@ -15,6 +15,7 @@ import {
   SvgLargeEdit,
   SvgOutlineVideo,
   SvgLargeUploadFailed,
+  SvgGlyphPlay,
 } from '@/shared/icons'
 import { transitions } from '@/shared/theme'
 import { formatDateAgo, formatDurationShort } from '@/utils/time'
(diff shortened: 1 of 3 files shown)


src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -345,6 +349,11 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
                   <SvgGlyphMore />
                 </KebabMenuIconContainer>
                 <ContextMenu contextMenuOpts={contextMenuOpts}>
+                  {onOpenInTabClick && (
+                    <ContextMenuItem icon={<SvgGlyphNewTab />} onClick={onOpenInTabClick}>
+                      Open in tab
+                    </ContextMenuItem>
```
kdembler: This still displays for drafts, we should display that only for published videos:
Screenshot 2021-06-10 at 10 08 57

src/shared/components/VideoPreviewBase/VideoPreviewBase.tsx
```diff
@@ -345,6 +349,11 @@ const VideoPreviewBase: React.FC<VideoPreviewBaseProps> = ({
                   <SvgGlyphMore />
                 </KebabMenuIconContainer>
                 <ContextMenu contextMenuOpts={contextMenuOpts}>
+                  {onOpenInTabClick && (
+                    <ContextMenuItem icon={<SvgGlyphNewTab />} onClick={onOpenInTabClick}>
```
kdembler: Thinking of it now, we will need to call this something else I think. We already use "Open in new tab" for opening video for edit (in new tab of video sheet). Let's try using `SvgGlyphPlay` instead and make text `Play in Joystream`

kdembler (approved):

PR #762, opened 2021-06-10
Remove pulsing animation on uploaded assets

mikkio-j:
Resolves: #759

### DIFF
--- a/src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.style.tsx
+++ b/src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.style.tsx
@@ -6,6 +6,7 @@ import { colors, sizes, media, transitions } from '@/shared/theme'
 
 type ProgressbarProps = {
   progress: number
+  isCompleted: boolean
 }
 
 type AssetsGroupUploadBarProps = {
@@ -62,12 +63,12 @@ export const ProgressBar = styled.div<ProgressbarProps>`
   transform-origin: 0 0;
   transform: scaleX(${({ progress }) => progress && `${progress / 100}`});
   transition: transform 1s linear;
-  animation: ${pulse} 2.5s infinite ease-in-out;
-  ${({ progress }) =>
-    progress === 100 &&
-    css`
-      animation: none;
-    `}
+  animation: ${({ isCompleted }) =>
+    isCompleted
+      ? 'none'
+      : css`
+          ${pulse} 2.5s infinite ease-in-out
+        `};
 `
 
 export const Thumbnail = styled.div`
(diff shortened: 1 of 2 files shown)


src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.style.tsx
```diff
@@ -62,6 +62,11 @@ export const ProgressBar = styled.div<ProgressbarProps>`
   transform: scaleX(${({ progress }) => progress && `${progress / 100}`});
   transition: transform 1s linear;
   animation: ${pulse} 2.5s infinite ease-in-out;
+  ${({ progress }) =>
+    progress === 100 &&
+    css`
+      animation: none;
+    `}
```
drillprop:
```suggestion
  animation: ${({ progress }) =>
    progress === 100
      ? 'none'
      : css`
          ${pulse} 2.5s infinite ease-in-out
        `};
```
I like this more

kdembler:
Maybe we should use status instead of progress?

drillprop (approved):

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.style.tsx
```diff
@@ -61,7 +62,12 @@ export const ProgressBar = styled.div<ProgressbarProps>`
   transform-origin: 0 0;
   transform: scaleX(${({ progress }) => progress && `${progress / 100}`});
   transition: transform 1s linear;
-  animation: ${pulse} 2.5s infinite ease-in-out;
+  animation: ${({ isCompleted }) =>
+    isCompleted
```
kdembler: Actually, I think that should also take into account errors for example. If both assets failed and are not uploading, the progress bar shouldn't be pulsating as well. I think the animation should happen only if at least one of the items is currently being uploaded, otherwise it doesn't make sense I think

src/views/studio/MyUploadsView/AssetsGroupUploadBar/AssetsGroupUploadBar.style.tsx
```diff
@@ -61,7 +62,12 @@ export const ProgressBar = styled.div<ProgressbarProps>`
   transform-origin: 0 0;
   transform: scaleX(${({ progress }) => progress && `${progress / 100}`});
   transition: transform 1s linear;
-  animation: ${pulse} 2.5s infinite ease-in-out;
+  animation: ${({ hasUploadingAsset }) =>
```
kdembler: I don't like this style of splitting the style declaration, let's keep it more concise. For example like this:
```ts
const pulseAnimationCss = (props: ProgressbarProps) =>
  props.hasUploadingAsset
    ? css`
        animation: ${pulse} 2.5s infinite ease-in-out;
      `
    : null

export const ProgressBar = styled.div<ProgressbarProps>`
  // ...
  ${pulseAnimationCss};
`

kdembler (approved):

Issue #757, opened 2021-06-10
Maximize video sheet when clicking a tab


kdembler:
If sheet is minimized, you see video tabs. Let's maximize the sheet if user clicks on one of them

PR #763, opened 2021-06-10
Video tab changes

eldiegod:
Close the video tab after publish was successful #758 
Maximize video sheet when clicking a tab #757


mikkio-j (approved):
Working good 👍  Only question I could think of is if we want to close the sheet if there are other videos in tabs. It makes more sense to leave it minimized imo so 👌

drillprop (commented):
I'm not sure if it completely resolves the problem.
First thing. I've noticed that there is a little delay between showing the dialog and closing the tab. I don't think it's intuitive. I think we should show the dialog and close the tab simultaneously(or close the tab first and then show the dialog - this would be perfect). It may be not easy to fix though - maybe we should create another issue fo this.
bug tab

Also, sometimes after I upload one video I still see a tab beneath My videos - unfortunately don't know the exact steps to reproduce. 
image

kdembler (changes requested):
@drillprop is obviously right here, if you have only one video tab open, the sheet will close after success and then go back to minimized when you click the `Close` button, it happens every time:
https://github.com/Joystream/atlas/blob/65a8a72768c3c01dc6ecae625fcaf31a53aac012/src/views/studio/EditVideoSheet/EditVideoSheet.tsx#L255

For now, let's not close the tab on success itself, let's wait for close click in confirmation dialog

eldiegod:
Removing the tab on close click on the dialog seems to have solved the problem

kdembler (approved):

Issue #751, opened 2021-06-09
Set up centralized logger


kdembler:
I think we could use some centralized logger to clean up our logs a bit.
Currently we just use `console.log/warn/error` whenever we see fit. Let's replace that with something a bit more custom so we can better control what get's logged to the console/sent to Sentry.

I think the api should be something like this
```ts
// instead of
console.log('Hello world', { details: 123 })
console.error('😢')

// let's do
import Logger from '@/somewhere/Logger'
Logger.log('Hello world', { details: 123 })
Logger.error('😢')
```

Probably the custom log methods should just call respective `console` methods for now

bedeho:
This would be server side logging? I think that would be great, I was actually going to make an issue about this if that is what you meant. The easiest way to detect problems in the infra is to have end user apps directly report the issue, and it should be publicly available in some sort of dashboard or whatever. Let me know if this is not what you had in mind.

kdembler:
Not exactly but would result mostly in the same thing. We already have kind of "server side logging", with Sentry. All errors happening in the app will get sent there. This is mostly about structuring that process a bit better on the app side so we have more control over what gets sent there.

kdembler:
However, for detecting problems with infra, we probably could have something else. Atlas could be calling some service that we create with messages like "I failed to get data from query node" or "I failed to get response from the faucet server". Not sure if this is what you mean?

PR #764, opened 2021-06-11
Centralized logger

mikkio-j:
Issue: #751


kdembler (commented):
Looks good!

Needs a rebase and there's one more thing I think we could do. Let's enable `no-console` ESLint rule to enforce not using `console` directly. Obviously let's disable that rule inside the logger

kdembler (approved):

PR #766, opened 2021-06-11
uploads improvements

kdembler:
Summary of changes:
- fix video delete
- fix videos sorting (broken on home page)
- maintain single list of storage providers, maintain list of storage providers that are failing so we don't use them
- improve uploads view handling of videos/channel, make sure we can add video without refetching the full list
- improve states of uploaded assets
- improve error handling on failed uploads


src/views/studio/MyUploadsView/MyUploadsView.tsx
```diff
@@ -1,23 +1,24 @@
...
 
-const MyUploadsView = () => {
-  const { activeChannelId } = useAuthorizedUser()
-  const { uploadsState, isLoading } = useUploadsManager(activeChannelId)
+const MyUploadsView: React.FC = () => {
+  const { uploadsState, isLoading } = useUploadsManager()
 
   const hasUploads = uploadsState.length > 0
 
+  const placeholderItems = Array.from({ length: 5 }).map((_, idx) => <AssetGroupUploadBarPlaceholder key={idx} />)
```
drillprop, line 15: Not sure which change cause it, but when I go to the uploads view I see:
image 
Blank ones probably from another accounts

src/hooks/useConnectionStatus.tsx
```diff
@@ -28,7 +28,7 @@ export const ConnectionStatusProvider: React.FC = ({ children }) => {
   const checkConnection = useCallback(async () => {
     try {
       const res = await withTimeout(
-        fetch('https://google.com', {
+        fetch('https://www.google.com', {
```
drillprop, line 31: Is does make the difference? Just curious

src/views/studio/MyUploadsView/MyUploadsView.tsx
drillprop: Probably something in `useUploadsManager.tsx`

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -222,15 +188,13 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
           raxConfig: {
             retry: RETRIES_COUNT,
             noResponseRetries: RETRIES_COUNT,
+            retryDelay: 1000,
```
drillprop: let's move it to some constant variable, maybe.  Like we did with RETRIES_COUNT

drillprop (commented):
Generally works good - tested it by uploading several videos.  I've experienced once #647 but this time video appeared in uploads(but I still saw warning dialog about closing tab), so I guess this PR will partly fix that. 

Found one issue with MyUploadsView, not sure what it's causing it. I have a lot of uploads saved in localstorage on chrome, so maybe some data are broken somehow, but on the other hand, it works on `dev`. 
Tested it also  on firefox(where I don't have a lot of uploads) and I didn't experience this issue

src/hooks/useConnectionStatus.tsx
kdembler: Request to `https://google.com` will return a redirect to `https://www.google.com` so we kept making 2 requests instead of 1

src/views/studio/MyUploadsView/MyUploadsView.tsx
kdembler: Those most likely are videos created on some different deployment? So for example on prod or other chain and the current chain doesn't know about those?

kdembler: I guess we could try to filter those out

mikkio-j: > Blank ones probably from another accounts

It seem like it's from another accounts. Same thing with the Channel group - there are 2 Channel groups on the screenshot. Normally all channel assets would stack into one Channel group since they would share the same `parentObjectId`. 

For the blank titles, even though assets would come from another channel, they still should be fetched properly?

kdembler: Yep, they should be fetched properly. Must like I said I think it may also be data from some other chain, since we may share local storage between those. Added code that will filter assets only owned by current channel

PR #767, opened 2021-06-11
Fix play/pause loop bug

### LINKED ISSUE #314: Video player enters a play/pause toggle infinite loop

drillprop:
Fix #314 
Tested it on my old laptop, without this change I could reproduce the bug almost every time on one particular video, so I hope it's enough. 🙏🏻


kdembler (commented):
I trust you that it fixes the bug indeed but it also breaks the play/pause on spacebar, that's why we have this piece of code

PR #768, opened 2021-06-11
AssetImage component

### LINKED ISSUE #749: Create single component for displaying all image assets
I think we need to do that to have more control over how we interact with storage providers. If you look at our current implementations, we usually use `getAssetUrl` on every render meaning that this URL can in fact change between renders, fetching the same image twice from different locations. If we create a single component that can be reused for all purposes, we can include any additional logic for storage providers that we would want (for example reporting storage provider not responding, etc)

rafalpawlow:
Closes #749


rafalpawlow:
@kdembler I tried several solutions but ultimately ended up with this one. I didn't do a full implementation, I'd like to know your opinion first so consider it as a draft. Also what I noticed, I'll need your PR #766.

src/shared/components/AssetImage/AssetImage.tsx
```diff
@@ -0,0 +1,56 @@
...
+const AssetImage = ({ entity, component }: AssetImageProps) => {
+  const [assetUrl, setAssetUrl] = useState<string | undefined>(undefined)
+  const { getStorageProvider } = useStorageProviders()
+  const assetData = useMemo(() => {
+    if (entity && entity.__typename === 'Channel') {
+      return {
+        title: entity.title,
+        availability: entity.coverPhotoAvailability || entity.avatarPhotoAvailability,
+        urls: entity.avatarPhotoUrls || entity.coverPhotoUrls,
+        dataObject: entity.coverPhotoDataObject || entity.avatarPhotoDataObject,
```
kdembler: `coverPhoto` and `avatarPhoto` are two separate assets so we need to somehow switch which one we use

src/shared/components/AssetImage/AssetImage.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type AssetImageProps = { entity?: AllChannelFieldsFragment | VideoFieldsFragment | null; component: ReactElement }
+
+const AssetImage = ({ entity, component }: AssetImageProps) => {
+  const [assetUrl, setAssetUrl] = useState<string | undefined>(undefined)
+  const { getStorageProvider } = useStorageProviders()
+  const assetData = useMemo(() => {
+    if (entity && entity.__typename === 'Channel') {
+      return {
+        title: entity.title,
```
kdembler: What do we need the title for?

rafalpawlow: Sorry, previously I tried with `img` element and I used it for `alt`

kdembler: :+1:

src/components/ChannelPreview.tsx
```diff
@@ -13,28 +13,24 @@ type ChannelPreviewProps = {
...
 export const ChannelPreview: React.FC<ChannelPreviewProps> = ({ id, className, onClick }) => {
+  const [{ url }, getAssetUrl] = useAsset()
   const { channel, loading } = useChannel(id ?? '', { fetchPolicy: 'cache-first', skip: !id })
   const { videoCount } = useChannelVideoCount(id ?? '', {
     fetchPolicy: 'cache-first',
     skip: !id,
   })
   const isLoading = loading || id === undefined
-  const { getAssetUrl } = useAsset()
+  getAssetUrl(channel?.avatarPhotoAvailability, channel?.avatarPhotoUrls, channel?.avatarPhotoDataObject)
```
kdembler: Two things here:
1. Do we actually need to call that? I think we could just pass those arguments directly into the hook? 
2. I think we still should use the pattern with `entity` and `type`, so we don't need to pass all the necessary fields explicitly

What do you think?

src/components/ChannelLink/ChannelLink.tsx
```diff
@@ -36,19 +36,22 @@ export const ChannelLink: React.FC<ChannelLinkProps> = ({
...
+  })
 
   const displayedChannel = overrideChannel || channel
 
-  const avatarPhotoUrl = getAssetUrl(
-    displayedChannel?.avatarPhotoAvailability,
-    displayedChannel?.avatarPhotoUrls,
-    displayedChannel?.avatarPhotoDataObject
-  )
+  useEffect(() => {
```
kdembler: Let's move that `useEffect` inside `useAsset`

src/hooks/useAsset/useAsset.tsx
```diff
@@ -0,0 +1,109 @@
+import { useCallback, useEffect, useMemo, useState } from 'react'
+
+import { AllChannelFieldsFragment, AssetAvailability, VideoFieldsFragment } from '@/api/queries'
+import { AssetType, useStorageProviders } from '@/hooks'
+import { createStorageNodeUrl } from '@/utils/asset'
+
+type UseAssetDataArgs =
+  | { entity?: VideoFieldsFragment | null; assetType: AssetType.THUMBNAIL | AssetType.MEDIA }
```
kdembler, line 10: We would normally use typescript string literal instead of an enum in case like this. Something like `assetType: 'thumbnail' | 'media'`. Do you think using enum has some advantage?

rafalpawlow: I don't like comparing strings, In may opinion this is antipattern. First of all, you can use enum as a type so you don't have to list all available strings one by one `assetType: AssetType` vs `assetType: 'thumbnail' | 'avatar' | 'media'`. Second, if needed you can always easily change enum value instead of finding every string in the code. I know that none of the above applies to this case but in my opinion this is just a good practice.

src/hooks/useAsset/useAsset.tsx
```diff
@@ -0,0 +1,116 @@
...
+    if (error) {
+      Logger.error(`Fail to load ${assetType}`)
+    }
+  }, [error, assetType])
+
+  const assetData = useMemo(() => {
+    if (entity && entity.__typename === 'Channel') {
+      return {
+        availability: assetType === AssetType.COVER ? entity.coverPhotoAvailability : entity.avatarPhotoAvailability,
+        urls: assetType === AssetType.COVER ? entity.coverPhotoUrls : entity.coverPhotoUrls,
```
kdembler:
```suggestion
        urls: assetType === AssetType.COVER ? entity.coverPhotoUrls : entity.avatarPhotoUrls,
```

src/hooks/useAsset/useAsset.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export const useAsset: UseAsset = ({ entity, assetType }) => {
+  const { getStorageProvider } = useStorageProviders()
+  const [error, setError] = useState<boolean>(false)
+  const [isLoading, setIsLoading] = useState<boolean | undefined>(undefined)
+  const [url, setUrl] = useState<string | undefined>(undefined)
+
+  useEffect(() => {
+    if (error) {
+      Logger.error(`Fail to load ${assetType}`)
```
kdembler: I know we talked about something else but let's actually store the error (not just a boolean) and log it here as well

src/hooks/useAsset/useAsset.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export const useAsset: UseAsset = ({ entity, assetType }) => {
+  const { getStorageProvider } = useStorageProviders()
+  const [error, setError] = useState<ErrorEvent | null>(null)
+  const [isLoading, setIsLoading] = useState<boolean | undefined>(undefined)
+  const [url, setUrl] = useState<string | undefined>(undefined)
+
+  useEffect(() => {
+    if (error) {
+      Logger.error(error.message)
```
kdembler:
```suggestion
      Logger.error(`Failed to load ${assetType}`, error.message)
```

kdembler (approved):

kdembler (changes requested):
Just noticed during final testing: this breaks changing the active channel in Studio. See:

https://user-images.githubusercontent.com/12646744/122758376-04f61f00-d299-11eb-8fc2-84bb0f88776e.mov

On `dev` avatar properly updates to the selected channel avatar

kdembler (approved):

PR #773, opened 2021-06-14
Check if browser is outdated

### LINKED ISSUE #760: API Feature check supported browsers
We will need a solution to feature check the browser to see if it supports APIs like ResizeObservers, IntersectionObserver... that we are using.

rafalpawlow:
Closes #760

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,14 @@ import { ApolloProvider } from '@apollo/client'
 import React from 'react'
 
 import { createApolloClient } from '@/api'
-import { ConnectionStatusProvider, OverlayManagerProvider, SnackbarProvider, StorageProvidersProvider } from '@/hooks'
+import {
+  useDialog,
+  DialogProvider,
+  ConnectionStatusProvider,
+  OverlayManagerProvider,
+  SnackbarProvider,
+  StorageProvidersProvider,
+} from '@/hooks'
 
 import MainLayout from './MainLayout'
 
@@ -17,7 +24,9 @@ export default function App() {
         <ConnectionStatusProvider>
           <OverlayManagerProvider>
             <StorageProvidersProvider>
-              <MainLayout />
+              <DialogProvider>
+                <MainLayout />
+              </DialogProvider>
             </StorageProvidersProvider>
           </OverlayManagerProvider>
         </ConnectionStatusProvider>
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -2,7 +2,14 @@ import { ApolloProvider } from '@apollo/client'
 import React from 'react'
 
 import { createApolloClient } from '@/api'
-import { ConnectionStatusProvider, OverlayManagerProvider, SnackbarProvider, StorageProvidersProvider } from '@/hooks'
+import {
+  useDialog,
```
kdembler: `useDialog` doesn't seem needed here?

src/MainLayout.tsx
```diff
@@ -21,6 +23,21 @@ const LoadableStudioLayout = loadable(() => import('./views/studio/StudioLayout'
 })
 
 const MainLayout: React.FC = () => {
+  const [openDialog, closeDialog] = useDialog({
+    title: 'Outdated browser detected',
+    description: 'It seems your browser is outdated, the Joystream app may not work properly',
```
kdembler: To be in line with similar dialog we already have:

```suggestion
    description: 'It seems the browser version you are using is not fully supported by Joystream. Some of the features may be broken or not accessible. For the best experience, please upgrade your browser to the latest version.
```

kdembler: I think it would also be nice to include a link with some instructions. For example https://www.whatismybrowser.com/guides/how-to-update-your-browser/auto will automatically show you instructions for your browser. That could be `Click here to see instructions.` at the end. Wdyt?

src/App.tsx
rafalpawlow: My bad but I think we should have eslint rule for that. WDYT?

rafalpawlow: I've noticed that it's turned off in `.eslintrc`, what is the reason for that?

kdembler: We most certainly should have a rule for that! It's turned off because of some old dependencies we're locked in with because of shared `eslint-config`. We should get rid of that constraint soon

src/MainLayout.tsx
eldiegod: Second this

eldiegod (approved):
Tested on chrome and firefox and functionality wise works great and shows the dialog when the apis are not available

kdembler (approved):

PR #779, opened 2021-06-15
Split prettier and eslint

### LINKED ISSUE #732: Split Prettier and ESLint
Our current setup uses Prettier plugin into ESLint so we can get style linting that way. That's a legacy approach however and Prettier docs suggest just using prettier directly for that. All IDEs now support Prettier so we probably should just split Prettier and ESLint linting. DM me if more details needed

drillprop:
Fix #732


.eslintrc.js
```diff
@@ -5,13 +5,38 @@ module.exports = {
...
-  extends: ['plugin:react-hooks/recommended', '@joystream/eslint-config'],
-  plugins: ['@emotion'],
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true,
+    },
+    ecmaVersion: 2019,
+    sourceType: 'module',
+  },
```
kdembler: Not sure if we need that section

__generated__/globalTypes.ts
```diff
@@ -1,17 +0,0 @@
...
-//==============================================================
-
-export enum VideoOrderByInput {
-  createdAt_ASC = "createdAt_ASC",
-  createdAt_DESC = "createdAt_DESC",
-}
-
-//==============================================================
-// END Enums and Input Objects
-//==============================================================
```
drillprop, line 17: I allowed myself to remove this file - we don't use it anymore

.eslintrc.js
drillprop: removed

package.json
```diff
@@ -21,8 +21,9 @@
     "build": "react-app-rewired build",
     "eject": "react-app-rewired eject",
     "lint:ts": "eslint --ext .js,.jsx,.ts,.tsx .",
+    "lint:prettier": "prettier --write .",
```
kdembler: This shouldn't use `--write` as that will automatically fix the errors. Purpose of this script is to check whether there are any errors

kdembler: We probably should use `--check`

drillprop: Did that in the hurry, just copied `lint-staged` - my mistake. Should be good now

kdembler (approved):

Issue #788, opened 2021-06-16
Split providers & hooks


kdembler:
Let's split current `hooks/` folder into `providers/` and `hooks/`. I think in `providers/` we should keep all the context providers (and hooks for access to them) and move everything else inside `hooks/`. We could also use some better naming for providers. For example `hooks/useSnackbar` would become `providers/snackbars`

PR #793, opened 2021-06-16
 Split providers & hooks 

eldiegod:
#788


eldiegod:
Not sure I got this 100% as we wanted it to, u let me know if I missed something @kdembler

kdembler (commented):
I think we can improve upon this a bit. My ideas:
1. Let's clean up the structure of `providers` a bit more:
  a) if the provider occupies a single file, let's make sure it doesn't have its own directory (we have some directories with just `index.ts` and the file)
  b) if the single file would be too long/too hard to read, let's split those into `provider.ts` `hook.ts` `types.ts`, etc, based on whatever is needed for this specific provider
  c) let's name providers accordingly to their purpose and not the hook name, for example `providers/useDialog` -> `providers/dialogs`
2. Let's keep `index.ts` inside `providers/`
3. Let's not export provider hooks from `hooks/`. `hooks/` should only export non-provider hooks (the ones defined in `hooks/`)

What do you think about those?

eldiegod:
Yep, made the fixes from the comment as I saw fit pretty much

eldiegod:
@kdembler lets try to merge this one soonish before it gets another million conflicts

kdembler (approved):

Issue #532, opened 2021-04-30
Make the uploads view handle unknown assets


kdembler:
It's possible for the user to start (but not finish) upload of some asset on different device than they're using currently to access the app. We should be able to pick up assets that haven't been uploaded, even if they aren't present in the local uploads state. Then we can use the data from the query node as the source of truth and only use local data as an additional flavour on top of it (for example upload progress, etc).

kdembler:
I've already done some work on this but haven't managed to finish. This is what I had in `useUploadsManager`:

```ts
  useEffect(() => {
    if (!activeChannelId) {
      return
    }

    const init = async () => {
      const videosWithPendingMediaPromise = client.query<GetVideosQuery, GetVideosQueryVariables>({
        query: GetVideosDocument,
        variables: { where: { mediaAvailability_eq: AssetAvailability.Pending, channelId_eq: activeChannelId } },
      })

      const videosWithPendingThumbnailPromise = client.query<GetVideosQuery, GetVideosQueryVariables>({
        query: GetVideosDocument,
        variables: {
          where: { thumbnailPhotoAvailability_eq: AssetAvailability.Pending, channelId_eq: activeChannelId },
        },
      })

      const [videosWithPendingMediaResponse, videosWithPendingThumbnailResponse] = await Promise.all([
        videosWithPendingMediaPromise,
        videosWithPendingThumbnailPromise,
      ])

      const videosWithPendingMedia = videosWithPendingMediaResponse.data.videos
      const videosWithPendingThumbnail = videosWithPendingThumbnailResponse.data.videos

      console.log({ videosWithPendingMedia, videosWithPendingThumbnail })

      const videosWithPendingMediaLookup = createLookup(videosWithPendingMedia || [])
      const videosWithPendingThumbnailLookup = createLookup(videosWithPendingThumbnail || [])
      // to remove any duplicates
      const videosLookup = {
        ...videosWithPendingMediaLookup,
        ...videosWithPendingThumbnailLookup,
      }
    }

    init()
  }, [activeChannelId, client])
  ```
  
I'm also happy to help if needed

PR #797, opened 2021-06-17
Handle unknown assets in Uploads

mikkio-j:
Issue: #532

### DIFF
--- a/src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
+++ b/src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
@@ -5,7 +5,8 @@ import { CSSTransition } from 'react-transition-group'
 import { NavItemType, SidenavBase } from '@/components/Sidenav/SidenavBase'
 import { absoluteRoutes } from '@/config/routes'
 import { useDisplayDataLostWarning } from '@/hooks'
-import { useAuthorizedUser, useDrafts, useEditVideoSheet, useUploadsManager } from '@/providers'
+import { useAuthorizedUser, useDrafts, useEditVideoSheet } from '@/providers'
+import { useUploadsStore } from '@/providers/uploadsManager/store'
 import { Button } from '@/shared/components'
 import { SvgGlyphAddVideo, SvgGlyphExternal, SvgNavChannel, SvgNavUpload, SvgNavVideos } from '@/shared/icons'
 import { transitions } from '@/shared/theme'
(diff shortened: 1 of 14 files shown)


src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -47,7 +57,7 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
   const [assetsFiles, setAssetsFiles] = useState<AssetFile[]>([])
   const { activeChannelId } = useUser()
   const { loading: channelLoading } = useChannel(activeChannelId ?? '')
-  const { loading: videosLoading } = useVideos(
+  const { loading: videosLoading, client } = useVideos(
```
kdembler: Why do we still need that fetch?

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -69,34 +81,154 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
...
+
+        const isMediaPresentInUploads = uploadsState.some(
+          (asset) => asset.contentId === video.mediaDataObject?.joystreamContentId
+        )
+        const isThumbnailPresentInUploads = uploadsState.some(
+          (asset) => asset.contentId === video.thumbnailPhotoDataObject?.joystreamContentId
+        )
+
+        if (isMediaPending && !isMediaPresentInUploads) {
+          addAsset({
```
kdembler: I'm wondering, what's the point of adding that all to the local storage?

The approach we take here currently is that we use local storage as the source of truth - that's what we will be using to display the data later on. And we do a sync at the beginning to make it updated. But what if we took a different approach and decided that the data from the query node is the source of truth and whatever we have in the local storage can be an addition to that?

kdembler: Does that make sense? I think in that case we wouldn't need to add those assets to the state at all

src/hooks/useUploadsManager/useUploadsManager.tsx
```diff
@@ -22,62 +31,195 @@ export const UploadManagerProvider: React.FC = ({ children }) => {
...
-  const { loading: channelLoading } = useChannel(activeChannelId ?? '')
-  const { loading: videosLoading } = useVideos(
-    {
-      where: {
-        id_in: uploadsStateIds,
-      },
-    },
-    { skip: !uploadsStateIds.length }
-  )
+  const [queryNodePendingAssets, setQueryNodePendingAssets] = useState<AssetUpload[]>([])
```
kdembler: Why not make this part of the store?

mikkio-j: I thought we don't want to make query node data part of the store. Can make that though - that will require some additional usage of `updateAsset` on initial sync up

kdembler: Ah okay I misunderstood what this is. Let me think about this

kdembler (approved):

Issue #792, opened 2021-06-16
Add some tool to monitor bundle size


kdembler:
Let's plug in something like `webpack-bundle-analyzer`

PR #799, opened 2021-06-17
Initial bundle visualization setup

eldiegod:
#792

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -18,7 +18,7 @@ module.exports = {
     addWebpackPlugin(
       new BundleAnalyzerPlugin({
         // use value from environment var STATS or 'disabled'
-        analyzerMode: process.env.STATS || 'disabled',
+        analyzerMode: process.env.STATS || 'static',
         generateStatsFile: true,
       })
     ),
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const path = require('path')
 const StylelintPlugin = require('stylelint-webpack-plugin')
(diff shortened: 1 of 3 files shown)


config-overrides.js
```diff
@@ -14,6 +15,13 @@ module.exports = {
   webpack: override(
     addBabelPlugin('@emotion/babel-plugin'),
     addWebpackPlugin(new StylelintPlugin({ files: './src/**/*.{tsx,ts}' })),
+    addWebpackPlugin(
+      new BundleAnalyzerPlugin({
+        // use value from environment var STATS or 'disabled'
+        analyzerMode: process.env.STATS || 'disabled',
```
kdembler: What would possible values for `STATS` be?

package.json
```diff
@@ -16,6 +16,7 @@
     "url": "https://github.com/Joystream/joystream/issues"
   },
   "scripts": {
+    "bundle-stats": "STATS=server npm run build",
```
kdembler, line 19: Does stats generation take some considerable time? I'm wondering whether maybe we should just enable it by default with a build?

eldiegod: I don't think it adds much more time but this opens a server and a tab on the browser, the build does make a file by default called stats.json which can be used as input on https://chrisbateman.github.io/webpack-visualizer/
or to open the the bundle analizer from the cli

config-overrides.js
eldiegod: These
> Default: server. In server mode analyzer will start HTTP server to show bundle report. In static mode single HTML file with bundle report will be generated. In json mode single JSON file with bundle report will be generated. In disabled mode you can use this plugin to just generate Webpack Stats JSON file by setting generateStatsFile to true.

kdembler (approved):

Issue #791, opened 2021-06-16
Remove usages of `css` prop


kdembler:
There are just few places in the app in which we use `css` prop. I think it doesn't bring us much value but it requires some special adjustments to the build process. My guess is it should be simple to get rid of those places. Let's try to do that

Just to be sure, this is not referring to places where we define some styles for later use.

**This is fine:**
```ts
const commonStyles = css`
  color: red;
`

const StyledComponent = styled.span`
  ${commonStyles};
`
```

**Let's remove that:**
```tsx
return (
  <div css={styles} />
)
```

PR #801, opened 2021-06-18
Remove css prop usage

mikkio-j:
Issue: #791


kdembler (commented):
Still couple more left

src/shared/components/ContextMenu/ContextMenu.style.ts
```diff
@@ -19,28 +19,7 @@ export const StyledContainer = styled.div<ContainerProps>`
   color: ${colors.white};
   padding: ${sizes(2)};
   word-break: break-all;
-`
```
kdembler, line 22: Cannot add a comment where I'd like to but there is unneeded import at the top of the file

src/shared/components/Text/Text.style.ts
```diff
@@ -14,126 +18,142 @@ export const secondaryTextStyles = css`
...
 export const styledVariants = {
-  hero: styled.h1`
+  hero: styled.h1<TextProps>`
     ${baseStyles};
 
     font-size: ${typography.sizes.hero};
     line-height: ${typography.lineHeights.hero};
     font-weight: ${typography.weights.bold};
     font-family: ${typography.fonts.headers};
+    ${({ isSecondary }) => isSecondary && secondaryTextStyles}
```
kdembler: I think this could be inside `baseStyles`?

kdembler (approved):

PR #804, opened 2021-06-18
add changing environment on the fly functionality

### LINKED ISSUE #658: Allow changing target environment on the fly
Dev mode needs some work (currently we are changing env variables manually, we want to go for dynamic mocking, staging etc.)

drillprop:
Fix #658

Not sure if it's enough - I didn't mess with the netlify plugin at all.


.env
```diff
@@ -1,16 +1,17 @@
 # This file is commited. Do not store secrets here
```
kdembler, line 1: Let's use the following scheme for _all_ the env variables: `REACT_APP_${ENV}_${NAME}`, so for example:
- `REACT_APP_PRODUCTION_QUERY_NODE_URL`
- `REACT_APP_STAGING_QUERY_NODE_URL`

src/config/urls.ts
```diff
@@ -1,4 +1,12 @@
+export const ENV_DEVELOPMENT = 'DEVELOPMENT'
+export const ENV_PRODUCTION = 'PRODUCTION'
+export const ENV_STAGING = 'STAGING'
```
kdembler: Let's not use predefined environments. Let's do a discovery based on available variables. So at the startup let's go through `process.env` and find all the prefixes. That should produce a list that for the time being will be `['production', 'staging']`

src/config/urls.ts
```diff
@@ -1,4 +1,12 @@
...
+
 const readEnv = (name: string): string | undefined => {
+  const env = window.localStorage.getItem('env')
```
kdembler: Let's move this and `process.env.REACT_APP_ENV` to some separate file where we clearly define what each environment means. It can be in `src/config/envs.ts`:
```ts
export const BUILD_ENV = process.env.REACT_APP_ENV || 'production'
export const TARGET_DEV_ENV = window.localStorage.getItem('target_env') || 'staging'
```

src/index.tsx
```diff
@@ -6,14 +6,14 @@ import ReactDOM from 'react-dom'
...
-import { SENTRY_DSN } from './config/urls'
+import { ENV_DEVELOPMENT, SENTRY_DSN } from './config/urls'
 
 type Env = 'production' | 'staging' | 'development'
 
 const env = (process.env.REACT_APP_ENV?.toLowerCase() || 'development') as Env
 
 const initApp = async () => {
-  if (env === 'development') {
+  if ((env === 'development' || env === 'staging') && window.localStorage.getItem('env') === ENV_DEVELOPMENT) {
```
kdembler: Let's call this environment `mocking` instead. Given the above points the condition for enabling it would be `if (BUILD_ENV !== 'production' && TARGET_DEV_ENV === 'mocking')`

src/config/urls.ts
```diff
@@ -1,4 +1,12 @@
...
+  if (env === ENV_PRODUCTION) {
+    return process.env[`${name}_${ENV_PRODUCTION}`]
+  }
   return process.env[name]
```
kdembler: Let's move prefix in here and make this function something along those lines:
```ts
const ENV_PREFIX = 'REACT_APP'

const readEnv = (name: string): string | undefined => {
  if (BUILD_ENV === 'production') {
    const fullName = `${ENV_PREFIX}_PRODUCTION_${name}`
    return process.env[fullName]
  }
  if (TARGET_DEV_ENV === 'mocking') {
    return `/${name}`
  }
  const fullName = `${ENV_PREFIX}_${TARGET_DEV_ENV.toUpperCase()}_${name}`
  return process.env[fullName]
}
```

kdembler (changes requested):

.env
```diff
@@ -1,16 +1,25 @@
...
+REACT_APP_ENV=development
 
-#REACT_APP_ENV=production
-#REACT_APP_QUERY_NODE_URL=https://hydra.joystream.org/graphql
-#REACT_APP_QUERY_NODE_SUBSCRIPTION_URL=wss://hydra.joystream.org/graphql
-#REACT_APP_ORION_URL=https://orion.joystream.org/graphql
-#REACT_APP_NODE_URL=wss://rome-rpc-endpoint.joystream.org:9944/
-#REACT_APP_FAUCET_URL=https://member-faucet.joystream.org/register
+# default is staging
+REACT_APP_STAGING_QUERY_NODE_URL=https://sumer-dev-2.joystream.app/query/server/graphql
```
kdembler: On a second thought, maybe let's call this env `development`? Staging would be actually a separate one, like we used `reckless` for staging before `sumer`.

.env
```diff
@@ -1,16 +1,25 @@
...
+REACT_APP_STAGING_QUERY_NODE_SUBSCRIPTION_URL=wss://sumer-dev-2.joystream.app/query/server/graphql
+REACT_APP_STAGING_ORION_URL=https://orion-staging.joystream.app/graphql
+REACT_APP_STAGING_NODE_URL=wss://sumer-dev-2.joystream.app/rpc
+REACT_APP_STAGING_FAUCET_URL=https://sumer-dev-2.joystream.app/members/register
+
+# these two url are only temporary. We don't want to break the app when switching to 'mocking'
```
kdembler: I don't understand this comment, why do we need valid values for query node urls?

.env
```diff
@@ -1,16 +1,25 @@
...
+REACT_APP_MOCKING_QUERY_NODE_URL=https://sumer-dev-2.joystream.app/query/server/graphql
+REACT_APP_MOCKING_QUERY_NODE_SUBSCRIPTION_URL=wss://sumer-dev-2.joystream.app/query/server/graphql
+
+REACT_APP_MOCKING_ORION_URL=/mocked-orion
+REACT_APP_MOCKING_NODE_URL=ws://127.0.0.1:9944
+REACT_APP_MOCKING_FAUCET_URL=/mocked-faucet
```
kdembler: Let's move those to the bottom

kdembler (commented):
Looking good!

kdembler:
And one more thing, I think we should delete the Netlify plugin now. So `/netlify-plugins` directory and mentions in `/netlify.toml`

kdembler (approved):

Issue #318, opened 2021-02-12
Video plays when typing on the searh bar


eldiegod:
Fairly self-explanatory, if I click on a video and then start typing on the search bar the video player will still capture all the keystrokes causing the typing to play/modify the state of the video player which is not intended.

PR #819, opened 2021-06-21
Fix video playing when typing on the searh bar

eldiegod:
#318

### DIFF
--- a/src/views/viewer/VideoView/VideoView.tsx
+++ b/src/views/viewer/VideoView/VideoView.tsx
@@ -4,7 +4,7 @@ import { useMatch, useParams } from 'react-router-dom'
 
 import { useAddVideoView, useVideo } from '@/api/hooks'
 import { ChannelLink, InfiniteVideoGrid } from '@/components'
-import { absoluteRoutes, QUERY_PARAMS } from '@/config/routes'
+import { absoluteRoutes } from '@/config/routes'
 import knownLicenses from '@/data/knownLicenses.json'
 import { useAsset, usePersonalData, useRouterQuery } from '@/hooks'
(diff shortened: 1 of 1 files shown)


src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -34,6 +35,7 @@ export const VideoView: React.FC = () => {
   const { state, updateWatchedVideos } = usePersonalData()
   const timestampFromQuery = Number(useRouterQuery('time'))
   const { getAssetUrl } = useAsset()
+  const searchRouteMatch = !!useMatch({ path: absoluteRoutes.viewer.search() })
```
kdembler: Maybe it'd be safer to have the condition the other way around? So to check whether we are on `VideoView`? Then if we add another overlay, this won't break

eldiegod: good point!

src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -34,7 +35,7 @@ export const VideoView: React.FC = () => {
   const { state, updateWatchedVideos } = usePersonalData()
   const timestampFromQuery = Number(useRouterQuery('time'))
   const { getAssetUrl } = useAsset()
-
+  const searchRouteMatch = useMatch({ path: absoluteRoutes.viewer.video(id) })
```
kdembler: I think we may need to adjust the name of the variable now, sorry :P

eldiegod: ah fk didnt see

kdembler (approved):

PR #820, opened 2021-06-21
Fix scroll issues with overlay

### LINKED ISSUE #812: Error dialog in edit video sheet unblocks scroll
Once you open the video sheet, we lock the body scroll. However, if you get an error dialog during publishing, after closing it, the scroll will unlock. That shouldn't happen because there is still an overlay open

drillprop:
Fix #812

### DIFF
--- a/src/components/SignInSteps/SignInStepsStepper.tsx
+++ b/src/components/SignInSteps/SignInStepsStepper.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import { useNavigate } from 'react-router'
 
 import { useRouterQuery } from '@/hooks'
@@ -12,15 +12,6 @@ import { Multistepper } from '../Dialogs'
 export const SignInStepsStepper: React.FC = () => {
   const navigate = useNavigate()
   const step = Number(useRouterQuery('step'))
-  const [isStepperOpen, setIsStepperOpen] = useState(false)
-
-  useEffect(() => {
-    if (step >= 1 || step > 3) {
-      setIsStepperOpen(true)
-    } else {
-      setIsStepperOpen(false)
-    }
-  }, [step])
 
   const steps = [
     {
@@ -41,7 +32,7 @@ export const SignInStepsStepper: React.FC = () => {
     <Multistepper
       currentStepIdx={step <= 0 ? 0 : step - 1}
       steps={steps}
-      showDialog={isStepperOpen}
+      showDialog={step >= 1}
       onExitClick={() => navigate('?step=0')}
     />
   )
--- a/src/components/SignInSteps/SignInStepsStepper.tsx
+++ b/src/components/SignInSteps/SignInStepsStepper.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { useNavigate } from 'react-router'
 
 import { useRouterQuery } from '@/hooks'
@@ -12,6 +12,15 @@ import { Multistepper } from '../Dialogs'
 export const SignInStepsStepper: React.FC = () => {
   const navigate = useNavigate()
   const step = Number(useRouterQuery('step'))
+  const [isStepperOpen, setIsStepperOpen] = useState(false)
+
+  useEffect(() => {
+    if (step >= 1 || step > 3) {
+      setIsStepperOpen(true)
+    } else {
+      setIsStepperOpen(false)
+    }
+  }, [step])
 
   const steps = [
     {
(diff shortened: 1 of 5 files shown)


src/components/SignInSteps/SignInStepsStepper.tsx
```diff
@@ -12,6 +12,15 @@ import { Multistepper } from '../Dialogs'
...
   const step = Number(useRouterQuery('step'))
+  const [isStepperOpen, setIsStepperOpen] = useState(false)
+
+  useEffect(() => {
+    if (step >= 1 || step > 3) {
+      setIsStepperOpen(true)
+    } else {
+      setIsStepperOpen(false)
+    }
+  }, [step])
```
kdembler: What's the purpose of that?

src/components/SignInSteps/SignInStepsStepper.tsx
```diff
@@ -12,6 +12,15 @@ import { Multistepper } from '../Dialogs'
 export const SignInStepsStepper: React.FC = () => {
   const navigate = useNavigate()
   const step = Number(useRouterQuery('step'))
+  const [isStepperOpen, setIsStepperOpen] = useState(false)
+
+  useEffect(() => {
+    if (step >= 1 || step > 3) {
```
kdembler: This condition is equivalent to just `step >= 1`

src/providers/overlayManager.tsx
```diff
@@ -72,7 +72,7 @@ export const useOverlayManager = () => {
   const incrementOverlaysOpenCount = useCallback(() => setOverlaysOpenCount((count) => count + 1), [
     setOverlaysOpenCount,
   ])
-  const decrementOverlaysOpenCount = useCallback(() => setOverlaysOpenCount((count) => (count > 0 ? count - 1 : 0)), [
+  const decrementOverlaysOpenCount = useCallback(() => setOverlaysOpenCount((count) => count - 1), [
```
kdembler: Why did we change this condition? It makes sense to not allow this to go negative right?

src/components/SignInSteps/TermsStep.tsx
```diff
@@ -53,7 +55,11 @@ export const TermsStep: React.FC = () => {
         </TermsOverlay>
       </TermsBox>
       <StepFooter>
-        <ContinueButton to={absoluteRoutes.studio.newMembership()} disabled={!hasScrolledToBottom}>
+        <ContinueButton
+          onClick={decrementOverlaysOpenCount}
```
kdembler: I don't like this very much, it leaks overlay implementation into a view that doesn't have anything to do with it. What if instead we made sure the `BaseDialog` will call this whatever happens?

One idea I have in mind is we could try the to keep the cleanup `useEffect` we have there but only make the call if the transition haven't fired? That should keep us safe, what do you think?

src/components/SignInSteps/SignInStepsStepper.tsx
drillprop: Without this `useEffect` if you just type in a browser `/studio/signin?step=1` the scroll will be not blocked, because `onEnter` will be not called.

src/providers/overlayManager.tsx
drillprop: oh, yeah - I set that only for testing purposes. I wanted to track how often decrement is called. I will get this condition back

src/components/SignInSteps/TermsStep.tsx
drillprop: > One idea I have in mind is we could try the to keep the cleanup `useEffect` we have there but only make the call if the transition haven't fired? That should keep us safe, what do you think?

Tried that earlier with no success. Do you have an idea how we could implement that? 

I've tried 
```js
// BaseDialog.tsx
const isOnExitCalled = useRef(false)
// ...
useEffect(() => {
    if (showDialog && !isOnExitCalled.current) {
      return () => {
        decrementOverlaysOpenCount()
      }
    }
  }, [decrementOverlaysOpenCount, showDialog])
//... 

      <CSSTransition
        in={showDialog}
        timeout={200}
        classNames={transitions.names.dialog}
        mountOnEnter
        unmountOnExit
        onExit={() => {
          isOnExitCalled.current = true
        }}
        onExited={() => {
          if (isOnExitCalled.current) {
            decrementOverlaysOpenCount()
          }
        }}
      >
```
and something analogous with using `useState`. If I close the dialog by clicking the `close button`, the decrementOverlaysOpenCount is called two times(from `onExited` and from the cleanup function).

src/components/SignInSteps/SignInStepsStepper.tsx
drillprop: Oh, actually this won't be needed, I'll remove that

drillprop:
Ok, I think I covered all of the scroll lock problems:
- when a user gets an error during publishing ✅ 
- when a user goes to `/signin?step=1`(paste the address in the browser) ✅ 
- when a user accept terms ✅ 
- when a user creates a membership ✅ 
- when a user opens two dialogs at once ✅ 
- all of the overlay logic is inside `BaseDialog` ✅

drillprop:
Draft for now. Will try one more approach.

kdembler (approved):
Glad we figured this out

PR #836, opened 2021-06-22
introduce common store utils

kdembler:
Looking for feedback!

### DIFF
--- a/src/views/playground/Playgrounds/PlaygroundCommonStore.tsx
+++ b/src/views/playground/Playgrounds/PlaygroundCommonStore.tsx
@@ -1,0 +1,66 @@
+import React, { useState } from 'react'
+
+import { createStore } from '@/store'
+
+type TestStoreState = {
+  newCount: number
+  test: number
+}
+
+type TestStoreActions = {
+  increment: (by: number) => void
+  decrement: (by: number) => void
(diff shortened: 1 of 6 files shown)


src/views/playground/Playgrounds/PlaygroundCommonStore.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type TestStoreActions = {
+  increment: (by: number) => void
+  decrement: (by: number) => void
+}
+
+const useStore = createStore<TestStoreState, TestStoreActions>(
+  {
+    state: { newCount: 0, test: 1 },
+    actionsFactory: (set) => ({
```
eldiegod, line 18: Why not just `actions`?

eldiegod (commented):
Looks really nice, love me some immer

src/views/playground/Playgrounds/PlaygroundCommonStore.tsx
kdembler: Wanted to make it clear that this isn't like `state` (just an object with properties being actions), but rather a function that given a `(set, get)` will return an object

drillprop (approved):

PR #838, opened 2021-06-22
Introduce cypress configuration with simple example test of sidebar

rafalpawlow:


cypress/support/index.js
```diff
@@ -0,0 +1,47 @@
...
+        get: () => new Promise((resolve) => resolve(Cypress.env('accounts'))),
+      },
+    }
+  }
+
+  window['injectedWeb3'] = {
+    'polkadot-js': {
+      enable,
+      version: '0.38.3',
+    },
```
kdembler, line 41: Why aren't we using `@polkadot-js/extension-inject`?

kdembler (commented):
Like mentioned, we should also include that in the CI

eldiegod:
Nice, we can finally start adding some e2e tests!

Let's make sure to add cypress-testing-library, which should make our life quite a bit easier.
https://testing-library.com/docs/cypress-testing-library/intro/

kdembler (approved):

PR #840, opened 2021-06-22
Fix useAsset for switching videos between uploaded and draft

rafalpawlow:


drillprop:
So after these changes when you click a video in MyVideos view to edit it, you'll not see any thumbnail on MultifileSelect. Thumbnail will just appear for a split second only when you click the `close sheet` button.  It works like this right now: 
bugbugbug

kdembler:
I think it happens also on `dev` so without this specific change. But yes, we should fix the above bug as well

drillprop:
> I think it happens also on `dev` so without this specific change. But yes, we should fix the above bug as well

with slightly different result - on `dev` you'll see the thumbnail, but it will be the image that belongs to the video which you previously clicked(if you didn't click any video before, you'll see nothing, of course)

src/providers/editVideoSheet/hooks.tsx
```diff
@@ -83,7 +83,7 @@ export const useEditVideoSheetTabData = (tab?: EditVideoSheetTab) => {
 
   return {
     tabData: normalizedData,
-    loading: tab.isDraft ? false : loading,
+    loading: tab.isDraft ? false : loading || !(thumbnailPhotoUrl && mediaUrl),
```
drillprop: I'm still getting weird results when I'm opening videos. 
For example: 
copy

I think this suggestion should resolve the problem
```suggestion
    loading: tab.isDraft ? false : loading || (!thumbnailPhotoUrl && !mediaUrl)
```
Could you confirm that? It works for me at least

rafalpawlow: This is super weird, for me it works well in both versions + those conditions are equal. But ok, I'll fix it anyway.

drillprop: I don't think they are equal
for example
```js
!(false && true) // this will return true
(!false && !true) // this will return false
```

rafalpawlow: Yes, sorry, you are right!

kdembler:
Superseded by #848

PR #843, opened 2021-06-23
Draft store

### LINKED ISSUE #833: Use zustand for drafts

eldiegod:
Use zustand for drafts #833


src/providers/drafts.ts
```diff
@@ -0,0 +1,118 @@
...
+  removeDrafts: (draftIds: string[]) => void
+  removeAllDrafts: (channelId: string) => void
+  removeAllUnseenDrafts: (channelId: string) => void
+}
+
+export const useDraftStore = createStore<DraftStoreState, DraftStoreActions>(
+  {
+    state: {
+      allDrafts: [], // includes drafts for different channels
+      allUnseenDrafts: [], // includes unseenDrafts for different channels
```
kdembler: I'm wondering, do we need to keep those two separate? Maybe some `seen: boolean` field on draft would do the trick? I remember there was some discussion like this originally, but don't remember what the arguments were

src/providers/drafts.ts
```diff
@@ -0,0 +1,118 @@
...
+    },
+    actionsFactory: (set, get) => ({
+      addDraft: (draft, explicitId) => {
+        const id = explicitId ?? createId()
+        const updatedAt = new Date().toISOString()
+        const newDraft: Draft = { ...draft, updatedAt, id }
+        set((draft) => {
```
kdembler: Let's use something else than `draft` here because it will cause a confusion between different types of `draft`. I suggest `state`. This applies to other actions as well

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -53,11 +54,18 @@ export const MyVideosView = () => {
...
+    const unseenDrafts = actions.getUnseenDraftsForChannel(activeChannelId)
+    return {
+      drafts:
+        sortVideosBy === VideoOrderByInput.CreatedAtAsc
+          ? draft.slice()?.sort((a, b) => new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime())
+          : draft.slice()?.sort((a, b) => new Date(a.updatedAt).getTime() - new Date(b.updatedAt).getTime()),
+      unseenDrafts,
+      ...actions,
+    }
+  }, shallow)
```
kdembler: `shallow` comparison won't do the trick here since `drafts` will be a new array reference on each run so this will never be equal

src/providers/drafts.ts
eldiegod: Yeah, before it was cumbersome but I don't see why keep them separate now, will merge

src/providers/drafts.ts
eldiegod: what about draftState?

src/providers/drafts.ts
kdembler: Great

src/providers/drafts.ts
```diff
@@ -0,0 +1,140 @@
...
+          }))
+        })
+      },
+    }),
+  },
+  {
+    persist: {
+      key: 'drafts',
+      whitelist: ['allDrafts'],
+      version: 2,
```
kdembler: We only had version `1` for this PR right? Since this wasn't ever deployed, I don't think we need to increment to `2` and migrate

src/providers/drafts.ts
```diff
@@ -0,0 +1,140 @@
...
+      migrate: (oldState, oldVersion, storageValue) => {
+        // migrate store before zustand was added
+        // console.log({ oldState, oldVersion, storageValue })
```
kdembler: For removal

src/providers/drafts.ts
```diff
@@ -0,0 +1,140 @@
...
+            unseenDrafts?.find((unseen) => unseen.draftId === draft.id)
+              ? { ...draft, seen: false }
+              : { ...draft, seen: true }
+          })
+          return {
+            allDrafts: drafts,
+          }
+          // combine unseen draft and drafts
+        } else if (oldVersion === 1) {
+          // console.log({ oldState, oldVersion, storageValue })
```
kdembler: Likewise

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -1,12 +1,13 @@
 import React, { useEffect, useRef, useState } from 'react'
 import { useNavigate } from 'react-router-dom'
+import shallow from 'zustand/shallow'
```
kdembler: Unused import

src/providers/drafts.ts
```diff
@@ -0,0 +1,140 @@
...
+  allDrafts: Draft[]
+  allUnseenDrafts: Array<{
+    draftId: string
+    channelId: string
+  }>
+}
+interface DraftStoreStateV2 {
+  allDrafts: Draft[]
+}
+interface DraftStoreActions {
```
kdembler: Let's use `type =`

src/providers/drafts.ts
```diff
@@ -0,0 +1,140 @@
...
+          if (idx >= 0) {
+            draftState.allDrafts[idx] = { ...draftState.allDrafts[idx], ...draftProps, updatedAt }
+          }
+        })
+      },
+      getDraft: (id) => get().allDrafts.find((draft) => draft.id === id),
+      getDraftsForChannel: (channelId) => get().allDrafts.filter((d) => d.channelId === channelId),
+      getUnseenDraftsForChannel: (channelId) =>
+        get().allDrafts.filter((d) => d.channelId === channelId && d.seen === false),
+      removeDrafts: (draftIds) => {
```
kdembler, line 69: Why did we change to remove an array of drafts? I think we always use this to delete only one draft?

src/providers/drafts.ts
eldiegod: True !!

src/providers/drafts.ts
eldiegod: Before we would accept a draftId or an array of draftIds and we had to maintain both scenarios and check if draftIds was an array or not; so I thought this way we only had to maintain one implementation and it wouldn't affect usage besides enclosing draftId in an array if removing only one.

Right now, there is not much use case for multiple removals so use case unclear, but before there was when we had the checkboxes, and possibly in the future there will be (?) seems worth having the option to remove multiple drafts without iterating an array but 🤷‍♂

src/providers/drafts.ts
```diff
@@ -0,0 +1,122 @@
...
+            Array<{
+              draftId: string
+              channelId: string
+            }>
+          >('unseenDrafts')
+          const drafts = [...(storageValue as Array<Draft>)].map((draft) => {
+            unseenDrafts?.find((unseen) => unseen.draftId === draft.id)
+              ? { ...draft, seen: false }
+              : { ...draft, seen: true }
+          })
```
drillprop, line 107: migrate doesn't work great for me. After I switched to your branch I got: 
image 
And my drafts looks like this: 
image

src/providers/drafts.ts
```diff
@@ -0,0 +1,122 @@
...
+    state: {
+      allDrafts: [], // includes drafts for different channels
+    },
+    actionsFactory: (set, get) => ({
+      addDraft: (draft, explicitId) => {
+        const id = explicitId ?? createId()
+        const updatedAt = new Date().toISOString()
+        const newDraft: Draft = { ...draft, updatedAt, id, seen: false }
+        set((draftState) => {
+          draftState.allDrafts = [newDraft, ...draftState.allDrafts]
```
drillprop:
```suggestion
          draftState.allDrafts.unshift(newDraft)
```

src/providers/drafts.ts
```diff
@@ -0,0 +1,122 @@
...
+
+type DraftStoreStateV1 = {
+  allDrafts: Draft[]
+}
+type DraftStoreActions = {
+  addDraft: (draft: RawDraft, explicitId?: string) => Draft
+  updateDraft: (draftId: string, draftProps: RawDraft) => void
+  getDraft: (draftId: string) => Draft | undefined
+  getDraftsForChannel: (draftIds: string) => Draft[]
+  getUnseenDraftsForChannel: (draftIds: string) => Draft[]
```
drillprop: I don't know - I don't like it. I always thought that actions should only modify data, not getting it. In my opinion, we should use selectors instead for all of these `get` actions.  For example:
```js
const drafts = useDraftStore(({ allDrafts }) => allDrafts.filter((d) => d.channelId === activeChannelId))
```

eldiegod: yeah I get the point I think we need to make separate selectors that do this but not live in the store actions

src/providers/drafts.ts
eldiegod: hmm I will need to check this with someone that still has the old data

drillprop: I guess you could just switch to `dev`, create several drafts and check it yourself ;). Just make sure to copy new drafts somewhere, to reuse them for testing.

src/providers/drafts.ts
```diff
@@ -0,0 +1,123 @@
...
+  updateDraft: (draftId: string, draftProps: RawDraft) => void
+  removeDrafts: (draftIds: string[]) => void
+  removeAllDrafts: (channelId: string) => void
+  markAllDraftsAsSeenForChannel: (channelId: string) => void
+}
+
+export const useDraftStore = createStore<DraftStoreStateV1, DraftStoreActions>(
+  {
+    state: {
+      allDrafts: [], // includes drafts for different channels
```
kdembler: Let's just use `drafts` here, we do the same in uploads, we don't call it `allUploads`

src/providers/drafts.ts
```diff
@@ -0,0 +1,123 @@
...
+      },
+    },
+  }
+)
+
+export const getDraft = (id: string) => (store: CommonStore<DraftStoreStateV1, DraftStoreActions>) =>
+  store.allDrafts.find((draft) => draft.id === id)
+export const getDraftsForChannel = (channelId: string) => (store: CommonStore<DraftStoreStateV1, DraftStoreActions>) =>
+  store.allDrafts.filter((d) => d.channelId === channelId)
+export const getUnseenDraftsForChannel = (channelId: string) => (
```
kdembler: I like this pattern but let's improve upon the naming here. I propose:
- `singleDraftSelector`
- `channelDraftsSelector`
- `chanelUnseenDraftsSelector`

src/providers/drafts.ts
```diff
@@ -0,0 +1,123 @@
...
+  language?: Language['iso'] | null
+  category?: VideoCategory['id'] | null
+  licenseCode?: License['code'] | null
+  licenseCustomText?: License['customText'] | null
+  licenseAttribution?: License['attribution'] | null
+} & CommonDraftProps
+
+export type RawDraft = Omit<Draft, 'id' | 'updatedAt'>
+
+type DraftStoreStateV1 = {
```
kdembler: Let's just call it `DraftStoreState`, it's the only version of that store so far

kdembler (changes requested):
Also there are some ESLint warnings

kdembler (approved):

PR #844, opened 2021-06-23
personalData store

### LINKED ISSUE #832: Use zustand for `personalData`

mikkio-j:
Issue: #832


src/providers/personalData/store.tsx
```diff
@@ -0,0 +1,99 @@
...
+    actionsFactory: (set) => ({
+      updateWatchedVideos: (__typename, id, timestamp) => {
+        set((state) => {
+          const currentVideo = state.watchedVideos.find((v) => v.id === id)
+          if (!currentVideo) {
+            const newVideo = __typename === 'COMPLETED' ? { __typename, id } : { __typename, id, timestamp }
+            return {
+              ...state,
+              watchedVideos: [...state.watchedVideos, newVideo],
+            }
```
kdembler: Why do we use destructuring instead of leveraging `immer` and use mutability?

src/providers/personalData/personalData.tsx
```diff
@@ -1,185 +1,34 @@
...
 type ContextValue = {
-  state: PersonalData
-  dispatch: Dispatch<Action>
+  state: PersonalDataStoreState
+  actions: PersonalDataStoreActions
 }
 const PersonalDataContext = React.createContext<ContextValue | undefined>(undefined)
 PersonalDataContext.displayName = 'PersonalDataContext'
 
+const PersonalDataProvider: React.FC = ({ children }) => {
```
kdembler: What do we keep this provider for?

src/providers/personalData/store.tsx
```diff
@@ -0,0 +1,93 @@
...
+        if (!oldVersion && oldVersion !== 0) {
+          return {
+            personalData: storageValue,
+          }
+        } else if (oldVersion === 0) {
+          return {
+            personalData: oldState,
+          }
+        }
+      },
```
kdembler: The legacy store doesn't use `personalData` key for local storage. It stores all the fields as separate entries. So `storageValue `will be `undefined` here. We need to make sure to properly migrate this data

kdembler: Also, since overwrite won't be enough here, we should delete the old keys to cleanup the storage

src/providers/personalData/store.tsx
```diff
@@ -0,0 +1,109 @@
...
+  updateRecentSearches: (id: string, type: RecentSearchType) => void
+  updateDismissedMessages: (id: string, add?: boolean) => void
+  updatePlayerVolume: (volume: number) => void
+}
+
+const watchedVideos = readFromLocalStorage<WatchedVideo[]>('watchedVideos') ?? []
+const followedChannels = readFromLocalStorage<FollowedChannel[]>('followedChannels') ?? []
+const recentSearches = readFromLocalStorage<RecentSearch[]>('recentSearches') ?? []
+const dismissedMessages = readFromLocalStorage<DismissedMessage[]>('dismissedMessages') ?? []
+const playerVolume = readFromLocalStorage<number>('playerVolume') ?? 1
```
drillprop, line 41:
```suggestion
```

I don't think we should use these. Persist should handle setting up state from localstorage

src/providers/personalData/store.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+export const usePersonalDataStore = createStore<PersonalDataStoreState, PersonalDataStoreActions>(
+  {
+    state: {
+      watchedVideos,
+      followedChannels,
+      recentSearches,
+      dismissedMessages,
+      playerVolume,
+    },
```
drillprop, line 51:
```suggestion
    state: {
      watchedVideos: [],
      followedChannels: [],
      recentSearches: [],
      dismissedMessages: [],
      playerVolume: 1,
    },
```

as above

src/providers/personalData/store.tsx
mikkio-j: That's the workaround for the migration. Since we're introducing new key in localStorage (`personalData`) - migrate will not fire. That's why I'm populating all personalData-related fields to initial state and then cleaning storage up. Alternative solution would be creating separate store for each field

src/providers/personalData/store.tsx
```diff
@@ -0,0 +1,109 @@
...
+        set((state) => {
+          state.playerVolume = volume
+        }),
+    }),
+  },
+  {
+    persist: {
+      key: 'personalData',
+      whitelist: WHITELIST,
+      version: 1,
```
kdembler:
```suggestion
      version: 0,
```

kdembler (approved):

PR #846, opened 2021-06-23
Orion improvements

### LINKED ISSUE #38: Orion integration improvements
Currently, the Orion integration is done the simple way:
- the video views are queried synchronously, so only after the query node request finishes
- view are queried on one by one basis - even though we fetch the list of videos from the query node and Orion allows batched lookup, we do that separately for every video

Going forward, we should:
- do the query node and Orion queries in parallel to reduce the request time
- use the batched lookup to reduce the number of requests to Orion

drillprop:
Fix #38

### DIFF
--- a/src/api/client/resolvers.ts
+++ b/src/api/client/resolvers.ts
@@ -63,7 +63,7 @@ export const queryNodeStitchingResolvers = (
   },
   Video: {
     views: async (parent, args, context, info) => {
-      if (parent.views !== null) {
+      if (parent.views != null) {
         return parent.views
       }
       try {
--- a/src/api/client/resolvers.ts
+++ b/src/api/client/resolvers.ts
@@ -5,12 +5,13 @@ import { GraphQLSchema } from 'graphql'
 import { Logger } from '@/utils/logger'
 
 import {
+  ORION_BATCHED_VIEWS_QUERY_NAME,
   ORION_FOLLOWS_QUERY_NAME,
   RemoveQueryNodeFollowsField,
   RemoveQueryNodeViewsField,
+  TransformBatchedOrionViewsField,
   TransformOrionFollowsField,
 } from './transforms'
-import { BATCHED_ORION_VIEWS_QUERY_NAME, TransformBatchedOrionViewsField } from './transforms/orionViews'
 
 import { VideoEdge } from '../queries'
 
@@ -63,8 +64,9 @@ export const queryNodeStitchingResolvers = (
       const batchedVideoViews = await delegateToSchema({
         schema: orionSchema,
         operation: 'query',
+        // operationName has to be manually kept in sync with the query name used
         operationName: 'GetBatchedVideoViews',
-        fieldName: BATCHED_ORION_VIEWS_QUERY_NAME,
+        fieldName: ORION_BATCHED_VIEWS_QUERY_NAME,
         args: {
           videoIdList: parent.edges.map((edge: VideoEdge) => edge.node.id),
         },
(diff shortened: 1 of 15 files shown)


src/api/client/resolvers.ts
```diff
@@ -57,29 +58,30 @@ export const queryNodeStitchingResolvers = (
...
-  Video: {
-    // TODO: Resolve the views count in parallel to the videosConnection query
-    // this can be done by writing a resolver for the query itself in which two requests in the same fashion as below would be made
-    // then the results could be combined
-    views: async (parent, args, context, info) => {
-      try {
-        return await delegateToSchema({
-          schema: orionSchema,
-          operation: 'query',
-          // operationName has to be manually kept in sync with the query name used
```
kdembler: Maybe let's keep this comment?

src/api/client/resolvers.ts
```diff
@@ -57,29 +58,30 @@ export const queryNodeStitchingResolvers = (
...
+        info,
+        transforms: [TransformBatchedOrionViewsField],
+      })
+
+      return parent.edges.map((edge: VideoEdge) => ({
+        ...edge,
+        node: {
+          ...edge.node,
+          views:
+            batchedVideoViews.find((videoView: { id: string; views: number }) => videoView?.id === edge.node.id)
```
kdembler: I think we could create a lookup for `batchedVideoViews` to avoid n^2 complexity? You can use `createLookup` from utils

src/api/queries/videos.graphql
```diff
@@ -106,6 +106,13 @@ query GetVideoViews($videoId: ID!) {
   }
 }
 
+query GetBatchedVideoViews($videoIdList: [ID!]!) {
```
kdembler, line 110: There's a comment before the query above, let's also include it here

src/api/client/resolvers.ts
```diff
@@ -57,29 +58,30 @@ export const queryNodeStitchingResolvers = (
       RemoveQueryNodeFollowsField,
     ]),
   },
-  Video: {
```
kdembler, line 60: I think this may still be needed though? For example, if you open video view, we only fetch a single video there, so without this that video won't contain the number of views. Currently what happens is that fetching "More videos from this channel" will actually end up populating that number but we shouldn't rely on that

kdembler (commented):
Mostly minor stuff, other than that seems to work great!

src/api/client/resolvers.ts
drillprop: I think this is more problematic - if I bring this back and query videosConnection, `GetVideoViews` will be also called for every video.
Not sure if there is a way to not call that.

kdembler: Are you sure? I'd expect `GetVideoViews` to be called only if video is missing the `views` field

drillprop: Bringing back this resolver was not enough, I had to add one if statement. I think it should work just fine. Let me know what you think

kdembler (commented):
That does the trick for single video and it looks good in general, but I just noticed if you have some videos in `Continue watching` and those won't be part of `videosConnection` loaded, they will be missing the views count. Not sure why that is happening though, since the `Video.views` resolver should kick in. Please take a look into that

drillprop:
> That does the trick for single video and it looks good in general, but I just noticed if you have some videos in `Continue watching` and those won't be part of `videosConnection` loaded, they will be missing the views count. Not sure why that is happening though, since the `Video.views` resolver should kick in. Please take a look into that

@kdembler it should be resolved now. Mocking should also works(featured videos and interrupted videos)

kdembler (approved):
Great job with that! It seems if you're scrolling the grid, videos are quite quicker to load now because of just one request being fired :+1:

PR #849, opened 2021-06-24
refactor transaction manager

kdembler:
Should fix problems with callbacks and transactions once and for all.

Changed the approach a bit, `TransactionManager` isn't a context anymore, it just manages the display of `TransactionDialog` and monitor state of query node sync, calling callback if needed.

All the other stuff is happening inside an unstateful hook now so there's no chance of callbacks clashing/not resetting properly between separate transactions.


src/providers/transactionManager/store.ts
```diff
@@ -0,0 +1,43 @@
...
+  addCallback: (callback: TransactionSyncCallback) => void
+  removeOldCallbacks: (currentBlock: number) => void
+  setDialogStep: (step: TransactionDialogStep) => void
+}
+
+export const useTransactionManagerStore = createStore<TransactionManagerStoreState, TransactionManagerStoreActions>({
+  state: { syncCallbacks: [], dialogStep: null },
+  actionsFactory: (set) => ({
+    addCallback: (callback) =>
+      set((draft) => {
```
drillprop: Why it's called `draft`? Why not just `state`?

src/providers/transactionManager/store.ts
```diff
@@ -0,0 +1,43 @@
+import { ExtrinsicStatus } from '@/joystream-lib'
+import { createStore } from '@/store'
+
+type TransactionSyncCallback = {
+  targetBlock: number
+  callback: () => void
+}
```
drillprop, line 7: Maybe it's just a preference but don't like how it's called, I would expect only a callback from this type. It's a little bit confusing. Also, I've noticed that you doing something like this: 
```tsx
cb.callback()
``` 
which is also in my opinion non-intuitive

src/providers/transactionManager/useTransaction.ts
```diff
@@ -0,0 +1,119 @@
...
+            } catch (e) {
+              Logger.error('Failed transaction sync callback', e)
+            }
+          }
+          resolve()
+        }
+        addCallback({ callback: syncCallback, targetBlock: block })
+      })
+
+      return new Promise<boolean>((resolve) => {
```
drillprop:
```suggestion
      return new Promise((resolve) => {
```
I don't think you need this

drillprop (approved):
Overall great work, much-needed PR ❤️  
Did several uploads with no problems, so it should resolve the issue.

src/providers/transactionManager/store.ts
kdembler: That's the default naming that `immer` uses but changed that to `state` now.

src/providers/transactionManager/store.ts
kdembler: Hmm yeah I just couldn't come up with better names. Changed that now, hopefully it's better

src/providers/transactionManager/useTransaction.ts
kdembler: removed :+1:

src/providers/transactionManager/store.ts
drillprop: Yep, I think it's much better now  👍🏻

drillprop (approved):

PR #854, opened 2021-06-25
Improve e2e

eldiegod:


eldiegod:
Idea: we should probably look at https://docs.cypress.io/guides/dashboard/introduction#Features to be able to debug tests that work on local but fail on CI

cypress/support/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+declare namespace Cypress {
+  interface Chainable {
+    /**
+     * Custom command to login into Atlas with a test membership.
+     */
+    login(): Chainable<void>
+    /**
+     * Custom command to make the polkadot api available and add accounts, memberships, etc.. to it.
+     */
+    injectPolkadot(accounts: any[]): Chainable<void>
```
kdembler, line 39: Should probably use types from polkadot: https://github.com/polkadot-js/extension#api-interface

cypress/support/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+    window['injectedWeb3'] = {
+      'polkadot-js': {
+        enable,
+        version: '0.38.3',
+      },
+    }
+  })
+})
+
+Cypress.Commands.add('login', () => {
```
rafalpawlow, line 61: I'm not sure about this, It seems non-semantic to me. Injecting Polkadot extension is not equal to log user in. WDYT?

eldiegod: Yeah, I wasn't sure how to do it, as the injected polkadot gets the accounts on that enable fn, I don't know if it's possible to add them post-injection? that should way better

could possibly be something you could add to the signing transaction PR if you have some insight on it?

rafalpawlow: Sure, let's do it this way

rafalpawlow (approved):

cypress/support/commands.ts
eldiegod: Lets do that in another PR to not block the other work as I didn't manage to add it without breaking intellisense right now

PR #855, opened 2021-06-25
Animations playground lottie

### LINKED ISSUE #845: Set up Lottie test
Designers suggested we use [Lottie](https://airbnb.io/lottie/#/) library for animations in Atlas. Let's set up some test in playground to see how it works. Check out https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=8840%3A230852, you can download the animation files there

eldiegod:
Set up Lottie test #845


kdembler (commented):
Those are just beautiful, I can't wait to get those integrated in the app.

However, there's an issue I see currently. The `json` files for animations are quite huge and I wouldn't want to include them in the bundle, that's very costly. We need to either:
a) somehow hot load them, that should put them in their own separate chunks
b) host those statically somewhere outside of the bundle altogether.

I'm not sure how a) would work with caching between deployments but my guess is it'd be easier to do

eldiegod:
Loading the animations on-demand now, seems to be working and putting them in separate chunks

src/views/playground/Playgrounds/Animations.tsx
```diff
@@ -0,0 +1,41 @@
+import styled from '@emotion/styled'
+import loadable from '@loadable/component'
+import React, { useEffect, useState } from 'react'
+
+const LoadableLottie = loadable(() => import('react-lottie-player'), {
+  fallback: <>LOADING LOTTIE</>,
+})
+export const Animations = () => {
+  const [XL, setXL] = useState<any>(null)
```
kdembler: If we really need `any` for all of those, let's disable the eslint warning

kdembler (changes requested):

src/views/playground/Playgrounds/Animations.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const [L, setL] = useState<any>(null)
+  const [M, setM] = useState<any>(null)
+  const [S, setS] = useState<any>(null)
+  const [XS, setXS] = useState<any>(null)
```
drillprop:
```suggestion
  const [XL, setXL] = useState<Record<string, unknown> | null>(null)
  const [L, setL] = useState<Record<string, unknown> | null>(null)
  const [M, setM] = useState<Record<string, unknown> | null>(null)
  const [S, setS] = useState<Record<string, unknown> | null>(null)
  const [XS, setXS] = useState<Record<string, unknown> | null>(null)
```
should fix it

drillprop: I overcomplicated this - we can write just `unknown | null` :P

kdembler (approved):

Issue #805, opened 2021-06-18
Should meta info on video tile be positioned consistently?


mochet:
Problem: some video titles wrap onto a second line, when this happens in both the videos page and individual channel view, the channel name and some other elements are not aligned anymore. I have highlighted the problem in the below screen capture.

![image](https://user-images.githubusercontent.com/50830161/122552792-d0d7ef80-d047-11eb-84b5-01b9b81755aa.png)

kdembler:
I'm not sure if this is a bug. That's how it works on YouTube and I think that's also what our designs assume. Will let the designers decide what to do about that

KubaMikolajczyk:
@kdembler is right! This is not a bug - we do allow the title to wrap to the second line and truncate it if it exceeds the second line limit. This will create a difference in height between one line and two line titled videos. It won't be a problem with a sufficient space left below all tiles so the ones with the longer title will still stay in grid and still have a sufficient margin between them and the ones below.

kdembler:
Since this is working as desired, I'm gonna close this

PR #862, opened 2021-06-25
Fix video sizing bug

eldiegod:


drillprop:
I believe it will fix the issue, but I'm not sure if we want to limit titles to just one line: 
image

related to https://github.com/Joystream/atlas/issues/805

eldiegod:
> I believe it will fix the issue, but I'm not sure if we want to limit titles to just one line:
> image
> 
> related to #805

Then I guess the same solution but for multiple lines

eldiegod:
Nvm, doesn't quite work, we might have to patch it this way if we want to release with the fix and find a better solution after @kdembler ?

PR #848, opened 2021-06-24
useAsset fix

drillprop:

PR #876, opened 2021-06-25
refactor useAsset, fix form problems

kdembler:
Turns out the fix for the problem with thumbnail updates in video sheet form we encountered in #848 was rather trivial after all. However I thought there was a problem with implementation of `useAsset` and did a refactor only to discover the problem is somewhere else. However, I think the refactored version is quite more readable so decided to open a PR for it anyway. This should also fix all the issues with the form.

### DIFF
--- a/src/hooks/useAsset/useAsset.tsx
+++ b/src/hooks/useAsset/useAsset.tsx
@@ -1,33 +1,17 @@
-import isEqual from 'lodash/isEqual'
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { useEffect, useState } from 'react'
 
-import { AllChannelFieldsFragment, AssetAvailability, VideoFieldsFragment } from '@/api/queries'
-import { AssetType } from '@/hooks'
 import { useStorageProviders } from '@/providers'
-import { createStorageNodeUrl } from '@/utils/asset'
 import { Logger } from '@/utils/logger'
 
-type UseAssetDataArgs =
-  | { entity?: VideoFieldsFragment | null; assetType: AssetType.THUMBNAIL | AssetType.MEDIA }
-  | {
-      entity?: Partial<AllChannelFieldsFragment> | null
-      assetType: AssetType.COVER | AssetType.AVATAR
-    }
+import { getAssetUrl, readAssetData, testAssetDownload } from './helpers'
+import { UseAsset } from './types'
 
-type UseAssetData = {
-  url?: string
-  error: ErrorEvent | null
-  isLoading?: boolean
-}
-
-type UseAsset = ({ entity, assetType }: UseAssetDataArgs) => UseAssetData
-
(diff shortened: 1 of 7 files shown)


drillprop:
Working much better now. Unfortunately, there is still one problem left, which wasn't there before:

If you change the thumbnail of the video and publish changes you'll not see the updated image in the `MyVideosView`. When you click again on the edited video, you'll also see the old thumbnail
asset bug

drillprop:
@kdembler I opened PR to your branch. It should resolve the issue, not sure if you like how I've done that, let me know what you think.

src/hooks/useAsset/useAsset.tsx
```diff
@@ -1,120 +1,60 @@
...
+export const useAsset: UseAsset = ({ entity, assetType }, opts = {}) => {
   const { getStorageProvider } = useStorageProviders()
   const [error, setError] = useState<ErrorEvent | null>(null)
-  const [isLoading, setIsLoading] = useState<boolean | undefined>(undefined)
+  const [isLoading, setIsLoading] = useState<boolean>(true)
   const [url, setUrl] = useState<string | undefined>(undefined)
-  const cachedAssetData = useRef({})
+  const [cachedAssetData, setCachedAssetData] = useState<AssetData | null | undefined>(undefined)
+
+  const assetData = entity ? readAssetData(entity, assetType) : null
```
drillprop:
```suggestion
  const assetData = readAssetData(entity, assetType) 
```
I don't think it's needed - if an entity is null `readAssetData` returns null anyway

src/hooks/useAsset/useAsset.tsx
```diff
@@ -1,120 +1,60 @@
...
+  const assetDataNotChanged = entity ? isEqual(assetData, cachedAssetData) : false
 
   useEffect(() => {
-    if (error) {
-      Logger.error(`Failed to load ${assetType}`, error)
+    if (!entity || assetDataNotChanged || opts.skip) {
```
drillprop: I could be wrong, but do we actually need `opts.skip`? I removed that from the code and didn't notice any difference, everything should work just fine. Tested updating thumbnails and switching tabs, with no problems

drillprop (commented):
Found two small things in the code. Besides that everything works really well.

src/hooks/useAsset/useAsset.tsx
drillprop: pretty much the same applies to the line underneath

src/hooks/useAsset/useAsset.tsx
kdembler: Yeah seems it wasn't doing us any good, removed

src/hooks/useAsset/useAsset.tsx
kdembler: 👍 removed

PR #880, opened 2021-06-28
add export/import of localStorage

### LINKED ISSUE #847: Add ability to export/import local state
I think it'd be a nice feature to enable some kind of portability between different instances of Atlas. We could add an option to export local state (everything saved in localstorage) as a JSON file in `/admin`. Likewise, there could be an option to load a JSON file to load state

mikkio-j:
Issue: #847

### DIFF
--- a/src/views/admin/AdminView.tsx
+++ b/src/views/admin/AdminView.tsx
@@ -1,4 +1,3 @@
-import { css } from '@emotion/react'
 import React, { useState } from 'react'
 import { Link } from 'react-router-dom'
 
@@ -11,9 +10,8 @@ const items = availableEnvs().map((item) => ({ name: item, value: item }))
 export const AdminView = () => {
   const env = availableEnvs().includes(TARGET_DEV_ENV) ? TARGET_DEV_ENV : null
   const [value, setValue] = useState<null | string>(env)
-  const [file, setFile] = useState<File | null>(null)
 
-  const handleChange = (value?: string | null | undefined) => {
+  const handleEnvironmentChange = (value?: string | null | undefined) => {
     if (!value) {
       return
     }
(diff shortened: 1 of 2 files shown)


src/views/admin/AdminView.tsx
```diff
@@ -23,20 +24,63 @@ export const AdminView = () => {
...
+      <div>
+        <Text variant="h2">Choose environment</Text>
+        <Select
+          items={items}
+          onChange={handleChange}
+          value={value}
+          css={css`
+            margin-top: 20px;
+            max-width: 400px;
+          `}
```
kdembler: Not your code, but we need to remove this, we no longer use this prop

src/views/admin/AdminView.tsx
```diff
@@ -23,20 +24,63 @@ export const AdminView = () => {
...
+
+    const json = JSON.stringify(state)
+    const blob = new Blob([json], { type: 'application/json' })
+    const href = URL.createObjectURL(blob)
+    const link = document.createElement('a')
+    link.href = href
+    link.download = fileName + '.json'
+    document.body.appendChild(link)
+    link.click()
+    document.body.removeChild(link)
```
kdembler: This won't fully work because you're treating read items as just regular text and those are already JSON objects. Values will be raw strings in that JSON file and not proper objects. Probably doing `JSON.parse` on `locaStorage.getItem()` result may do the trick

kdembler: Take a look at scripts I've prepared here: https://gist.github.com/kdembler/a77b32bc1ee7b8898e3d07cf34e26d26#file-atlas-dev-test-md

src/views/admin/AdminView.tsx
```diff
@@ -23,20 +24,63 @@ export const AdminView = () => {
...
+            margin-top: 20px;
+            max-width: 400px;
+          `}
+        />
+      </div>
+      <div>
+        <Text variant="h2">Import/Export Local state</Text>
+        <Button onClick={handleExportClick}>Export local state</Button>
+        <div style={{ margin: '20px 0' }}>
+          <input type="file" onChange={handleFileChange} accept="application/json" />
```
kdembler: Could we make this input "virtual" or hidden so it's not visible on the screen but gets triggered when the user clicks import button? Should make the screen more readable

src/views/admin/AdminView.tsx
mikkio-j: Did this correctly at the beginning but for some reason it didn't work. Same with the scripts that you've provided - turned out if there is `target_env` present in the localStorage the scripts will fail. Added fix for storing env, so should be good now

src/views/admin/AdminView.tsx
```diff
@@ -23,20 +22,59 @@ export const AdminView = () => {
...
+        acc[key] = JSON.parse(rawValue)
+      }
+
+      return acc
+    }, {} as Record<string, unknown>)
+    const jsonStorage = JSON.stringify(storage)
+
+    const linkElement = document.createElement('a')
+    linkElement.setAttribute('href', 'data:application/json;charset=utf-8,' + encodeURIComponent(jsonStorage))
+    linkElement.setAttribute('download', 'atlas-export.json')
```
drillprop: Just an idea, but I think it could be useful - maybe we could add formatted date to the name of the file? For example `atlas-export-29-06-2021-1307.json`

src/views/admin/AdminView.tsx
```diff
@@ -23,20 +22,59 @@ export const AdminView = () => {
...
+    const file = (e.target as HTMLInputElement).files?.[0]
+    if (!file) return
+    const fileText = await file.text()
+    const storage = JSON.parse(fileText)
+    Object.keys(storage).forEach((key) => {
+      window.localStorage.setItem(key, JSON.stringify(storage[key]))
+    })
+  }
+
+  const handleImportClick = async () => {
```
drillprop: Is it async?
Also, we could add some error handling here. For example, if you import damaged JSON, we could display information that imported JSON is not correct.

drillprop (approved):

src/views/admin/AdminView.tsx
```diff
@@ -23,20 +25,72 @@ export const AdminView = () => {
...
+        acc[key] = JSON.parse(rawValue)
+      }
+
+      return acc
+    }, {} as Record<string, unknown>)
+    const jsonStorage = JSON.stringify(storage)
+
+    const linkElement = document.createElement('a')
+    linkElement.setAttribute('href', 'data:application/json;charset=utf-8,' + encodeURIComponent(jsonStorage))
+    linkElement.setAttribute('download', `atlas-export-${new Date().toLocaleString()}.json`)
```
kdembler: This will produce a file name without spaces and forbidden characters (`:`)

```suggestion
    linkElement.setAttribute('download', `atlas-export-${new Date().toISOString()}.json`)
```

kdembler (approved):

Issue #879, opened 2021-06-27
In-App feedback 


bedeho:
Could it be a good idea to introduce a really low bar way of allowing anyone to provide feedback while we are in the development stage? There are plenty of off the shelf solutions for this, like (Mopinion)[https://mopinion.com/], so it should very low cost to integrate maintain.

dmtrjsg:
usersnap is a contender; while mopinion can be added to the discovery mix on MIro board. Ask @dmtrjsg

PR #884, opened 2021-06-28
Add feeder.sh in-app feedback

phaitonican:
I'm adding this module:
https://feeder.sh/

#879
It enables user feedback, we have to create an account and change the projectId to one of yours. Right now I have a own account and made projectId, but its not a blockchain solution. (Preview is also on the website)

Was not sure about the .css file. Without .d.ts type file, it would give an error.

Was also not sure if I would implement in App.tsx or Mainlayout, just chose MainLayout for now.

It's kind of hard to see, because the background is black too, but works well.

Picture:
image

Received feedback on feeder.sh:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "date-fns": "^2.15.0",
     "downshift": "^6.1.0",
     "emotion-normalize": "~11.0.0",
+    "feeder-react-feedback": "^0.0.6",
     "glider-js": "^1.7.3",
     "graphql": "^15.3.0",
     "graphql-tag": "^2.11.0",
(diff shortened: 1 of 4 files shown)


kdembler:
Hey there!

Thanks for your work on this, but we're probably gonna use a different solution to achieve that, taking into consideration other needs we may have.

In general, it's a good idea to drop a comment in an issue you are interested about before starting the work on it. That will allow you to get any necessary info that may not be captured in Github or just make sure that this is a way we want to approach a certain topic.

Since we'll be using something else, I'm gonna close the PR

Thanks!

Issue #828, opened 2021-06-21
Introduce in-player social features


toiletgranny:
The scope of this task is to introduce an experience in the player component to allow the user to:
- Copy/generate a link to the currently watched video,
	- Include a timestamp,
- Share the currently watched video on social media platforms (the list of platforms to be finalized through research).

---

_Based on: #785_

toiletgranny:
Designs for this flow can be found here — [LINK](https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=9981%3A257700).

PR #896, opened 2021-06-28
Add react share module and implement component in VideoView

phaitonican:
Sorry, redone the PR to fix commits and alphabetical order. Added Facebook and Twitter share under video license.

I am not able to find a good design implementation, could you help? Also whats happening with the yarn.lock debug things? Sorry for inconvenience...


src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -205,6 +206,13 @@ export const VideoView: React.FC = () => {
             <Placeholder height={12} width={200} />
           )}
         </LicenseContainer>
+        <SocialShareContainer>
+          {video ? (
+            <SocialShare url={window.location.href} message={'I am watching ' + video.title || ''} hashtags={['']} />
```
phaitonican, line 211: Not happy here with || and empty hashtags, wrong approach .. ?

kdembler:
Thanks for the PR.

While it seems to work fine, this will certainly require designers input. If you take a look at our backlog, you will find https://github.com/Joystream/atlas/issues/828 which is exactly for that purpose. So until the design for how that feature may look like is ready, we will have to postpone any implementation.

Like mentioned in the other PR, make sure to drop a comment first, then we can make sure you don't waste your time because you don't know what I said above.

I'm gonna close this PR for the reasons outlined, but if you're interested, I can give you a ping once we have the designs for that done and we're ready to start implementation

PR #916, opened 2021-06-29
Use global store for connection status

### LINKED ISSUE #913: Use global store for connection status provider

mikkio-j:
Issue: #913

### DIFF
--- a/src/providers/connectionStatus.tsx
+++ b/src/providers/connectionStatus.tsx
@@ -1,37 +1,41 @@
 import React, { useCallback, useEffect, useRef } from 'react'
-import create from 'zustand'
 
 import { useSnackbar } from '@/providers/snackbars'
+import { createStore } from '@/store'
 
 export type ConnectionStatus = 'connected' | 'disconnected' | 'connecting'
 
-type ConnectionStatusValue = {
+type ConnectionStatusStoreState = {
   nodeConnectionStatus: ConnectionStatus
-  isUserConnectedToInternet: boolean
+  internetConnectionStatus: ConnectionStatus
+}
+
+type ConnectionStatusStoreActions = {
   setNodeConnection: (connection: ConnectionStatus) => void
-  setInternetConnection: (isConnected: boolean) => void
+  setInternetConnection: (connection: ConnectionStatus) => void
 }
 
 const SNACKBAR_TIMEOUT = 15000
 
-export const useConnectionStatusStore = create<ConnectionStatusValue>((set) => ({
-  isUserConnectedToInternet: true,
-  nodeConnectionStatus: 'connecting',
-  setNodeConnection: (connection) => {
-    set((state) => {
-      state.nodeConnectionStatus = connection
-    })
-  },
-  setInternetConnection: (isConnected) => {
-    set((state) => {
-      state.isUserConnectedToInternet = isConnected
-    })
-  },
-}))
+export const useConnectionStatusStore = createStore<ConnectionStatusStoreState, ConnectionStatusStoreActions>({
+  state: { internetConnectionStatus: 'connected', nodeConnectionStatus: 'connecting' },
+  actionsFactory: (set) => ({
+    setNodeConnection: (connection) => {
+      set((state) => {
+        state.nodeConnectionStatus = connection
+      })
+    },
+    setInternetConnection: (connection) => {
+      set((state) => {
+        state.internetConnectionStatus = connection
+      })
+    },
+  }),
+})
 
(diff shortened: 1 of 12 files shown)


src/providers/connectionStatus.tsx
```diff
@@ -1,23 +1,39 @@
...
-  const [nodeConnectionStatus, setNodeConnection] = useState<ConnectionStatus>('connecting')
-  const [isUserConnectedToInternet, setIsUserConnectedToInternet] = useState(true)
+export const useConnectionStatusStore = create<ConnectionStatusValue>((set) => ({
+  isUserConnectedToInternet: true,
+  nodeConnectionStatus: 'connecting',
+  setNodeConnection: (connection) => {
+    set((state) => ({
+      ...state,
+      nodeConnectionStatus: connection,
+    }))
```
drillprop:
```suggestion
    set((state) => {
      state.nodeConnectionStatus = connection
    })
```

src/providers/connectionStatus.tsx
```diff
@@ -1,23 +1,39 @@
...
+  },
+  setInternetConnection: (value) => {
+    set((state) => ({
+      ...state,
+      isUserConnectedToInternet: value,
+    }))
```
drillprop:
```suggestion
    set((state) => {
      state.isUserConnectedToInternet = value
    })
```

drillprop (commented):
Works good, but you forgot that we are using `immer`. Added two suggestions

src/providers/connectionStatus.tsx
```diff
@@ -1,23 +1,37 @@
...
 
-const ConnectionStatusContext = React.createContext<undefined | ConnectionStatusValue>(undefined)
-ConnectionStatusContext.displayName = 'ConnectionStatusContext'
-
 type ConnectionStatusValue = {
-  setNodeConnection: (connection: ConnectionStatus) => void
   nodeConnectionStatus: ConnectionStatus
   isUserConnectedToInternet: boolean
+  setNodeConnection: (connection: ConnectionStatus) => void
+  setInternetConnection: (value: boolean) => void
```
drillprop:
```suggestion
  setInternetConnection: (isConnected: boolean) => void
```
last small thing - I think we should be more semantic here, let's just rename it

src/providers/connectionStatus.tsx
```diff
@@ -1,23 +1,37 @@
...
+  isUserConnectedToInternet: true,
+  nodeConnectionStatus: 'connecting',
+  setNodeConnection: (connection) => {
+    set((state) => {
+      state.nodeConnectionStatus = connection
+    })
+  },
+  setInternetConnection: (value) => {
+    set((state) => {
+      state.isUserConnectedToInternet = value
```
drillprop:
```suggestion
  setInternetConnection: (isConnected) => {
    set((state) => {
      state.isUserConnectedToInternet = isConnected
```

drillprop (approved):

src/providers/connectionStatus.tsx
```diff
@@ -1,23 +1,37 @@
...
+  setNodeConnection: (connection: ConnectionStatus) => void
+  setInternetConnection: (isConnected: boolean) => void
 }
 
 const SNACKBAR_TIMEOUT = 15000
 
-export const ConnectionStatusProvider: React.FC = ({ children }) => {
-  const [nodeConnectionStatus, setNodeConnection] = useState<ConnectionStatus>('connecting')
-  const [isUserConnectedToInternet, setIsUserConnectedToInternet] = useState(true)
+export const useConnectionStatusStore = create<ConnectionStatusValue>((set) => ({
```
kdembler: This is using the default `zustand` `create`, not from out store utilities. We should use the common helper. This one also doesn't have `immer` so my guess is it's not fully working

kdembler (changes requested):

src/views/studio/SignInView/SignInView.tsx
```diff
@@ -22,7 +22,7 @@ import {
 export const SignInView = () => {
   const navigate = useNavigate()
   const { activeChannelId, setActiveUser, memberships } = useUser()
-  const { nodeConnectionStatus } = useConnectionStatus()
+  const nodeConnectionStatus = useConnectionStatusStore((state) => state.nodeConnectionStatus)
```
drillprop, line 25: Previously if there was no internet connection all the tiles and sign-in button were disabled. Right now if you disconnect from wifi, you'll still be able to click those.

kdembler (approved):

PR #917, opened 2021-06-29
Open studio/viewer in new tab

### LINKED ISSUE #911: Studio/viewer link inside the sidenav should open a new tab
When you move to studio/viewer experience, that should open a new tab instead of navigating the current one.

drillprop:
Fix #911

### DIFF
--- a/src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
+++ b/src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
@@ -1,5 +1,4 @@
 import React, { useState } from 'react'
-import { useNavigate } from 'react-router'
 import { CSSTransition } from 'react-transition-group'
 
 import { NavItemType, SidenavBase } from '@/components/Sidenav/SidenavBase'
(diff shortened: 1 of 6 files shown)


src/utils/browser.ts
```diff
@@ -14,3 +14,8 @@ export const isChromiumBased = (): boolean => !!window.chrome
 export const isAllowedBrowser = () => isFirefox() || isChromiumBased()
 
 export const isBrowserOutdated = !('ResizeObserver' in window) || !('IntersectionObserver' in window)
+
+export const openInNewTab = (url: string, local?: boolean) => {
+  const origin = window.location.origin
+  window.open(`${local ? origin : ''}${url}`, '_blank')?.focus()
+}
```
mikkio-j, line 21: since you've added that utility fn it might make sense to use that here:
https://github.com/Joystream/atlas/blob/dev/src/components/VideoPreview.tsx#L90

not in the scope though

drillprop: done

src/components/Sidenav/StudioSidenav/StudioSidenav.tsx
```diff
@@ -60,12 +59,13 @@ export const StudioSidenav: React.FC = () => {
       openWarningDialog({
```
kdembler: Do we still need that warning though? You don't lose anything by opening a new tab

drillprop: right, I guess we don't need this anymore.

drillprop: removed

mikkio-j (approved):

kdembler (approved):

PR #929, opened 2021-06-29
introduce persistent asset resolution

kdembler:
This should fix all the problems with assets in all cases (but then again, I thought that the last time as well)

This changes the approach to asset resolution slightly: there's a global resolution map, mapping `contentId` to `{ url, blob }`. This ensures that one asset will not be fetched from multiple storage providers if present in more than one location. Also, both edit video form and channel edit form now resolve assets by themselves, not relying on asset finishing loading to present correct form states. One additional benefit of global resolution map is that we can get rid of the workaround with writing asset url into cache to use local URL. Now we can just add an entry to the resolution map pointing to the local URL and it's working perfectly.

Please test the hell out of it...


src/providers/assets/helpers.ts
```diff
@@ -4,11 +4,11 @@ import {
   BasicChannelFieldsFragment,
   VideoFieldsFragment,
 } from '@/api/queries'
-import { AssetType } from '@/hooks'
 import { createStorageNodeUrl } from '@/utils/asset'
 import { Logger } from '@/utils/logger'
 
-import { AssetData } from './types'
+import { AssetType } from './constants'
```
drillprop: Can we move `AssetType` to `types` after all? I know that we can use enums as a value, but I'm getting the impression that we have already too many files in this folder.

drillprop (commented):
Found two issues:
- problem with sizing, when the video thumbnails load: 
thumbnail title sizes
- If you update the video thumbnail the first time, everything goes fine. But if you update the thumbnail of the same video a second time, you'll not see the updated image in MyVideosView. I think it could be connected to your caching changes.

Besides that, I think everything works, so this is a step in the right direction.

src/providers/assets/helpers.ts
kdembler: Yes, good idea

drillprop (approved):

eldiegod (approved):
Just tested it quite a bit and I didn't get the issues @drillprop mentioned I assume those are probably fixed (?) 

So everything seems to work for me

kdembler:
Yes, already fixed those :+1:

PR #973, opened 2021-07-01
Banner component update

### LINKED ISSUE #943: Update banner component
We currently have `DismissibleMessage` component. There are new designs for it, found in https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=6397%3A23

At the most basic level, we should update the variant we have but we also could implement the other variants specified by the design. For reference: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=7260%3A204477

mikkio-j:
Resolves: #943


src/shared/components/DismissibleMessage/DismissibleMessage.style.ts
```diff
@@ -30,9 +31,33 @@ export const MessageTitle = styled(Text)`
...
   margin-top: ${sizes(2)};
   line-height: ${sizes(5)};
   color: ${colors.gray[300]};
   word-break: break-word;
 `
+
+export const MessageWrapper = styled.div<MessageProps>`
+  position: relative;
+  padding: ${sizes(4)};
+  box-shadow: ${({ variant }) => variant === 'primary' && `inset 0px 0px 0px 1px ${colors.gray[700]}`};
```
drillprop: Im getting: image

```suggestion
  box-shadow: ${({ variant }) => variant === 'primary' && `inset 0 0 0 1px ${colors.gray[700]}`};
```

If this suggestion introduces any problems you should just add stylelint ignore.

drillprop (commented):
Looks good 👍🏻 . I think we should also rename this component - `Banner` will be a more suitable name than `DismissibleMessage`. What do you think?

mikkio-j:
> Looks good 👍🏻 . I think we should also rename this component - `Banner` will be a more suitable name than `DismissibleMessage`. What do you think?

You're right, but the rename won't be enough. The component itself operates on state of dismissible messages, so I will extract styles into `Banner` instead (and use that `Banner` inside `DismissibleMessage`)

eldiegod (approved):

PR #976, opened 2021-07-03
Channel view layout

### LINKED ISSUE #935: Channel view layout
Doesn't include tabs, search or sorting dropdown

### LINKED ISSUE #942: Channel view Tabs component

### LINKED ISSUE #946: Channel view pagination

### LINKED ISSUE #947: Channel view about layout

### LINKED ISSUE #948: update channel cover component

### LINKED ISSUE #955: Channel view sorting dropdown

eldiegod:
#935
#948
https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=8433%3A2894


src/components/LimitedWidthContainer.tsx
```diff
@@ -0,0 +1,17 @@
...
+  position: relative;
+  margin: 0 auto;
+`
+export const LimitedWidthContainer = styled.div`
+  --max-inner-width: calc(1440px - var(--sidenav-collapsed-width) - calc(2 * var(--global-horizontal-padding)));
+
+  max-width: var(--max-inner-width);
+  position: relative;
+  margin: 0 auto;
+`
```
drillprop:
```suggestion
export const LimitedWidthContainer = styled.div`
  ${limitedWidthContainerStyle}
`
```
Let's not repeat yourself. Also, I'm generally okay with it, but not sure why you've changed the name.

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -83,15 +82,15 @@ export const ChannelView: React.FC = () => {
             )}
           </TitleContainer>
           <StyledButtonContainer>
-            <Button variant={isFollowing ? 'secondary' : 'primary'} onClick={handleFollow} size="large">
+            <Button icon={''} variant={isFollowing ? 'secondary' : 'primary'} onClick={handleFollow} size="large">
```
drillprop:
```suggestion
            <Button variant={isFollowing ? 'secondary' : 'primary'} onClick={handleFollow} size="large">
```

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -77,11 +47,9 @@ export const Title = styled(Text)`
 
 export const SubTitle = styled(Text)`
   ${fluidRange({ prop: 'fontSize', fromSize: '14px', toSize: '18px' })};
-
   padding: ${sizes(1)} ${sizes(2)};
-  margin-top: ${sizes(2)};
-  color: ${colors.white};
-  background-color: ${colors.gray[800]};
+  margin-top: ${sizes(1)};
```
drillprop, line 62:
```suggestion

  margin-top: ${sizes(1)};
```
Stylelint says: ` Expected empty line before declaration (declaration-empty-line-before)`

src/shared/components/ChannelCover/ChannelCover.style.ts
```diff
@@ -45,87 +34,11 @@ export const CoverImage = styled.div<CoverImageProps>`
...
-  ${media.xxlarge} {
-    background-image: linear-gradient(
-        0deg,
-        black 0%,
-        black ${CONTENT_OVERLAP_MAP.XXLARGE - GRADIENT_OVERLAP}px,
-        rgba(0, 0, 0, 0) ${CONTENT_OVERLAP_MAP.XXLARGE - GRADIENT_OVERLAP + GRADIENT_HEIGHT}px
-      ),
-      url(${({ $src }) => $src});
-  }
+  background-image: url(${({ $src }) => $src});
```
drillprop: Stylelint says here: `Unexpected empty line before declaration (declaration-empty-line-before)`. Let's just remove the empty line above this line.

drillprop (changes requested):
The video grid is not looking great on viewport > 1900. There are 6 videos in a row(not 4 as in the design) and the thumbnails are really small. 
image
Not sure, but I think you'll need to do some changes on `InfiniteGrid`

src/components/LimitedWidthContainer.tsx
drillprop: Oh - ok. I know why - because we are applying it to Studio and Viewer views.

drillprop:
`why is lint not doing this automatically`  😄 
To fix these, you would have to install stylelint extension for vscode or run `lint:css --fix` if I'm not mistaken.

eldiegod:
> `why is lint not doing this automatically` 😄
> To fix these, you would have to install stylelint extension for vscode or run `lint:css --fix` if I'm not mistaken.

I knew I was missing something

drillprop:
Initially, we wanted to fix these problems with prettier, but I couldn't find a working plugin, so unfortunately you need to rely on console warnings if you don't use extension/command

src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -0,0 +1,111 @@
+import React from 'react'
+import { useParams } from 'react-router'
+
+import { useChannel } from '@/api/hooks'
+import { languages } from '@/config/languages'
+import { Avatar, Button, Text } from '@/shared/components'
```
drillprop:
```suggestion
import { Button, Text } from '@/shared/components'
```

src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <Details>
+          <Text variant="caption" secondary>
+            Num. of views
+          </Text>
+          <Text variant="h6">7 245 345</Text>
+        </Details>
+
+        <Details>
+          <Text variant="caption" secondary>
+            Num. of videos
```
drillprop, line 64: I think we don't need dummy value here, we can get number of videos, right? I see it in `/channels`

src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <DetailsContainer>
+        <DetailsText variant="h4">Details</DetailsText>
+
+        <Details>
+          <Text variant="caption" secondary>
+            Owned by member
+          </Text>
+          <AvatarContainer>
+            <StyledAvatar />
+            <Text variant="h6">Garry Covin</Text>
```
drillprop: Im almost sure that we can get member name too.

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -83,15 +140,61 @@ export const ChannelView: React.FC = () => {
...
 }
+
+const usePagination = (currentTab: number) => {
+  const [currentPage, setCurrentPage] = useState(0)
+  // reset the pagination when changing tabs
+  useEffect(() => {
+    setCurrentPage(0)
+  }, [currentTab])
+  return { currentPage, setCurrentPage }
+}
```
drillprop, line 200: Not sure what's causing it, I tried to investigate this, but when I go to the channel from `/channels` I often see this: 
image
For some reason videos not loading correctly.  This is weird because it doesn't happen on /studio/videos

src/shared/components/ChannelCover/ChannelCover.style.ts
```diff
@@ -44,88 +33,11 @@ export const CoverImage = styled.div<CoverImageProps>`
...
-    margin-bottom: -${CONTENT_OVERLAP_MAP.LARGE}px;
-  }
-
-  ${media.xlarge} {
-    margin-bottom: -${CONTENT_OVERLAP_MAP.XLARGE}px;
-  }
-
-  ${media.xxlarge} {
-    margin-bottom: -${CONTENT_OVERLAP_MAP.XXLARGE}px;
-  }
```
drillprop, line 128: on `/studio/channel` I spotted this when my viewport was > 1900
image

src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const { channel } = useChannel(id)
+
+  return (
+    <Container>
+      <div>
+        <TextContainer>
+          <Text variant="h4">Description</Text>
+          <Text variant="body1" secondary>
+            {channel?.description}
+          </Text>
```
drillprop: Don't know what designers think about it but I think we should omit description if there is no any.
This looks little bit weird. 
image
I mean, you don't need to change that, just want to raise an issue

kdembler:
This uses mocked data for `About` section at the moment, is there some separate issue to enable real data?

src/shared/components/ChannelCover/ChannelCover.style.ts
eldiegod: I think i fixed this

src/shared/components/ChannelCover/ChannelCover.style.ts
```diff
@@ -44,88 +42,11 @@ export const CoverImage = styled.div<CoverImageProps>`
   background-position: center;
   background-attachment: local;
   background-size: cover;
```
kdembler: Cannot comment on the line below for some reason but cannot we make this a regular `img` now? Should be simpler than background for `div`

src/shared/components/ChannelPreviewBase/ChannelPreviewBase.tsx
```diff
@@ -74,7 +74,7 @@ export const ChannelPreviewBase: React.FC<ChannelPreviewBaseProps> = ({
                       timeout={parseInt(transitions.timings.loading) * 0.5}
                       classNames={transitions.names.fade}
                     >
-                      <VideoCount variant="subtitle2">{videoCount ? `${videoCount} Uploads` : '⠀'}</VideoCount>
+                      <VideoCount variant="subtitle2">{`${videoCount ?? 0} Uploads`}</VideoCount>
```
kdembler: We resolve video count asynchronously so we shouldn't fallback to 0. This will lead to preview showing "0 uploads" and then changing to "X uploads" once loading is finished

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -1,27 +1,38 @@
...
+import { VideoOrderByInput } from '@/api/queries'
+import { LimitedWidthContainer, VideoPreview, ViewWrapper } from '@/components'
 import { AssetType, useAsset, usePersonalDataStore } from '@/providers'
-import { Button, ChannelCover } from '@/shared/components'
+import { ChannelCover, Grid, Pagination, Select, Tabs, Text } from '@/shared/components'
+import { SvgGlyphCheck, SvgGlyphPlus } from '@/shared/icons'
 import { transitions } from '@/shared/theme'
 import { Logger } from '@/utils/logger'
 import { formatNumberShort } from '@/utils/number'
+import { SORT_OPTIONS } from '@/views/studio/MyVideosView/MyVideosView'
```
kdembler: We should never import into view from another view as this can easily create unexpected import hierarchy. If `SORT_OPTIONS` is needed app-wide, let's extract it from My Videos and define in some generic place

src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -0,0 +1,113 @@
...
+        {!!channel?.description && (
+          <TextContainer>
+            <Text variant="h4">Description</Text>
+            <Text variant="body1" secondary>
+              {channel.description}
+            </Text>
+          </TextContainer>
+        )}
+        {/* hidding this until we can add links to channels */}
+        {/* <LinksContainer>
```
kdembler: Let's just remove this altogether for now

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -30,11 +41,28 @@ export const ChannelView: React.FC = () => {
   const followedChannels = usePersonalDataStore((state) => state.followedChannels)
   const updateChannelFollowing = usePersonalDataStore((state) => state.actions.updateChannelFollowing)
   const [isFollowing, setFollowing] = useState<boolean>()
+  const [currentVideosTab, setCurrentVideosTab] = useState(0)
+  const currentTabName = TABS[currentVideosTab]
+  const [sortVideosBy, setSortVideosBy] = useState<typeof SORT_OPTIONS[number]['value'] | undefined>(
```
kdembler: Why not just use `VideoOrderByInput` as type of state?

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -55,20 +83,50 @@ export const ChannelView: React.FC = () => {
...
+    { length: loadingVideos ? videosPerPage - (videos ? videos.length : 0) : 0 },
+    () => ({
+      id: undefined,
+      progress: undefined,
+    })
+  )
+  const videosWithPlaceholders = [...(videos || []), ...placeholderItems]
+  const mappedTabs = TABS.map((tab) => ({ name: tab, badgeNumber: 0 }))
+
+  console.log({ channel, videosWithPlaceholders, videos, loadingVideos, edges, videosError })
```
kdembler: Shouldn't be here I suppose?

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -83,15 +141,61 @@ export const ChannelView: React.FC = () => {
...
+                      itemsPerPage={videosPerPage}
+                      totalCount={totalCount}
+                    />
+                  </PaginationContainer>
+                </>
+              )
+            case 'About':
+              return <ChannelAbout />
+          }
+        })()}
```
kdembler: We shouldn't use a pattern like that, with invoking an anonymous function right away. Let's move that condition out of `return` entirely

kdembler (changes requested):
Let's extract logic for channel details out of this PR

src/shared/components/ChannelPreviewBase/ChannelPreviewBase.tsx
eldiegod: I don't think its the case cause the component transitions in when videoCount is truthy but we also don't need the 0 anyways

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: I think is a clean way to not have too many conditional operators, moving it out is also fine

kdembler:
Should also be rebased off `dev` to include component for animated loaders

src/utils/sorting.ts
```diff
@@ -0,0 +1,6 @@
+import { VideoOrderByInput } from '@/api/queries'
+
+export const SORT_OPTIONS = [
+  { name: 'Newest first', value: VideoOrderByInput.CreatedAtDesc },
+  { name: 'Oldest first', value: VideoOrderByInput.CreatedAtAsc },
+]
```
kdembler: I think this should rather go in `/config/`, it doesn't provide any utility

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -83,15 +165,36 @@ export const ChannelView: React.FC = () => {
             )}
           </TitleContainer>
           <StyledButtonContainer>
-            <Button variant={isFollowing ? 'secondary' : 'primary'} onClick={handleFollow} size="large">
+            <StyledButton
```
kdembler, line 168: Seems this is missing the confirmation from the designs? Can be no problem done in a separate PR, just want to make sure we don't miss that. https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=8171%3A210947

kdembler (changes requested):
(github lost my comment, will repost in a sec)

src/views/viewer/ChannelView/ChannelAbout.style.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+  ${media.small} {
+    gap: 79px;
+    grid-template-rows: initial;
+    grid-template-columns: minmax(0, 1fr) minmax(206px, 228px);
+  }
+  ${media.medium} {
+    gap: 94px;
+    grid-template-columns: minmax(0, 1fr) minmax(206px, 260px);
+  }
```
kdembler, line 32: Where do those specific pixel values (142, 79, 206, etc) come from? I would expect our design to use some sensible values from theme for spacing, is that not the case?

kdembler (changes requested):
here

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: lets do it in another PR

src/views/viewer/ChannelView/ChannelAbout.style.tsx
eldiegod: adjusted by a couple of pixels the ones that were not matching initially it works the same

src/views/viewer/ChannelView/ChannelView.tsx
kdembler: Sure. In cases like this, when followup is needed, we should make sure there is an issue in the milestone tracking this required work

src/views/viewer/ChannelView/ChannelAbout.style.tsx
```diff
@@ -0,0 +1,77 @@
...
+  ${media.small} {
+    gap: ${sizes(20)};
+    grid-template-rows: initial;
+    grid-template-columns: minmax(0, 1fr) minmax(${sizes(52)}, ${sizes(57)});
+  }
+  ${media.medium} {
+    gap: ${sizes(24)};
+    grid-template-columns: minmax(0, 1fr) minmax(${sizes(52)}, ${sizes(65)});
+  }
```
kdembler: When I brought up the magic numbers, I didn't mean replacing them with `sizes()`, just the fact that we have need for those values. `sizes(57)` is just as much a magic number as `228px` is :P I just expected the design to follow some generic spacing to achieve that value but looking at Figma, seems like this is really needed. In that case I think actually using those values directly may be more clear than multiples of 4.

kdembler (changes requested):
Some visual comments:

1. Seems the channel name can be cut off: <br>Screenshot 2021-07-21 at 10 55 48

1. Seems we're lacking some spacing for some details values (should have 8px margin) <br>Screenshot 2021-07-21 at 11 09 09

1. On mobile in "About", description should be below the details according to the design 

1. Seems we're using incorrect font style for details values (design has Inter there)

1. Design uses "Information" for the tab name instead of "About"

Other than above and one comment I think we're good to go. If you want we can merge as is (just with direct pixel values) to wrap up this work and tackle above points in followup

src/views/viewer/ChannelView/ChannelAbout.style.tsx
eldiegod: Yeah that was my thought as well

eldiegod:
Pushed the feedback fixes

kdembler (approved):

PR #980, opened 2021-07-05
Adjust carousel styles according to new designs

### LINKED ISSUE #954: Update carousel according to designs
https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=8915%3A4376

rafalpawlow:
Close #954


src/shared/components/Gallery/Gallery.tsx
```diff
@@ -10,15 +13,35 @@ export type GalleryProps = {
   className?: string
 } & CarouselProps
 
+interface ImperativeHandleData {
+  getPrevArrowProps: () => object
+  getNextArrowProps: () => object
+}
```
drillprop: We are using `type` everywhere, so I think we should stick to it for consistency.
Also, not sure, but we could replace `object` with `React.ComponentProps<typeof Arrow>`
```tsx
type ImperativeHandleData = {
  getPrevArrowProps: () => React.ComponentProps<typeof Arrow>
  getNextArrowProps: () => React.ComponentProps<typeof Arrow>
}
```
(sorry, wanted to add a suggestion, but for some reason I don't see the button on github)

src/shared/components/Carousel/Carousel.tsx
```diff
@@ -11,41 +9,60 @@ export type CarouselProps = {
...
+      arrowPosition,
+      slidesToShow = 'auto',
+      dotsVisible,
+      prevArrowRef,
+      nextArrowRef,
+      ...gliderOptions
+    },
+    ref
+  ) => {
+    // TODO: this is the only place in the app that requires refs to buttons. Once we refactor this component, we can remove forwardRef from buttons
```
drillprop: We should move this comment to `Gallery` if it still valid, right?

src/components/VideoGallery.tsx
```diff
@@ -82,16 +70,15 @@ export const VideoGallery: React.FC<VideoGalleryProps> = ({
       paddingLeft={sizes(2, true)}
       paddingTop={sizes(2, true)}
       responsive={breakpoints}
-      itemWidth={MIN_VIDEO_PREVIEW_WIDTH}
-      arrowPosition={arrowPosition}
+      itemWidth={281}
```
drillprop: Let's store this magic number in some constant variable, above component declaration.

drillprop (commented):
The code generally looks good, but the hover animation it's a little bit different in the design. We should update this also
Nagranie z ekranu 2021-07-6 o 08 45 20

src/shared/components/Gallery/Gallery.tsx
rafalpawlow: You are of course right, it was temporary and tbh I forgot about it 😉. I think your suggestion might do a job

src/shared/components/Carousel/Carousel.style.ts
```diff
@@ -75,3 +49,25 @@ export const GliderContainer = styled.div<HasPadding>`
...
+
+  .glider-dot {
+    background-color: ${colors.gray[700]};
+    width: ${sizes(8)};
+    height: ${sizes(1)};
+    border-radius: 0;
+    margin: 0 ${sizes(1)};
+    transition: all ${transitions.timings.regular} ${transitions.easing};
+
+    &:hover {
```
drillprop:
```suggestion
    &:hover:not(.active) {
```
This will make sure that we will not trigger the hover effect on buttons that are already active. I think it makes sense to do it that way.

drillprop (commented):
One last thing - is it possible to expand the clickable/hoverable area of `.glider-dot`? 
Right now it's really hard to click this(because it's so small).
I was trying to do that, but in the end, I started doing some dirty hacks with the linear gradient.
```tsx 
  .glider-dot {
    background: linear-gradient(transparent 40%, ${colors.gray[700]} 40% 60%, transparent 60%);
    width: ${sizes(8)};
    height: ${sizes(5)};
    border-radius: 0;
    margin: 0 ${sizes(1)};

    &:hover:not(.active) {
      background: linear-gradient(transparent 40%, ${colors.gray[50]} 40% 60%, transparent 60%);
      transform: translateY(-2px);
    }

    &.active {
      background: linear-gradient(transparent 40%, ${colors.gray[300]} 40% 60%, transparent 60%);
    }
  }
```
I mean, it works quite well ¯\_(ツ)_/¯
I leave it up to you, I can live without it

src/shared/components/Carousel/Carousel.style.ts
```diff
@@ -75,3 +49,37 @@ export const GliderContainer = styled.div<HasPadding>`
...
+  margin-top: ${sizes(13)};
+
+  .glider-dot {
+    background-color: transparent;
+    width: ${sizes(10)};
+    border-radius: 0;
+    padding: ${sizes(1)};
+    margin: 0;
+
+    &::after {
```
rafalpawlow, line 63: @drillprop I thought it might be more elegant than linear gradient. WDYT?

drillprop: yeah, looks good 👍🏻

drillprop (approved):

Issue #945, opened 2021-06-30
Buttons update


kdembler:
Buttons got a significant update inside Figma, with some new variants and some general changes. Let's make sure our code reflects that. This should include some API design first that will be discussed with the rest of the team

eldiegod:
Had to add some extra code to implement the icon only button

PR #983, opened 2021-07-07
Buttons update

eldiegod:
Issue: #945

### DIFF
--- a/src/shared/components/ButtonBase/ButtonBase.style.ts
+++ b/src/shared/components/ButtonBase/ButtonBase.style.ts
@@ -55,7 +55,8 @@ const variantStyles = ({ variant, textOnly, iconOnly }: ButtonBaseStyleProps): S
         color: ${colors.gray[50]};
 
         &:hover {
-          box-shadow: inset 0 0 0 2px ${colors.gray[50]};
+          box-shadow: inset 0 0 0 2px ${colors.gray[300]};
+          border-color: ${colors.gray[300]};
         }
 
         &:active {
--- a/src/shared/components/ButtonBase/ButtonBase.style.ts
+++ b/src/shared/components/ButtonBase/ButtonBase.style.ts
@@ -32,7 +32,8 @@ const variantStyles = ({ variant, textOnly, iconOnly }: ButtonBaseStyleProps): S
           fill: ${textOnly && colors.blue[300]};
         }
 
-        &:hover {
+        &:hover,
+        &:focus {
           background-color: ${colors.blue[700]};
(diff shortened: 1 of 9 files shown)


eldiegod:
Had to add some extra code to implement the icon only button

drillprop:
I just wanted to let you know:
You didn't have to close Mikołaj's PR, Klaudiusz gave you access to push someone else's PRs 😄   If you want to rerun checks you can do it by going to checks and clicking Re-run jobs(you should also have access to do that) 

Anyway, I fixed the problem with failing deploys

eldiegod:
> I just wanted to let you know:
> You didn't have to close Mikołaj PR, Klaudiusz gave you access to push someone PRs 😄 If you want to rerun checks you can do it by going to checks and clicking Re-run jobs(you should also have access to do that)

@drillprop yep you are 100% right 😅

Also thanks for fixing that annoying TS bug I was about to ask for help

src/shared/components/ButtonBase/ButtonBase.style.ts
```diff
@@ -4,44 +4,75 @@ import styled from '@emotion/styled'
...
     case 'secondary':
       return css`
         /* 1px inset border */
         box-shadow: inset 0 0 0 1px ${colors.gray[500]};
+        border-color: ${colors.gray[50]};
+        color: ${colors.gray[50]};
 
         &:hover {
-          box-shadow: inset 0 0 0 1px ${colors.gray[50]};
+          box-shadow: inset 0 0 0 2px ${colors.gray[50]};
```
drillprop:
```suggestion
          box-shadow: inset 0 0 0 2px ${colors.gray[300]};
          border-color: ${colors.gray[300]};
```
In the design colors are different for hover and active state. I'm a little confused why we are using box-shadow AND border. Maybe we should stick to just one of it(I think box-shadow would be preferable, because it doesn't change the size of button)

drillprop (commented):
Besides one thing looks good to me. Just FYI I also opened #984, I think we should do something about it at some point.

eldiegod:
@drillprop I think we don't even need the borders, they don't seem to be doing anything

eldiegod:
Added the focus states meanwhile we get response from design, might even be enough

drillprop:
I just added focus state to`textOnly` variant(didn't work earlier). Could you make sure I didn't break anything? I think I didn't.

eldiegod:
Looks like everything is working as expected

drillprop (approved):

PR #987, opened 2021-07-09
Adjust Video Preview component to new designs

### LINKED ISSUE #982: Update video tile design

rafalpawlow:
Close #982

### DIFF
--- a/src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
+++ b/src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
@@ -1,9 +1,9 @@
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
-import { fluidRange, transparentize } from 'polished'
+import { fluidRange } from 'polished'
 import { Link } from 'react-router-dom'
 
-import { colors, media, sizes, transitions, typography, zIndex } from '@/shared/theme'
+import { colors, media, sizes, square, transitions, typography, zIndex } from '@/shared/theme'
 
 import { Avatar } from '../Avatar'
 import { IconButton } from '../IconButton'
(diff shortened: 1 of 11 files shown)


src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -57,7 +64,11 @@ export const CoverContainer = styled.div<ClickableProps>`
   transition: all ${transitions.timings.regular} ${transitions.easing};
   cursor: ${(props) => (props.clickable ? 'pointer' : 'auto')};
 
-  :hover {
+  :active {
+    ${() => clickableAnimation(false)}
```
drillprop, line 68:
```suggestion
    ${clickableAnimation(false)}
```

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -208,22 +219,49 @@ export const CoverVideoPublishingStateOverlay = styled.div`
...
 export const PublishingStateText = styled(Text)`
   margin-left: ${sizes(1.5)};
 `
 
+export const KebabMenuIconContainer = styled.div<{ isActive?: boolean }>`
+  display: ${({ isActive }) => (isActive ? 'flex' : 'none')};
+  align-items: center;
+  justify-content: center;
+  width: 36px;
+  height: 36px;
```
drillprop: let's maybe use `sizes` here
```suggestion
  width: ${sizes(9}};
  height: ${sizes(9}};
```

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
```diff
@@ -208,22 +219,49 @@ export const CoverVideoPublishingStateOverlay = styled.div`
...
+  path {
+    transition: all ${transitions.timings.regular} ${transitions.easing};
+  }
+
+  &:hover {
+    path:not([fill='none']) {
+      fill: ${colors.white};
+    }
+
+    background-color: ${transparentize(1 - 0.06, colors.white)};
```
drillprop: I think we could use here `colors.transparentWhite[6]`. Also probably not your choice, but this particular color used in this context is barely visible. Do we plan changing background around video tiles in the future? Personally I don't like it,  maybe we could change it to more brighter variant?

drillprop (commented):
Generally looks good, found some small things

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
drillprop: My bad, I think I know why you did that. 
Hmm, let's just ignore this then

src/shared/components/VideoPreviewBase/VideoPreviewBase.styles.tsx
rafalpawlow: Good catch. After investigation, it turned out, there was a change in background color.

src/shared/theme/sizes.ts
```diff
@@ -8,11 +10,18 @@ export function sizes(n: number, raw?: boolean) {
...
   header: 100,
   sheetOverlay: 150,
   nearSheetOverlay: 160,
   sideNav: 200,
-  overlay: 10,
-  nearOverlay: 20,
   globalOverlay: 999,
 }
+
+export function square(size: string | number): SerializedStyles {
```
drillprop: nice utility function!
```suggestion
export function square(size: string | number) {
```
but we don't need to specify return type here.

drillprop (approved):

PR #994, opened 2021-07-13
Adjust InfiniteVideoGrid to handle query on demand, add LoadMoreButton component

rafalpawlow:
Close (#957)


drillprop (commented):
1. Space between tiles and button for some viewports should be bigger.
Yours
image
Design: 
image

2. I think we shouldn't show button if videos are still loading
Nagranie z ekranu 2021-07-14 o 17 44 54

drillprop (approved):

PR #996, opened 2021-07-14
Ending, loading & error player screen

### LINKED ISSUE #933: Update player loading screen
Design: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=9942%3A83 
Use Lottie animation, don't forget to add overlay(right now we don't have any).

### LINKED ISSUE #938: Add player error screen
Design: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=9931%3A252692 
Use Lottie animation

### LINKED ISSUE #939: Add player ending screen
Design:  https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=9981%3A259067
Probably will change in near future, needs more information

drillprop:
fix #939, #933, #938
Figma: https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=9977%3A263879
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=9164%3A0


kdembler:
It seems when there's an error, the player will go through different steps (loading, default videojs error message) and only then show our custom error message. Could we improve that?

https://user-images.githubusercontent.com/12646744/126159815-8a380b78-0949-4ee9-85ab-b75514b22cd2.mov

src/shared/components/VideoPlayer/ControlsIndicatorManager.tsx
```diff
@@ -0,0 +1,152 @@
...
+    return () => {
+      player.off(indicatorEvents, handler)
+    }
+  }, [cachedPlayerVolume, player])
+
+  return (
+    <ControlsIndicatorTransitions>
+      <CSSTransition
+        in={indicator?.isVisible}
+        timeout={indicator?.isVisible ? 0 : 750}
```
kdembler: Why do we need to switch timeout? Doesn't `in={indicator?.isVisible}` do the job?

src/shared/components/VideoPlayer/ControlsIndicatorManager.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const ControlsIndicatorManager: React.FC<ControlsIndicatorManagerProps> = ({ player, cachedPlayerVolume }) => {
+  const [indicator, setIndicator] = useState<EventState | null>(null)
+  useEffect(() => {
+    if (!player) {
+      return
+    }
+    const indicatorEvents = Object.values(CustomVideojsEvents)
+    const handler = (e: Event) => {
+      const playerVolume = e.type === CustomVideojsEvents.Unmuted ? cachedPlayerVolume || VOLUME_STEP : player.volume()
```
kdembler: Out of curiosity, why do we need `cachedPlayerVolume`, doesn't `player.volume()` do the trick?

src/shared/components/VideoPlayer/ControlsIndicatorManager.style.ts
```diff
@@ -0,0 +1,83 @@
...
+  ${media.small} {
+    top: calc(50% - ${sizes(16)});
+    left: calc(50% - ${sizes(16)});
+  }
+`
+
+export const ControlsIndicator = styled.div`
+  width: ${sizes(20)};
+  height: ${sizes(20)};
+  backdrop-filter: blur(${sizes(6)});
```
kdembler: What do we need this blur for? Seems support for `backdrop-filter` isn't great

src/shared/components/VideoPlayer/VideoOverlayManager.tsx
```diff
@@ -0,0 +1,75 @@
...
+    where: {
+      channelId_eq: channelId,
+      isPublic_eq: true,
+      mediaAvailability_eq: AssetAvailability.Accepted,
+    },
+  })
+
+  const getRandomNumber = useCallback((videosLength: number) => {
+    return getRandomIntInclusive(0, videosLength - 1)
+  }, [])
```
kdembler: This seems like just `getRandomIntExclusive`? We should either move it out of the component or directly to `utils`

src/shared/components/VideoPlayer/VideoOverlayManager.tsx
```diff
@@ -0,0 +1,75 @@
...
+}) => {
+  const { id } = useParams()
+  const [randomNextVideo, setRandomNextVideo] = useState<VideoFieldsFragment | null>(null)
+  const { videos } = useVideos({
+    where: {
+      channelId_eq: channelId,
+      isPublic_eq: true,
+      mediaAvailability_eq: AssetAvailability.Accepted,
+    },
+  })
```
kdembler: Is proposing next video to play part of the MVP? If that's the case we cannot longer keep `VideoPlayer` inside `shared`

src/shared/components/CircularProgressbar/CircularProgressbar.tsx
```diff
@@ -10,6 +10,7 @@ export const VIEWBOX_CENTER_Y = VIEWBOX_HEIGHT / 2
 
 export type CircularProgressbarProps = {
   value: number
+  trailColor?: string
```
kdembler: Maybe we can just define some variants instead of providing the color explicitly

src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.style.ts
```diff
@@ -0,0 +1,142 @@
...
+import { ChannelLink } from '@/components/ChannelLink'
+import { breakpoints, colors, media, sizes, zIndex } from '@/shared/theme'
+
+import { Button } from '../../Button'
+import { CircularProgressbar } from '../../CircularProgressbar'
+import { IconButton } from '../../IconButton'
+import { Text } from '../../Text'
+
+type OverlayBackgroundProps = {
+  thumbnail?: string | null
```
kdembler: Let's add `Url` suffix

src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -0,0 +1,125 @@
...
+  RestartButton,
+  StyledChannelLink,
+  StyledCircularProgressBar,
+  SubHeading,
+  VideoInfo,
+} from './EndingOverlay.style'
+
+type EndingOverlayProps = {
+  channelId?: string
+  currentThumbnail?: string | null
```
kdembler: Let's add `Url` suffix

src/shared/components/VideoPlayer/VideoOverlayManager.tsx
```diff
@@ -0,0 +1,75 @@
...
+  channelId?: string
+  currentThumbnail?: string | null
+}
+export const VideoOverlayManager: React.FC<VideoOverlayManagerProps> = ({
+  playerState,
+  onPlay,
+  channelId,
+  currentThumbnail,
+}) => {
+  const { id } = useParams()
```
kdembler: I think this could easily come from props as well right? That way we don't need to make this component connected to routing

src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -0,0 +1,125 @@
...
+  channelId,
+  currentThumbnail,
+  randomNextVideo,
+  isEnded,
+}) => {
+  const navigate = useNavigate()
+  const [countdownProgress, setCountdownProgress] = useState(0)
+  const [isCountDownStarted, setIsCountDownStarted] = useState(false)
+
+  const { url: randomNextVideoThumbnail } = useAsset({
```
kdembler: `Url`

src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  useEffect(() => {
+    if (!randomNextVideo || !isCountDownStarted) {
+      return
+    }
+    const timeout = setTimeout(() => {
+      setCountdownProgress(countdownProgress + 1)
+    }, 100)
+
+    if (countdownProgress === 100) {
```
kdembler: Ideally we'd have some kind of `NEXT_VIDEO_TIMEOUT` that we can change at top of the file

src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -0,0 +1,125 @@
...
+      setCountdownProgress(0)
+      setIsCountDownStarted(false)
+    }
+
+    return () => {
+      clearTimeout(timeout)
+    }
+  }, [countdownProgress, isCountDownStarted, isEnded, navigate, randomNextVideo])
+
+  const handleCountDownmButton = () => {
```
kdembler: Typo `Downm`

src/shared/components/VideoPlayer/VideoOverlays/ErrorOverlay.tsx
```diff
@@ -0,0 +1,37 @@
...
+  Heading,
+  InnerContainer,
+  OverlayBackground,
+  StyledAnimatedError,
+  StyledDiscordButton,
+} from './ErrorOverlay.style'
+
+import { Button } from '../../Button'
+
+export const ErrorOverlay = () => {
```
kdembler: Lacks `React.FC` typing

src/shared/components/VideoPlayer/VideoOverlays/ErrorOverlay.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <AnimationWrapper>
+          <StyledAnimatedError />
+        </AnimationWrapper>
+        <Heading variant="h3">Aw, shucks!</Heading>
+        <ErrorMessage variant="body1" secondary>
+          The video could not be loaded because of an error. Please try again later. If the issue persists, reach out to
+          our community on Discord.
+        </ErrorMessage>
+        <ButtonGroup>
+          <StyledDiscordButton variant="secondary" to="https://discord.gg/DE9UN3YpRP">
```
kdembler: Let's move that URL somewhere to `config/`

kdembler (changes requested):
Looks really good overall!

src/shared/components/VideoPlayer/ControlsIndicatorManager.tsx
drillprop: Maybe Im doing something wrong but I was trying to not switch that and end up with something like this: 
https://user-images.githubusercontent.com/51168865/126515803-537cec70-06b3-448b-85e6-484f6c9114f8.mov
If you look at video you'll see that animation continue when you spam the space button, only the icon is changing.

I don't  think it's desirable, we need cut animation somehow, and that's why I used this condition.

src/shared/components/VideoPlayer/ControlsIndicatorManager.tsx
drillprop: Because I was getting the stale value from `player.volume()`. 
If I replace this with just `player.volume` and click `M` to mute video and then click `M` again to unmute, the indicator will show `0%` volume(but will show correct sound on icon). 
I could workaround this with setTimeout:
```tsx
  useEffect(() => {
    if (!player) {
      return
    }
    let timeout: NodeJS.Timeout
    const indicatorEvents = Object.values(CustomVideojsEvents)

    const handler = (e: Event) => {
      timeout = setTimeout(() => {
        const indicator = createIndicator(e.type as VideoEvent, player.volume(), player.muted())
        if (indicator) {
          setIndicator({ ...indicator, isVisible: true })
        }
      }, 0)
    }
    player.on(indicatorEvents, handler)

    return () => {
      clearTimeout(timeout)
      player.off(indicatorEvents, handler)
    }
  }, [player])
```

Let me know what you think

src/shared/components/VideoPlayer/VideoOverlays/ErrorOverlay.tsx
drillprop: I read some time ago that `React.FC` is considered as bad practice. You can read about why here: https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/  Even CRA team removed that from the template: https://github.com/facebook/create-react-app/pull/8177 (we also don't have React.FC in App.tsx 😅 )

I'll fix that, but maybe it's worth to discuss this?

drillprop: Maybe we could use `React.FC` only when it's actually needed e.g when we are using `children` in a component?

src/shared/components/VideoPlayer/ControlsIndicatorManager.style.ts
drillprop: We used backdrop blur on several elements: buttons, tooltips, indicators: https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=9977%3A263957 
It looks pretty cool on browsers which support it. I would keep it. On firefox(which is only one major browser where we cant use backdrop-filter) user will see just the background with reduced opacity, which is not much different and it don't break anything in layout in terms of look or positioning.
Chrome:
image
Firefox:
image

src/shared/components/VideoPlayer/VideoOverlayManager.tsx
drillprop: We decided with Adam to add the next video feature as part of MVP after all. I think it was agreed with Bedeho also. We will need to refactor this in the future of course. 
Don't understand why I can't hold VideoPlayer in `shared` though? Where should I move it? To the `viewer/VideosView`?

src/shared/components/VideoPlayer/VideoOverlays/ErrorOverlay.tsx
kdembler: I don't really buy the reasoning from the article with exception of the "forced children" point. I don't really see drawback that `React.FC` brings but it adds some value. We can discuss further but for now let's just stay consistent and type everything

src/shared/components/VideoPlayer/VideoOverlayManager.tsx
kdembler: Regarding including it in the MVP, sounds good!

Regarding keeping it in `shared`, we shouldn't have any components there that rely on our data layer, as that's Atlas-specific and those components aren't generic anymore. But actually let's ignore that for the moment, as I think we need to do a bigger cleanup of `shared` anyway

src/shared/components/VideoPlayer/ControlsIndicatorManager.style.ts
kdembler: Okay, I buy it

src/shared/components/VideoPlayer/ControlsIndicatorManager.tsx
kdembler: Yeah that's a good argument, let's keep it

src/shared/components/VideoPlayer/VideoOverlayManager.tsx
drillprop: that makes sense. I'll leave the component inside shared for now.

src/shared/components/VideoPlayer/ControlsIndicator.tsx
```diff
@@ -0,0 +1,155 @@
...
+    let timeout: NodeJS.Timeout
+    const indicatorEvents = Object.values(CustomVideojsEvents)
+
+    const handler = (e: Event) => {
+      timeout = setTimeout(() => {
+        const indicator = createIndicator(e.type as VideoEvent, player.volume(), player.muted())
+        if (indicator) {
+          setIndicator({ ...indicator, isVisible: true })
+        }
+      }, 0)
```
kdembler: I think this works fine, let's just add a comment on top to explain why we do it this way

src/shared/components/VideoPlayer/ControlsIndicator.tsx
```diff
@@ -0,0 +1,155 @@
...
+  player: VideoJsPlayer | null
+}
+
+export const ControlsIndicator: React.FC<ControlsIndicatorProps> = ({ player }) => {
+  const [indicator, setIndicator] = useState<EventState | null>(null)
+  useEffect(() => {
+    if (!player) {
+      return
+    }
+    let timeout: NodeJS.Timeout
```
kdembler: We're not in nodejs environment, let's not use this type. `setTimeout` returns a regular `number`

drillprop: hmm, number doesn't work.
image
image
I'll just use `ReturnType<typeof setTimeout>`

drillprop: Oh I see, to make this work I would have to write `window.setTimeout`, will stick to ReturnType

kdembler: Actually, thinking about this now, we actually should use `window.setTimeout`, `setTimeout` isn't a global function and it only works because by default CRA webpack will polyfill node globals

kdembler (approved):
Great stuff!

PR #1011, opened 2021-07-19
don't retry asset indefinitely

### LINKED ISSUE #1005: Infinite loop of error messages
When visiting https://play.joystream.org/video/1962, I will get an error message about some assets failing to load and the errors will keep piling up, it seems we have some infinite loop there

kdembler:
fix #1005


drillprop (approved):
Looks good to me!

PR #1017, opened 2021-07-20
add storage provider fallback, add asset response timeout

### LINKED ISSUE #1002: Implement storage provider fallback for assets
I am getting a substantial share of videos looking like this

Screenshot 2021-07-17 at 21 05 49

I was told by senior community people that the cause is that some storage providers are having trouble with all the new load, and they are re-synching with greater capacity (which apparently takes many days already!!), and Atlas does not deal properly with this.

Could we instead introduce some behaviour where Atlas tries another storage provider if there is a failure?

kdembler:
Resolves #1002 

Adds fallback to different storage provider if the first tried failed to respond. Also adds a timeout that forces storage provider to respond within 5 seconds, otherwise other storage provider is tried automatically. This was added to alleviate the problem with some storage providers taking even 30secs to respond

### DIFF
--- a/src/config/assets.ts
+++ b/src/config/assets.ts
@@ -1,0 +1,1 @@
+export const ASSET_RESPONSE_TIMEOUT = 5000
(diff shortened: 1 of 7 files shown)


drillprop (approved):
Looks and works really well now! 👍🏻

PR #1021, opened 2021-07-21
Custom timeline & timeline tooltip

### LINKED ISSUE #937: Add tooltip to player controls
Design: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=10005%3A263359

drillprop:
Fix #937 
Figma: https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=9977%3A263879
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=9164%3A0

I really hate the fact that I had to patch the video.js library, but it works surprisingly well ¯\_(ツ)_/¯


patches/video.js+7.13.3.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/video.js/dist/video.es.js b/node_modules/video.js/dist/video.es.js
+index a07aa77..3f7bc30 100644
+--- a/node_modules/video.js/dist/video.es.js
++++ b/node_modules/video.js/dist/video.es.js
+@@ -13959,7 +13959,7 @@ var ProgressControl = /*#__PURE__*/function (_Component) {
+     }
+ 
+     var seekBarEl = seekBar.el();
+-    var seekBarRect = findPosition(seekBarEl);
++    var seekBarRect = getBoundingClientRect(seekBarEl);
```
drillprop: This should fix the video.js bug, where the timeline tooltip is getting out of bounds on the right. Found it here https://github.com/videojs/video.js/pull/7308

patches/video.js+7.13.3.patch
```diff
@@ -1,8 +1,33 @@
...
+     // further to the right to center it over the `seekBarPoint`.
+ 
+     var pullTooltipBy = tooltipRect.width / 2; // Adjust the `pullTooltipBy` distance to the left or right depending on
+@@ -13349,7 +13349,8 @@ var MouseTimeDisplay = /*#__PURE__*/function (_Component) {
+ 
+     var time = seekBarPoint * this.player_.duration();
+     this.getChild('timeTooltip').updateTime(seekBarRect, seekBarPoint, time, function () {
+-      _this2.el_.style.left = seekBarRect.width * seekBarPoint + "px";
++      _this2.el_.style.right = seekBarRect.width * (1 -seekBarPoint) + "px";
++      _this2.el_.style.width = seekBarRect.width * seekBarPoint + "px";
```
drillprop: this will increase the size of the mouse time display(small 1px vertical line on timeline bar which is visible when you moving the cursor) 
Before
image
Now
image

Design
image

drillprop:
Draft. I will create our custom timeline bar and replace video.js one

src/shared/components/VideoPlayer/CustomTimeline.tsx
```diff
@@ -0,0 +1,155 @@
...
+      // position of playProgressThumb
+    }, UPDATE_INTERVAL)
+    return () => {
+      clearInterval(interval)
+    }
+  }, [isScrubbing, player, playerState])
+
+  const handleMouseAndTouchMove = (e: React.MouseEvent | React.TouchEvent) => {
+    const seekBar = e.currentTarget.querySelector(SeekBar.toString())
+    const mouseDisplayTooltip = e.currentTarget.querySelector(MouseDisplayTooltip.toString())
```
kdembler: Why? If we cannot go without those components, let's just use ref

src/shared/components/VideoPlayer/CustomTimeline.tsx
```diff
@@ -0,0 +1,155 @@
...
+    if (!seekBar || !mouseDisplayTooltip || !player) {
+      return
+    }
+    // this will prevent hiding controls when scrubbing on mobile
+    player.enableTouchActivity()
+
+    const duration = player.duration()
+
+    // position of seekBar
+    const { x: seekbarPosition, width: seekBarWidth } = seekBar.getBoundingClientRect()
```
kdembler: Let's use common naming case - `seekBarPosition`

src/shared/components/VideoPlayer/CustomTimeline.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const handleJumpToTime = (e: React.MouseEvent) => {
+    if (isScrubbing) {
+      return
+    }
+    const seekBar = e.currentTarget.querySelector(SeekBar.toString())
+    if (!seekBar) {
+      return
+    }
+
+    const { x: seekbarPosition, width: seekBarWidth } = seekBar.getBoundingClientRect()
```
kdembler: Likewise, `seekBarPosition`

src/shared/components/VideoPlayer/CustomTimeline.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  useEffect(() => {
+    if (!player) {
+      return
+    }
+    const handler = () => {
+      const duration = player.duration()
+      const bufferedEnd = player.bufferedEnd()
+      const loadProgressPercentage = round((bufferedEnd / duration) * 100, 2)
+      setLoadProgressWidth(loadProgressPercentage)
```
kdembler: Are we sure this logic is correct? I've noticed weird behaviour, even if I scrub to time that is allegedly already buffered, it still has to load:

https://user-images.githubusercontent.com/12646744/127139664-b85f32b6-0331-40d1-b5ac-e26ee1987c2d.mov

This could be just buffering behaviour of video.js but if I look at the old player, clicking on the buffered area doesn't cause any loading, may be worth investigating

kdembler (changes requested):

kdembler:
Some other UX things I just noticed:
1. When the user is scrubbing I don't think we should start playback until they let go of the thumb.
2. When the user is scrubbing, we likely shouldn't let the timeline collapse, this can lead to some weird experience:

https://user-images.githubusercontent.com/12646744/127141245-42d5f4eb-4e80-4207-a341-b72bc0199ce9.mov

3. It would be nice to ensure we don't select text when scrubbing but not sure how to achieve that, doesn't have to be included

src/shared/components/VideoPlayer/CustomTimeline.tsx
drillprop: should be fixed. It's not super perfect, but the native `video.js` timeline is not perfect either. https://github.com/videojs/video.js/blob/4238f5c1d88890547153e7e1de7bd0d1d8e0b236/src/js/control-bar/progress-control/load-progress-bar.js This could direct you why I did that. Basically player.buffer() returns an object with just two methods(start, end) and property `length`. `Length` tells how many buffered time ranges are present in the video timeline, I just need to find and take the one which is close to the current video time.

drillprop:
> Some other UX things I just noticed:
> 
> 1. When the user is scrubbing I don't think we should start playback until they let go of the thumb.
> 2. When the user is scrubbing, we likely shouldn't let the timeline collapse, this can lead to some weird experience:
> 
>  Screen.Recording.2021-07-27.at.12.42.16.mov 
> 1. It would be nice to ensure we don't select text when scrubbing but not sure how to achieve that, doesn't have to be included

1. Fixed
2. Fixed, basically when you hold mouse down, you can now move mouse below the timeline bar and on the video. Kinda wanted to mimic youtube behavior. I mean, it's not exactly like youtube, but I think it's enough
3. You can still find some edge cases, but it's generally fixed.

kdembler (approved):

PR #1028, opened 2021-07-22
Fix orion requests issue in video ending view

drillprop:


src/api/hooks/video.ts
```diff
@@ -52,3 +55,15 @@ export const useAddVideoView = (opts?: AddVideoViewOpts) => {
     ...rest,
   }
 }
+
+type VideosWithBasicInformationOpts = QueryHookOptions<GetVideosWithBasicInformationQuery>
+export const useVideosWithBasicInformation = (
```
kdembler: Maybe just `useBasicVideos`? We already use a similar name - `useBasicChannel`

src/api/queries/videos.graphql
```diff
@@ -11,6 +11,16 @@ fragment LicenseFields on License {
   customText
 }
 
+fragment VideoBasicFields on Video {
```
kdembler: Likewise, we currently use `BasicChannelFields`

kdembler (approved):

PR #1031, opened 2021-07-22
Search channel view

### LINKED ISSUE #953: Channel view search feature
Includes updating our input component to optionally accept an icon

eldiegod:
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=10854%3A314594
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=10474%3A305579
closes: #953


src/shared/icons/svgs/outline-search2.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M11 11L14 14" stroke="#F4F6F8" stroke-width="2"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M8 14C11.3137 14 14 11.3137 14 8C14 4.68629 11.3137 2 8 2C4.68629 2 2 4.68629 2 8C2 11.3137 4.68629 14 8 14ZM8 12C10.2091 12 12 10.2091 12 8C12 5.79086 10.2091 4 8 4C5.79086 4 4 5.79086 4 8C4 10.2091 5.79086 12 8 12Z" fill="#F4F6F8"/>
+</svg>
```
drillprop:
```suggestion
</svg>

```
I see on github:
image

I really dont know what's the consequences of not fixing that, but let's just fix it 😄

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -126,15 +133,82 @@ export const PaginationContainer = styled.div`
...
+    border-bottom: solid 1px ${colors.gray[800]};
+  }
+  ${media.small} {
+    border-bottom: none;
+  }
   ${media.medium} {
-    display: grid;
-    grid-template-columns: 1fr 250px;
+    grid-area: initial;
   }
```
drillprop, line 189: Not sure if it's in the scope of this PR, but I see in the design hover state for Tabs
image

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -126,15 +133,82 @@ export const PaginationContainer = styled.div`
...
+
+type TextFieldProps = {
+  isOpen?: boolean
+}
+export const StyledTextField = styled(TextField)<TextFieldProps>`
+  transition: all ${transitions.timings.regular} ${transitions.easing};
+  will-change: max-width;
+  width: 100%;
+  align-items: center;
+  max-width: ${({ isOpen }) => (isOpen ? '192px' : '0px')};
```
drillprop, line 198: Input is much bigger in the design 
image

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -126,15 +133,82 @@ export const PaginationContainer = styled.div`
...
+    grid-template-areas:
+      'tabs tabs '
+      'search  sort';
+  }
+  ${media.medium} {
+    grid-template-areas: initial;
+    gap: ${sizes(8)};
+    grid-template-rows: 1fr;
+    grid-template-columns: auto 1fr 250px;
+  }
```
drillprop, line 158: Not sure if it's in the scope of this PR, but I see in the design hover state for Tabs
image

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -126,15 +133,82 @@ export const PaginationContainer = styled.div`
...
+  transition: all ${transitions.timings.regular} ${transitions.easing};
+  will-change: max-width;
+  width: 100%;
+  align-items: center;
+  max-width: ${({ isOpen }) => (isOpen ? '192px' : '0px')};
+
+  > input {
+    padding: 10px 16px 10px 42px;
+    border: ${({ isOpen }) => isOpen == false && 'none !important'};
+
```
drillprop, line 209: we should update :focus state for input. Border in the design is white

src/shared/components/Button/Button.tsx
```diff
@@ -8,7 +8,7 @@ import { TextVariant } from '../Text'
 export type ButtonProps = {
   icon?: React.ReactNode
   iconPlacement?: IconPlacement
-  children: string
+  children?: string
```
drillprop: I don't think you should edit this. You should use `IconButton` instead of regular `Button` and pass outline as children like this 
```js
// ChannelView.tsx
              <SearchButton
                onClick={() => {
                  setIsSearchingInputOpen(true)
                  searchInputRef.current?.focus()
                }}
                variant="tertiary"
              >
                <SvgOutlineSearch2 />
              </SearchButton>
```
```js
//ChannelView.style.ts

export const SearchButton = styled(IconButton)`
  position: absolute;
`
```

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -176,8 +216,35 @@ export const ChannelView: React.FC = () => {
...
+                helperText={null}
+              />
+              <SearchButton
+                onClick={() => {
+                  setIsSearchingInputOpen(true)
+                  searchInputRef.current?.focus()
+                }}
+                variant="tertiary"
+                icon={<SvgOutlineSearch2 />}
+              ></SearchButton>
```
drillprop: as I mentioned before, use IconButton
```suggestion
              <SearchButton
                onClick={() => {
                  setIsSearchingInputOpen(true)
                  searchInputRef.current?.focus()
                }}
                variant="tertiary"
              >
                <SvgOutlineSearch2 />
              </SearchButton>
```

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -192,9 +259,60 @@ export const ChannelView: React.FC = () => {
...
+type useSearchVideosParams = {
+  id: string
+}
+const useSearchVideos = ({ id }: useSearchVideosParams) => {
+  const [isSearchInputOpen, setIsSearchingInputOpen] = useState(false)
+  const [searchQuery, setSearchQuery] = useState('')
+  const [triggeredSearchQuery, setTriggeredSearchQuery] = useState('')
+  const [isSearching, setIsSearching] = useState(false)
+  const searchInputRef = useRef<HTMLInputElement>(null)
+  // TODO: bug this is not querying by channel Id and results don't seem accurate
```
drillprop: I wanted to pointing this out. It seems that we are also getting results from other channels

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -192,9 +259,60 @@ export const ChannelView: React.FC = () => {
...
+  const { data: searchData, loading: loadingSearch, error: errorSearch } = useSearch({
+    text: triggeredSearchQuery,
+    whereVideo: {
+      isPublic_eq: true,
+      mediaAvailability_eq: AssetAvailability.Accepted,
+      thumbnailPhotoAvailability_eq: AssetAvailability.Accepted,
+    },
+    whereChannel: {
+      id_in: [id],
```
drillprop: you should update `extendedQueryNode` and add `id_eq` field. Maybe it will fix the issue

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -85,33 +103,55 @@ export const ChannelView: React.FC = () => {
...
     }
   }
   const handleOnResizeGrid = (sizes: number[]) => setVideosPerRow(sizes.length)
   const handleChangePage = (page: number) => {
-    setCurrentPage(page)
+    isSearching ? setCurrentSearchPage(page) : setCurrentPage(page)
   }
-  const handleSetCurrentTab = async (tab: number) => {
-    setCurrentVideosTab(tab)
+  const handleKeyPress = (event: React.KeyboardEvent<HTMLInputElement>) => {
```
drillprop: Maybe we could wrap this input in form, I think we will not need all these keyboard events If Im not mistaken

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -176,8 +216,35 @@ export const ChannelView: React.FC = () => {
...
         <TabsContainer>
-          <Tabs initialIndex={0} tabs={mappedTabs} onSelectTab={handleSetCurrentTab} />
+          <StyledTabs
+            selected={isSearching ? -1 : undefined}
+            initialIndex={0}
+            tabs={mappedTabs}
+            onSelectTab={handleSetCurrentTab}
+          />
           {currentTabName === 'Videos' && (
+            <SearchContainer>
```
drillprop, line 247: As above - this could be a form element

src/views/viewer/ChannelView/ChannelView.style.tsx
kdembler: Good catch. Doesn't have to be in this PR but let's create a followup issue to tackle that @eldiegod

kdembler: One thing I'd like to request here, I think the Information tab should have it's own URL so you can link directly to channel details. Maybe something like `/channel/1` / `/channel/1?tab=videos` for videos and `/channel/1?tab=info` for information?

Again, probably not in scope for this PR but let's create an issue for that as well

src/views/viewer/ChannelView/ChannelView.style.tsx
eldiegod: I think thats an old design:
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=10474%3A305579

src/views/viewer/ChannelView/ChannelView.style.tsx
eldiegod: Made a new issue for the hover states and the URL paths

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: fixed now

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: It didn't work at the end we did need the keypresses

drillprop: but why?
I've just applied to your code: 
```tsx
// SearchContainer needs to be form element
          <SearchContainer
              onSubmit={(e) => {
                e.preventDefault()
                if (searchQuery.trim() === '') {
                  setSearchQuery('')
                  setIsSearching(false)
                } else {
                  search()
                  setIsSearching(true)
                }
              }}
            >
              <StyledTextField
                ref={searchInputRef}
                isOpen={isSearchInputOpen}
                value={searchQuery}
                onBlur={() => {
                  setIsSearchingInputOpen(false)
                  setSearchQuery('')
                }}
                onChange={(e) => setSearchQuery(e.target.value)}
                // onKeyDown={handleKeyPress}
                placeholder="Search"
                type="search"
                helperText={null}
              />

```
and everything works just fine. I've added onBlur event too. Not sure why we need an event that is firing on the ESC key, but I guess in the end we could keep it. 

~~Also, I still think we should use apollo's `lazy query` here~~.

drillprop:
There is one more UX-related thing. If you search videos, there is no way to reset that search and see all the videos. You need to erase input and hit enter or click `videos` again. I found this very non-intuitive.
Maybe, after all, we should fire search every time user is writing, ie. during an onChange event? I think it could work, but we need to make sure to use debouncing here

src/views/viewer/ChannelView/ChannelView.tsx
drillprop: My bad didn't notice that you used lazy query after all 😅

eldiegod: Yep ESC stuff was by design on the figma prototype

eldiegod:
> There is one more UX-related thing. If you search videos, there is no way to reset that search and see all the videos. You need to erase input and hit enter or click `videos` again. I found this very non-intuitive.
> Maybe, after all, we should fire search every time user is writing, ie. during an onChange event? I think it could work, but we need to make sure to use debouncing here

I think you can just click on the videos tab to see all the videos

drillprop:
Yeah, I know - but it is still misleading. 
For example - I want to search videos with a golden retriever. I go to the channel click the search button type `golden` and hit enter and I see this: 
image
Everything is fine so far. 
But now I need to come back to the main videos. I click videos and I see this. And I'm still not sure if I'm in `search mode` or not. I still see typed `golden` in input.  What if the first ten videos were about golden retrievers? How I could know if I'm not searching? 
image

Not sure what to do with that. We could reset input in this case but I don't think it's enough. Maybe it's worth discussing this with a designer, or maybe I'm just exaggerating and see problems where there are none

eldiegod:
> Yeah, I know - but it is still misleading.
> For example - I want to search videos with a golden retriever. I go to the channel click the search button type `golden` and hit enter and I see this:
> image
> Everything is fine so far.
> But now I need to come back to the main videos. I click videos and I see this. And I'm still not sure if I'm in `search mode` or not. I still see typed `golden` in input. What if the first ten videos were about golden retrievers? How I could know if I'm not searching?
> image
> 
> Not sure what to do with that. We could reset input in this case but I don't think it's enough. Maybe it's worth discussing this with a designer, or maybe I'm just exaggerating and see problems where there are none

While is not *super* clear that the user is in search or on a normal videos tab what design did for this is when the user is searching no tab will be active and when they click back on videos videos tab is highlighted which gives away they are no longer on the search ghost tab, it's similar to youtube except youtube also shows a slightly different video layout

kdembler:
@KubaMikolajczyk want to chime in here?

eldiegod:
@kdembler @drillprop very valid discussion but I do think we should probably merge and figure out what to do on a follow-up?

KubaMikolajczyk:
> Yeah, I know - but it is still misleading.
> For example - I want to search videos with a golden retriever. I go to the channel click the search button type `golden` and hit enter and I see this:
> image
> Everything is fine so far.
> But now I need to come back to the main videos. I click videos and I see this. And I'm still not sure if I'm in `search mode` or not. I still see typed `golden` in input. What if the first ten videos were about golden retrievers? How I could know if I'm not searching?
> image
> 
> Not sure what to do with that. We could reset input in this case but I don't think it's enough. Maybe it's worth discussing this with a designer, or maybe I'm just exaggerating and see problems where there are none

From my pov. when we go back to videos tab we could reset and close the search input (in this case clear the word golden and close the input)- this way user is aware that it is no longer a part of searching experience.

eldiegod:
I like that as well

src/shared/icons/svgs/outline-search2.svg
```diff
@@ -0,0 +1,8 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<circle cx="10" cy="10" r="8" stroke="#F4F6F8" stroke-width="2"/>
+<path d="M22 22L18 18" stroke="#F4F6F8" stroke-width="2" stroke-miterlimit="10"/>
+</svg>
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M11 11L14 14" stroke="#F4F6F8" stroke-width="2"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M8 14C11.3137 14 14 11.3137 14 8C14 4.68629 11.3137 2 8 2C4.68629 2 2 4.68629 2 8C2 11.3137 4.68629 14 8 14ZM8 12C10.2091 12 12 10.2091 12 8C12 5.79086 10.2091 4 8 4C5.79086 4 4 5.79086 4 8C4 10.2091 5.79086 12 8 12Z" fill="#F4F6F8"/>
+</svg>
```
kdembler: Why does this contain 2 `svg` elements? This doen't seem right, we should only have one which is 16x16

It also shouldn't be named `outline-search2`, this variant is a glyph

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -126,15 +132,93 @@ export const PaginationContainer = styled.div`
...
+export const StyledTextField = styled(TextField)<TextFieldProps>`
+  transition: all ${transitions.timings.regular} ${transitions.easing};
+  will-change: max-width;
+  width: 100%;
+  align-items: center;
+  max-width: ${({ isOpen }) => (isOpen ? '192px' : '0px')};
+
+  > input {
+    padding: 10px 16px 10px 42px;
+    border: ${({ isOpen }) => (isOpen == false ? 'none !important' : `1px solid ${colors.gray[300]};`)};
```
kdembler: Why do we need those? Shouldn't this use defaults from the `TextField` component?

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -126,15 +132,93 @@ export const PaginationContainer = styled.div`
...
+    caret-color: ${colors.blue[500]};
+
+    &:focus {
+      border: 1px solid ${colors.white};
+    }
```
kdembler, line 208: Likewise, this likely should be coming from the component?

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -71,18 +99,25 @@ export const ChannelView: React.FC = () => {
...
           error: true,
-          description: `Do you want to unfollow ${channel?.title}?`,
+          title: 'Would you consider staying?',
+          description: (
+            <>
+              {channel?.title} released <DialogAccentText>{videosLastMonth} new videos </DialogAccentText>
+              this month.
+              <br /> Cancel to follow for more fresh content!
+            </>
+          ),
```
kdembler, line 111: Where does this come from? I don't see this in the design

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -50,7 +75,9 @@ export const ChannelView: React.FC = () => {
     entity: channel,
     assetType: AssetType.COVER,
   })
-  const { currentPage, setCurrentPage } = usePagination(0)
+  const [dateOneMonthPast] = useState(subMonths(new Date(), 1))
```
kdembler: I guess you could move it outside of the component as a constant

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -100,33 +135,55 @@ export const ChannelView: React.FC = () => {
...
   }
   const videosPerPage = ROWS_AMOUNT * videosPerRow
 
-  const videos = edges
-    ?.map((edge) => edge.node)
-    .slice(currentPage * videosPerPage, currentPage * videosPerPage + videosPerPage)
+  const videos = ((isSearching ? searchVideos : edges?.map((edge) => edge.node)) ?? []).slice(
+    (isSearching ? currentSearchPage : currentPage) * videosPerPage,
+    (isSearching ? currentSearchPage : currentPage) * videosPerPage + videosPerPage
+  )
```
kdembler: It's difficult to understand this logic, let's split it into some more steps

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -207,9 +292,65 @@ export const ChannelView: React.FC = () => {
...
   const [currentPage, setCurrentPage] = useState(0)
+  const [currentSearchPage, setCurrentSearchPage] = useState(0)
   // reset the pagination when changing tabs
   useEffect(() => {
     setCurrentPage(0)
   }, [currentTab])
-  return { currentPage, setCurrentPage }
+  useEffect(() => {
+    setCurrentSearchPage(0)
+  }, [currentSearchPage])
```
kdembler: This hook says "whenever search page change, set search page to 0"

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -207,9 +292,65 @@ export const ChannelView: React.FC = () => {
...
+  return { currentPage, setCurrentPage, currentSearchPage, setCurrentSearchPage }
+}
+
+type useSearchVideosParams = {
```
kdembler: type name should be capitalized

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -207,9 +292,65 @@ export const ChannelView: React.FC = () => {
...
+      variables: {
+        text: searchQuery,
+        whereVideo: {
+          isPublic_eq: true,
+          mediaAvailability_eq: AssetAvailability.Accepted,
+          thumbnailPhotoAvailability_eq: AssetAvailability.Accepted,
+          channelId_eq: id,
+        },
+      },
+    })
```
kdembler: This needs to specify a limit. The default for the search is 5 videos so this is how much we will show currently. Also, we need to somehow get items count to enable proper pagination

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -207,9 +292,65 @@ export const ChannelView: React.FC = () => {
...
+  useEffect(() => {
+    setCurrentSearchPage(0)
+  }, [currentSearchPage])
+  return { currentPage, setCurrentPage, currentSearchPage, setCurrentSearchPage }
+}
+
+type useSearchVideosParams = {
+  id: string
+}
+const useSearchVideos = ({ id }: useSearchVideosParams) => {
```
kdembler: What's the exact purpose of this hook? I understand that it's supposed to make the component look cleaner and extract some logic out of it. It seems though that it's not really defined what goes into it and what doesn't. For example, search state and refs are defined here but the keypress is defined in the component. I suggest you figure out what you want from this hook and clean this up

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -207,9 +292,65 @@ export const ChannelView: React.FC = () => {
...
+    })
+  }
+  const getVideosFromSearch = (loading: boolean, data: SearchQuery['search'] | undefined) => {
+    if (loading || !data) {
+      return { channels: [], videos: [] }
+    }
+    const results = data
+    const searchVideos = results.flatMap((result) => (result.item.__typename === 'Video' ? [result.item] : []))
+    return { searchVideos }
+  }
```
kdembler: This could be defined as constant outside of hook, right?

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -207,9 +292,65 @@ export const ChannelView: React.FC = () => {
...
+          channelId_eq: id,
+        },
+      },
+    })
+  }
+  const getVideosFromSearch = (loading: boolean, data: SearchQuery['search'] | undefined) => {
+    if (loading || !data) {
+      return { channels: [], videos: [] }
+    }
+    const results = data
```
kdembler: What do we need this variable for?

kdembler (changes requested):

src/views/viewer/ChannelView/ChannelView.style.tsx
eldiegod: This is so we can collapse the input when closed but the second part of the ternary operator we don't need so I will remove

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -126,15 +132,93 @@ export const PaginationContainer = styled.div`
 `
 
 export const TabsContainer = styled.div`
```
kdembler, line 134: Don't know where exactly the problem is but in implementation the search input goes to the second row on a bigger screen size than in the design

src/views/viewer/ChannelView/ChannelView.style.tsx
eldiegod: If you look at this prototype of the searchbar it has some on focus styles that we don't have on the base component
https://www.figma.com/proto/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?page-id=10474%3A305579&node-id=10586%3A292658&viewport=496%2C1603%2C0.4263942539691925&scaling=min-zoom

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: It's spread over several files sadly https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=10582%3A289758

src/views/viewer/ChannelView/ChannelView.style.tsx
eldiegod: not 100% sure this is the case for what you are mentioning here, but I do have noticed our media query breakpoints don't match designs so that causes differences

src/views/viewer/ChannelView/ChannelView.tsx
kdembler: @KubaMikolajczyk let's try to keep prototype consistent with the design 🙏

kdembler: Also I'm a bit unsure about this message. The point of the dialog is to make sure user doesn't unfollow by accident. This looks more like trying to keep the user to follow. If they didn't click the button by accident why would we try to persuade them to not unfollow? I personally don't like this type of communication from the app and I think it should be used only if the user wants to stop paying us or something, if ever.

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: notin

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: Limit where? the SearchResults component doesn't do this?

src/views/viewer/ChannelView/ChannelView.style.tsx
kdembler: Hmm maybe yeah, we don't have a breakpoint close to `768px` like the design does

src/views/viewer/ChannelView/ChannelView.tsx
KubaMikolajczyk: @kdembler I promise to make it consistent - where is it inconsistent? 

Firstly - there was a wording mistake on my end "keep following for.." instead "cancel to follow". sorry :)

If it comes to the dialog itself - I hear you - we could only ask if the user wants to unfollow and thats it. The reason I wanted to introduce this pattern was the fact that currently each engaged user is really valuable - when user decides to unfollow - both viewer and publisher loses some added value to their experience with the platform.

Please notice that we don't use any dark pattern here like switching the primary action to keep following and confusing user - just asking question and showing some additional information that may be helpful for the viewer.

Maybe @bedeho could add something to this discussion?

I made a second version also: [Link](https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=11921%3A424903)

image

kdembler: Regarding inconsistency, [here](https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=8171%3A210739) a different dialog is used.

I like the second one more, but I see your point as well. I think we will need a tiebreaker @bedeho :D

src/views/viewer/ChannelView/ChannelView.tsx
kdembler: Default comes from the query node and we will need to supply it inside the search args. You're right that `SearchResults` doesn't do this and it's a bug (https://github.com/Joystream/atlas/issues/1055)

eldiegod: hmm, not sure what to do in this PR then? increase the scope to add the limit thing?

kdembler: Well, channel search won't work properly so we need to add the limit. If you manage to fix the regular search at the same time, even better

eldiegod: Made limit 100 which is more than enough for now but we might have to properly refactor the search so it gives totalCount and can be paginated properly

kdembler (approved):

PR #1033, opened 2021-07-22
fix URL assets, fix mocking issue

kdembler:


drillprop (approved):
Works good, glad to have mocking back!

PR #1037, opened 2021-07-23
Change videos query orderBy argument to array

rafalpawlow:


kdembler:
Converting to draft for now since query node deployment was reverted and we will likely also need to adjust other queries

drillprop (approved):
Looks good to me!

PR #1046, opened 2021-07-26
Add ranking number to carousel

rafalpawlow:


src/components/VideoGallery.tsx
```diff
@@ -57,7 +45,33 @@ export const VideoGallery: React.FC<VideoGalleryProps> = ({
...
+  hasRanking = true,
 }) => {
+  const breakpoints = useMemo(() => {
+    return breakpointsOfGrid({
+      breakpoints: 6,
+      minItemWidth: 300,
+      gridColumnGap: 24,
+      viewportContainerDifference: 64,
+    }).map((breakpoint, idx) => {
+      if (breakpoint <= 688 && hasRanking) {
```
kdembler: What is this `688`?

kdembler (changes requested):
Seems the placement of the number could be improved on some screen sizes:
Screenshot 2021-07-26 at 15 42 41
Could we make the overlap fixed and instead add some spacing to the left if needed?

Also, if we get to 10, it doesn't look good:
Screenshot 2021-07-26 at 15 42 33

src/components/VideoGallery.tsx
rafalpawlow: This value comes from this function. 
```
364
688
1012
1336
1660
1984
```

rafalpawlow:
@kdembler About values above 10, this is because `letter-spacing: -0.25em;` which comes from figma. I've talked to Kuba about this, and he asked to make it overlap like this. I'll try to spread it out a little bit.

src/components/VideoGallery.tsx
kdembler: This doesn't tell me anything at all :P

kdembler (approved):
Looking good!

drillprop (commented):
I think we should change letter-spacing a bit. It's still not looking great: 
image

`-0.1em` maybe?

PR #1051, opened 2021-07-26
Channel view

eldiegod:
PR for Channel view update milestone


src/shared/components/ChannelPreviewBase/ChannelPreviewBase.tsx
```diff
@@ -74,7 +74,7 @@ export const ChannelPreviewBase: React.FC<ChannelPreviewBaseProps> = ({
                       timeout={parseInt(transitions.timings.loading) * 0.5}
                       classNames={transitions.names.fade}
                     >
-                      <VideoCount variant="subtitle2">{videoCount ? `${videoCount} Uploads` : '⠀'}</VideoCount>
+                      <VideoCount variant="subtitle2">{`${videoCount} Uploads`}</VideoCount>
```
kdembler: This can read `undefined Uploads` at times, needs to be fixed

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -83,15 +233,134 @@ export const ChannelView: React.FC = () => {
...
+              <StyledTextField
+                ref={searchInputRef}
+                isOpen={isSearchInputOpen}
+                value={searchQuery}
+                onChange={(e) => setSearchQuery(e.target.value)}
+                onKeyDown={handleSearchInputKeyPress}
+                placeholder="Search"
+                type="search"
+                helperText={null}
+              />
```
kdembler: We need to better separate concerns in this component. Because this is all a single component and the data flows the way it does, every keystroke inside the search field will cause the entire `ChannelView` to re-render. Try holding a single key inside the input and see what happens - the page becomes unresponsive. If you change the data flow and keep state of this input inside a separate component and only pass value to `ChannelView` on submit, we can avoid a lot of unnecessary re-rendering

kdembler (changes requested):

src/api/client/resolvers.ts
```diff
@@ -115,6 +126,29 @@ export const queryNodeStitchingResolvers = (
   },
 
   Channel: {
+    views: async (parent, args, context, info) => {
```
kdembler, line 217: There has been some improvements made to these resolvers on `dev`. We should rebase this branch and do something similar for channel views. So make sure that we don't need to fetch them on one-to-one basis but use the batched query instead. This will save us a lot of requests to Orion

eldiegod: So far we only query views for individual channels for their respective views, we for sure will need this but I think it makes more sense in a different PR issue?

kdembler: If you go to `/channels` on your branch, there will be a separate Orion request made for each channel, to fetch views. We should fix this before merging

kdembler: @drillprop did some clean up on `dev` for those so should be very straightforward to introduce batched query

src/api/client/resolvers.ts
```diff
@@ -17,10 +17,18 @@ import {
...
+  ORION_CHANNEL_VIEWS_QUERY_NAME,
+  TransformBatchedChannelOrionViewsField,
+  TransformOrionChannelViewsField,
+} from './transforms/orionViews'
+import { RemoveQueryNodeChannelViewsField } from './transforms/queryNodeViews'
 
 import { Channel, ChannelEdge, Video, VideoEdge } from '../queries'
 
 const BATCHED_VIDEO_VIEWS_QUERY_NAME = 'GetBatchedVideoViews'
+const BATCHED_CHANNEL_VIEWS_QUERY_NAME = 'GetBatchedVideoViews'
```
kdembler: This value doesn't seem right

kdembler (approved):

PR #1052, opened 2021-07-27
Tabs hover styles

### LINKED ISSUE #1038: Hover state for Tabs component
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=8543%3A217

eldiegod:
solves: #1038

### DIFF
--- a/src/shared/components/Tabs/Tabs.styles.tsx
+++ b/src/shared/components/Tabs/Tabs.styles.tsx
@@ -33,7 +33,7 @@ export const TabsGroup = styled.div`
 `
 
 export const Tab = styled.div<TabProps>`
-  transition: all 0.125s ease;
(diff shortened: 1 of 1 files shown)


src/shared/components/Tabs/Tabs.styles.tsx
```diff
@@ -33,6 +33,7 @@ export const TabsGroup = styled.div`
 `
 
 export const Tab = styled.div<TabProps>`
+  transition: all 0.125s ease;
```
drillprop: I think better practice is not to use `all` here, let's just add separate transitions to: `color` and `border-bottom`.

drillprop (approved):
Looks good to me!

Issue #1039, opened 2021-07-23
Channel view tabs should be reflected in URL


eldiegod:
Something like /channel/1?tab=info or /channel/1?tab=videos, maybe we could even add search as well

PR #1053, opened 2021-07-27
Channel view tabs reflected on URL

eldiegod:
solves: #1039

### DIFF
--- a/src/views/viewer/ChannelView/ChannelView.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.tsx
@@ -49,7 +49,7 @@ const ROWS_AMOUNT = 4
 export const ChannelView: React.FC = () => {
   const [openUnfollowDialog, closeUnfollowDialog] = useDialog()
   const { id } = useParams()
-  const [searchParams, setSearchParams] = useSearchParams({ tab: [...TABS] })
+  const [searchParams, setSearchParams] = useSearchParams()
   const { channel, loading, error } = useChannel(id)
   const {
     searchVideos,
@@ -205,7 +205,7 @@ export const ChannelView: React.FC = () => {
 
   // At mount set the tab from the search params
   useEffect(() => {
-    const tabIndex = TABS.findIndex((t) => t === searchParams.get('tab'))
+    const tabIndex = TABS.findIndex((t) => t === currentTabName)
     if (tabIndex === -1) setSearchParams({ 'tab': 'Videos' }, { replace: true })
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
@@ -245,7 +245,7 @@ export const ChannelView: React.FC = () => {
         </TitleSection>
         <TabsContainer>
           <StyledTabs
-            selected={isSearching ? -1 : TABS.findIndex((x) => x === searchParams.get('tab'))}
+            selected={isSearching ? -1 : TABS.findIndex((x) => x === currentTabName)}
             initialIndex={0}
             tabs={mappedTabs}
             onSelectTab={handleSetCurrentTab}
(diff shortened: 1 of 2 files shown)


eldiegod:
to merge after search-channel-view

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -68,8 +69,8 @@ export const ChannelView: React.FC = () => {
   const followedChannels = usePersonalDataStore((state) => state.followedChannels)
   const updateChannelFollowing = usePersonalDataStore((state) => state.actions.updateChannelFollowing)
   const [isFollowing, setFollowing] = useState<boolean>()
-  const [currentVideosTab, setCurrentVideosTab] = useState(0)
-  const currentTabName = TABS[currentVideosTab]
+  const [currentTabIndex, setCurrentTabIndex] = useState(0)
```
kdembler: Why do we need this index? I think we can just use the state from the search params?

eldiegod: I think thats right, removed it

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -49,6 +49,7 @@ const ROWS_AMOUNT = 4
 export const ChannelView: React.FC = () => {
   const [openUnfollowDialog, closeUnfollowDialog] = useDialog()
   const { id } = useParams()
+  const [searchParams, setSearchParams] = useSearchParams({ tab: [...TABS] })
```
kdembler: What is this parameter supposed to do? Why are we passing an array?

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -203,6 +203,13 @@ export const ChannelView: React.FC = () => {
     }
   }
 
+  // At mount set the tab from the search params
+  useEffect(() => {
+    const tabIndex = TABS.findIndex((t) => t === searchParams.get('tab'))
+    if (tabIndex === -1) setSearchParams({ 'tab': 'Videos' }, { replace: true })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kdembler, line 210: Why disable the deps? If `searchParams` gets updated externally, we can still navigate, right?

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -203,6 +203,13 @@ export const ChannelView: React.FC = () => {
     }
   }
 
+  // At mount set the tab from the search params
+  useEffect(() => {
+    const tabIndex = TABS.findIndex((t) => t === searchParams.get('tab'))
```
kdembler: We already created `currentTabName`, let's use that

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -238,7 +245,7 @@ export const ChannelView: React.FC = () => {
         </TitleSection>
         <TabsContainer>
           <StyledTabs
-            selected={isSearching ? -1 : undefined}
+            selected={isSearching ? -1 : TABS.findIndex((x) => x === searchParams.get('tab'))}
```
kdembler: We already created `currentTabName`, let's use that

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: thought it might enter an infinite loop since it's changed inside the effect but actually no, it works fine

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: I was trying something but I noticed it isn't actually needed so I'm removing it

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: nvm, removing deps again because it bugs out when clicking a video, we only need this effect to run on mount and set the searchParams 'tab' to 'Videos' in case there is  no searchParams set or it's an incorrect one so after mount we don't do anything in it

kdembler (approved):

Issue #1036, opened 2021-07-23
Current player issues


drillprop:
This is applying for the `video-player-update` branch. Creating this issue to not lose those. 

- [x] Animation of the loader is a little bit different in the design. We should make sure that we follow design here(use scale, and opacity)
- [x]  If it's possible, we should hide loader when shortcuts indicators showing up.
- [x] For some reason current tooltip doesn't have `backdrop-filter` on Chrome. It works on safari though, so it could be a chrome-related bug. Possibly we cannot apply backdrop-filter to `::before` or children, but it's worth investigating this. If it's impossible to fix, let's just leave it as it is.
- [x] When we focus on the player control and then move the cursor to another player button we should hide the tooltip on the previous player control.
- [x] There is weird vertical padding on the indicator's tooltip
- [x] Tooltips: 
- font-size, font-family and line-height are different
- padding is different - it should be 8px
- color is different - we should use `colors.gray[50]`
- [x]  When the screen with the big play button is showing up, we should be able to play the video by clicking the video container, not just the big play button. 
- [x] Saving volume in local storage doesn't work perfectly. If the user mutes the video, information about setting volume to 0 will not be saved. We should fix this

PR #1054, opened 2021-07-27
Tooltip fixes

drillprop:
Partly resolves: #1036 

Should fix: 

> - [ ] Tooltips: 
> - font-size, font-family and line-height are different
> - padding is different - it should be 8px
> - color is different - we should use `colors.gray[50]`

and: 

> - [ ] When we focus on the player control and then move the cursor to another player button we should hide the tooltip on the previous player control.

Unfortunately, the backdrop-filter problem is a chrome bug. https://stackoverflow.com/questions/62454801/css-backdrop-filter-not-functioning-on-child-element-in-slideout-menu 
I could work around this by making a tooltip a sibling, but it will require some refactoring and will make our HTML code even more complicated. Not sure if it's worth it. Let me know what you think?

### DIFF
--- a/src/shared/components/VideoPlayer/PlayerControlButton.style.ts
+++ b/src/shared/components/VideoPlayer/PlayerControlButton.style.ts
@@ -89,7 +89,7 @@ export const ControlButton = styled.button<ControlButtonProps>`
     }
   }
 
-  ${focusStyles}
+  ${focusStyles};
 `
 
 type ControlButtonTooltipProps = {
@@ -103,7 +103,7 @@ export const ControlButtonTooltip = styled.div<ControlButtonTooltipProps>`
   opacity: 0;
   pointer-events: none;
   position: absolute;
-  bottom: calc(3em);
+  bottom: 3em;
   background: ${colors.transparentBlack[54]};
   backdrop-filter: blur(${sizes(8)});
   display: flex;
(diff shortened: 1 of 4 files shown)


src/shared/components/VideoPlayer/PlayerControlButton.style.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  :focus:not(:focus-visible):not(:hover) {
+    ${() => ControlButtonTooltip} {
+      opacity: 0;
+    }
+  }
+
+  ${focusStyles}
```
kdembler: Should have semicolon at line end

src/shared/components/VideoPlayer/PlayerControlButton.style.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export const ControlButtonTooltip = styled.div<ControlButtonTooltipProps>`
+  ${({ tooltipPosition }) => tooltipPosition === 'left' && 'left: 0'};
+  ${({ tooltipPosition }) => tooltipPosition === 'right' && 'right: 0'};
+
+  opacity: 0;
+  pointer-events: none;
+  position: absolute;
+  bottom: calc(3em);
```
kdembler: Why `calc`?

kdembler (commented):
> Unfortunately, the backdrop-filter problem is a chrome bug

I would leave it out

src/shared/components/VideoPlayer/PlayerControlButton.style.ts
drillprop: Decided to remove this. Did some refactoring also

kdembler (approved):

PR #1061, opened 2021-07-28
Player mobile view

### LINKED ISSUE #1006: Update player compact version
Designs here: 
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=11253%3A349268
Make sure to hide all controls which are not needed for mobile, and every functionality work as it should be(changing volume, playing/pausing on tap, etc)

drillprop:
fix #1006 

I will open this as a draft for now. Some things differs from original design because they just didn't work good and I need to discuss them with a designer.

### DIFF
--- a/src/shared/components/VideoPlayer/ControlsIndicator.style.ts
+++ b/src/shared/components/VideoPlayer/ControlsIndicator.style.ts
@@ -5,12 +5,12 @@ import { colors, media, sizes, transitions } from '@/shared/theme'
 import { Loader } from '../Loader'
 
 export const ControlsIndicatorWrapper = styled.div`
-  position: absolute;
-  width: 100%;
-  height: 100%;
-  pointer-events: none;
   display: none;
-  ${media.small} {
+  @media (hover: hover) {
+    pointer-events: none;
+    width: 100%;
+    height: 100%;
+    position: absolute;
     display: flex;
     flex-direction: column;
     align-items: center;
--- a/src/shared/components/VideoPlayer/ControlsIndicator.style.ts
+++ b/src/shared/components/VideoPlayer/ControlsIndicator.style.ts
@@ -92,9 +92,11 @@ export const ControlsIndicatorTransitions = styled.div`
     ${() => StyledLoader} {
       opacity: 1;
       transform: scale(1);
-      transition: opacity 150ms ease-in, transform 150ms ease-in;
+      transition: opacity, transform;
+      transition-timing-function: ease-in;
+      transition-duration: 150ms;
       @media (hover: hover) {
-        transition: opacity 150ms 600ms ease-in, transform 150ms 600ms ease-in;
+        transition-delay: ${INDICATOR_TIMEOUT - 150}ms;
       }
     }
   }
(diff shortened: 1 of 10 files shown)


kdembler (approved):

src/shared/components/VideoPlayer/ControlsIndicator.style.ts
```diff
@@ -63,30 +81,33 @@ export const ControlsIndicatorTooltip = styled.div`
...
 
-  .indicator-enter-active {
+  .${INDICATOR_TRANSITION}-enter-active {
     ${() => StyledLoader} {
       opacity: 1;
       transform: scale(1);
-      transition: opacity ${transitions.timings.sharp} ease-in, transform ${transitions.timings.sharp} ease-in;
+      transition: opacity 150ms ease-in, transform 150ms ease-in;
+      @media (hover: hover) {
+        transition: opacity 150ms 600ms ease-in, transform 150ms 600ms ease-in;
```
kdembler: Is this valid value for `transition`? The syntax is `property name | duration | easing function | delay`, what's that second `ms` value supposed to represent?

kdembler: Btw, maybe at this point it would be more clear to explicitly define specific properties, then you don't have to repeat the values inside the media query. For example:
```css
transition-property: opacity, transform;
transition-duration: 150ms;
transition-timing-function: ease-in;
@media (hover: hover) {
  transition-delay: 600ms;
}
```
(Not sure if this is exactly what you wanted to achieve but you get the point)

src/shared/components/VideoPlayer/CustomTimeline.style.ts
```diff
@@ -41,35 +46,38 @@ export const ProgressControl = styled.div<ProgressControlProps>`
...
+    ${({ isScrubbing }) => isScrubbing && `opacity: 1;`}
   }
   ${() => MouseDisplayTooltip} {
-    ${({ isScrubbing }) => isScrubbing && `transform: translateY(-0.5em) !important`}
+    ${({ isScrubbing }) => isScrubbing && `transform: translateY(-0.5em);`}
   }
   ~ ${CustomControls} {
-    ${({ isScrubbing }) => isScrubbing && `opacity: 0; transform: translateY(0.5em) !important`};
+    ${({ isScrubbing }) => isScrubbing && `opacity: 0; transform: translateY(0.5em);`}
   }
```
kdembler, line 69: Maybe it would be more clear to define those scrubbing styles in a separate block? Like so:

```jsx
const scrubbingStyles = css`
  ${() => SeekBar} {
    height: 0.5em;
  }
  ${() => MouseDisplayWrapper}, ${() => PlayProgressThumb} {
    opacity: 1;
  }
  // ...
`

export const ProgressControl = styled.div<ProgressControlProps>`
  ${({ isScrubbing }) => isScrubbing && scrubbingStyles};
`
```

What do you think?

drillprop: I don't think it's possible with emotion `css`. Typescript is throwing me errors, and if I ignored these errors I would get:
image

To make this work I would have to pass component as an argument to `scrubbingStyles`, something like this: 
```tsx
const scrubbingStyles = (seekBar: typeof SeekBar, /* other components here */ isFullScreen?: boolean) => css`
  height: 100vh;
  bottom: ${isFullScreen ? 0 : '-200px'};
  padding-bottom: ${isFullScreen ? `1.5em 1.5em` : '200px'};
  ${seekBar} {
    height: 0.5em;
  }

  /* other components here */
`
```

PR #1069, opened 2021-07-29
Pagination fix

### LINKED ISSUE #1035: Paginated video grid not fetching additional items
When I'm in channel videos and go to a page that contains the last of 50 initially fetched videos, nothing happens and more videos don't get loaded. I guess it's possible the same happens in Studio->My videos. We should likely extract those to a common component and make sure it works properly

eldiegod:
fixes: #1035

### DIFF
--- a/src/views/viewer/ChannelView/ChannelView.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.tsx
@@ -97,6 +97,9 @@ export const ChannelView: React.FC = () => {
       where: {
         channelId_eq: id,
         isPublic_eq: true,
+        isCensored_eq: false,
+        thumbnailPhotoAvailability_eq: AssetAvailability.Accepted,
+        mediaAvailability_eq: AssetAvailability.Accepted,
       },
     },
     { notifyOnNetworkStatusChange: true }
--- a/src/views/viewer/ChannelView/ChannelView.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.tsx
@@ -9,7 +9,13 @@ import {
   useUnfollowChannel,
   useVideosConnection,
 } from '@/api/hooks'
-import { AssetAvailability, SearchQuery, VideoOrderByInput, useSearchLazyQuery } from '@/api/queries'
+import {
+  AssetAvailability,
+  SearchQuery,
+  VideoFieldsFragment,
+  VideoOrderByInput,
+  useSearchLazyQuery,
+} from '@/api/queries'
 import { LimitedWidthContainer, VideoPreview, ViewWrapper } from '@/components'
 import { SORT_OPTIONS } from '@/config/sorting'
 import { AssetType, useAsset, useDialog, usePersonalDataStore } from '@/providers'
@@ -75,7 +81,16 @@ export const ChannelView: React.FC = () => {
     assetType: AssetType.COVER,
   })
   const { currentPage, setCurrentPage, currentSearchPage, setCurrentSearchPage } = usePagination(0)
-  const { edges, totalCount, loading: loadingVideos, error: videosError, refetch } = useVideosConnection(
+  const {
+    edges,
+    totalCount,
+    loading: loadingVideos,
+    error: videosError,
+    fetchMore,
+    refetch,
+    variables,
+    pageInfo,
+  } = useVideosConnection(
     {
       first: INITIAL_FIRST,
       orderBy: sortVideosBy,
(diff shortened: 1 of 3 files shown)


kdembler (changes requested):
This contained other changes previously that were removed I think? It doesn't fix the problem

eldiegod:
> This contained other changes previously that were removed I think? It doesn't fix the problem

hmm, wonder where did my commit go 🤷‍♂️

eldiegod:
In the end decided to handle this on the pagination handler instead of the useEffect as it makes more sense to do it there

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -155,7 +170,20 @@ export const ChannelView: React.FC = () => {
...
-    isSearching ? setCurrentSearchPage(page) : setCurrentPage(page)
+    if (isSearching) {
+      setCurrentSearchPage(page)
+    } else {
+      setCurrentPage(page)
+      if (!!edges && page * videosPerPage + videosPerPage > edges?.length) {
+        fetchMore({
+          variables: {
+            ...variables,
+            first: page * videosPerPage + videosPerPage * 3,
```
kdembler: What is the reasoning behind that logic? Why not always fetch a constant amount? Currently the number of fetched items will depend on page and can get quite high

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -86,7 +101,7 @@ export const ChannelView: React.FC = () => {
         isPublic_eq: true,
       },
     },
-    { notifyOnNetworkStatusChange: true, fetchPolicy: 'cache-and-network' }
+    { notifyOnNetworkStatusChange: true }
```
kdembler, line 105: Why do we need `notifyOnNetworkStatusChange`?

kdembler (changes requested):

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: Needed because when `fetchingMore` we need the loading flag to show the placeholders; without `notifyOnNetworkStatusChange` we don't get the loading flag

src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: It's an issue with how `first` variable works, it just starts from the beginning, unless I'm doing something wrong, I have to tell it how many videos I need from page 1 to the page I'm currently at, otherwise, if I go to let's say page 100 and it only loads 50 more videos, so 50 initial + 50 now, it won't be enough videos to populate the page, so that's why I'm not using a constant amount.

Not sure I explained this the best but there it is

eldiegod: Also, the `* 3` is there because I load videos for the current page and 2 pages ahead to avoid showing loading components at every page change

kdembler: `first` shouldn't work this way, you are specifying `after` after all. With this logic you first fetch 50, then 50 + 50, then 50 + 50 + 50. Each fetch should be a separate 50

eldiegod: so how does it looks when we start on page 1 then go to page 8 and then go back to page 4? assuming 16 vids per page

kdembler (changes requested):
The pagination looks good now but I think it introduced a new error. Try going to https://deploy-preview-1069--atlas-dev.netlify.app/channels and click any channel with existing uploads. What I see is that the channel doesn't have any videos, you can see just pagination:
Screenshot 2021-08-04 at 09 51 57

If you refresh, or go directly to channel link, everything works fine

eldiegod:
> If you refresh, or go directly to channel link, everything works fine

@kdembler 
Hard to debug, something obscure is going on under the hood:

- I'm only able to reproduce it from the channels view everywhere else that I click on a channel it just works.

- The issue disappears if I set the `fetchPolicy` to `network-only`

- The graphql query response looks like this: {"data":{"videosConnection":{"totalCount":1,"__typename":"VideoConnection"}}} A.K.A no edges come with it, hence the issue.

Looks like a cache issue, could it be that the cache thinks it has the data for the channel videos since it probably fetched something in channelsViews and doesn't fetches the videos because of this?

eldiegod:
Pushed a fix changing the fetch policy and not using fetch more which works good but there is definitely something going on with the queries unrelated to the PR

Issue #1027, opened 2021-07-22
Extend Dialog component to accept new button props /design


eldiegod:
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=8171%3A210947

PR #1070, opened 2021-07-29
Dialog api improvements

eldiegod:
solves: #1027 
And although it is no longer needed I think it makes the dialog API clearer by separating the props for the primary and secondary buttons so worth taking a look 
cc @kdembler

### DIFF
--- a/src/components/Dialogs/ActionDialog/ActionDialog.stories.tsx
+++ b/src/components/Dialogs/ActionDialog/ActionDialog.stories.tsx
@@ -34,6 +34,8 @@ export default {
 
 type StoryProps = ActionDialogProps & {
   showAdditionalAction?: boolean
+  primaryButtonText?: string
+  secondaryButtonText?: string
 }
 
 const additionalActionNode = (
@@ -49,7 +51,7 @@ const content = (
   </div>
 )
 
-const RegularTemplate: Story<StoryProps & any> = ({
+const RegularTemplate: Story<StoryProps> = ({
   showAdditionalAction,
   primaryButtonText,
   secondaryButtonText,
@@ -67,7 +69,7 @@ const RegularTemplate: Story<StoryProps & any> = ({
 }
 export const Regular = RegularTemplate.bind({})
 
-const ContentTemplate: Story<StoryProps & any> = ({
+const ContentTemplate: Story<StoryProps> = ({
   showAdditionalAction,
   primaryButtonText,
   secondaryButtonText,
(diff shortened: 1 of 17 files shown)


src/shared/components/Button/Button.tsx
```diff
@@ -8,7 +8,7 @@ import { TextVariant } from '../Text'
 export type ButtonProps = {
   icon?: React.ReactNode
   iconPlacement?: IconPlacement
-  children: string
+  children?: string
```
drillprop: I would still keep the children required and modify props type in `ActionDialog` by omitting `children` like this: 
```tsx
// ActionDialog.tsx
type DialogButtonProps = {
  text: string
  disabled?: boolean
  onClick?: (e: React.MouseEvent) => void
} & Omit<ButtonProps, 'children'>
```
I think it makes sense to do it that way. We don't need to pass children to primary button(or secondary button), because we already have a `text` property

drillprop (commented):
I really like that change! Found one thing to consider.

src/shared/components/Button/Button.tsx
eldiegod: Yeah I think this is a good idea

drillprop (approved):

src/components/Dialogs/ActionDialog/ActionDialog.tsx
```diff
@@ -11,32 +11,30 @@ import {
...
+} & Omit<ButtonProps, 'children'>
+
 export type ActionDialogProps = {
   additionalActionsNode?: React.ReactNode
-  primaryButtonText?: string
-  secondaryButtonText?: string
-  primaryButtonDisabled?: boolean
-  secondaryButtonDisabled?: boolean
-  onPrimaryButtonClick?: (e: React.MouseEvent) => void
-  onSecondaryButtonClick?: (e: React.MouseEvent) => void
```
kdembler, line 21: We need to make necessary changes in the storybook to make the story still usable. After this change you will see only text input for the buttons, with no clear way of making changes

kdembler (changes requested):

src/components/Dialogs/ActionDialog/ActionDialog.stories.tsx
```diff
@@ -47,30 +49,59 @@ const content = (
   </div>
 )
 
-const RegularTemplate: Story<StoryProps> = ({ showAdditionalAction, ...args }) => {
+const RegularTemplate: Story<StoryProps & any> = ({
```
kdembler: 1. I'd rather have us type this properly
2. `StoryProps & any` is just `any`
3. If we must use `any`, we should disable the ESLint warning

kdembler (approved):

PR #1081, opened 2021-08-02
Reduce orion requests

### LINKED ISSUE #1029: Reduce the number of requests to Orion when using useVideos
The problem which is mentioned in https://github.com/Joystream/atlas/issues/38 is still present for `Videos` query.
When we are running useVideos, we are fetching separetely views from orion for each one video.
We should update out resolvers like we did with `VideosConnection`.

drillprop:
Fix #1029
Also:
- created a playground for Orion testing.
- updated channels query(added limit and offset arguments)
- updated `channels` resolver and add resolver for `channelConnection`'s edges to reduce orion request when fetching channel's follows
- updated mocking


src/mocking/accessors/filtering.ts
```diff
@@ -80,11 +80,11 @@ const createPredicate = (key: string, value: any, testData: GenericData): Predic
 
 export const genericSort = <TData extends GenericData>(data: TData[], variables: SortingArgs): TData[] => {
   const { orderBy } = variables
-  if (!orderBy) {
+  if (!orderBy?.length) {
     return data
   }
 
-  const [field, direction] = orderBy.split('_')
+  const [field, direction] = orderBy[0].split('_')
```
drillprop, line 87: I'm a little confused why `orderBy` is now array? This works for now because we are passing only one element to `orderBy` everywhere, but I assume it's probably wrong.

kdembler: My guess is it became an array so we can do multiple rules for sorting. So if the first rule places the items in the same position, you can use the second to determine the final order

kdembler: But for mocking we can safely just use the first one I think

src/api/client/resolvers.ts
```diff
@@ -114,8 +169,38 @@ export const queryNodeStitchingResolvers = (
...
+        node: {
+          ...edge.node,
+          follows: followsLookup[edge.node.id]?.views || 0,
+        },
+      }))
+    },
+  },
   Channel: {
     follows: async (parent, args, context, info) => {
+      if (parent.follows != null) {
```
rafalpawlow: Is this double comparison instead of triple is intentional?

drillprop: It is, but on second thought, I think it could be confusing. I will update that.

kdembler (approved):
Nice job on that, good idea to introduce an Orion playground

PR #1093, opened 2021-08-02
Add popular videos section on popular page

rafalpawlow:


rafalpawlow:
I've converted it to draft because I realized it should use most popular videos of all time query.

src/api/hooks/video.ts
```diff
@@ -109,3 +112,50 @@ export const useMostViewedVideos = (variables?: GetMostViewedVideosQueryVariable
...
+  opts?: MostViewedVideosAllTimeOpts
+) => {
+  const { mostViewedVideosAllTime } = useMostViewedVideosAllTimeIds(variables, opts)
+
+  const mostViewedVideosAllTimeIds = useMemo(() => {
+    if (mostViewedVideosAllTime) {
+      return mostViewedVideosAllTime.map((item) => item.id)
+    }
+    return null
+  }, [mostViewedVideosAllTime])
```
kdembler: `useMemo` likely doesn't make much sense here, the performance cost of memoization will likely be higher than the computation being done

src/api/hooks/video.ts
```diff
@@ -109,3 +112,50 @@ export const useMostViewedVideos = (variables?: GetMostViewedVideosQueryVariable
...
+      where: {
+        id_in: mostViewedVideosAllTimeIds,
+      },
+    },
+    { skip: !mostViewedVideosAllTimeIds }
+  )
+
+  const sortedVideos = useMemo(() => {
+    if (videos) {
+      return [...videos].sort((a, b) => (b.views && a.views ? b.views - a.views : 0))
```
kdembler: If I'm not mistaken, because of `b.views && a.views` condition, if one of the items doesn't have views but the other does, the sort may be incorrect right?

src/api/hooks/video.ts
```diff
@@ -109,3 +112,50 @@ export const useMostViewedVideos = (variables?: GetMostViewedVideosQueryVariable
...
+  opts?: MostViewedVideosAllTimeOpts
+) => {
+  const { data, ...rest } = useGetMostViewedVideosAllTimeQuery({ ...opts, variables })
+  return {
+    mostViewedVideosAllTime: data?.mostViewedVideosAllTime,
+    ...rest,
+  }
+}
+
+export const useMostViewedVideosAllTime = (
```
kdembler: Where is this hook used?

kdembler (approved):

PR #1104, opened 2021-08-04
Fix player's seeking issue on IOS 

### LINKED ISSUE #1091: Changing time in mobile player causes delay
Take a look at this recording. Note that when changing the time, I didn't do any scrubbing, just tapping the timeline:

https://user-images.githubusercontent.com/12646744/127871131-4c0052d8-11fd-4844-b84a-c9ae0454682b.MP4

Not only there seems to be a delay before the "jump" happens, also the loader doesn't seem to be visible when the new video chunk is loading

drillprop:
Fix #1091

For some reason on ios, the player doesn't trigger the `waiting` event when the `seeking` event occurs. 
Need to be tested on an IOS device, I have access only to the Xcode simulator(ios emulator) and am not 100% sure if it resolves the problem.

### DIFF
--- a/src/shared/components/VideoPlayer/CustomTimeline.tsx
+++ b/src/shared/components/VideoPlayer/CustomTimeline.tsx
@@ -43,14 +43,26 @@ export const CustomTimeline: React.FC<CustomTimelineProps> = ({ player, isFullSc
   const [playedBefore, setPlayedBefore] = useState(false)
 
   useEffect(() => {
-    if (!player || !playedBefore) {
+    if (!player) {
       return
     }
-    if (isScrubbing) {
-      player.pause()
-    } else {
-      player.play()
-      setPlayedBefore(false)
+    const handler = (event: Event) => {
+      if (event.type === 'seeking') {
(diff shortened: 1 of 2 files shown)


kdembler (approved):
I think this is a great improvement but it seems there still are some problems. They may be separate from this specific one though so possibly not part of this PR:

There seems to be some lag after I scrub in which the old frame is frozen for a second but you can already hear the new part of the video playing:

https://user-images.githubusercontent.com/12646744/128152549-4a207aac-48fb-4c2f-9952-8cad333d3d4f.MP4

There in general seems to be some delay after scrub:

https://user-images.githubusercontent.com/12646744/128152780-c20a4e68-388e-449e-805d-9da5911670a4.MP4

We can merge this change and investigate 👆 in separate issues

drillprop:
> There seems to be some lag after I scrub in which the old frame is frozen for a second but you can already hear the new part of the video playing:

Damn it, I was hoping this delay was caused by my emulator. I think we could merge this and deal with it in another PR. I will create a new issue

PR #1120, opened 2021-08-05
Top 10 channels

### LINKED ISSUE #1099: [Browse] Top 10 channels
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=10080%3A310778

drillprop:
Fix #1099


src/components/TopTenChannels.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from '@emotion/styled'
+import React from 'react'
+
+import { useMostViewedChannels } from '@/api/hooks'
+import { sizes } from '@/shared/theme'
+
+import { ChannelGallery } from './ChannelGallery'
+
+export const TopTenChannels = () => {
+  const { channels, loading } = useMostViewedChannels({ viewedWithinDays: 10, limit: 10 })
```
drillprop: Just FYI, we should use here `useMostFollowedChannels`, but https://github.com/Joystream/orion/pull/17 is not merged yet, so temporarily I used `useMostViewedChannels` here

src/components/ChannelCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { url } = useAsset({ entity: channel, assetType: AssetType.AVATAR })
+
+  const { followChannel } = useFollowChannel()
+  const { unfollowChannel } = useUnfollowChannel()
+  const followedChannels = usePersonalDataStore((state) => state.followedChannels)
+  const updateChannelFollowing = usePersonalDataStore((state) => state.actions.updateChannelFollowing)
+
+  const isFollowing = followedChannels.some((channel) => channel.id === id)
+
+  const handleFollow = (event: React.MouseEvent) => {
```
rafalpawlow: As far as I remember we use this method in 3 different places. Maybe we could move this to some utils?

src/components/ChannelGallery.tsx
```diff
@@ -1,40 +1,65 @@
...
-  const createClickHandler = (id?: string) => () => id && onChannelClick && onChannelClick(id)
-
   const placeholderItems = Array.from({ length: loading ? PLACEHOLDERS_COUNT : 0 }, () => ({ id: undefined }))
   return (
-    <Gallery title={title} itemWidth={220} exactWidth={true} paddingLeft={sizes(2, true)} paddingTop={sizes(2, true)}>
+    <Gallery
+      title={title}
+      responsive={breakpoints}
+      itemWidth={350}
+      paddingLeft={sizes(2, true)}
```
rafalpawlow: I know this is not your change, but we really need paddings as props?

rafalpawlow (changes requested):

src/components/ChannelGallery.tsx
drillprop: I think we can safely just remove those. We don't need to set padding via props at all for now.

src/components/ChannelCard.tsx
drillprop: I created a separate hook to handle all this logic, let me know what you think.

drillprop:
Draft, needs some cleaning and refactoring after rebase

src/api/hooks/channel.ts
```diff
@@ -164,8 +167,62 @@ export const useMostViewedChannels = (
...
+  opts?: MostPopularChannelsOpts
+) => {
+  const { mostFollowedChannelsAllTime } = useMostFollowedChannelsAllTimeIds(variables, opts)
+
+  const mostFollowedChannelsAllTimeIds = useMemo(() => {
+    if (mostFollowedChannelsAllTime) {
+      return mostFollowedChannelsAllTime.map((item) => item.id)
+    }
+    return null
+  }, [mostFollowedChannelsAllTime])
```
kdembler: This doesn't need `useMemo`

drillprop:
Draft again. Just found a variant with SkeletonLoader: https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=8958%3A745. I need to update this

src/components/ChannelCard.tsx
```diff
@@ -1,39 +1,44 @@
...
 import { ChannelCardBase } from '@/shared/components'
 
-type ChannelCardProps = {
+export type ChannelVariant = 'primary' | 'secondary'
+
+export type ChannelCardProps = {
   id?: string
+  rankingNumber?: number
+  isLoading?: boolean
+  variant?: ChannelVariant
```
kdembler: What does this variant refer to?

src/hooks/useHandleFollowChannel.tsx
```diff
@@ -0,0 +1,31 @@
...
+      } else {
+        updateChannelFollowing(id, true)
+        followChannel(id)
+      }
+    } catch (error) {
+      Logger.warn('Failed to update Channel following', { error })
+    }
+  }
+  return {
+    followChannel: _followChannel,
```
kdembler: We shouldn't use that name if the function both follows and unfollows. Better one would be something like `toggleFollowing`

src/components/ChannelCard.tsx
drillprop: this is `primary` variant
image
and this is `secondary` 
image

src/api/client/resolvers.ts
```diff
@@ -298,14 +298,14 @@ export const queryNodeStitchingResolvers = (
         info
       )
 
-      const followsLookup = createLookup<{ id: string; views: number }>(batchedChannelFollows || [])
+      const followsLookup = createLookup<{ id: string; follows: number }>(batchedChannelFollows || [])
```
kdembler: This change should be done via rebase of `video-content-pages` right?

src/api/hooks/channel.ts
kdembler: @drillprop this comment was resolved but the change wasn't made

src/components/ChannelCard.tsx
kdembler: I'm a bit unsure whether those should be the same component but let's leave that for now

src/api/hooks/channel.ts
drillprop: Hmm, I'm pretty sure, that I've changed that, will update in a second.

drillprop: Okay, I see, I've updated only `useMostViewedChannelsAllTime` not `mostFollowedChannelsAllTimeIds`. Should be fixed

src/api/client/resolvers.ts
drillprop: I just wanted to test this quickly locally without rebasing and committed this change by mistake. Should I remove that?

kdembler: I see. If you rebase `video-content-pages` with `dev` and then your branch with `video-content-pages` it should go away

drillprop: done

kdembler (approved):

PR #1121, opened 2021-08-05
Fix player loader animation & some tiny visual improvements

### LINKED ISSUE #1109: Lottie animation on player loader should always start from begining
https://user-images.githubusercontent.com/51168865/128174652-ac1bf816-160c-4b0b-9bcf-2da95aa8866d.mov

Currently, when the user is holding the right arrow on the keyboard, release the key and then the loading occurs, Lottie animation starts from a different moment. 
This could lead to weird experiences when a user sees the ending animation of the loader when he releases the button. It would be nice to keep the loader consistent in this scenario.

drillprop:
Fix #1109


eldiegod (approved):
Works good 👌

PR #1130, opened 2021-08-09
Video content pages

rafalpawlow:


.env
```diff
@@ -8,15 +8,18 @@ REACT_APP_DEVELOPMENT_QUERY_NODE_SUBSCRIPTION_URL=wss://sumer-dev-2.joystream.ap
 REACT_APP_DEVELOPMENT_ORION_URL=https://sumer-dev-2.joystream.app/orion/graphql
 REACT_APP_DEVELOPMENT_NODE_URL=wss://sumer-dev-2.joystream.app/rpc
 REACT_APP_DEVELOPMENT_FAUCET_URL=https://sumer-dev-2.joystream.app/members/register
+REACT_APP_DEVELOPMENT_OFFICIAL_JOYSTREAM_CHANNEL=12
```
kdembler: Let's add `_ID` at the end

src/api/hooks/video.ts
```diff
@@ -64,3 +71,53 @@ export const useBasicVideos = (variables?: GetBasicVideosQueryVariables, opts?:
...
+
+export const useMostViewedVideos = (variables?: GetMostViewedVideosQueryVariables, opts?: MostViewedVideosOpts) => {
+  const { mostViewedVideos } = useMostViewedVideosIds(variables, opts)
+
+  const mostViewedVideosIds = useMemo(() => {
+    if (mostViewedVideos) {
+      return mostViewedVideos.map((item) => item.id)
+    }
+    return null
+  }, [mostViewedVideos])
```
kdembler: Let's get rid of `useMemo` here

src/api/hooks/languages.ts
```diff
@@ -0,0 +1,13 @@
...
+
+type LanguagesOpts = QueryHookOptions<GetLanguagesQuery>
+export const useLanguages = (variables?: GetLanguagesQueryVariables, opts?: LanguagesOpts) => {
+  const { data, ...rest } = useGetLanguagesQuery({ ...opts, variables })
+
+  return {
+    languages: data?.languages,
+    ...rest,
+  }
+}
```
kdembler: This is used for the language select right? I'm wondering what's the benefit of getting this data from the query node. Query node will only return list of languages that have been used in some way (as a channel or video language), but that doesn't guarantee that they will have any public videos to be displayed in the app. So this list will have kind of random entries and there's no guarantee that those will have any videos. So maybe instead let's just use the list we have defined in the app (`config/languages.ts`) and skip the fetch?

src/views/viewer/ChannelsView/ChannelsView.tsx
```diff
@@ -1,17 +1,53 @@
...
-import { transitions } from '@/shared/theme'
-import { Header, StyledViewWrapper } from '@/views/viewer/VideosView/VideosView.style'
+import { DiscoverChannels, InfiniteChannelWithVideosGrid, TopTenChannels, ViewWrapper } from '@/components'
+import { absoluteRoutes } from '@/config/routes'
+import { CallToActionButton, CallToActionWrapper, Text } from '@/shared/components'
+import { SvgNavHome, SvgNavNew, SvgNavPopular } from '@/shared/icons'
+import { sizes } from '@/shared/theme'
 
-export const ChannelsView: React.FC = () => {
+export const ChannelsView = () => {
```
kdembler: Should have `: React.FC`

kdembler (changes requested):
Let's fix those in a separate PR to not have any direct commits to this branch

src/api/hooks/languages.ts
rafalpawlow: I added this query because I needed id of each language. Question is, is to ok to hard code those ids in `languages.ts`

kdembler:
I think we haven't added the `categoryId` when adding a video view, right @wyrdrender ?

rafalpawlow:
> I think we haven't added the `categoryId` when adding a video view, right @wyrdrender ?

Yes, you are right. But on the other hand I just realized we lost "Most popular categories" section because of changes in design. I think we can remove this query from orion. WDYT?

kdembler:
Yeah I think we can delete the queries, just let's do that in a separate PR so it's easy to rollback if needed

kdembler:
Also, let's keep the `categoryId` for `addVideoView` so we collect the data anyway

kdembler (approved):

PR #1137, opened 2021-08-09
Add promising channels section

### LINKED ISSUE #969: [New & noteworthy] Add Promising new channels section

drillprop:
Fix #969 
https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=10114%3A276936

### DIFF
--- a/src/api/client/cache.ts
+++ b/src/api/client/cache.ts
@@ -7,6 +7,7 @@ import { parseISO } from 'date-fns'
 import {
   AllChannelFieldsFragment,
   AssetAvailability,
+  GetChannelsConnectionQueryVariables,
   GetVideosConnectionQueryVariables,
   GetVideosQueryVariables,
   Query,
@@ -33,6 +34,16 @@ const getVideoKeyArgs = (args: Record<string, GetVideosQueryVariables['where']> 
   }
 
   return `${onlyCount}:${channelId}:${categoryId}:${channelIdIn}:${createdAtGte}:${isPublic}:${idEq}:${idIn}:${isFeatured}`
+}
+
+const getChannelKeyArgs = (args: Record<string, GetChannelsConnectionQueryVariables['where']> | null) => {
+  // make sure queries asking for a specific category are separated in cache
+  const languageId = args?.where?.languageId_eq || ''
+  const idIn = args?.where?.id_in || []
+  const orderBy = args?.orderBy || []
+  const first = args?.first || ''
+
+  return `${first}:${languageId}:${idIn}:${orderBy}`
 }
 
 const createDateHandler = () => ({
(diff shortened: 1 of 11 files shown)


src/api/client/cache.ts
```diff
@@ -34,6 +35,16 @@ const getVideoKeyArgs = (args: Record<string, GetVideosQueryVariables['where']>
   return `${onlyCount}:${channelId}:${categoryId}:${channelIdIn}:${createdAtGte}:${isPublic}:${idEq}:${idIn}`
 }
 
+const getChannelKeyArgs = (args: Record<string, GetChannelsConnectionQueryVariables['where']> | null) => {
```
kdembler, line 39: Why exactly do we need this?

src/components/InfiniteGrids/useInfiniteGrid.ts
```diff
@@ -130,7 +132,12 @@ export const useInfiniteGrid = <
     return () => window.removeEventListener('scroll', scrollHandler)
   }, [isReady, loading, allItemsLoaded, onScrollToBottom, onDemand])
 
-  const displayedEdges = data?.edges.slice(skipCount, targetLoadedItemsCount) ?? []
+  const edges = sortByViews
+    ? [...(data?.edges || [])].sort((a, b) => {
+        return ((b?.node as ChannelEdge['node']).views || 0) - ((a?.node as ChannelEdge['node']).views || 0)
+      })
+    : data?.edges
```
kdembler: 1. This needs to be split somehow, it's not readable at the moment
2. You assume `ChannelEdge` but this will handle videos as well

src/components/InfiniteGrids/InfiniteChannelWithVideosGrid.tsx
```diff
@@ -75,11 +82,12 @@ export const InfiniteChannelWithVideosGrid: FC<InfiniteChannelWithVideosGridProp
     query: GetChannelsConnectionDocument,
     isReady: languageSelector ? isReady && !!selectedLanguage : isReady,
     skipCount,
-    orderBy: ChannelOrderByInput.CreatedAtAsc,
+    orderBy: orderBy || ChannelOrderByInput.CreatedAtAsc,
```
kdembler: May be more clear to provide a default value for `orderBy` instead

src/api/client/cache.ts
drillprop: because when we switch from `/channels` to `/new` we will get old results.

kdembler: Hmm okay, although I don't think we need `first` here

drillprop: removed

src/components/InfiniteGrids/useInfiniteGrid.ts
drillprop: I changed the approach a bit, I replaced the `sortByViews` boolean with the sort function(`additionalSortFn`). It will make `useInfniteGrid` more flexible I think. Still not sure how to better type this, but it should be more readable now.

kdembler (approved):

PR #1138, opened 2021-08-09
pause asset resolution when providers are being fetched

kdembler:
This fixes situation in which the providers list wasn't fetched yet and the app tries to resolve an asset. Previously that resolution would just fail, now it will wait for providers fetch to finish

### DIFF
--- a/src/api/hooks/workers.ts
+++ b/src/api/hooks/workers.ts
@@ -22,15 +22,21 @@ export const useWorker = (id: string, opts?: WorkerOpts) => {
 }
 
 type WorkersOpts = QueryHookOptions<GetWorkersQuery>
+export const storageWorkersVariables: GetWorkersQueryVariables = {
+  where: {
+    metadata_contains: 'http',
+    isActive_eq: true,
+    type_eq: WorkerType.Storage,
+  },
+}
 export const useStorageWorkers = (variables: GetWorkersQueryVariables, opts?: WorkersOpts) => {
   const { data, loading, ...rest } = useGetWorkersQuery({
     ...opts,
     variables: {
+      ...storageWorkersVariables,
       ...variables,
       where: {
-        metadata_contains: 'http',
-        isActive_eq: true,
-        type_eq: WorkerType.Storage,
+        ...storageWorkersVariables.where,
         ...variables.where,
       },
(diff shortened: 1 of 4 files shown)


src/api/hooks/workers.ts
```diff
@@ -22,15 +22,21 @@ export const useWorker = (id: string, opts?: WorkerOpts) => {
 }
 
 type WorkersOpts = QueryHookOptions<GetWorkersQuery>
+export const storageWorkersVariables: GetWorkersQueryVariables = {
```
drillprop, line 25: we could just create `storageWorkersVariablesWhere` and don't pass `... storageWorkersVariables` below. Do we plan to add more properties here in the future?

src/providers/storageProviders.tsx
```diff
@@ -59,19 +74,16 @@ export const useStorageProviders = () => {
...
+  const { storageProvidersPromiseRef, notWorkingStorageProvidersIds, setNotWorkingStorageProvidersIds } = ctx
+
+  const getStorageProviders = useCallback(async () => {
+    let storageProviders: BasicWorkerFieldsFragment[] = []
+    try {
+      const storageProvidersData = await storageProvidersPromiseRef.current
+      storageProviders = storageProvidersData?.data.workers || []
+    } catch {
+      // error is handled by the context
+      return []
```
drillprop, line 86: I could be totally wrong here, but do we actually need `try-catch` in this situation?

kdembler: We actually do, but good question. Check this out: https://codesandbox.io/s/nice-brattain-g0ftz?file=/src/index.js

kdembler: (open the sandbox console in bottom middle-right)

src/api/hooks/workers.ts
kdembler: True, but I don't think it hurts us in any way to do this and we can always add something later

src/providers/storageProviders.tsx
drillprop: ok, thanks for clarifying this!

drillprop (approved):

PR #1142, opened 2021-08-10
update channel unfollow dialog

kdembler:


eldiegod (approved):
LGTM those styles are not needed anymore

Issue #1042, opened 2021-07-26
Create grid system in the app


kdembler:
Create a proposed grid system to be used inside the implementation, following requirements from #1110

PR #1145, opened 2021-08-10
Grid system

eldiegod:
Fixes: #1042


src/shared/components/Grid2/Grid2.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type ReponsivenessArray = Array<number | null>
+
+type GridProps = {
+  gap?: number | ReponsivenessArray
+  cols?: number | ReponsivenessArray
+}
+
+// Applies styles to the corresponding media queries given by the values passed to the ReponsivenessArray
+const applyStyles = <T extends Record<string, unknown & { toString: () => string }>>(
```
kdembler: Can you explain what's going on here? What's the high level explanation of how this component works?

eldiegod:
I.e for the props, values or array of values can be used, each value in the array will be applied to a media query starting from the smaller one so gap={[8,24]} will apply grid-gap: 8px; to media query base and grid-gap: 24px; to media query compact
@kdembler

eldiegod:
the component needs a bit more testing but wanted to have an opinion on the API and about what should we name it since we have a Grid component already (not sure if replaceable)

src/shared/components/GridLayout/GridLayout.tsx
```diff
@@ -0,0 +1,71 @@
+import styled from '@emotion/styled'
+
+import { media } from '@/shared/theme'
+
+type ReponsivenessObject = Partial<Record<keyof typeof media, number>>
+
+export const GridLayout = styled.div`
```
kdembler: Let's call it `LayoutGrid` 🙏

src/shared/components/GridLayout/GridLayout.tsx
```diff
@@ -0,0 +1,71 @@
...
+`
+
+type GridItemProps = {
+  colStart?: number | 'initial' | ReponsivenessObject
+  colSpan?: number | ReponsivenessObject
+  rowStart?: number | 'initial' | ReponsivenessObject
+  rowSpan?: number | ReponsivenessObject
+}
+
+function isResponsiveObject(prop?: number | 'initial' | ReponsivenessObject): prop is ReponsivenessObject {
```
kdembler: I guess this should be `isResponsivenessObject`

src/shared/components/GridLayout/GridLayout.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return !!prop && typeof prop !== 'number' && prop !== 'initial'
+}
+
+export const GridItem = styled.div<GridItemProps>`
+  ${media.base} {
+    ${({ colStart }) => isResponsiveObject(colStart) && `grid-column-start: ${colStart.base};`}
+    ${({ colSpan }) => isResponsiveObject(colSpan) && `grid-column-end: span ${colSpan.base};`}
+    ${({ rowStart }) => isResponsiveObject(rowStart) && `grid-row-start: ${rowStart.base};`}
+    ${({ rowSpan }) => isResponsiveObject(rowSpan) && `grid-row-end: span ${rowSpan.base};`}
+  }
```
kdembler: We can still use something to help create those. How about:

```suggestion

const createBreakpointGridRules = (breakpointKey: keyof ReponsivenessObject) => ({ colStart, colSpan, rowStart, rowSpan }: GridItemProps) => css`
  ${isResponsiveObject(colStart) && `grid-column-start: ${colStart[breakpointKey]};`}
  // ...
`

export const GridItem = styled.div<GridItemProps>`
  ${media.base} {
    ${createBreakpointGridRules('base')}
  }
```

kdembler (approved):

PR #1151, opened 2021-08-11
Extract rankingNumber out of VideoTileBase

### LINKED ISSUE #1143: Extract `rankingNumber` out of VideoTileBase/ChannelCardBase
Let's not repeat ourselves, we could:
> a) create wrapper component, called NumberTile or alike that wraps something and adds a number to it
b) just make the number part of a carousel that's generic and accepts both video tiles and channel cards?

drillprop:
Fix #1143 

just FYI, I used [css-counters](https://caniuse.com/css-counters). You can read more about them here: https://css-tricks.com/almanac/properties/c/counter-increment/ 
(We are also using them in sign in view)


src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -203,196 +200,197 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
...
-      <SwitchTransition>
-        <CSSTransition
-          key={isLoading ? 'placeholder' : `content-${contentKey}`}
-          timeout={parseInt(transitions.timings.sharp)}
-          classNames={transitions.names.fade}
-        >
-          <InfoContainer main={main}>
-            {displayChannel && (
-              <AvatarContainer scalingFactor={scalingFactor}>
+    <div>
```
kdembler: What's this empty `div` for?

drillprop: In the end - I moved this div to `VideoGallery` because we still need this. The Carousel is messing with the width of slide parent(it's applying`.glider-slide` to it) and I still want to keep 100% or 78%(when it has `rankingNumber`) width on the VideoTile container:
image

kdembler (changes requested):
The numbers looks a bit differently now.

Before:
Screenshot 2021-08-13 at 13 29 48


After:
Screenshot 2021-08-13 at 13 29 39

You can see that numbers aren't glued to the tiles, as they used to. They should always overlap. I like that the numbers are bigger now but I think we still need to fine tune it a bit. For example, at `1700px` screen, the numbers are slightly too big and overflow the tile:
Screenshot 2021-08-13 at 13 33 59

kdembler (approved):

PR #1156, opened 2021-08-11
Video content carousel adjustments

### LINKED ISSUE #1131: Video Content Pages implementation feedback
Some visual issues I've noticed so far: (preview accessible at https://deploy-preview-1130--atlas-dev.netlify.app)

- [x] Language select is too tall:
Screenshot 2021-08-09 at 12 54 11
Screenshot 2021-08-09 at 12 54 21

- [x] "Channels in your language" isn't properly aligned (text + select):
Screenshot 2021-08-09 at 12 54 24
Screenshot 2021-08-09 at 12 54 29

- [x] I don't think all the empty area next to channel preview should be clickable. Probably only the avatar and the title? Not sure, @KubaMikolajczyk your feedback would be appreciated here. (click anywhere in this area will trigger navigation)
Screenshot 2021-08-09 at 12 59 27

- [x] Seems the follower count isn't fetched properly for channels:

https://user-images.githubusercontent.com/12646744/128696691-3ebf066d-dec3-4104-a8b1-64f576161bd2.mov

- [x] Videos in the ranked carousel on mobile seem to have very restricted area for title+meta:
IMG_5420E307823E-1

- [x] When doing an initial page load, the "Official Joystream updates" section flashes on top of the screen only to be replaced by "Popular on Joystream". Where does this come from?

https://user-images.githubusercontent.com/12646744/128697497-bb20b194-cfe6-4749-8da9-d04400ce1ab9.mov

- [x] The carousel interaction when the last page is non-full is a bit misleading to me. Take a look at this scenario, the last (4th page) has only one item. When you get to the end and go back, looking at the indicator at the bottom, it seems like you've moved two pages of content back. If you press again, it seems like you're on the first page but even then you can press once more to reveal the first item. This will likely need some feedback from @KubaMikolajczyk as well. Maybe when at the end of the content, the "back button" should make the previous page to be fully shown and not always move by a fixed amount of items?

https://user-images.githubusercontent.com/12646744/128698082-413787e5-53b4-4eef-90d3-c1975b436300.mov

- [x] Mobile carousel shows only one item, the design specified a behaviour in which a part of the next item is also visible to show the user the content is scrollable. If it's complex, we can tackle that as part of post-sprint responsiveness update:
IMG_C31F4B2ACE88-1

rafalpawlow:
Resolves #1131


patches/glider-js+1.7.7.patch
```diff
@@ -0,0 +1,20 @@
...
+ 
+-        if (backwards) slide -= _.opt.slidesToScroll
++        if (backwards) {
++          var itemsRemainder = _.slides.length % _.opt.slidesToScroll
++          if (_.arrows.next.classList.contains('disabled') && itemsRemainder) {
++            slide -= itemsRemainder
++          } else {
++            slide -= _.opt.slidesToScroll
++          }
++        }
```
kdembler: Couldn't we just adjust `slidesToScroll` dynamically instead of changing the library code?

kdembler (commented):
Could you explain a bit what's happening here now? I have hard time understanding. Also what's the purpose of `FRACTIONAL_LEVEL` and `FRACTIONAL_LEVEL_RANKING`

rafalpawlow:
`FRACTIONAL_LEVEL` and `FRACTIONAL_LEVEL_RANKING` are for fractional displaying of slides, and they are different for carousel with ranking and without it. https://nickpiscitelli.github.io/Glider.js/#glider-cut

About fix for carousel
1. on each end of animation I take `slidesToScroll` for current breakpoint from `gliderOptions`
2. calculate if last page is fully filled by slides, meaning if number of slides on last page is equal to `slidesToShow`
3. if we are on the last page and it's not fully filled, I set `slidesToScroll` to the number of slides on the last page
4. otherwise I set `slidesToScroll` to previously stored value
5. if there is no stored `slidesToScroll` I set it to value from `gliderOptions`
6. if user resize window, we need to clear stored `slidesToScroll` because they are different for each breakpoint

Now, what might be confusing is the refs thing. It's not fully clear to me but when I manually set `slidesToScroll` inside `onAnimated` function, such value begins to come from `gliderOptions`. That's why I can't rely on this, and store initial value as ref.

kdembler (approved):

Issue #1112, opened 2021-08-04
Use grid system for channel details view


kdembler:
As in the design, this should use grid: https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=8672%3A247519

PR #1158, opened 2021-08-12
Channel view grid

eldiegod:
Channel view layout using LayoutGrid

Fixes:  #1112

To merge after grid system changes


src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -32,8 +31,8 @@ export const ChannelAbout = () => {
             </Text>
           </TextContainer>
         )}
-      </div>
-      <DetailsContainer>
+      </GridItem>
+      <GridItem colSpan={{ base: 12, small: 3 }} colStart={{ small: -3 }}>
```
kdembler: This makes the details section span only 2 columns, not 3

src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -22,8 +21,8 @@ export const ChannelAbout = () => {
   const { channel } = useChannel(id)
   const { videoCount } = useChannelVideoCount(id)
   return (
-    <Container>
-      <div>
+    <StyledLayoutGrid>
```
kdembler, line 24: Spacing on mobile in information tab is wrong (between tabs and content). This was here before but let's fix it now
Screenshot 2021-08-13 at 11 33 48

kdembler (commented):
Let's also try using the Grid instead of `limitedWidthContainer` and see how it works

eldiegod:
Not sure if we should make the grid responsible for the LimitedWidthContainer job feels a bit odd

kdembler:
Why is that? We don't have to think about it as a grid, we can think about it as a layout wrapper or something like this. Why do you think it's weird? You'd rather use limited width container explicitly?

eldiegod:
I think having `LimitedWidthContainer` inside the Grid is nice but I think the Grid will be used more times inside already existing layouts than as LayoutWrapper itself, so what feels weird is just using the Grid as layout wrapper for the `LimitedWidthContainer` utility only.

So, I would keep them both and use one or the other as needed

kdembler:
> So, I would keep them both and use one or the other as needed

I see that as a problem though, we encourage inconsistency and then you need to think about which one to choose. We also need to keep those in sync so they produce the same content.

Let's just keep the `LimitedWidthContainer` for now and move the grid sizing inside it.

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -263,58 +263,60 @@ export const ChannelView: React.FC = () => {
...
-            onSelectTab={handleSetCurrentTab}
-          />
-          {currentTabName === 'Videos' && (
-            <Search
-              searchInputRef={searchInputRef}
-              isSearchInputOpen={isSearchInputOpen}
-              setIsSearchingInputOpen={setIsSearchingInputOpen}
-              setIsSearching={setIsSearching}
-              search={search}
+      <LayoutGrid>
```
kdembler: We need to revert that change, right?

src/components/LimitedWidthContainer.tsx
```diff
@@ -1,13 +1,18 @@
...
 import styled from '@emotion/styled'
 
-export const limitedWidthContainerStyle = css`
-  --max-inner-width: calc(1440px - var(--sidenav-collapsed-width) - calc(2 * var(--global-horizontal-padding)));
+type LimitedWidthContainerProps = { big?: boolean }
+
+export const limitedWidthContainerStyle = (props: LimitedWidthContainerProps) => css`
+  --max-inner-width: calc(
+    ${props.big ? '2284' : '1440'}px - var(--sidenav-collapsed-width) - calc(2 * var(--global-horizontal-padding))
+  );
```
kdembler: This value isn't correct, it's `1304px`, should be `1376px`, we basically shouldn't subtract sidenav width

kdembler (approved):

PR #1159, opened 2021-08-12
Disable the `L` key and `right arrow` key when a video is over

### LINKED ISSUE #1157: When a video ends user shouldn't be able to change time with the right arrow key
Right now when a video ends and the ending screen is showing up, the user still can change time with the right arrow key(or `L` key), but this interaction is a little bit misleading. After the key is pressed, the video just starts from the beginning. 
Let's just disable the right arrow key and `L` key when the video is ended.

drillprop:
Fix #1157

### DIFF
--- a/src/shared/components/VideoPlayer/CustomTimeline.tsx
+++ b/src/shared/components/VideoPlayer/CustomTimeline.tsx
@@ -23,11 +23,17 @@ type CustomTimelineProps = {
   player?: VideoJsPlayer | null
   isFullScreen?: boolean
   playerState: PlayerState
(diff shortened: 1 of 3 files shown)


drillprop:
~~Draft. Doesn't work for all videos, unfortunately~~

drillprop:
Should also fix #1178

kdembler (approved):
At this point I have no idea what's going on in the `VideoPlayer` but I trust that you do :P

Seems to fix the problem instead, I'm not able to tell if it fixes #1178 because I wasn't able to reproduce it but if you say it does, let's close it 👍

Issue #1148, opened 2021-08-10
Improve scrolling in video workspace 


kdembler:
We should implement outcomes of #795 in the app

PR #1174, opened 2021-08-13
Improve scrolling in video workspace

eldiegod:
Fixes: #1148

### DIFF
--- a/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
+++ b/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
@@ -2,6 +2,7 @@ import { formatISO, isValid } from 'date-fns'
 import { debounce } from 'lodash'
 import React, { useEffect, useRef, useState } from 'react'
 import { Controller, DeepMap, FieldError, FieldNamesMarkedBoolean, useForm } from 'react-hook-form'
+import useMeasure from 'react-use-measure'
 
 import { useCategories } from '@/api/hooks'
 import { License } from '@/api/queries'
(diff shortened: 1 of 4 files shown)


src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -413,219 +417,225 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
-              'title',
-              textFieldValidation({ name: 'Video Title', minLength: 3, maxLength: 40, required: true })
+      <FormScrolling>
+        <FormWrapper>
+          <Controller
+            name="assets"
+            control={control}
+            render={() => (
+              <div ref={fileSelectContainerRef}>
+                <MultiFileSelectWrapper width={fileSelectContainerBounds.width}>
```
kdembler: Why is this needed?

eldiegod: it's so the multi file selector sizes itself propperly as it now has to have position: fixed we need to defined it's width manually

kdembler: I've missed that it's `fixed` before. Understandable then, but does it really need to be `fixed`? I did a quick experiment with `position: sticky; top: 32px` and it seems to be working fine without any manual width setting. Can you try that?

kdembler:
One additional thing I think we should fix while we're at it, is the inputs container kind of spanning outside of visible area. You can see that in the video:

https://user-images.githubusercontent.com/12646744/129569747-6d9a041c-09d5-4b89-92de-aa015b45c227.mov

The scrollbar goes under the action bar, which is not really intuitive. Instead it should end when the action bar begins.

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
eldiegod: sticky worked as well 👍

kdembler (approved):

PR #1183, opened 2021-08-16
refactor error handling

kdembler:
Key changes:
- all errors are now captured explicitly with additional message and source
- errors are no longer captured from console, going through there meant losing a lot of details of the captured errors
- not everything logged will be captured to sentry
- errors will now contain much more useful data
- we're not `throw`ing any errors, always handling them inside the view. this way, if something is thrown and captured by the boundary, we know that we have some unknown problem
- cleaned up the error screen
- removed some old illustrations that we don't use anymore


src/components/ViewErrorFallback.tsx
```diff
@@ -1,12 +1,55 @@
...
+    } else {
+      navigate(absoluteRoutes.viewer.index())
+    }
+  }
+
+  return (
+    <Container>
+      <AnimatedError />
+      <Message>
+        <Text variant="h3">Oops! An error occurred.</Text>
```
drillprop: I didn't find a design for this screen, but I think we could at least add some margin below `h3`.

drillprop (commented):
Looks good, added just one small suggestion.

PR #1186, opened 2021-08-16
Create Video file encoding.md

KubaMikolajczyk:
Started a new docs for video file encoding

### DIFF
--- a/docs/Video file encoding.md
+++ b/docs/Video file encoding.md
@@ -1,0 +1,13 @@
+# Video file encoding
+
+Current implementation of our file select component allows users to upload video files using only encoding which is accepted by the browser thay are using. 
+If you seek for more informations about what video formats your browser accepts follow the guide below.
+
+### Supported formats & codecs
+
+| Browser  | Chrome | Firefox | Internet Explorer | Safari | iPhone | Android | Operna | Chrome Android |
+| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
+| Video Formats  | WebM, MP4, Theora |                | MP4 | MP4 | MP4 | WebM, MP4 | WebM, MP4, Theora | WebM, MP4 |
+| Video Codecs  | 	H.264, VP8, VP9 | H.264, VP8, VP9 | H.264 | H.264 | H.264 | H.264, VP8 | H.264, VP8 | H.264, VP8, VP9 |
+
+*Above is not final - has to be checked and adjusted


kdembler (commented):
I like the idea of creating this document, however I'm not sure how true this data is. First off, it's likely not enough to mention the browser name, it also depends on its version, but I guess we could assume "latest". Second off, the data doesn't seem to be fully truthful, Safari supports WebM and VP8/9 for some time now.

drillprop:
I think the most accurate data about browser support should be here: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs 

[This](https://videojs.com/html5-video-support/)  also could be helpful

KubaMikolajczyk:
> I like the idea of creating this document, however I'm not sure how true this data is. First off, it's likely not enough to mention the browser name, it also depends on its version, but I guess we could assume "latest". Second off, the data doesn't seem to be fully truthful, Safari supports WebM and VP8/9 for some time now.

@kdembler Yes it might not be accurate - I will update it with the examples from @drillprop 


> I think the most accurate data about browser support should be here: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs
> 
> [This](https://videojs.com/html5-video-support/) also could be helpful

The table from video js about supported codecs is awesome - we can include it here - do you know any simmilar resources about video formats (I think we should include that as well in separate table except the codecs, or am I wrong and there is a better way to describe which formats are accepted?)

kdembler:
That resource from MDN looks great but I think it's not fully up to date as well, as it mentions Safari not supporting VP8/VP9 but it does :P I'm wondering whether we actually need a table like this. After all, users don't want to upload a video that works only for their browser, they want to upload files that will be playable by all users. Maybe instead of this doc we just condense this knowledge and place it directly in the file picker? We could just suggest using `MP4/H.264` or `WEBM/VP8,VP9` and call it a day?

KubaMikolajczyk:
> That resource from MDN looks great but I think it's not fully up to date as well, as it mentions Safari not supporting VP8/VP9 but it does :P I'm wondering whether we actually need a table like this. After all, users don't want to upload a video that works only for their browser, they want to upload files that will be playable by all users. Maybe instead of this doc we just condense this knowledge and place it directly in the file picker? We could just suggest using `MP4/H.264` or `WEBM/VP8,VP9` and call it a day?

Alright this sounds okish :) at least we would inform about possibilities right away - do you have any suggestions for the image file extensions of thumbnails as well to make it consistent?

kdembler:
For images JPEG and WEBP should be good

KubaMikolajczyk:
OK so Im closing this and deleting the branch we will use this solution!

PR #1188, opened 2021-08-17
Fix layout shift on popular page

### LINKED ISSUE #1187: Layout change when loading Popular page
https://user-images.githubusercontent.com/12646744/129579850-f2b7de59-9be6-4dc3-aac4-2e79273c3d73.mov

As you can see, when you load the Popular page there's a change of content. There's a `Popular videos` section visible for a split second only to be pushed down by `Top 10 this month`. I think we should show a placeholder for the first section right away to avoid this jump

drillprop:
Fix #1187
Also: 
- fix the skeleton loader in InfiniteVideoGrid
- fix the follow button on ChannelWithVideos(the app was navigating to channel when user clicks follow/unfollow)

### DIFF
--- a/src/components/InfiniteGrids/InfiniteVideoGrid.tsx
+++ b/src/components/InfiniteGrids/InfiniteVideoGrid.tsx
@@ -7,10 +7,10 @@ import {
   GetVideosConnectionQueryVariables,
   VideoWhereInput,
 } from '@/api/queries'
-import { Grid, GridHeadingContainer, LoadMoreButton, SkeletonLoader, Text } from '@/shared/components'
+import { Grid, GridHeadingContainer, LoadMoreButton, SkeletonLoader, Text, TitleContainer } from '@/shared/components'
 import { SvgGlyphChevronRight } from '@/shared/icons'
 
-import { AdditionalLink, LoadMoreButtonWrapper, TitleWrapper } from './InfiniteGrid.style'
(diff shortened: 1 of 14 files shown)


src/components/InfiniteGrids/InfiniteVideoGrid.tsx
```diff
@@ -173,7 +173,11 @@ export const InfiniteVideoGrid: React.FC<InfiniteVideoGridProps> = ({
       <TitleWrapper>
         {title && (
           <GridHeadingContainer>
-            {!ready ? <SkeletonLoader height={23} width={250} /> : <Text variant="h4">{title}</Text>}
+            {loading || !displayedItems.length ? (
```
kdembler: We're ignoring the `ready` props now. Was that the plan?

kdembler (changes requested):
Unfortunately there are still some issues with jumping. Take a look at this (New view):

https://user-images.githubusercontent.com/12646744/129763714-93e54b93-e5c5-4c61-81fc-4d41caa02080.mov

You can see 2 jumps here:
1. The placeholder jumps down and back up a bit (not a biggie, we can skip if hard to fix)
2. The border under the title jumps down significantly after the skeleton disappears, we need to improve that

drillprop:
Fixed, It should be good now

kdembler (approved):

Issue #1113, opened 2021-08-04
Update responsiveness of video hero component


kdembler:
As seen in the screens [here](https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=11545%3A402820) there has been some new behaviours defined for the video hero.

Firstly, the rules for the title width have been defined with columns. Other than that, spacing of controls inside the hero have been defined, but also spacing between the hero and the content. We should make sure we follow the design there

PR #1190, opened 2021-08-17
Video Hero Updates

eldiegod:
fixes:  #1113


src/components/VideoHero/VideoHero.style.ts
```diff
@@ -1,26 +1,22 @@
...
+  XXLARGE: 24,
 }
-const GRADIENT_OVERLAP = 50
-const GRADIENT_HEIGHT = 250
-const INFO_BOTTOM_MARGIN = 100
-const BUTTONS_HEIGHT_PX = '54px'
 
 export const Container = styled.section`
   position: relative;
+  max-height: 100vh;
```
kdembler: Why do we need that?

src/components/VideoHero/VideoHero.style.ts
```diff
@@ -1,26 +1,22 @@
...
 
   /* because of the fixed aspect ratio, as the viewport width grows, the media will occupy more height as well
```
kdembler: Why would we keep this section and all the styles below it? It's just the same style now repeated 5 times

kdembler (changes requested):
There's a number of inconsistencies with the design:
1. The channel title text has different styles
2. The video hero has only vertical gradient now, design specifies 3 overlays
3. The spacing of content inside the video hero and also content below the video hero is inconsistent across different breakpoints. Please refer to Home view design in Figma
4. Video content is generally not properly sized, it features black bars on a lot of screen sizes, it should always span full width:
Screenshot 2021-08-18 at 11 53 00

src/components/VideoHero/VideoHero.style.ts
eldiegod: It's so on widescreens the hero section doesn't grow way bigger than the screen height and the info about the video channel doesn't get outside the screen, i.e
imagen

vs 

imagen

src/components/ChannelLink/ChannelLink.tsx
```diff
@@ -7,13 +7,14 @@ import { AssetType, useAsset } from '@/providers'
...
 
-import { Container, Handle, HandleSkeletonLoader } from './ChannelLink.style'
+import { Container, Handle, HandleSkeletonLoader, StyledText } from './ChannelLink.style'
 
 type ChannelLinkProps = {
   id?: string
   onClick?: (e: React.MouseEvent<HTMLDivElement>) => void
   hideHandle?: boolean
   hideAvatar?: boolean
+  heroTitle?: boolean
```
kdembler: Let's avoid specific naming like this, we could use `variant: 'primary' | 'secondary'` instead

src/components/ChannelLink/ChannelLink.tsx
```diff
@@ -48,7 +50,13 @@ export const ChannelLink: React.FC<ChannelLinkProps> = ({
       {!hideAvatar && <Avatar loading={!displayedChannel} size={avatarSize} assetUrl={avatarPhotoUrl} />}
       {!hideHandle &&
         (displayedChannel ? (
-          <Handle withAvatar={!hideAvatar}>{displayedChannel.title}</Handle>
+          heroTitle ? (
+            <StyledText withAvatar={!hideAvatar} secondary variant="button2">
+              {displayedChannel.title}
+            </StyledText>
+          ) : (
+            <Handle withAvatar={!hideAvatar}>{displayedChannel.title}</Handle>
```
kdembler: I think we could actually use `StyledText` for this as well and get rid of `Handle`? The default has `h6`

src/components/VideoHero/VideoHero.tsx
```diff
@@ -111,19 +112,15 @@ export const VideoHero: React.FC = () => {
             appear
           >
             <ButtonsContainer>
```
kdembler, line 118: We need to handle this appearing, right now there's a jump in the title. Before this change, the space for buttons was preserved, so when they appeared nothing changed places

kdembler (changes requested):

kdembler:
One more thing for the control buttons:
1. Seems they don't use the correct icons as specified in the design
2. Seems they don't have the correct sizes

eldiegod:
Small detail is that the variant of the button size is not a responsive prop thus only thing that would be missing now, I think, not a big deal but at the same time we probably should look at making our props accept responsive objs or make that process straight forward so we can go around changing variants according to breakpoints I think it's something that will probably encounter more and more in design as it gets more complex

src/components/ChannelLink/ChannelLink.style.ts
```diff
@@ -1,7 +1,7 @@
 import styled from '@emotion/styled'
 import { Link } from 'react-router-dom'
 
-import { SkeletonLoader } from '@/shared/components'
+import { SkeletonLoader, Text } from '@/shared/components'
 import { colors, sizes, typography } from '@/shared/theme'
```
kdembler: eslint warnings here

src/components/VideoHero/VideoHero.tsx
```diff
@@ -111,19 +113,15 @@ export const VideoHero: React.FC = () => {
...
-              <PlayButton
-                onClick={handlePlayPauseClick}
-                icon={videoPlaying ? <SvgPlayerPause /> : <SvgPlayerPlay />}
-                size="large"
-              >
+              <PlayButton onClick={handlePlayPauseClick} icon={videoPlaying ? <SvgPlayerPause /> : <SvgPlayerPlay />}>
```
kdembler: We're still using wrong icons, both for play/pause and sound on/off

src/components/VideoHero/VideoHero.style.ts
```diff
@@ -206,21 +116,18 @@ export const TitleSkeletonLoader = styled(SkeletonLoader)`
...
 `
 
+export const ButtonsSpaceKeeper = styled.div`
+  min-height: ${BUTTONS_HEIGHT}px;
+`
+
 export const PlayButton = styled(Button)`
-  width: 140px;
-  height: ${BUTTONS_HEIGHT_PX};
+  width: 97px;
```
kdembler: Let's use the following width: once you update the icons, let's remove explicit width and see how wide the button is when "Pause" is visible. Then let's use this value

src/shared/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -1,5 +1,6 @@
+import { InterpolationWithTheme, css } from '@storybook/theming'
```
kdembler: eslint warnings

kdembler (changes requested):

kdembler (approved):

Issue #1180, opened 2021-08-14
Use new empty states for channel and video views


kdembler:
We currently show only a string "Channel/Video not found" if user opens an URL with wrong id. We should use the new `EmptyFallback` there with ability to go back to home screen

PR #1191, opened 2021-08-18
Not found video and channel fallbacks

eldiegod:
Fixes: #1180

### DIFF
--- a/src/views/viewer/ChannelView/ChannelView.style.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.style.tsx
@@ -227,3 +227,7 @@ export const UnfollowDescriptionAccentText = styled.span`
   font-weight: ${typography.weights.regular};
   color: ${colors.gray[50]};
 `
+
+export const NotFoundChannelContainer = styled.div`
+  transform: translateY(50%);
+`
--- a/src/views/viewer/ChannelView/ChannelView.style.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.style.tsx
@@ -1,8 +1,8 @@
 import styled from '@emotion/styled'
 import { fluidRange } from 'polished'
 
-import { ChannelLink } from '@/components'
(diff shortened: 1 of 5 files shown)


src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -227,3 +227,7 @@ export const UnfollowDescriptionAccentText = styled.span`
   font-weight: ${typography.weights.regular};
   color: ${colors.gray[50]};
 `
+
+export const NotFoundChannelContainer = styled.div`
```
drillprop, line 231: Instead creating new container, maybe we could apply these styles to `EmptyFallback` itself - like:
```tsx
const StyledEmptyFallback = styled(EmptyFallback)` ...`
```
But firstly we would have to pass `className` prop to EmptyFallback if I'm not mistaken

kdembler: Also, let's use a regular margin instead of transform

eldiegod: Since not gonna be using transform did it with a flex container that's the cleanest

Did add the className to EmptyFallback anyway because why not

kdembler (approved):

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -1,8 +1,8 @@
 import styled from '@emotion/styled'
 import { fluidRange } from 'polished'
 
-import { ChannelLink } from '@/components'
-import { Button, IconButton, SkeletonLoader, Tabs, Text, TextField } from '@/shared/components'
+import { ChannelLink, TOP_NAVBAR_HEIGHT } from '@/components'
+import { Button, EmptyFallback, IconButton, SkeletonLoader, Tabs, Text, TextField } from '@/shared/components'
```
kdembler: eslint warning here

eldiegod: fixed

PR #1195, opened 2021-08-18
Remove language query from InfiniteChannelWithVideosGrid

drillprop:


drillprop:
I've also added `EmptyFallback`: 
image

There is no design for this if I'm not mistaken, but I think it's worth it. Let me know what you think

kdembler (approved):

PR #1198, opened 2021-08-19
Step component

### LINKED ISSUE #1087: File step and multistepper step should use shared generic component
If you take a look at [the design](https://www.figma.com/file/TGUpBXgXf9ceMvHbRB5BIx/Joystream-Atlas?node-id=5610%3A2096), it has one base `Step` component that's used both inside the file stepper and multistepper header. We have 2 separate implementations for those, let's create a shared version

drillprop:
Fix #1087


src/shared/components/Stepper/Stepper.tsx
```diff
@@ -1,47 +1,50 @@
...
+  onDelete?: () => void
 }
-
-export const FileStep: React.FC<FileStepProps> = ({
-  stepNumber = 1,
-  active,
-  isFileSet,
-  type,
-  onDelete,
+export const Stepper: React.FC<StepperProps> = ({
```
kdembler: I think this should be named `Step`. `Stepper` suggests it's a collection of steps and this is only a single one

src/shared/components/Stepper/Stepper.tsx
```diff
@@ -1,47 +1,50 @@
...
-  isFileSet?: boolean
+export type StepperProps = {
+  title: string
+  variant?: 'file' | 'default'
+  completed?: boolean
   thumbnailUrl?: string | null
-  onSelect?: (step: FileType) => void
-  disabled?: boolean
   isLoading?: boolean
+  type?: 'video' | 'image'
```
kdembler: How about we remove this prop and replace it with something generic like `icon: React.ReactNode`?

src/shared/components/Stepper/Stepper.tsx
```diff
@@ -1,47 +1,50 @@
...
+  isFileSet?: boolean
```
kdembler: If this component is to be generic this prop doesn't make much sense

kdembler (changes requested):

kdembler (approved):

PR #1199, opened 2021-08-19
Fix play interrupted error

### LINKED ISSUE #1139: Fix error `the play() request was interrupted by a call to pause().` when playing video
More information on why it's happening: https://developers.google.com/web/updates/2017/06/play-request-was-interrupted

rafalpawlow:
Closes #1139


drillprop:
The code generally looks good, and it mostly works well, but: 
- you cannot anymore pause the video when loading. I think users should be able to do it
- I found that some of the videos are just not playing on IOS, for instance video/5793(production), It just stuck on the `pause` icon.  
image
Tried this on an emulator, maybe it's not reproducible on a real device, but I need confirmation. 

Not sure yet, what's happening. I'll try to look at it later and maybe propose some changes as well.

drillprop (approved):

Issue #1182, opened 2021-08-16
Use skeleton loader for assets that are being loaded


kdembler:
Currently `VideoTile` (and I think `ChannelCard` as well) shows the skeleton loader only when the data about video is being fetched from the query node. Once that's done we will display fixed grayed out placeholder for the thumbnail till it loads. As a conclusion from #877, we should keep displaying the skeleton loader for the thumbnail until the image is ready. This also applies to the channel avatar.

PR #1200, opened 2021-08-19
Img loading skeleton holders

eldiegod:
fixes: #1182


eldiegod:
I think this solves the issue but it's hard to tell

drillprop:
Didn't look at the code yet, but I've noticed when the channel has no avatar we will see the skeleton loader all the time
image

src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -283,7 +283,7 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
           <InfoContainer main={main}>
             {displayChannel && (
               <AvatarContainer scalingFactor={scalingFactor}>
-                {isLoading || hasAvatarImgLoaded === false ? (
+                {isLoading || (hasAvatarImgLoaded === false && channelAvatarUrl) ? (
```
drillprop: Now, it's not working :D

I mean, I think I know how to fix that. You could just change this line to: 
```tsx
{isLoading || (hasAvatarImgLoaded === false && channelAvatarUrl !== undefined) ? ( ...
```
This should be working because if the user hasn't set his avatar we're getting `null` instead of `undefined` if I'm not mistaken. 
Also , I think you could move this condition inside your custom hook

eldiegod: yeah it makes sense

eldiegod: Since this can be used anywhere else I think it's a bit risky to assume the null condition will be true everywhere else so I think I will keep it outside the hook for now

src/hooks/usePreloadImg.tsx
```diff
@@ -0,0 +1,14 @@
+import { useEffect, useState } from 'react'
+
+export const usePreloadImg = (thumbnailUrl: string | undefined) => {
```
kdembler: Why do we need this? AssetsManager already does that via `testAssetDownload`

src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -140,12 +140,6 @@ export const CoverImage = styled.img<CoverImageProps>`
   ${({ darkenImg }) => darkenImg && `filter: brightness(45%);`}
 `
 
-export const CoverNoImage = styled.div`
-  width: 100%;
-  height: 100%;
-  background: linear-gradient(125deg, rgba(16, 18, 20, 1) 30%, rgba(34, 36, 38, 1) 65%, rgba(16, 18, 20, 1) 100%);
-`
```
kdembler: I think we still need this for drafts? We don't want those to keep loading as there's nothing to load

kdembler (changes requested):

src/hooks/usePreloadImg.tsx
eldiegod: wanted to preload the image somehow should we use `testAssetDownload` instead?

kdembler: but the image is already loaded by the assets manager

eldiegod: then I'm not sure what is to be done here exactly

eldiegod:
I think this is good to go now but awkward to test in the state of the chain as of now

kdembler:
You can always change to `production` for testing

eldiegod:
looks good in prod

kdembler (changes requested):
Doesn't work like it's supposed to, there are 2 problems I found:

1. There are some weird flickers of loading that weren't there before:

https://user-images.githubusercontent.com/12646744/130811115-baa603b1-64b8-4369-a494-8d167590830c.mov



2. It seems that when the image is loading the full tile is in loading state while it should be only the images: Screenshot 2021-08-25 at 16 37 09

drillprop:
Maybe I'm exaggerating a little here, but I also see some inconsistencies. If a new batch of videos ends initial loading this set blinks for a split second(fade out and then fade in) and then image loading occurs. Not sure how to fix this though, probably you'll need to look at Switch and CssTransition inside `VideoTile`.
Possibly we could live with it, don't know, do we need to work on that @kdembler ?  

https://user-images.githubusercontent.com/51168865/130912320-e832ca4d-cc35-48cc-9df6-72bae59f5125.mov

eldiegod:
@drillprop Yeah @kdembler pointed this out as well in the chats and I was also trying to figure out why this is happening but haven't found a solution either

drillprop (approved):
Looks good to me, great work guys!

kdembler (approved):

PR #1202, opened 2021-08-20
Fix circular dependencies

### LINKED ISSUE #1194: Fix circular dependency problem
This is a long-standing issue, which caused a lot of problems in the past. For instance, it caused not working storybook or it causing throwing errors during hot refresh. 
If you experienced an error that says: `Uncaught Error: You are trying to create a styled element with an undefined component`, it's probably caused by this issue.
Using https://www.npmjs.com/package/circular-dependency-plugin could be really helpful to investigate that.

drillprop:
Fix #1194


drillprop:
We still have some circular dependency problems, but I'm not sure how to fix this yet. It's caused by `getUserInfo` in `src/utils/logs/shared.ts`. 
I will set it as a draft for now.

src/views/playground/Playgrounds/GridTesting.tsx
```diff
@@ -1,7 +1,7 @@
 import styled from '@emotion/styled'
 import React from 'react'
 
-import { GridItem, LayoutGrid } from '@/shared/components'
+import { GridItem, LayoutGrid } from '@/shared/components/LayoutGrid/LayoutGrid'
```
kdembler: Shouldn't need that additional `LayoutGrid` at the end

kdembler (approved):

PR #1203, opened 2021-08-20
Add error handling for batched orion queries

rafalpawlow:


src/api/client/resolvers.ts
```diff
@@ -193,24 +193,28 @@ export const queryNodeStitchingResolvers = (
...
+
+        return parent.edges.map((edge: VideoEdge) => ({
+          ...edge,
+          node: {
+            ...edge.node,
+            views: viewsLookup[edge.node.id]?.views || 0,
+          },
+        }))
+      } catch (error) {
+        Logger.warn('Failed to resolve views field', { error })
```
kdembler: let's rebase and use `SentryLogger.error` here

src/api/client/resolvers.ts
```diff
@@ -273,42 +277,47 @@ export const queryNodeStitchingResolvers = (
...
+        return parent.edges.map((edge: ChannelEdge) => ({
+          ...edge,
+          node: {
+            ...edge.node,
+            follows: followsLookup[edge.node.id]?.follows || 0,
+            views: viewsLookup[edge.node.id]?.views || 0,
+          },
+        }))
+      } catch (error) {
+        Logger.warn('Failed to resolve follows or views field', { error })
```
kdembler: same

kdembler (changes requested):

src/api/client/resolvers.ts
```diff
@@ -84,7 +84,7 @@ export const queryNodeStitchingResolvers = (
         const viewsLookup = createLookup<{ id: string; views: number }>(batchedVideoViews || [])
         return videos.map((video: Video) => ({ ...video, views: viewsLookup[video.id]?.views || 0 }))
       } catch (error) {
-        ConsoleLogger.warn('Failed to resolve videos field', { error })
+        SentryLogger.error('Failed to resolve videos field', 'batchedVideoViewsResolver', error)
```
kdembler: Let's split this try-catch. I think we should call `const videos = await videosResolver(parent, args, context, info)` without `try` to let the error bubble up so it can be handled by the views. Let's only include orion query inside `try` and then change the message to `Failed to resolve video views`

kdembler: Let's update the source so we can know from which resolver the error came. Let's use `videos resolver` here

src/api/client/resolvers.ts
```diff
@@ -84,7 +84,7 @@ export const queryNodeStitchingResolvers = (
...
         return null
```
kdembler: If we fail to resolve views but managed to resolve videos, we should return them

src/api/client/resolvers.ts
```diff
@@ -143,7 +143,7 @@ export const queryNodeStitchingResolvers = (
           views: viewsLookup[channel.id]?.views || 0,
         }))
       } catch (error) {
```
kdembler, line 146: Let's use a similar strategy as with the previous resolver:
- channels fetch outside of try
- separate independent try-catch for views and follows with separate messages
- `channels resolver` as source

src/api/client/resolvers.ts
```diff
@@ -179,7 +179,7 @@ export const queryNodeStitchingResolvers = (
           info
         )
       } catch (error) {
-        ConsoleLogger.warn('Failed to resolve views field', { error })
+        SentryLogger.error('Failed to resolve views field', 'batchedChannelViews', error)
```
kdembler:
```suggestion
        SentryLogger.error('Failed to resolve video views', 'Video.views resolver', error)
```

src/api/client/resolvers.ts
```diff
@@ -193,24 +193,28 @@ export const queryNodeStitchingResolvers = (
...
-      }))
+        return parent.edges.map((edge: VideoEdge) => ({
+          ...edge,
+          node: {
+            ...edge.node,
+            views: viewsLookup[edge.node.id]?.views || 0,
+          },
+        }))
+      } catch (error) {
+        SentryLogger.error('Failed to resolve views field', 'batchedVideoViewsResolver', error)
```
kdembler:
```suggestion
        SentryLogger.error('Failed to resolve video views', 'VideoConnection.edges resolver', error)
```

src/api/client/resolvers.ts
```diff
@@ -235,7 +239,7 @@ export const queryNodeStitchingResolvers = (
           info
         )
       } catch (error) {
-        ConsoleLogger.warn('Failed to resolve views field', { error })
+        SentryLogger.error('Failed to resolve views field', 'orionViewsResolver', error)
```
kdembler:
```suggestion
        SentryLogger.error('Failed to resolve channel views', 'Channel.views resolver', error)
```

kdembler: I think you get the idea, we should do the same for all resolvers, not gonna add a comment for the rest

kdembler (changes requested):

src/api/client/resolvers.ts
```diff
@@ -179,7 +186,7 @@ export const queryNodeStitchingResolvers = (
           info
         )
       } catch (error) {
-        ConsoleLogger.warn('Failed to resolve views field', { error })
+        SentryLogger.error('Failed to resolve views views', 'Video.views resolver', error)
```
kdembler:
```suggestion
        SentryLogger.error('Failed to resolve video views', 'Video.views resolver', error)
```

src/api/client/resolvers.ts
```diff
@@ -135,17 +142,17 @@ export const queryNodeStitchingResolvers = (
...
-        return channels.map((channel: Channel) => ({
-          ...channel,
-          follows: followsLookup[channel.id]?.follows || 0,
-          views: viewsLookup[channel.id]?.views || 0,
-        }))
+        viewsLookup = createLookup<{ id: string; views: number }>(batchedChannelViews || [])
       } catch (error) {
-        ConsoleLogger.warn('Failed to resolve channels field', { error })
-        return null
+        SentryLogger.error('Failed to resolve channels views', 'channels resolver', error)
```
kdembler:
```suggestion
        SentryLogger.error('Failed to resolve channel views', 'channels resolver', error)
```

src/api/client/resolvers.ts
```diff
@@ -117,15 +127,12 @@ export const queryNodeStitchingResolvers = (
           context,
           info
         )
-        const followsLookup = createLookup<{ id: string; follows: number }>(batchedChannelFollows || [])
+        followsLookup = createLookup(batchedChannelFollows || [])
+      } catch (error) {
+        SentryLogger.error('Failed to resolve channels follows', 'channels resolver', error)
```
kdembler:
```suggestion
        SentryLogger.error('Failed to resolve channel follows', 'channels resolver', error)
```

kdembler (approved):

PR #1204, opened 2021-08-20
Reduce requests to orion when searching

### LINKED ISSUE #1090: Reduce number of requests to orion when using `search`
Similar to #38 and #1081

drillprop:
Fix #1090

### DIFF
--- a/src/api/client/resolvers.ts
+++ b/src/api/client/resolvers.ts
@@ -6,24 +6,22 @@ import { createLookup } from '@/utils/data'
 import { SentryLogger } from '@/utils/logs'
 
 import {
+  ORION_BATCHED_CHANNEL_VIEWS_QUERY_NAME,
   ORION_BATCHED_FOLLOWS_QUERY_NAME,
   ORION_BATCHED_VIEWS_QUERY_NAME,
+  ORION_CHANNEL_VIEWS_QUERY_NAME,
   ORION_FOLLOWS_QUERY_NAME,
   ORION_VIEWS_QUERY_NAME,
-  RemoveQueryNodeFollowsField,
-  RemoveQueryNodeViewsField,
-  TransformBatchedOrionFollowsField,
-  TransformBatchedOrionViewsField,
-  TransformOrionFollowsField,
-  TransformOrionViewsField,
-} from './transforms'
-import {
-  ORION_BATCHED_CHANNEL_VIEWS_QUERY_NAME,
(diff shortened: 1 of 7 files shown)


src/api/client/resolvers.ts
```diff
@@ -152,10 +152,99 @@ export const queryNodeStitchingResolvers = (
...
+      )
+
+      const batchedChannelViewsResolver = createResolverWithTransforms(
+        orionSchema,
+        ORION_BATCHED_CHANNEL_VIEWS_QUERY_NAME,
+        [TransformBatchedChannelOrionViewsField],
+        // operationName has to be manually kept in sync with the query name used
+        BATCHED_CHANNEL_VIEWS_QUERY_NAME
+      )
+      const batchedChannelViews = await batchedChannelViewsResolver(
```
kdembler: We should actually await both of those at the same time to reduce the request time. Like so:
```ts
const batchedChannelFollowsPromise = batchedChannelFollowsResolver(...)
const batchedChannelViewsPromise = batchedChannelViewsResolver(...)
const [batchedChannelFollows, batchedChannelViews] = await Promise.all([
  batchedChannelFollowsPromise,
  batchedChannelViewsPromise
])
```

I think the current resolver for `channels` query has the same problem, it'd be great if you could fix that as well

kdembler (changes requested):

src/api/client/resolvers.ts
drillprop: good idea, I'll refactor this

src/api/client/resolvers.ts
```diff
@@ -142,27 +132,131 @@ export const queryNodeStitchingResolvers = (
...
+        const followsLookup = createLookup<{ id: string; follows: number }>(batchedChannelFollows || [])
+        const viewsLookup = createLookup<{ id: string; views: number }>(batchedChannelViews || [])
+
+        return channels.map((channel: Channel) => ({
+          ...channel,
+          follows: followsLookup[channel.id]?.follows || 0,
+          views: viewsLookup[channel.id]?.views || 0,
+        }))
       } catch (error) {
         SentryLogger.error('Failed to resolve channel views', 'channels resolver', error)
```
kdembler: We should likely update the message here as it will catch both views&follows errors
```suggestion
        SentryLogger.error('Failed to resolve channel views or follows', 'channels resolver', error)
```

kdembler (approved):

PR #1235, opened 2021-08-23
Add ability to change joystream node

### LINKED ISSUE #1044: Allow changing joystream node you're connected to
Currently Atlas allows using only one pre-specified (via build env variables) URL for node to connect to. It'd be nice for the user to have ability to switch to a different node or provide their own URL

rafalpawlow:
Closes #1044


src/providers/environment/environment.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const contextValue: EnvironmentContextValue = {
+    selectedNode,
+    setSelectedNode,
+    setTargetEnv,
+    targetEnv,
+  }
+
+  return <EnvironmentContext.Provider value={contextValue}>{children}</EnvironmentContext.Provider>
+}
```
kdembler: Why do we need this provider at all? I think all this info could be extracted directly from store?

src/config/envs.ts
```diff
@@ -1,16 +1,11 @@
 type BuildEnv = 'production' | 'development'
 
 export const BUILD_ENV = (process.env.REACT_APP_ENV || 'production') as BuildEnv
-const target_env = window.localStorage.getItem('target_env')
-export const TARGET_DEV_ENV = target_env ? JSON.parse(target_env) : 'development'
+const environment = window.localStorage.getItem('environment')
```
kdembler: Let's use the store hook here. You can do `useEnvironmentStore.getState()`

kdembler (changes requested):

src/providers/environment/environment.tsx
rafalpawlow: I did this because at first I thought we use it in several places. But you are right, it's better to get rid of this.

src/providers/environment/environment.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return { setTargetEnv, targetEnv }
+}
+
+export const useNode = () => {
+  const { setSelectedNode, selectedNode } = useEnvironmentStore((state) => ({
+    ...state.actions,
+    ...state,
+  }))
+  return { setSelectedNode, selectedNode }
+}
```
kdembler: We still create context here. I'm pretty sure we don't need it. Also, do we need those wrapper hooks for the store? I think we could just access the store directly in components that need it

src/views/admin/AdminView.tsx
```diff
@@ -1,33 +1,96 @@
...
+  {
+    name: 'Sumer Dev',
+    value: 'wss://sumer-dev-2.joystream.app/rpc',
+  },
+  {
+    name: 'Local node',
+    value: 'ws://127.0.0.1:9944',
+  },
+]
+
```
kdembler: Let's stick this into `@/config`

src/providers/environment/store.ts
```diff
@@ -0,0 +1,44 @@
...
+const LOCAL_STORAGE_KEY = 'environment'
+
+export type EnvironmentState = {
+  selectedNode: string
+  targetEnv: string
+}
+
+const INITIAL_STATE: EnvironmentState = {
+  targetEnv: 'development',
+  selectedNode: 'wss://sumer-dev-2.joystream.app/rpc',
```
kdembler: Why do we use this value as default? If we deployed this on prod, then everyone would be connecting to the wrong node. If overwrite wasn't provided we still need to connect to the correct node. I suggest we treat this only as an overwrite, so by default would be empty and we would still use values from the environment. Only if someone manually selected a different node, should we use that one instead.

kdembler (changes requested):

kdembler (approved):

PR #1248, opened 2021-08-24
Wrap all video content pages into single template

rafalpawlow:


src/components/templates/VideoContentTemplate.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <CallToActionButton
+          key={`cta-${idx}`}
+          label={CTA_MAP[item].label}
+          to={CTA_MAP[item].to}
+          colorVariant={CTA_MAP[item].colorVariant}
+          icon={CTA_MAP[item].icon}
+        />
+      )),
+    [cta]
+  )
```
kdembler: We don't need this `useMemo`, let's get rid of it. This component will re-render very rarely (most likely never) so optimization isn't needed

src/components/templates/VideoContentTemplate.tsx
```diff
@@ -0,0 +1,93 @@
...
+  font-size: ${typography.sizes.h3};
+  font-weight: ${typography.weights.bold};
+
+  ${media.large} {
+    margin: ${sizes(16)} 0 ${sizes(17)};
+    font-size: ${typography.sizes.h2};
+  }
+`
+
+const StyledViewWrapper = styled(ViewWrapper)`
```
kdembler: Why do we need this `ViewWrapper`? I think it's something quite old, it contains only `position: relative;`. We should use `LimitedWidthContainer` instead to make sure the pages don't exceed their max width

src/components/templates/index.ts
```diff
@@ -0,0 +1 @@
+export * from './VideoContentTemplate'
```
kdembler: Let's not use `index.ts` here and import directly, like with the rest of the components

kdembler (changes requested):

src/components/templates/VideoContentTemplate.tsx
```diff
@@ -0,0 +1,90 @@
...
+  popular: {
+    label: 'Popular on Joystream',
+    to: absoluteRoutes.viewer.popular(),
+    colorVariant: 'red',
+    icon: <SvgNavPopular />,
+  },
+}
+
+export const VideoContentTemplate: React.FC<VideoContentTemplateProps> = ({ children, title, cta }) => {
+  const renderCta =
```
kdembler: This shouldn't be called `renderCta` as this isn't a render function, it's just an already computed value. Something like `ctaContent` or alike would be a better name

src/components/templates/VideoContentTemplate.tsx
```diff
@@ -0,0 +1,90 @@
...
+  )
+}
+
+const Header = styled(Text)`
+  margin: ${sizes(16)} 0 ${sizes(18)};
+  font-size: ${typography.sizes.h3};
+  font-weight: ${typography.weights.bold};
+
+  ${media.large} {
+    margin: ${sizes(16)} 0 ${sizes(17)};
```
kdembler: margins aren't correct here, design specifies constant `margin: 64px 0;`

src/components/templates/VideoContentTemplate.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <StyledViewWrapper big>
+      {title && <Header>{title}</Header>}
+      {children}
+      {cta && <CallToActionWrapper>{renderCta}</CallToActionWrapper>}
+    </StyledViewWrapper>
+  )
+}
+
+const Header = styled(Text)`
```
kdembler, line 71: This is missing some of necessary typography styles

kdembler (changes requested):

src/components/templates/VideoContentTemplate.tsx
rafalpawlow: Did I select wrong layer?
Zrzut ekranu 2021-08-26 o 10 57 57
Zrzut ekranu 2021-08-26 o 10 58 04

kdembler: You're measuring it a bit incorrectly. You're trying to find distance between two text nodes, but in reality `Top 10 channels` actually has also some container around it so if you take 68/72 and apply it as margin, the padding around `Top 10 channels` will cause the distance between the text nodes to be more like 80 or something.

2 tips for measuring in Figma:
1. You can always use the nodes list on the left to make sure you have a proper container selected instead of a child of it.
2. If you hold Cmd (on Mac, I guess it's Ctrl for others), Figma will instead measure distance to the closest container it can find along the way, not to the node you're hovering above.

You can see those in this video (in the second part I'm pressing Cmd):

https://user-images.githubusercontent.com/12646744/130935411-6ca32013-3429-421d-b7a7-ea5284874d29.mov

kdembler (approved):

PR #1250, opened 2021-08-24
Reduce CTA's font size on small screens

rafalpawlow:


src/shared/components/CallToActionButton/CallToActionButton.tsx
```diff
@@ -29,7 +29,7 @@ export const CallToActionButton: FC<CallToActionButtonProps> = ({
     <StyledContainer {...linkProps} onClick={onClick} colorVariant={colorVariant}>
       <ContentWrapper>
         <IconWrapper colorVariant={colorVariant === 'blue' ? 'lightBlue' : colorVariant}>{icon}</IconWrapper>
-        <BodyWrapper variant="h5">
+        <BodyWrapper>
```
kdembler: Let's add `variant="h6"` to make sure we apply any additional header-related styles

src/shared/components/CallToActionButton/CallToActionButton.style.ts
```diff
@@ -41,6 +41,12 @@ export const BodyWrapper = styled(Text)`
   display: flex;
   align-items: center;
   justify-content: space-between;
+  font-size: ${typography.sizes.h6};
+  font-weight: ${typography.weights.bold};
```
kdembler: If you use the variant, you don't have to add this explicitly

src/shared/components/CallToActionButton/CallToActionButton.style.ts
```diff
@@ -41,6 +41,12 @@ export const BodyWrapper = styled(Text)`
...
+
+  ${media.compact} {
+    font-size: ${typography.sizes.h5};
```
kdembler, line 46: We should also add `line-height` here

kdembler (changes requested):

src/shared/components/CallToActionButton/CallToActionButton.style.ts
```diff
@@ -41,6 +41,12 @@ export const BodyWrapper = styled(Text)`
   display: flex;
   align-items: center;
   justify-content: space-between;
+  font-size: ${typography.sizes.h6};
```
kdembler:
```suggestion
```

kdembler (approved):

PR #1251, opened 2021-08-24
Change size of video hero controls and make them always visible

### LINKED ISSUE #1243: VideoHero tweaks
1. Buttons should use 48px
2. Let's make the buttons visible right away and not appear later. Let's just disable them when hero is not ready

rafalpawlow:
Closes #1243

### DIFF
--- a/src/components/VideoHero/VideoHero.tsx
+++ b/src/components/VideoHero/VideoHero.tsx
@@ -34,7 +34,6 @@ export const VideoHero: React.FC = () => {
   const coverVideo = useVideoHero()
 
   const [videoPlaying, setVideoPlaying] = useState(false)
-  const [disabledControls, setDisabledControls] = useState(true)
   const [soundMuted, setSoundMuted] = useState(true)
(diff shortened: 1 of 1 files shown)


src/components/VideoHero/VideoHero.tsx
```diff
@@ -45,7 +43,7 @@ export const VideoHero: React.FC = () => {
 
   const handlePlaybackDataLoaded = () => {
     setTimeout(() => {
-      setDisplayControls(true)
+      setDisabledControls(false)
```
kdembler: After testing it out I've noticed that after the most recent change to the video hero this doesn't make much sense. Previously this button was controlling the playback inside the video hero but now it only navigates you to the video. So I think we could actually make the button enabled always (i.e. as long as the `coverVideo` is loaded). Let's remove the logic for disabling this button

kdembler (changes requested):

kdembler (approved):

PR #1252, opened 2021-08-25
Adjust video tile variant's font sizes and avatar

### LINKED ISSUE #1246: Scaling of VideoTile
It seems that updated scaling of VideoTile wasn't implemented yet. We still do the linear scaling based on width that applies to all texts. Please refer to the VideoTile design which explains how this should behave

rafalpawlow:
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A126073

Closes #1246

### DIFF
--- a/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
+++ b/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
@@ -11,6 +11,7 @@ import { Text } from '../Text'
 
 export const HOVER_BORDER_SIZE = '2px'
 const KEBAB_MENU_WIDTH = sizes(8)
+const AVATAR_WITH_MARGIN = sizes(13)
 
 type SizeProps = {
   size?: 'small' | 'big'
@@ -97,18 +98,13 @@ export const Container = styled.article`
 
 export const InfoContainer = styled.div`
   min-height: 86px;
-  display: flex;
+  display: grid;
+  grid-template-columns: ${AVATAR_WITH_MARGIN} auto ${KEBAB_MENU_WIDTH};
   margin-top: ${sizes(3)};
 `
 
 export const AvatarContainer = styled.div`
   ${square(sizes(10))};
-
-  margin-right: ${sizes(3)};
-`
-
-export const TextContainer = styled.div`
-  width: calc(100% - ${KEBAB_MENU_WIDTH});
 `
 
 type MetaContainerProps = { noMarginTop: boolean }
--- a/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
+++ b/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
@@ -5,12 +5,18 @@ import { Link } from 'react-router-dom'
 
 import { colors, media, sizes, square, transitions, typography, zIndex } from '@/shared/theme'
 
+import { TileSize } from './VideoTileBase'
+
 import { Avatar } from '../Avatar'
 import { IconButton } from '../IconButton'
 import { SkeletonLoader } from '../SkeletonLoader'
 import { Text } from '../Text'
 
 export const HOVER_BORDER_SIZE = '2px'
+
+type SizeProps = {
+  size?: TileSize
+}
 
 type MainProps = {
   main: boolean
@@ -22,10 +28,6 @@ type ChannelProps = {
 
 type ClickableProps = {
   clickable: boolean
-}
-
-type ScalesWithCoverProps = {
-  scalingFactor: number
 }
 
 export const CoverWrapper = styled.div<MainProps>`
@@ -120,10 +122,9 @@ export const InfoContainer = styled.div<MainProps>`
   ${({ main }) => main && mainInfoContainerCss};
 `
 
-export const AvatarContainer = styled.div<ScalesWithCoverProps>`
-  width: calc(40px * ${(props) => props.scalingFactor});
-  min-width: calc(40px * ${(props) => props.scalingFactor});
-  height: calc(40px * ${(props) => props.scalingFactor});
+export const AvatarContainer = styled.div`
+  ${square(sizes(10))};
+
   margin-right: ${sizes(3)};
 `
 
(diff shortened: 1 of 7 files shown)


src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -5,13 +5,19 @@ import { Link } from 'react-router-dom'
 
 import { colors, media, sizes, square, transitions, typography, zIndex } from '@/shared/theme'
 
+import { TileSize } from './VideoTileBase'
```
kdembler: Let's not do this - it will create a circular dependency between `VideoTileBase.styles` and `VideoTileBase`. Let's either define `TileSize` here or move types to a separate file

src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -275,11 +276,11 @@ export const StyledAvatar = styled(Avatar)<ChannelProps>`
...
-export const TitleHeader = styled(Text)<MainProps & ScalesWithCoverProps & ClickableProps>`
+export const TitleHeader = styled(Text)<MainProps & ClickableProps & SizeProps>`
   margin: 0;
   margin-bottom: ${sizes(2)};
   font-weight: ${typography.weights.bold};
-  font-size: calc(${(props) => props.scalingFactor} * ${typography.sizes.h6});
+  font-size: ${({ size }) => (size === 'small' ? typography.sizes.h6 : typography.sizes.subtitle1)};
   ${({ main }) => main && fluidRange({ prop: 'fontSize', fromSize: '24px', toSize: '40px' })};
 
   line-height: ${({ main }) => (main ? 1 : 1.25)};
```
kdembler: We also need to update this to use `h6` and `subtitle1` values respectively

src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -275,11 +276,11 @@ export const StyledAvatar = styled(Avatar)<ChannelProps>`
...
 `
 
-export const TitleHeader = styled(Text)<MainProps & ScalesWithCoverProps & ClickableProps>`
+export const TitleHeader = styled(Text)<MainProps & ClickableProps & SizeProps>`
   margin: 0;
   margin-bottom: ${sizes(2)};
   font-weight: ${typography.weights.bold};
-  font-size: calc(${(props) => props.scalingFactor} * ${typography.sizes.h6});
+  font-size: ${({ size }) => (size === 'small' ? typography.sizes.h6 : typography.sizes.subtitle1)};
   ${({ main }) => main && fluidRange({ prop: 'fontSize', fromSize: '24px', toSize: '40px' })};
```
kdembler: Let's actually remove everything `main` related. We still use this variant in search results, but let's just get rid of it there.

src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -335,7 +329,7 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
                   {isLoading ? (
                     <SpacedSkeletonLoader height={main ? 16 : 12} width={main ? '40%' : '80%'} />
                   ) : createdAt ? (
-                    <MetaText variant="subtitle2" main={main} scalingFactor={scalingFactor} secondary>
+                    <MetaText variant="subtitle2" main={main} secondary>
```
kdembler: Both this and `ChannelHandle` should use `body2` variant according to design

src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -291,15 +292,14 @@ export const TitleHeader = styled(Text)<MainProps & ScalesWithCoverProps & Click
   overflow-wrap: break-word;
 `
 
-export const ChannelHandle = styled(Text)<ChannelProps & ScalesWithCoverProps>`
-  font-size: calc(${(props) => props.scalingFactor} * ${typography.sizes.subtitle2});
+export const ChannelHandle = styled(Text)<ChannelProps>`
+  font-size: ${typography.sizes.subtitle2};
```
kdembler: If you set the proper variant for the component we don't need this style at all

src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -291,15 +292,14 @@ export const TitleHeader = styled(Text)<MainProps & ScalesWithCoverProps & Click
...
   display: inline-block;
   cursor: ${({ channelClickable }) => (channelClickable ? 'pointer' : 'auto')};
 `
 
-export const MetaText = styled(Text)<MainProps & ScalesWithCoverProps>`
-  font-size: ${({ main, scalingFactor }) =>
-    main ? typography.sizes.h6 : `calc(${scalingFactor}*${typography.sizes.subtitle2}) `};
+export const MetaText = styled(Text)<MainProps>`
+  font-size: ${({ main }) => (main ? typography.sizes.h6 : typography.sizes.subtitle2)};
```
kdembler: Once we remove `main` we can remove this as well

src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -155,16 +150,22 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
...
+  const [tileSize, setTileSize] = useState<TileSize>(undefined)
+
   const { ref: imgRef } = useResizeObserver<HTMLImageElement>({
     onResize: (size) => {
       const { width: videoTileWidth, height: videoTileHeight } = size
+      if (videoTileWidth && !main) {
+        if (tileSize !== 'small' && videoTileWidth <= SMALL_SIZE_WIDTH) {
+          setTileSize('small')
+        }
+        if (tileSize !== 'big' && videoTileWidth > SMALL_SIZE_WIDTH) {
```
kdembler: Let's change the inclusivity of this condition and follow mobile-first. So if width === 300 let's already use the bigger version (I know design says otherwise but we talked about this with designers and this should be approach we're using)

src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -155,16 +150,22 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
...
+          setTileSize('big')
+        }
+      }
       if (onCoverResize) {
         onCoverResize(videoTileWidth, videoTileHeight)
       }
```
kdembler: I'm pretty sure this isn't used everywhere, let's get rid of this prop

kdembler (changes requested):

kdembler:
I also found this issue that was there before but I think it's a good time to fix it as well:

https://user-images.githubusercontent.com/12646744/130937216-46df900b-fbb3-4f46-8736-cdc4fa649c4c.mov

Seems we reserve too little space for the icon button. I see that we're setting explicit `width` for TextContainer by doing some calc. I think we could instead just have some `padding-right` on `InfoContaine`?

rafalpawlow:
@kdembler Adding padding to `InfoContainer` will not work because there is Kebab Menu inside.

kdembler (approved):

PR #1253, opened 2021-08-25
Detect finalized upload

### LINKED ISSUE #1239: Detect when upload is finalized
Currently we mark upload as finished when the upload request to storage provider is done. However, that's not the full picture. Instead, once that request is done, we should start polling the query node to detect when the status of that dataObject changes to `ACCEPTED`. At this point the upload is finalized and if user were to open the video, it would play just fine. For now let's just prepare the logic for that and use some temporary UI indication

drillprop:
Fix #1239 

For now, UploadGroup will be just displaying `Processing` when the upload is finalizing and `CircularProgress` will be just stuck on 100% for a few seconds. Also, didn't update the snack bar logic. I will deal with those in separate PRs


src/api/hooks/dataObject.ts
```diff
@@ -0,0 +1,19 @@
...
+  const [getDataObjectAvailability, { data, ...rest }] = useGetDataObjectAvailabilityLazyQuery(opts)
+
+  return {
+    getDataObjectAvailability: (contentId: string) =>
+      getDataObjectAvailability({
+        variables: {
+          joystreamContentIdEq: contentId,
+        },
+      }),
+    dataObjectAvailability: data?.dataObjects.length ? data.dataObjects[0].liaisonJudgement : undefined,
```
kdembler: btw you can also use optional chaining here:
```ts
data.dataObjects?.[0].liaisonJudgement
```

src/providers/uploadsManager/useStartFileUpload.tsx
```diff
@@ -29,6 +30,39 @@ export const useStartFileUpload = () => {
...
+
+  useEffect(() => {
+    if (
+      !variables?.joystreamContentIdEq ||
+      uploadStatus[variables?.joystreamContentIdEq]?.lastStatus !== 'proccessing'
+    ) {
+      return
+    }
+    if (dataObjectAvailability === 'PENDING') {
+      startPolling?.(3000)
```
kdembler: let's move this to `config` and let's call it something like `ASSET_POLLING_INTERVAL` and make this value 2000 for now

kdembler (changes requested):
Other than the comments I've made, there is also a much bigger issue that will likely not be trivial to solve. You currently do the polling inside `useStartFileUpload` which is kind of ephemeral. You use this hook to start a download but you cannot expect that this hook will keep being mounted. For example, if you use it in `EditVideoSheet`, it will likely work because when we close the sheet, we actually just move it below the screen but it's still mounted so the hook will persist and the polling will continue. However, if you use in `CreateEditChannelView`, it looks different. Once you finish editing your channel you leave the view and so all the hooks mounted there, including `useStartFileUpload` and consequentially `useDataObjectAvailabilityLazy` will get unmounted and the polling will stop. You can check that as long as you are on channel edit view the polling will happen but once you leave it will stop.

Now, what can we do about it. I think we may need to change our approach to uploads a bit. For this reason but also for others. How I imagine this could work: when view wants to schedule an upload, they don't call any hook but just some store action that will add something to "pending uploads". Then, some long-living component (like `UploadsManager`) handles new entries to pending uploads and actually executes upload logic for them (HTTP request, etc). This way we can be sure that when upload is scheduled, its full logic will get executed no matter what happens to the view that scheduled it. This is a rough draft of how I think this could work, I'm open to suggestions. Also let me know if you have any questions

src/api/hooks/dataObject.ts
```diff
@@ -0,0 +1,26 @@
+import { QueryHookOptions } from '@apollo/client'
+import { useCallback } from 'react'
+
+import { GetDataObjectAvailabilityQuery, useGetDataObjectAvailabilityLazyQuery } from '@/api/queries'
+
+type DataObjectOpts = QueryHookOptions<GetDataObjectAvailabilityQuery>
```
kdembler: Let's avoid creating a separate type if not needed, we can supply `QueryHookOptions<GetDataObjectAvailabilityQuery>` directly in the fn definition

src/providers/uploadsManager/types.ts
```diff
@@ -4,7 +4,7 @@ import { AssetDimensions, ImageCropData } from '@/types/cropper'
 type AssetType = 'video' | 'thumbnail' | 'cover' | 'avatar'
 export type AssetParent = 'video' | 'channel'
 
-export type AssetUploadStatus = 'completed' | 'inProgress' | 'error' | 'reconnecting'
+export type AssetUploadStatus = 'completed' | 'inProgress' | 'error' | 'reconnecting' | 'proccessing'
```
kdembler: should be `processing`, with single `c`

src/views/studio/MyUploadsView/MyUploadsView.style.tsx
```diff
@@ -1,11 +1,13 @@
 import styled from '@emotion/styled'
 
 import { LimitedWidthContainer } from '@/components/LimitedWidthContainer'
+import { TOP_NAVBAR_HEIGHT } from '@/components/TopbarBase'
 import { Text } from '@/shared/components/Text'
 import { sizes } from '@/shared/theme'
 
 export const UploadsContainer = styled(LimitedWidthContainer)`
   padding-bottom: 120px;
+  min-height: calc(100vh - ${TOP_NAVBAR_HEIGHT}px - ${sizes(8)});
```
kdembler: Why do we make this change?

kdembler (commented):
Other than those minor changes, seems to be working fine! 👍

src/views/studio/MyUploadsView/MyUploadsView.style.tsx
drillprop: Hmm, I think that I wanted to check something and committed this by mistake. Will erase these lines.

kdembler (approved):

Issue #1230, opened 2021-08-23
General Category Card component


eldiegod:

toiletgranny:
@eldiegod, as with the Featured Category Card, this one looks stunning, too!

There are also some minor details that don't quite match the designs. Here they are:
1. **AS IS:** The pie chart is not centered vertically with the "123 videos" label.
**TO BE:** It should be vertically centered. I think `align-items: center` on the `.VideosNumberContainer` should do the trick.
2. **AS IS:** The grayscale filter on the cover image doesn't animate on hover.
**TO BE:** If possible, the grayscale filter should be animated with the same timing and easing as all the other parameters on hover.
3. **AS IS:** It seems like the blue-ish color overlay is missing on the cover image, the image is just grayscale.
**TO BE:** In Figma, I have this additional "Color overlay" rectangle applied on top of the cover image. Its color is set to `Neutral / 800` and blending mode to `Color`. If possible, it would be great to have this in implementation as well. Happy to jump on a quick call to discuss details around this.
4. **AS IS:** _(It's the same comment as in the Featured Category Card component)_ It seems like the "Science & Technology" icon has been exported without the surrounding `24x24` container around it.
**TO BE:** All icons should be exported with the `24x24` container so that they are positioned based on this wrapper.

PR #1255, opened 2021-08-26
Discovery page cards

eldiegod:
fixes:  #1230

https://www.figma.com/file/UdBMxrLspjcf6IVYYJS8YP/%5BPage%5D-Discover-%26-Video-category?node-id=1%3A30617


src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,134 @@
...
+  grid-template-rows: auto auto 1fr;
+  padding: ${({ variantCategory }) => (variantCategory === 'default' ? sizes(8) : sizes(4))};
+`
+
+export const IconCircle = styled.div<ColorProps>`
+  display: flex;
+  margin-bottom: ${sizes(4)};
+  align-items: center;
+  justify-content: center;
+  width: 40px;
```
rafalpawlow: Let's use `sizes` helper, or even better `square` helper

eldiegod:
API is likely to change once I start implementing the layout of the discovery page so this is mostly for styles purposes

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,134 @@
...
+  background: url(${({ bgImgUrl }) => bgImgUrl});
+  background-position: center;
+  background-size: cover;
+  filter: grayscale(100%);
+`
+
+const defaultStyles = ({ variantCategory }: VariantProps) =>
+  variantCategory === 'default' &&
+  css`
+    height: 228px;
```
rafalpawlow:
```suggestion
    height: ${sizes(57)};
```

rafalpawlow (changes requested):

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
kdembler: I actually prefer the current version

src/shared/components/VideoCategoryCard/VideoCategoryCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+              }
+            ></CoverImg>
+          )}
+        </GeneralContainer>
+      </CSSTransition>
+    </SwitchTransition>
+  )
+}
+
+export const FeaturedVideoCategoryCard: React.FC<VideoCategoryCardProps> = ({
```
kdembler: Let's move this to a separate file

src/shared/components/VideoCategoryCard/VideoCategoryCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+              )}
+            </VideoCountContainer>
+          </Content>
+
+          {variant === 'default' && !loading && (
+            <CoverImg
+              bgImgUrl={
+                'https://eu-central-1.linodeobjects.com/atlas-assets/category-images/science-and-technology.webp'
+              }
+            ></CoverImg>
```
kdembler: If you have an empty tag it doesn't make sense to include it's closing tag. Just use `<CoverImg ... />`

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,180 @@
+import { css } from '@emotion/react'
+import styled from '@emotion/styled'
+import { transparentize } from 'polished'
+
+import { colors, sizes, transitions } from '@/shared/theme'
+
+import { VideoCategoryCardProps } from '.'
```
kdembler: You will create circular dependencies this way, let's avoid it

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,180 @@
...
+import { Text } from '../Text'
+
+type ColorProps = { color: string }
+type LoadingProps = { loading?: VideoCategoryCardProps['loading'] }
+type VariantProps = { variantCategory?: VideoCategoryCardProps['variant'] }
+export const CoverImg = styled.div<{ bgImgUrl: string }>`
+  background: url(${({ bgImgUrl }) => bgImgUrl});
+  background-position: center;
+  background-size: cover;
+  filter: grayscale(100%);
```
kdembler: Why are we using a filter instead of a color overlay like the design does?

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,180 @@
...
+  variantCategory === 'default' &&
+  css`
+    height: 228px;
+    grid-template-columns: 2fr 1fr;
+  `
+export const GeneralContainer = styled(Container)`
+  ${generalStyles}
+`
+
+export const PlayerContainer = styled.div`
```
kdembler: Let's extract styles for `FeaturedVideoCategoryCard` to a separate file

kdembler (changes requested):
Also some general points:
1. Your icons are exported the wrong way. You can easily see they vary in widths and heights - `width={18} height={21}`, `width={20} height={20}`, etc. They should all be 24x24. If you need help with that, ping me.
2. Implementation of the card hover effect isn't exactly correct. You use a disappearing border on the card which creates a different visual effect than what the design describes. There is a [separate section in Figma](https://www.figma.com/file/UdBMxrLspjcf6IVYYJS8YP/Page-Discover-and-Video-category?node-id=1%3A30625) that describes how to build this component to achieve this effect, why not follow it?
3. On hover there should be a transition on the image colors, currently there is a jump. If you follow Figma structure and use color overlay instead of a filter, that shouldn't be an issue

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
eldiegod: not needed anymore the overlay is there

eldiegod:
I think everything here should be already addressed; for point number 2 I discussed it with Adam and we get to save on that extra complexity for the component

imagen

src/shared/components/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import { SkeletonLoader } from '../SkeletonLoader'
+import { Text } from '../Text'
+import { VideoPlayer } from '../VideoPlayer'
+
+export type FeaturedVideoCategoryCardProps = {
+  variant?: 'default' | 'compact'
+  loading?: boolean
+  color: string
+} & React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>
```
kdembler: Why do we need those details props here?

src/shared/components/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+                <Text variant={variant === 'default' ? 'h3' : 'h4'}>Science & Techology</Text>
+              )}
+            </div>
+
+            {!loading && (
+              <FeaturedVideoTitleContainer variantCategory={variant}>
+                <FeaturedVideoText variant={'caption'} secondary>
+                  Featured video
+                </FeaturedVideoText>
+                <Text variant={'h6'}>KOIOS Blockchain Week</Text>
```
kdembler: Why not just `variant="caption"`? Seems cleaner :D

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,147 @@
...
+type VariantProps = { variantCategory?: 'default' | 'compact' }
+
+export const CoverImg = styled.div<{ bgImgUrl: string }>`
+  position: relative;
+  width: 100%;
+  height: 100%;
+  background: url(${({ bgImgUrl }) => bgImgUrl});
+  background-position: center;
+  background-size: cover;
+`
```
kdembler, line 20: I'm wondering - is there any benefit to using div with `background` instead of `img` here?

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,147 @@
...
+    border-left: 0 solid ${color};
+    transform: translate(-${sizes(2)}, -${sizes(2)});
+    box-shadow: ${sizes(2)} ${sizes(2)} 0 ${color};
+  `
+
+export const Container = styled.div<ColorProps & VariantProps & LoadingProps>`
+  transition: all ${transitions.timings.regular} ${transitions.easing},
+    border ${transitions.timings.sharp} ${transitions.easing};
+  display: grid;
+  cursor: pointer;
```
kdembler: Maybe we should apply this only when not loading?

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export const IconCircle = styled.div<ColorProps>`
+  display: flex;
+  margin-bottom: ${sizes(4)};
+  align-items: center;
+  justify-content: center;
+  width: ${sizes(10)};
+  height: ${sizes(10)};
+  border-radius: 100%;
+  background: ${({ color }) => transparentize(0.8, color)};
```
kdembler, line 85: Could we just apply regular color and add `opacity`? I'd like to avoid `polished` helpers where possible

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
```diff
@@ -0,0 +1,147 @@
...
+`
+
+export const Title = styled(Text)<VariantProps>`
+  margin-bottom: ${({ variantCategory }) => (variantCategory === 'default' ? sizes(6) : sizes(4))};
+`
+
+// ref https://codeburst.io/how-to-pure-css-pie-charts-w-css-variables-38287aea161e
+export const PieChart = styled.div`
+  margin-right: ${sizes(2)};
+  background: ${transparentize(0.8, colors.gray[300])};
```
kdembler, line 99: Likewise, this could just use `opacity`

kdembler (commented):
Looking really good! Mostly some questions left

src/shared/components/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
eldiegod: It's nice to have i.e. supports className and style props for the container div

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
eldiegod: Tried an img first but it didn't fit well `background-position: center;` and `background-size: cover;` help

src/shared/components/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
kdembler: I think it's more clean to just include `className` directly as we do for other components, I don't think we should need any other prop, for example we shouldn't use the `style` prop directly

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
eldiegod: hmm no because then the child elements get transparent as well and we don't want that

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
eldiegod: as I mention above child elements get transparent with opacity, option would be to hard code this color but that would be worse imo

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
kdembler: True, you would need to change the layout like the design has. It has a container that has 2 children - background and icon

src/shared/components/VideoCategoryCard/VideoCategoryCard.style.ts
kdembler: I won't be able to tell much without any details

kdembler (approved):

PR #1257, opened 2021-08-27
Adjust video grids responsivness

### LINKED ISSUE #1245: Responsive improvements in video grids
Our current approach to displaying items in video grids doesn't work very well. We always start with 2 rows of videos - on wide desktop that means 6*2 videos which is fine. But on mobile this means only 1*2 which isn't enough. The design specifies the policy for managing that - the number of rows changes based on the screen width. This also applies to sections like `Discover new channels` where we only show 1 row (so only 1 video on mobile). Please refer to the design for how this should look like.

In terms of implementation, we should also be smarter about how many videos we fetch. If we only show 3 videos at a time, why only fetch 3 and make the user wait for next 3? We could easily fetch like 20 in advance and improve that experience

rafalpawlow:
Closes #1245


src/components/ChannelWithVideos/ChannelWithVideos.tsx
```diff
@@ -26,21 +27,48 @@ type ChannelWithVideosProps = {
...
+      return
+    }
+    if (window.innerWidth >= breakpoints.small) {
+      setVideoRows(2)
+      return
+    }
+    if (window.innerWidth >= breakpoints.base) {
+      setVideoRows(3)
+    }
+  }
```
kdembler: Let's use new `useMediaMatch` for those

src/shared/theme/breakpoints.ts
```diff
@@ -1,12 +1,12 @@
...
-  xlarge: '1920px',
-  xxlarge: '2560px',
+  base: 320,
+  compact: 450,
+  small: 600,
+  smalldium: 768,
+  medium: 1024,
+  large: 1440,
+  xlarge: 1920,
+  xxlarge: 2560,
```
kdembler: Is this necessary? I think we use values with `px` more often than we use the raw ones?

rafalpawlow: I guess this won't be needed when I use `useMediaMatch`, right?

kdembler: yep!

src/hooks/useVideoGridRows.tsx
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from 'react'
+
+import { useMediaMatch } from '@/hooks/useMediaMatch'
+
+type UseVideoGridRowsArgs = 'videos' | 'videosInChannel'
```
kdembler: How about this?
```suggestion
type VideoGridRowsVariant = 'default' | 'compact'
```

kdembler: I'm also wondering whether we should use the exact same policy for channel view - we previously used 4 rows by default there, now it will use 2 on biggest screen size, like content sections. Waiting for reply from Kuba as to what he thinks we should do there

rafalpawlow: > How about this?

I wanted to use more generic names but I wasn't sure if we could actually call it a default, especially if we get one more variant.

kdembler: Yeah let's wait for now

kdembler: Okay, let's have a separate variant for channel view in which we will use 4 rows on the biggest screens. You can also adjust other breakpoints for that as you see fit

src/components/InfiniteGrids/InfiniteVideoGrid.tsx
```diff
@@ -70,6 +70,10 @@ export const InfiniteVideoGrid: React.FC<InfiniteVideoGridProps> = ({
   titleLoader,
 }) => {
   const [videosPerRow, setVideosPerRow] = useState(INITIAL_VIDEOS_PER_ROW)
+  const rowsToLoad = useVideoGridRows()
+  const [_targetRowsCount, setTargetRowsCount] = useState<number>(rowsToLoad)
```
drillprop:
```suggestion
  const [_targetRowsCount, setTargetRowsCount] = useState(rowsToLoad)
```
not needed

drillprop (approved):

PR #1263, opened 2021-08-27
Hide Show more videos header when there is no more videos on channel

### LINKED ISSUE #1220: In `VideoView` "More from ..."  header shouldn't be displayed when there is no more videos on Channel
Just noticed that:
image
It's a little bit misleading. I think we could just safely remove this section when a channel doesn't contain more videos.

rafalpawlow:
Closes #1220

### DIFF
--- a/src/views/viewer/VideoView/VideoView.tsx
+++ b/src/views/viewer/VideoView/VideoView.tsx
@@ -2,7 +2,7 @@ import { throttle } from 'lodash'
 import React, { useCallback, useEffect, useState } from 'react'
 import { useParams } from 'react-router-dom'
 
-import { useAddVideoView, useVideo, useVideos } from '@/api/hooks'
+import { useAddVideoView, useVideo } from '@/api/hooks'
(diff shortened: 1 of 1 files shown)


src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -72,6 +72,11 @@ export const VideoView: React.FC = () => {
   }, [video?.duration, timestampFromQuery])
 
   const channelId = video?.channel.id
+  const { videos } = useVideos(
+    { where: { channelId_eq: channelId, isPublic_eq: true, isCensored_eq: false } },
+    { skip: !channelId }
+  )
+  const shouldShowMoreVideos = videos && videos.length > 1
```
kdembler: Do we need to fetch data manually here? I think `InfiniteVideoGrid` can handle it on it's own if pass it the title?

kdembler (changes requested):

kdembler (approved):

PR #1268, opened 2021-08-30
replace CRA with vite, upgrade libraries

kdembler:
### Left to do:

- [x] Fix Polkadot's issues with Buffers
- [x] Update prod build commands
- [x] Check if prod build is correct
- [x] Update Storybook to use Vite as well
- [x] Think about strategy for codecheck - vite doesn't do typechecking
- [x] Update docs

### Summary of changes:
- Get rid of `CRA` and webpack completely, replace with vite
- Switch to yarn v3
- Upgrade libraries - `react`, `eslint`, `typescript`
- Remove `patch-package` and use built-in yarn patching
- Remove some unused components
- Flatten out structure of `components/Topbar` and `Sidenav`


vite.config.ts
```diff
@@ -0,0 +1,77 @@
...
+  //         '@polkadot/api-derive',
+  //         '@polkadot/rpc-core',
+  //         '@polkadot/rpc-provider',
+  //         '@joystream/types',
+  //         '@joystream/content-metadata-protobuf',
+  //       ],
+  //     },
+  //   },
+  // },
+  // },
```
kdembler: I may still tweak this part so left as comment for time being

.gitignore
```diff
@@ -1,76 +1,26 @@
...
-*.tgz
-
-# cache and distribution folders
 .parcel-cache
 .cache/
 .now
 dist/
 storybook-static
-# Mac files
-.DS_Store
```
drillprop, line 61: We should keep this line: 
image

package.json
```diff
@@ -10,33 +10,26 @@
...
   "scripts": {
-    "bundle-stats": "webpack-bundle-analyzer dist/bundle-stats.json -m static -r dist/stats.html",
-    "start": "react-app-rewired start",
-    "dev": "yarn start",
-    "build": "react-app-rewired build --stats && yarn bundle-stats",
-    "eject": "react-app-rewired eject",
+    "start": "vite",
+    "dev": "vite",
+    "build": "vite build",
+    "storybook": "start-storybook -p 6006",
```
drillprop, line 22: storybook not working for me. I'm getting: 
image

vite.config.ts
```diff
@@ -0,0 +1,77 @@
...
+import { defineConfig } from 'vite'
+import checker from 'vite-plugin-checker'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [
+    reactRefresh(),
+    checker({
+      typescript: true,
+      eslint: { files: ['./src'], extensions: ['.js', '.jsx', '.ts', '.tsx'] },
```
drillprop, line 16: I'm getting this overlay for eslint warnings: 
image
image
For better developer experience I think we should see this only when an error occurs(not warning)

src/views/playground/Playgrounds/OrionTesting.tsx
```diff
@@ -18,15 +18,13 @@ export const OrionTesting = () => {
...
-    },
-    fetchPolicy: 'network-only',
-  })
+  const [getVideosConnection, { data: videosConnectionData, loading: videosConnectionLoading }] =
+    useGetVideosConnectionLazyQuery({
+      variables: {
+        first: 5,
+      },
+      fetchPolicy: 'network-only',
+    })
```
drillprop, line 27: Why there are so many prettier-related changes? 
Also when I hit `ctrl + s` on these files my editor autocorrects me and brings back prev version.

https://user-images.githubusercontent.com/51168865/131966510-f11bf37a-56cf-453f-b515-0c238a1f5cd6.mov

src/shared/components/VideoPlayer/videoJsPlayer.ts
```diff
@@ -1,5 +1,6 @@
 import { RefObject, useEffect, useRef, useState } from 'react'
-import videojs, { VideoJsPlayer, VideoJsPlayerOptions } from 'video.js'
+import { VideoJsPlayer, VideoJsPlayerOptions } from 'video.js'
+import videojs from 'video.js/dist/alt/video.core.novtt'
```
drillprop, line 3: Just out of curiosity - this will reduce bundle size?

.gitignore
kdembler: Those should be in your global gitignore for your system IMO but sure, can bring it back

src/views/playground/Playgrounds/OrionTesting.tsx
kdembler: There are changes because I've upgraded to a newer version and those format the code differently. Your editor probably still uses some cached old version of prettier. Make sure you select the correct one

src/shared/components/VideoPlayer/videoJsPlayer.ts
kdembler: Yes, this stripped around 700kb (uncompressed) from the bundle

src/views/playground/Playgrounds/OrionTesting.tsx
drillprop: Yeah sorry, my mistake. Restarted extension host and everything works fine now!

vite.config.ts
kdembler: Yeah this is painful. I don't think our current plugin actually lets use configure this but will look into it

kdembler: Do you think just disabling this overlay altogether and having the errors in the console would be okay?

drillprop: Hmm, nope, I'm a little worried that it could be very misleading. I'd rather keep this overlay

drillprop: Just checked how it works, and changed my mind 😅  maybe in the end let's keep it that way!

drillprop: Also, one thing which I don't like in this solution, that warnings will show only on my terminal, not in the browser console.
But I think we could disable this overlay for now and not focus on this. Let's maybe create an issue to bring back this in the future, wdyt?

package.json
drillprop: Works great now, it's super fast 🚀

vite.config.ts
kdembler: Yeah, I agree it should be in the console as well. But I also think it works well enough for now (you still have warnings in the build output) and I'd love to tackle this at some later point

drillprop:
Found one more thing. When I go to `studio` I'm getting errors about invalid hook calls in EditVideoSheet, not sure why it's happening.

eldiegod (commented):
Holy shit it loads so quick after making changes, this is huge 🤯

.storybook/preview.jsx
```diff
@@ -31,19 +31,6 @@ export const decorators = [StylesWrapperDecorator]
...
-      {
-        name: 'black',
-        value: '#000000',
-      },
-      {
-        name: 'gray',
-        value: '#272D33',
-      },
-    ],
-  },
```
eldiegod, line 46: why we removing the backgrounds aren't those the same bgs as our app?

kdembler: Storybook was warning me about this option being deprecated. Removing it didn't have any effect, the background for canvas is still black, because of our `GlobalStyle`

drillprop (approved):

eldiegod (approved):

src/shared/icons/index.tsx
```diff
@@ -1,5 +1,4 @@
-// THIS FILE WAS AUTOGENERATED BY SVGR. DO NOT MODIFY IT MANUALLY.
-export * from './ActionPause'
+// THIS FILE WAS AUTOGENERATED BY SVGR. DO NOT MODIFY IT MANUALLYexport * from './ActionPause'
```
drillprop: Something went wrong here 😅

src/shared/illustrations/index.tsx
```diff
@@ -1,5 +1,4 @@
-// THIS FILE WAS AUTOGENERATED BY SVGR. DO NOT MODIFY IT MANUALLY.
-export * from './AccountCreationIllustration'
+// THIS FILE WAS AUTOGENERATED BY SVGR. DO NOT MODIFY IT MANUALLYexport * from './AccountCreationIllustration'
```
drillprop: also here

drillprop (changes requested):

rafalpawlow (approved):

drillprop (approved):

PR #1284, opened 2021-08-30
fix overflowing video title

kdembler:
Screenshot 2021-08-30 at 11 40 10

### DIFF
--- a/src/shared/components/Grid/Grid.tsx
+++ b/src/shared/components/Grid/Grid.tsx
@@ -57,7 +57,7 @@ const maxColumnsCss = ({ maxColumns }: ContainerProps) =>
   maxColumns
     ? css`
         ${media.xlarge} {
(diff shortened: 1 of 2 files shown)


drillprop (approved):

kdembler:
hmm it still breaks when the tile is the last one in the row
Screenshot 2021-08-30 at 12 43 19

drillprop:
Good catch, didn't notice that at all. Also, it seems that it's not happening on certain viewports

https://user-images.githubusercontent.com/51168865/131329477-c8b8be6a-da69-4d67-9003-609e7b43a977.mov

kdembler:
Yep, it was because of the `maxColumnCss` style, should be fixed now

drillprop (approved):
Works fine now 👍🏻

Issue #1256, opened 2021-08-26
Vidoe listing on channel broken


bedeho:
This seems totally broken, the set of visible videos, including pagination, changes as you resize the viewport. As always, I am on FireFox.

kdembler:
Can you explain a bit more, possibly with some example? I'm not really sure what the problem is

kdembler:
Nevermind, managed to reproduce. Seems the problem happens when you enter the channel view linked from some other view. Sometimes only some of the videos will show up and if you refersh, more will appear

bedeho:
Yes sorry, I should have perhaps described that part also. I guess I just assumed it was a universal problem, not depending on what path you took, really weird.

PR #1285, opened 2021-08-30
fix broken videos on channel view

kdembler:
Resolves https://github.com/Joystream/atlas/issues/1256

### DIFF
--- a/src/api/queries/videos.graphql
+++ b/src/api/queries/videos.graphql
@@ -94,23 +94,6 @@ query GetVideos($offset: Int, $limit: Int, $where: VideoWhereInput, $orderBy: Vi
   }
 }
 
-query GetChannelVideosPreview($limit: Int = 10, $channelId: ID) {
-  videos(
-    offset: 0
-    limit: $limit
-    where: {
-      channelId_eq: $channelId
-      isPublic_eq: true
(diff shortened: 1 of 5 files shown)


src/api/queries/videos.graphql
```diff
@@ -94,6 +94,23 @@ query GetVideos($offset: Int, $limit: Int, $where: VideoWhereInput, $orderBy: Vi
...
+      isCensored_eq: false
+      thumbnailPhotoAvailability_eq: ACCEPTED
+      mediaAvailability_eq: ACCEPTED
+    }
+    orderBy: [createdAt_DESC]
+  ) {
+    ...VideoFields
+  }
+}
+
```
drillprop: I'm not sure if we actually need this. We could just use regular `GetVideos` here, we are creating a separate hook anyway.

drillprop (commented):
Works well, just one thing to consider.

src/api/queries/videos.graphql
kdembler: Agreed. I thought separate query name would separate the entries in cache but it actually doesn't care so I guess it doesn't make sense to introduce a separate one. Removed now

drillprop (approved):

PR #1286, opened 2021-08-30
fix skeleton loaders in channel view

kdembler:
Before this change, skeleton loaders for videos in channel view kept restarting - their animation wasn't continuous but jumped back to start a couple of times:

https://user-images.githubusercontent.com/12646744/131324041-8978512a-c771-4519-b56d-7bc2eafe2875.mov

We ensure the placeholders are kept in place by using `key={idx}` on them so React preserves them between renders. However, because `TabContent` component was used, it was changing between each render so content inside it couldn't be preserved.


eldiegod (approved):
Looks good, shame I really liked the swtich statement pattern

PR #1292, opened 2021-08-31
Avoid assets upload error when creating first channel

rafalpawlow:


src/providers/uploadsManager/uploadsManager.tsx
```diff
@@ -34,7 +34,7 @@ export const UploadsManager: React.FC = () => {
 
   useEffect(() => {
     // do this only on first render or when active channel changes
-    if (!activeChannelId || cachedActiveChannelId === activeChannelId || isSyncing) {
+    if (!activeChannelId || cachedActiveChannelId === activeChannelId || cachedActiveChannelId === null || isSyncing) {
```
kdembler, line 37: After giving it a second thought, I'm not so sure this is a good solution. This condition will actually kick in also when you log out and log back in to your channel. We should show the notification in that case :(

rafalpawlow:
I'm closing this PR because #1239 may provide some solution to fix this problem.

PR #1303, opened 2021-08-31
Discovery page layout

eldiegod:
To merge after the cards:   #1255

### DIFF
--- a/src/views/viewer/DiscoverView/DiscoverView.style.tsx
+++ b/src/views/viewer/DiscoverView/DiscoverView.style.tsx
@@ -1,0 +1,42 @@
+import styled from '@emotion/styled'
+
+import { LimitedWidthContainer } from '@/components/LimitedWidthContainer'
+import { colors, media, sizes } from '@/shared/theme'
+
+export const StyledLimitedWidthContainer = styled(LimitedWidthContainer)`
+  margin: ${sizes(16)} auto;
+`
+
+export const FeaturedCategoriesContainer = styled.div`
+  display: grid;
+  gap: ${sizes(4)};
+  margin: ${sizes(16)} 0;
+
+  ${media.sm} {
+    grid-template-columns: 1fr 1fr;
+  }
+
+  ${media.xl} {
+    grid-template-columns: repeat(3, 1fr);
+  }
+`
+
+export const CategoriesContainer = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  gap: ${sizes(6)};
+  margin: ${sizes(12)} 0 ${sizes(16)} 0;
+
+  ${media.lg} {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  ${media.xl} {
+    grid-template-columns: repeat(4, 1fr);
+  }
+`
+
+export const BorderTextContainer = styled.div`
+  padding-bottom: ${sizes(5)};
+  border-bottom: 1px solid ${colors.gray[700]};
+`
--- a/src/views/viewer/DiscoverView/DiscoverView.tsx
+++ b/src/views/viewer/DiscoverView/DiscoverView.tsx
@@ -1,0 +1,58 @@
+import React from 'react'
+
+import { useVideosConnection } from '@/api/hooks'
+import { useMediaMatch } from '@/hooks/useMediaMatch'
+import { GridItem } from '@/shared/components/LayoutGrid'
+import { Text } from '@/shared/components/Text'
+import { FeaturedVideoCategoryCard, VideoCategoryCard } from '@/shared/components/VideoCategoryCard'
+
+import {
+  BorderTextContainer,
+  CategoriesContainer,
+  FeaturedCategoriesContainer,
+  StyledLimitedWidthContainer,
+} from './DiscoverView.style'
+import { featuredVideoCategories, videoCategories } from './data'
+
+export const DiscoverView: React.FC = () => {
+  const { videosConnection } = useVideosConnection({
+    first: 0,
+  })
+  const isMdBreakpoint = useMediaMatch('md')
+  return (
+    <StyledLimitedWidthContainer big>
+      <Text variant="h2">Discover</Text>
+      <FeaturedCategoriesContainer>
+        {featuredVideoCategories.map((category, i) => (
+          <GridItem key={i} colSpan={{ sm: i === 0 ? 2 : 1, xl: 1 }}>
+            <FeaturedVideoCategoryCard
+              variant={isMdBreakpoint ? 'default' : 'compact'}
+              title={category.title}
+              videoTitle={category.videoTitle}
+              videoUrl={category.videoUrl}
+              color={category.color}
+              icon={category.icon}
+            />
+          </GridItem>
+        ))}
+      </FeaturedCategoriesContainer>
+      <BorderTextContainer>
+        <Text variant="h4">All categories</Text>
+      </BorderTextContainer>
+      <CategoriesContainer>
+        {Object.values(videoCategories).map((category, i) => (
+          <VideoCategoryCard
+            key={i}
+            title={category.title}
+            coverImg={category.coverImg}
+            categoryId={category.id}
+            color={category.color}
+            icon={category.icon}
+            videosTotalCount={videosConnection?.totalCount}
+            variant={isMdBreakpoint ? 'default' : 'compact'}
+          />
+        ))}
+      </CategoriesContainer>
+    </StyledLimitedWidthContainer>
+  )
+}
--- a/src/views/viewer/DiscoverView/data.tsx
+++ b/src/views/viewer/DiscoverView/data.tsx
@@ -1,12 +1,5 @@
-import styled from '@emotion/styled'
 import React from 'react'
 
-import { useVideosConnection } from '@/api/hooks'
-import { LimitedWidthContainer } from '@/components/LimitedWidthContainer'
-import { useMediaMatch } from '@/hooks/useMediaMatch'
-import { GridItem } from '@/shared/components/LayoutGrid'
-import { Text } from '@/shared/components/Text'
-import { FeaturedVideoCategoryCard, VideoCategoryCard } from '@/shared/components/VideoCategoryCard'
 import {
   SvgVideoCategoriesAutosAndVehicles,
   SvgVideoCategoriesComedy,
@@ -24,10 +17,9 @@ import {
   SvgVideoCategoriesSports,
   SvgVideoCategoriesTravelAndEvents,
 } from '@/shared/icons'
-import { colors, media, sizes } from '@/shared/theme'
 
 type VideoCategoryData = { title: string; icon: React.ReactNode; color: string; coverImg: string }
-const videoCategories = {
+export const videoCategories = {
   'science-and-Technology': {
     id: '14',
     title: 'Science & Technology',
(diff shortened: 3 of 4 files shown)


toiletgranny:
@eldiegod, this looks perfect from a design standpoint. Sizes, margins, gaps, and RWD behavior matches designs 100% from what I can see.👌 

There's one thing that I might have missed yesterday when reviewing the [general category card component](https://github.com/Joystream/atlas/issues/1230), which is the fact that the cover image should be of a `.2` opacity in its default state. By the way, let me know if this is a good place for me to report that.

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  ${media.xl} {
+    grid-template-columns: repeat(3, 1fr);
+  }
+`
+
+const CategoriesContainer = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  gap: ${sizes(4)};
```
drillprop: I think it should be 24px i.e. `${sizes(6)};`

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,235 @@
...
+  )
+}
+
+const StyledLimitedWidthContainer = styled(LimitedWidthContainer)`
+  margin: ${sizes(16)} auto;
+`
+
+const FeaturedCategoriesContainer = styled.div`
+  display: grid;
+  gap: ${sizes(4)};
```
drillprop:
```suggestion
  gap: ${sizes(6)} ${sizes(7)};
```
design: 
image

image

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,235 @@
...
+    videoTitle: 'Anderson .Paak and The Free Nationals',
+  },
+]
+
+export const DiscoverView: React.FC = () => {
+  const { videosConnection } = useVideosConnection(
+    {
+      first: 0,
+    }
+    // { skip: !anyFollowedChannels, onError: (error) => SentryLogger.error('Failed to fetch videos', 'HomeView', error) }
```
drillprop: do we need that?

drillprop (changes requested):
Also, I'm getting: 
image
I'm guessing it's coming from previous changes.

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,235 @@
...
+import { SvgVideoCategoriesGaming } from '@/shared/icons/VideoCategoriesGaming'
+import { SvgVideoCategoriesHowtoAndStyle } from '@/shared/icons/VideoCategoriesHowtoAndStyle'
+import { SvgVideoCategoriesMusic } from '@/shared/icons/VideoCategoriesMusic'
+import { SvgVideoCategoriesNewsAndPolitics } from '@/shared/icons/VideoCategoriesNewsAndPolitics'
+import { SvgVideoCategoriesNonprofitsAndActivism } from '@/shared/icons/VideoCategoriesNonprofitsAndActivism'
+import { SvgVideoCategoriesPeopleAndBlogs } from '@/shared/icons/VideoCategoriesPeopleAndBlogs'
+import { SvgVideoCategoriesPetsAndAnimals } from '@/shared/icons/VideoCategoriesPetsAndAnimals'
+import { SvgVideoCategoriesScienceAndTechnology } from '@/shared/icons/VideoCategoriesScienceAndTechnology'
+import { SvgVideoCategoriesSports } from '@/shared/icons/VideoCategoriesSports'
+import { SvgVideoCategoriesTravelAndEvents } from '@/shared/icons/VideoCategoriesTravelAndEvents'
```
drillprop: Also, why do you import icons like that? I don't see all these icons in `shared/icons/index`, why you didn't add that? Just run: `yarn codegen:svgs && yarn prettier . --write` and you should be able to do: 
```tsx
import {
  SvgVideoCategoriesAutosAndVehicles,
  SvgVideoCategoriesComedy
   ...
} from '@/shared/icons'
```

src/views/viewer/DiscoverView.tsx
eldiegod: nop

src/views/viewer/DiscoverView.tsx
eldiegod: it's just `gap: ${sizes(6)}` if you look closely you had the horizontal measurement over the cards border that is 4px wide

drillprop: My mistake, you're right 👍🏻

drillprop (approved):
This is just beautiful! Looks good to me!

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,234 @@
...
+    ...videoCategories['education'],
+    videoUrl: 'https://eu-central-1.linodeobjects.com/atlas-assets/featured-videos/3941.mp4',
+    videoTitle: 'ФИЛОСОФИЯ СТОИЦИЗМА',
+  },
+  {
+    ...videoCategories['film-and-animation'],
+    videoUrl: 'https://eu-central-1.linodeobjects.com/atlas-assets/featured-videos/5984.mp4',
+    videoTitle: 'taxi driver edit',
+  },
+]
```
kdembler: Let's move this out of this file. Not sure where, so maybe just an adjacent `config.ts` (although you will need `DiscoverView` directory)

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,234 @@
...
+            videosTotalCount={videosConnection?.totalCount}
+            variant={isMdBreakpoint ? 'default' : 'compact'}
+          />
+        ))}
+      </CategoriesContainer>
+    </StyledLimitedWidthContainer>
+  )
+}
+
+const StyledLimitedWidthContainer = styled(LimitedWidthContainer)`
```
kdembler: Let's move all styles to a separate file

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+export const DiscoverView: React.FC = () => {
+  const { videosConnection } = useVideosConnection({
+    first: 0,
+  })
+  const isMdBreakpoint = useMediaMatch('md')
+  return (
+    <StyledLimitedWidthContainer big>
+      <Text variant="h2">Discover</Text>
+      <FeaturedCategoriesContainer>
```
kdembler: Why aren't we using our `LayoutGrid` component for this and `CategoriesContainer`?

kdembler (changes requested):

src/views/viewer/DiscoverView.tsx
eldiegod: now that I think of it it could very well be used, it just still feels very awkward to have it set to 12 fixed columns and then doing the conversion to what the actual columns that I want.

kdembler: I'm not sure why this is awkward as you can easily see the number of required columns in the design. Let's try to follow the design as closely as possible and use columns grid

src/shared/components/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
```diff
@@ -18,65 +17,70 @@ import { SkeletonLoader } from '../SkeletonLoader'
 import { Text } from '../Text'
 import { VideoPlayer } from '../VideoPlayer'
 
+export type Variant = 'default' | 'compact'
```
kdembler: If this is exported, it should have a more descriptive name, like `FeaturedVideoCategoryCardVariant`

src/shared/components/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
```diff
@@ -18,65 +17,70 @@ import { SkeletonLoader } from '../SkeletonLoader'
...
-  loading,
+  title,
+  icon,
+  videoUrl,
+  videoTitle,
   color,
+  variant = 'default',
 }) => {
   const [hoverRef, isVideoHovering] = useHover<HTMLDivElement>()
+  const isLoading = false
```
kdembler: Why not make this come from props?

src/shared/components/VideoCategoryCard/VideoCategoryCard.tsx
```diff
@@ -21,67 +21,83 @@ import { SkeletonLoader } from '../SkeletonLoader'
...
+  videosTotalCount,
+  coverImg,
+  color,
+}) => {
+  const { videosConnection, loading: loadingVidConnection } = useVideosConnection({
+    first: 0,
+    where: {
+      categoryId_eq: categoryId,
+    },
+  })
```
kdembler: This component gets the total count via props but gets its category count by itself. I don't think it's worth making this component smart. Maybe let's make this come via props as well?

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,234 @@
...
+  SvgVideoCategoriesPeopleAndBlogs,
+  SvgVideoCategoriesPetsAndAnimals,
+  SvgVideoCategoriesScienceAndTechnology,
+  SvgVideoCategoriesSports,
+  SvgVideoCategoriesTravelAndEvents,
+} from '@/shared/icons'
+import { colors, media, sizes } from '@/shared/theme'
+
+type VideoCategoryData = { title: string; icon: React.ReactNode; color: string; coverImg: string }
+const videoCategories = {
```
kdembler: This was supposed to be moved to a separate file, really makes this file harder to read

src/views/viewer/DiscoverView.tsx
kdembler: All the styles were supposed to be moved to a separate file.

src/views/viewer/DiscoverView.tsx
kdembler: This discussion was resolved but the code is still old?

src/views/viewer/DiscoverView.tsx
```diff
@@ -0,0 +1,234 @@
...
+  },
+  {
+    ...videoCategories['film-and-animation'],
+    videoUrl: 'https://eu-central-1.linodeobjects.com/atlas-assets/featured-videos/5984.mp4',
+    videoTitle: 'taxi driver edit',
+  },
+]
+
+export const DiscoverView: React.FC = () => {
+  const { videosConnection } = useVideosConnection({
```
kdembler: What about a loading state and error handling for this query?

kdembler: We should run a separate query that will fetch only the page info, without all the other video details.

src/shared/components/VideoCategoryCard/VideoCategoryCard.tsx
kdembler: Also, like with the total count, this needs to use a query that fetches only the necessary data.

kdembler (changes requested):
Left some comments for stuff that need improvement.

Once you make the changes, I suggest to open this view with prod environment, to check the response times. Currently you can see this takes forever to load, it shouldn't

src/shared/components/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
eldiegod: in what case is would this be needed?

src/shared/components/VideoCategoryCard/VideoCategoryCard.tsx
eldiegod: I think quite the opposite, moving all those queries to the view component would add unneeded pollution + maintenance; where this way it's clean and doesn't require maintenance for each id as it comes from props

PR #1307, opened 2021-09-01
Calculate dynamically number of visible pages to always fit on screen

### LINKED ISSUE #1300: Improve Pagination component scaling across screens
Our Pagination component currently either displays `maxPaginationLinks` or 1 item on smallest screens. We should improve it so that it adapts to available space on page and is responsive. The number of items should still be capped by `maxPaginationLinks` but we should display the max number of items that will fit. For example on smaller screen, we should display 2-3 items if they fit and on bigger screens it should be more. This will likely require some measuring

rafalpawlow:
Closes #1300

### DIFF
--- a/src/shared/components/Pagination/Pagination.tsx
+++ b/src/shared/components/Pagination/Pagination.tsx
@@ -13,6 +13,9 @@ export type PaginationProps = {
   page: number
 }
 
+const PAGINATION_BUTTON_WIDTH = 48
+const VIEWPORT_PADDING = 32
+
 // Codewise component works with index starting from 0 but it's rendered with index starting from 1
 export const Pagination: React.FC<PaginationProps> = ({
   itemsPerPage = 0,
@@ -25,15 +28,11 @@ export const Pagination: React.FC<PaginationProps> = ({
   const [paginationLength, setPaginationLength] = useState(defaultPaginationLength)
 
   const calculatePaginationLength = () => {
-    const paginationButtonWidth = 48
-    const arrowsWithMargins = paginationButtonWidth + 32 * 4
-    const viewPortHorizontalPadding = 32
+    const arrowsWithMargins = PAGINATION_BUTTON_WIDTH + 32 * 4
     const viewPortDifference =
-      window.innerWidth -
-      viewPortHorizontalPadding -
-      (paginationButtonWidth * defaultPaginationLength + arrowsWithMargins)
+      window.innerWidth - VIEWPORT_PADDING - (PAGINATION_BUTTON_WIDTH * defaultPaginationLength + arrowsWithMargins)
     if (viewPortDifference < 0) {
-      setPaginationLength(defaultPaginationLength - Math.ceil(Math.abs(viewPortDifference) / paginationButtonWidth))
+      setPaginationLength(defaultPaginationLength - Math.ceil(Math.abs(viewPortDifference) / PAGINATION_BUTTON_WIDTH))
     } else {
       setPaginationLength(defaultPaginationLength)
     }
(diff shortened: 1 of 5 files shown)


src/shared/components/Pagination/Pagination.tsx
```diff
@@ -7,25 +8,51 @@ import { ChevronButton, PaginationButton, PaginationWrapper, ThreeDotsWrapper }
...
   totalCount = 0,
-  maxPaginationLinks = 5,
+  defaultPaginationLength = 5,
   page = 0,
   onChangePage,
 }) => {
+  const [paginationLength, setPaginationLength] = useState(defaultPaginationLength)
+
+  const calculatePaginationLength = () => {
+    const arrowsWithMargins = PAGINATION_BUTTON_WIDTH + 32 * 4
```
drillprop:
```suggestion
    const arrowsWithMargins = PAGINATION_BUTTON_WIDTH + VIEWPORT_PADDING * 4
```

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -7,25 +8,51 @@ import { ChevronButton, PaginationButton, PaginationWrapper, ThreeDotsWrapper }
...
   itemsPerPage?: number
   totalCount?: number
-  maxPaginationLinks?: number
+  defaultPaginationLength?: number
   onChangePage: (page: number) => void
   page: number
 }
 
+const PAGINATION_BUTTON_WIDTH = 48
+const VIEWPORT_PADDING = 32
```
drillprop: If those apply to button styles, we should also use these values in `Pagination.styles`, wdyt?

drillprop (commented):
Btw - could you also apply ` user-select: none;` on the pagination button and three dots? 
image

Besides that, works fine!

drillprop (approved):

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -16,16 +18,39 @@ export type PaginationProps = {
...
+  const calculatePaginationLength = () => {
+    const arrowsWithMargins = PAGINATION_BUTTON_WIDTH + ARROWS_MARGINS * 4
+    const viewPortDifference =
+      window.innerWidth - VIEWPORT_PADDING - (PAGINATION_BUTTON_WIDTH * defaultPaginationLength + arrowsWithMargins)
+    if (viewPortDifference < 0) {
+      setPaginationLength(defaultPaginationLength - Math.ceil(Math.abs(viewPortDifference) / PAGINATION_BUTTON_WIDTH))
+    } else {
+      setPaginationLength(defaultPaginationLength)
+    }
+  }
```
kdembler: I'm not a fan of the amount of variables we need to take into account here. I think we could make `Pagination` take 100% width and run resize observer on it. Then we don't need to bother with viewport size/padding

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -16,16 +18,39 @@ export type PaginationProps = {
...
+  const debouncedPaginationLength = debounce(calculatePaginationLength, 100)
+
+  useEffect(() => {
+    debouncedPaginationLength()
+    window.addEventListener('resize', debouncedPaginationLength)
+
+    return () => {
+      window.removeEventListener('resize', debouncedPaginationLength)
+    }
+  }, [debouncedPaginationLength])
```
kdembler: `debouncedPaginationLength` will be different on every render, so that `useEffect` will run every time component renders. However my other comment may help with that

src/shared/components/Pagination/Pagination.tsx
```diff
@@ -16,16 +18,39 @@ export type PaginationProps = {
 export const Pagination: React.FC<PaginationProps> = ({
   itemsPerPage = 0,
   totalCount = 0,
-  maxPaginationLinks = 5,
+  defaultPaginationLength = 5,
```
kdembler: This still works as a maximum right? If it does, I think we should keep the old name or `maxPaginationLength` if you prefer

kdembler (approved):

PR #1308, opened 2021-09-01
Fix disappearing tab when exiting the channel

### LINKED ISSUE #1302: Channel view content when navigating out of it
https://user-images.githubusercontent.com/12646744/131502397-c86c269c-de04-4d77-b39a-a02d23862d1f.mov

Take a look at the video, when I navigate to `Home`, the content of ChannelView changes to the content of Information tab for a split second before the view disappears. I guess it's because of the tabs implementation. We should improve this to avoid unnecessary changes on the screen

rafalpawlow:
Closes #1302

### DIFF
--- a/src/views/viewer/ChannelView/ChannelView.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.tsx
@@ -266,11 +266,14 @@ export const ChannelView: React.FC = () => {
     )
 
   // At mount set the tab from the search params
+  const initialRender = useRef(true)
   useEffect(() => {
-    const tabIndex = TABS.findIndex((t) => t === currentTabName)
-    if (tabIndex === -1) setSearchParams({ 'tab': 'Videos' }, { replace: true })
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+    if (initialRender.current) {
+      const tabIndex = TABS.findIndex((t) => t === currentTabName)
+      if (tabIndex === -1) setSearchParams({ 'tab': 'Videos' }, { replace: true })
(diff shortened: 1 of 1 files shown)


drillprop (approved):
Looks good to me 👍🏻

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -271,6 +272,12 @@ export const ChannelView: React.FC = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
```
drillprop: Just as a sidenote - this is not related to the issue, but I think we could remove this comment and add instead: 
```tsx
  const initialRender = useRef(true)
  // At mount set the tab from the search params
  useEffect(() => {
    if (initialRender.current) {
      const tabIndex = TABS.findIndex((t) => t === currentTabName)
      if (tabIndex === -1) setSearchParams({ 'tab': 'Videos' }, { replace: true })
      initialRender.current = false
    }
  })
```

rafalpawlow: This way `useEffect` will be fired on every render

drillprop: I don't think so, keep in mind that I used here `useRef`. It should run only on mount. This is very similar to https://stackoverflow.com/a/53180013

rafalpawlow: Yeah, you are right. I didn't look closely, I only saw that there was no array 😉

kdembler (approved):

PR #1311, opened 2021-09-01
Fix channel view layout jumping on search

### LINKED ISSUE #1298: Channel view layout jump on mobile
When you search for something on channel view the `sort by` section disappears. On smallest screens this causes the layout to jump a bit. We should fix that

https://user-images.githubusercontent.com/12646744/131486935-904b8194-6e06-4da8-b4a7-cd476343b434.mov

rafalpawlow:
Closes #1298

### DIFF
--- a/src/views/viewer/ChannelView/ChannelView.style.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.style.tsx
@@ -141,12 +141,13 @@ export const TabsContainer = styled.div`
   grid-template-areas:
     'tabs tabs tabs'
     'search sort sort';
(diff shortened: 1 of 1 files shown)


drillprop (changes requested):
Jumps don't occur anymore, but select and input are badly aligned on screens with width  < ~400px
image

Can we fix this as well as part of this PR?

drillprop:
still see one issue: 
image

But I think we could just safely change `line 60` in `ChannelView.style` from `xs` to `sm`.

kdembler (approved):

PR #1313, opened 2021-09-01
Add new video tile

### LINKED ISSUE #1269: Add new video tile on MyVideosView
image

Requirements: 
- it should be always first in the videos grid, but we shouldn't see it on the next pages(pagination)
- it should open create video form

drillprop:
Fix #1269
Design: https://www.figma.com/file/mMGcjK0usOHzKXAwcBdwiv/10-Upload-flow-improvements?node-id=3%3A7225

### DIFF
--- a/src/views/studio/MyVideosView/MyVideos.styles.ts
+++ b/src/views/studio/MyVideosView/MyVideos.styles.ts
@@ -1,7 +1,10 @@
 import styled from '@emotion/styled'
+import { Link } from 'react-router-dom'
 
 import { DismissibleBanner } from '@/shared/components/DismissibleBanner'
-import { colors, media, sizes } from '@/shared/theme'
+import { Text } from '@/shared/components/Text'
+import { SvgGlyphAddVideo } from '@/shared/icons'
+import { colors, media, sizes, transitions } from '@/shared/theme'
 
 export const ViewContainer = styled.div`
   padding-top: ${sizes(8)};
@@ -38,4 +41,54 @@ export const SortContainer = styled.div`
   ${media.md} {
     display: grid;
   }
+`
+
+export const NewVideoTile = styled(Link)`
+  position: relative;
+  background: none;
+  border: 1px dashed ${colors.gray[500]};
+  color: ${colors.gray[500]};
+  transition: border ${transitions.timings.player} ease-out;
+  width: 100%;
+  height: 0;
+  padding-bottom: 56.25%;
+  cursor: pointer;
+
+  :hover {
+    border: 1px dashed ${colors.gray[200]};
+    color: ${colors.gray[200]};
+    ${() => StyledIcon} {
+      > path {
+        fill: ${colors.gray[200]};
+      }
+    }
+    ${() => StyledText} {
+      color: ${colors.gray[200]};
+    }
+  }
+`
+
+export const TextAndIconWrapper = styled.div`
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+`
+
+export const StyledIcon = styled(SvgGlyphAddVideo)`
+  > path {
+    transition: fill ${transitions.timings.player} ease-out;
+    fill: ${colors.gray[500]};
+  }
+`
+
+export const StyledText = styled(Text)`
+  text-align: center;
+  margin-top: ${sizes(2)};
+  color: ${colors.gray[500]};
+  transition: color ${transitions.timings.player} ease-out;
 `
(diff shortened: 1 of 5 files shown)


rafalpawlow:
It's a 2px too long on desktop. `padding-bottom: 55.47%;` works a little bit better for desktop
Zrzut ekranu 2021-09-2 o 16 48 48

drillprop:
> It's a 2px too long on desktop. `padding-bottom: 55.47%;` works a little bit better for desktop
> Zrzut ekranu 2021-09-2 o 16 48 48

Good catch! Should be fixed now

rafalpawlow (approved):

src/views/studio/MyVideosView/MyVideos.styles.ts
```diff
@@ -39,3 +42,55 @@ export const SortContainer = styled.div`
...
+
+  :hover {
+    border: 1px dashed ${colors.gray[200]};
+    color: ${colors.gray[200]};
+    ${() => StyledIcon} {
+      > path {
+        fill: ${colors.gray[200]};
+      }
+    }
+    ${() => StyledText} {
```
kdembler: If we can, let's avoid this type of selectors. In this case I think you can just move `StyledIcon` and `StyledText` above this declaration, right?

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -297,6 +301,14 @@ export const MyVideosView = () => {
...
             )}
             <Grid maxColumns={null} onResize={handleOnResizeGrid}>
+              {currentPage === 0 && (
+                <NewVideoTile to={absoluteRoutes.studio.editVideo()}>
+                  <TextAndIconWrapper>
+                    <StyledIcon />
+                    <StyledText variant="body2">Upload new video</StyledText>
+                  </TextAndIconWrapper>
+                </NewVideoTile>
+              )}
```
kdembler: This should likely handle loading state in some way. Currently this is clickable once other items are in loading state. For uniformity I would display some kind of placeholder for it as well

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -297,6 +301,14 @@ export const MyVideosView = () => {
               />
             )}
             <Grid maxColumns={null} onResize={handleOnResizeGrid}>
+              {currentPage === 0 && (
+                <NewVideoTile to={absoluteRoutes.studio.editVideo()}>
```
kdembler: This kind of breaks the pagination on this page - now the first page will have items for 4 rows + 1 item in 5th row

kdembler (changes requested):

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -337,7 +357,8 @@ export const MyVideosView = () => {
                 onChangePage={handleChangePage}
                 page={currentPage}
                 itemsPerPage={videosPerPage}
-                totalCount={isDraftTab ? drafts.length : totalCount}
+                // +1 is for new video tile
+                totalCount={isDraftTab ? drafts.length + 1 : (totalCount || 0) + 1}
```
kdembler: But `drafts.length` already includes that item so this will not be correct

kdembler (changes requested):
I'm pretty sure we should still use the current empty states when there are no items to display for a tab. After this PR we show:

Screenshot 2021-09-07 at 13 30 11

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -367,3 +388,32 @@ const getPublicness = (currentTabName: typeof TABS[number]) => {
...
+                <StyledIcon />
+                <StyledText variant="body2">Upload new video</StyledText>
+              </TextAndIconWrapper>
+            </NewVideoTileLink>
+          )}
+        </NewVideoTileWrapper>
+      </CSSTransition>
+    </SwitchTransition>
+  )
+}
```
kdembler: Let's move this to a separate file, along with its styles. It can be here or in one of the `components` folders

drillprop:
> I'm pretty sure we should still use the current empty states when there are no items to display for a tab. After this PR we show:
> 

Updated. I decided to not show `NewVideoTile` when there are no videos/drafts. Not sure if you meant that. I just think it looks odd with `EmptyFallback` together.

kdembler (approved):

KubaMikolajczyk:
Love the implementation! And yes I wouldn't use it on pages like empty for drafts and empty for unlisted - those should container only empty fallback - so thats perfect!
One minor thing:
image
On the screen above I added 48px margin to this new tile - as it looks really cluttered when those are squished together currently theyre 24px apart and its tight

Great job!

PR #1314, opened 2021-09-01
Center dialogs vertically

### LINKED ISSUE #1297: Center dialogs vertically
Currently dialogs are displayed 128px from top on desktop. We should instead center them vertically

rafalpawlow:
Closes #1297


drillprop (commented):
It definitely fixes the issue but I'm wondering - do we really want that @kdembler?. It looks really odd now, especially on larger screens.  Take a look on screenshot:
image

kdembler:
That's the direction the designers envisioned. I also think it makes sense, it's the center of the screen, place where you will likely focus your attention

kdembler (approved):

PR #1315, opened 2021-09-01
Keep select in focused state while its menu is open

### LINKED ISSUE #1296: Keep select in focused state while its menu is open
Original comment from Kuba:

> If we open the dropdown sort by the blue outline styling is an on click effect - so when it opens it immidiatelly dissapears and outline becomes white again - we could leave it blue as long as it is open - use it for focused state like in designs to indicate that this is the current action (also the blue outline stays on the select when we unclick it - could then turn white as it comes back to default state)

Basically we should use the same variant for select when its clicked and when its menu is open

rafalpawlow:
Closes #1296

### DIFF
--- a/src/shared/components/Select/Select.style.ts
+++ b/src/shared/components/Select/Select.style.ts
@@ -29,7 +29,8 @@ export const SelectButton = styled.button<SelectButtonProps>`
   display: flex;
   justify-content: space-between;
   align-items: center;
-  border: ${({ isOpen }) => (isOpen ? `1px solid ${colors.blue[500]} !important` : 0)};
+  border: ${({ isOpen, error, disabled }) =>
+    `1px solid ${isOpen ? colors.blue[500] : error && !disabled ? colors.error : colors.gray[200]}`};
 
   ${({ size }) => {
     switch (size) {
--- a/src/shared/components/Select/Select.style.ts
+++ b/src/shared/components/Select/Select.style.ts
@@ -23,13 +23,13 @@ export const SelectButton = styled.button<SelectButtonProps>`
   width: 100%;
   overflow: hidden;
   text-overflow: ellipsis;
-  border: none;
   background: none;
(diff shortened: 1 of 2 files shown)


drillprop (approved):

src/shared/components/Select/Select.style.ts
```diff
@@ -23,13 +23,13 @@ export const SelectButton = styled.button<SelectButtonProps>`
...
   overflow: hidden;
   text-overflow: ellipsis;
-  border: none;
   background: none;
   color: ${({ filled }) => (filled ? colors.gray[50] : colors.gray[300])};
   text-align: left;
   display: flex;
   justify-content: space-between;
   align-items: center;
+  border: ${({ isOpen }) => (isOpen ? `1px solid ${colors.blue[500]} !important` : 0)};
```
kdembler: Why is `!important` needed here?

rafalpawlow: Because without this it is overwritten by FormGroup styles
Zrzut ekranu 2021-09-3 o 11 03 56

kdembler: Oh, that's a bummer. I think if we tweaked `FormGroup` styles we could resolve this. Could you try removing that `:not(:focus)` selector in `InputBase.styles.ts:46` and just apply the border directly to `input, button, textarea`? I think that should reduce the specificity and make this work without `important`

rafalpawlow: Nah, still the same
Zrzut ekranu 2021-09-3 o 11 28 38

Maybe we could remove button selector from `FormGroup` styles and apply styles only to `Select`? Select is the only possible button in FormGroup and already has bunch of own styles.

kdembler: Okay, let's try that. But only for `border` styles

kdembler (approved):

PR #1316, opened 2021-09-01
Collapse search input on losing focus

### LINKED ISSUE #1295: Search input on channel view should collapse on losing focus
When you click it expands, when you click outside it should shrink back

rafalpawlow:
Closes #1295

### DIFF
--- a/src/views/viewer/ChannelView/ChannelView.style.tsx
+++ b/src/views/viewer/ChannelView/ChannelView.style.tsx
@@ -1,3 +1,4 @@
+import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 import { fluidRange } from 'polished'
 
@@ -5,6 +6,7 @@ import { ChannelLink } from '@/components/ChannelLink'
 import { TOP_NAVBAR_HEIGHT } from '@/components/TopbarBase'
 import { Button } from '@/shared/components/Button'
 import { IconButton } from '@/shared/components/IconButton'
+import { Select } from '@/shared/components/Select'
 import { SkeletonLoader } from '@/shared/components/SkeletonLoader'
 import { Tabs } from '@/shared/components/Tabs'
 import { Text } from '@/shared/components/Text'
@@ -15,6 +17,18 @@ const SM_TITLE_HEIGHT = '44px'
 const TITLE_HEIGHT = '51px'
 const SM_SUBTITLE_HEIGHT = '24px'
 const SUBTITLE_HEIGHT = '27px'
+
+const ActiveUnderline = css`
+  &::after {
+    content: '';
+    display: block;
+    position: absolute;
+    width: 100%;
+    height: 4px;
+    background-color: ${colors.blue[500]};
+    bottom: -${sizes(3)};
+  }
+`
 
 export const TitleSection = styled.div`
   display: grid;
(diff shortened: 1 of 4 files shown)


drillprop (commented):
I'm not sure if it resolves the problem. When you have text inside the input, it's not shrinking back. I think it should.  cc @KubaMikolajczyk

rafalpawlow:
I thought it was silly that you couldn't see the text you searched. But maybe we could disable shrinking only after you actually searched something. Either way let's wait for @KubaMikolajczyk

drillprop:
I agree that it should be visible when you searched videos i.e. you entered some input and hit `Enter` or clicked the search button. But what if you didn't run a search? I want to know what to do when the latter occurs.

KubaMikolajczyk:
Alright so after discussing with Bartosz we agreed on two things:
1. There is still a confusion about collapsing, focusing and different states of this component
2. The "searched" state is not prominent enough so the user might not see clearly that the search was fired 

**So** - I fixed this component and moved it to components file also added a prominent underline in all "searched" states + made a prototype with all different state of the search bar + described uncklicking, deleting and all interactions with additional descriptions

Component: [Link](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=693%3A94233)
Prototype: [Link](https://www.figma.com/proto/Pf31tuYpozYmpq163U2ho8/Web-Components?page-id=693%3A92364&node-id=693%3A98712&viewport=309%2C48%2C0.07&scaling=min-zoom&starting-point-node-id=693%3A98712)
**(Remember to use enter key for submiting and backspace key for deleting in prototype)**
All descriptions to prototype: [Link](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=693%3A98712)

If there will be additional question (or you would like a prototype tour :) ) at this point please DM me + let me review this after implementation.

rafalpawlow:
@drillprop FYI, I asked @KubaMikolajczyk about mobile version and he's working on it will has to wait

KubaMikolajczyk:
OK so I have two pages - hopefully now we have everything covered :) 
- Channel search component: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=849%3A139022
- Channel search flows & RWD: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=844%3A99726

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -16,6 +18,18 @@ const TITLE_HEIGHT = '51px'
 const SM_SUBTITLE_HEIGHT = '24px'
 const SUBTITLE_HEIGHT = '27px'
 
+const ActiveUnderline = css`
```
kdembler: PascalCase should be used only for components. Let's call this `activeUnderlineCss`

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -276,12 +276,20 @@ export const ChannelView: React.FC = () => {
               setIsSearchingInputOpen={setIsSearchingInputOpen}
               setIsSearching={setIsSearching}
               search={search}
+              isSearching={isSearching}
+              setCurrentTab={setCurrentTab}
             />
           )}
-          {currentTab === 'Videos' && !isSearching && (
+          {currentTab === 'Videos' && (
```
kdembler, line 283: Previously the sorting select wasn't displayed at all when in search results because we cannot change their order. The updated design added the disabled select with `Best match` selected but the implementation doesn't reflect that

Screenshot 2021-09-21 at 11 21 41

Screenshot 2021-09-21 at 11 19 50

kdembler (changes requested):

kdembler (approved):

PR #1317, opened 2021-09-02
Update TitleArea component

### LINKED ISSUE #1274: Update TitleArea component
image

drillprop:
Fix #1274 
Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A115740

### DIFF
--- a/src/shared/components/TitleArea/TitleArea.style.tsx
+++ b/src/shared/components/TitleArea/TitleArea.style.tsx
@@ -1,6 +1,8 @@
 import styled from '@emotion/styled'
 
-import { colors, media, sizes, typography } from '@/shared/theme'
+import { colors, sizes, typography } from '@/shared/theme'
+
+import { TitleAreaVariant } from './'
 
 import { Text } from '../Text'
 
@@ -10,6 +12,7 @@ export const Container = styled.div`
 
 type StyledTextAreaProps = {
   touchedAndEmpty?: boolean
+  variant?: TitleAreaVariant
 }
 
 type TitleAreaInfoProps = {
@@ -25,7 +28,6 @@ export const TitleAreaInfo = styled.div<TitleAreaInfoProps>`
 
 export const StyledTextArea = styled.textarea<StyledTextAreaProps>`
   caret-color: ${colors.blue[500]};
-  font-size: ${typography.sizes.h4};
   color: white;
   background-color: ${colors.transparent};
   font-family: ${typography.fonts.headers};
@@ -34,6 +36,14 @@ export const StyledTextArea = styled.textarea<StyledTextAreaProps>`
   width: 100%;
   resize: none;
   height: auto;
+  font-size: ${({ variant }) => {
+    if (variant === 'large') {
+      return typography.sizes.h4
+    }
+    if (variant === 'small') {
+      return typography.sizes.h2
+    }
+  }};
 
   &:hover {
     opacity: 75%;
@@ -52,9 +62,6 @@ export const StyledTextArea = styled.textarea<StyledTextAreaProps>`
     + ${TitleAreaInfo} {
       opacity: 1;
     }
-  }
-  ${media.sm} {
-    font-size: ${typography.sizes.h2};
   }
 `
 
--- a/src/shared/components/TitleArea/TitleArea.style.tsx
+++ b/src/shared/components/TitleArea/TitleArea.style.tsx
@@ -1,5 +1,4 @@
 import styled from '@emotion/styled'
-import { fluidRange } from 'polished'
 
 import { colors, media, sizes, typography } from '@/shared/theme'
 
@@ -7,60 +6,11 @@ import { Text } from '../Text'
 
 export const Container = styled.div`
   position: relative;
-  width: fit-content;
 `
 
 type StyledInputProps = {
   touchedAndEmpty?: boolean
 }
-
-export const StyledInput = styled.input<StyledInputProps>`
-  --input-max-width: 60vw;
-
-  caret-color: ${colors.blue[500]};
-  ${media.sm} {
-    --input-max-width: 400px;
-  }
-
-  ${media.md} {
-    --input-max-width: 600px;
-  }
-
-  line-height: 1;
-  padding: ${sizes(1)} 0 ${sizes(2)} ${sizes(2)};
-  ${fluidRange({ prop: 'fontSize', fromSize: '32px', toSize: '48px' })};
-
-  color: white;
-  background-color: ${colors.transparent};
-  white-space: nowrap;
-  text-overflow: ellipsis;
-  overflow: hidden;
-  font-family: ${typography.fonts.headers};
-  font-weight: ${typography.weights.bold};
-  border: none;
-  min-width: 100px;
-  max-width: var(--input-max-width);
-  height: 52px;
-
-  &:hover {
-    opacity: 75%;
-  }
-
-  ::placeholder {
-    color: ${colors.gray[500]};
-  }
-
-  + ${() => TitleAreaInfo} {
-    ${({ touchedAndEmpty }) => touchedAndEmpty && `opacity: 1`};
-  }
-
-  :focus,
-  :invalid {
-    + ${() => TitleAreaInfo} {
-      opacity: 1;
-    }
-  }
-`
 
 type TitleAreaInfoProps = {
   error?: boolean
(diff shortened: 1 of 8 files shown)


kdembler (changes requested):
1. The text counter doesn't update for me in video workspace:

Screenshot 2021-09-03 at 13 16 30

2. The component [docs](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A115732) specify that this component should actually wrap to two lines if the text doesn't fit. This makes sense for longer titles

Screenshot 2021-09-03 at 13 16 44

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -44,22 +40,51 @@ export const StyledInput = styled.input<StyledInputProps>`
...
   &:hover {
-    filter: brightness(80%);
+    opacity: 75%;
+  }
+
+  ::placeholder {
+    color: ${colors.gray[500]};
+  }
+
+  + ${() => TitleAreaInfo} {
```
kdembler: Let's avoid selector with fn inside

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -32,7 +28,7 @@ export const StyledInput = styled.input<StyledInputProps>`
 
   line-height: 1;
   padding: ${sizes(1)} 0 ${sizes(2)} ${sizes(2)};
-  ${fluidRange({ prop: 'fontSize', fromSize: '32px', toSize: '40px' })};
+  ${fluidRange({ prop: 'fontSize', fromSize: '32px', toSize: '48px' })};
```
kdembler: I don't think this matches the design anymore. The design specifies `large` and `small` variants for this component, let's follow that.

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -44,22 +40,51 @@ export const StyledInput = styled.input<StyledInputProps>`
   border: none;
   min-width: 100px;
   max-width: var(--input-max-width);
-  width: ${({ widthSize }) => (widthSize ? `${widthSize}ch` : '100%')};
-  height: ${sizes(13)};
+  height: 52px;
```
kdembler: Seems to me this should be 48px tall?

src/shared/components/TitleArea/TitleArea.tsx
```diff
@@ -1,39 +1,80 @@
...
+
+    useEffect(() => {
+      const textarea = containerRef.current?.querySelector('textarea')
+      if (!textarea) {
+        return
+      }
+      textarea.style.height = 'initial'
+      const scrollHeight = textarea.scrollHeight
+      textarea.style.height = scrollHeight + 'px'
+    })
```
drillprop: I really don't like it but I need this `useEffect`. The component needs to update its height somehow when you switch video tabs.
To be honest, I ran out of ideas on how to do this better.

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -1,65 +1,77 @@
...
-  }
+type TitleAreaInfoProps = {
+  error?: boolean
+}
 
-  line-height: 1;
+export const TitleAreaInfo = styled.div<TitleAreaInfoProps>`
+  display: grid;
+  opacity: 0;
+  max-width: 600px;
```
kdembler: Why does it need `max-width`?

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -1,65 +1,77 @@
...
   padding: ${sizes(1)} 0 ${sizes(2)} ${sizes(2)};
```
kdembler: Why do we have padding on `TitleAreaInfo`? Design doesn't specify any

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -1,65 +1,77 @@
...
-  ${fluidRange({ prop: 'fontSize', fromSize: '32px', toSize: '40px' })};
+  grid-template-columns: 1fr 1fr;
+  grid-gap: ${sizes(9)};
```
kdembler: Margin between `min/max` section and current counter is 32px in design

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -1,65 +1,77 @@
...
+`
 
+export const StyledInput = styled.textarea<StyledInputProps>`
+  caret-color: ${colors.blue[500]};
+  line-height: 1;
+  padding: ${sizes(1)} 0 ${sizes(2)} ${sizes(2)};
+  font-size: ${sizes(8)};
```
kdembler: 1. Why do we specify font-size using `sizes`? Let's use `typography` instead
2. This doesn't seem to follow the design

src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -1,65 +1,77 @@
...
+  }
+
+  :focus,
+  :invalid {
+    + ${TitleAreaInfo} {
+      opacity: 1;
+    }
+  }
+  ${media.sm} {
+    font-size: ${sizes(12)};
```
kdembler: Let's do what the design does and not specify any queries inside the component. Instead let's a `variant` prop which determines sizing and so on. Also, this seems to big compared to the design

kdembler (changes requested):

src/shared/components/TitleArea/TitleArea.tsx
kdembler: I don't understand why we need this in the first place, could you describe what problem this is solving?

drillprop: Basically, we need these lines to solve this issue: 
> The component docs specify that this component should actually wrap to two lines if the text doesn't fit. This makes sense for longer titles

When you typing in `textarea` and if the text starts going to the next line,`textarea` doesn't change its size. It just adds a scrollbar on the right. So, to solve that, my first attempt was to update the height of this textarea on every change - basically, I just put analogical code inside `handleOnChange`, but it wasn't enough. If you open for example one tab with a video that has a title that expands to two lines and then you change to another video that has a very short title, the TitleArea will not update itself and you'll have two-row input. I hope it makes sense. 

https://user-images.githubusercontent.com/51168865/132349073-02d8d427-2ad9-4587-96dd-89c51cf4528c.mov

kdembler: Ah okay, I forgot text areas don't expand by themselves. Have you seen this: https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/ ? Looks interesting and doesn't need any javascript

drillprop: Woah! That looks cool. I don't know how I missed that! Will try to implement this.

drillprop: > Ah okay, I forgot text areas don't expand by themselves. Have you seen this: https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/ ? Looks interesting and doesn't need any javascript

Unfortunately, this won't fix all the issues. When I'm switching tabs I still don't know what input is inside the textarea, because we're using uncontrolled components here, so basically, I don't know what the `value` is. If I don't know the `value` I cannot set `data-replicated-value` , this method is using javascript  after all - If you look at HTML in the link you'll see: 
```tsx
  <div class="grow-wrap">
    <textarea name="text" id="text" onInput="this.parentNode.dataset.replicatedValue = this.value"></textarea>
  </div>
```

I'm wondering. Can I just pass `value` like this?  

```tsx
// EditVideoForm line 438
            <StyledTitleArea
              {...register(
                'title',
                textFieldValidation({ name: 'Video Title', minLength: 3, maxLength: 40, required: true })
              )}
              value={getValues('title')} // can I just add this?
              min={3}
              max={40}
              placeholder="Video title"
            />
```
I guess it will rerender TitleArea on every input? Maybe with delay, because it looks like `getValues` is debounced under the hood(?)

KubaMikolajczyk:
I did the review for this component and I have some thoughts below:
image
The spacing in implementation is 16px - we could make it a bit more spacey like in designs with bumping it to 32px.
I understand that this is because of the fact that we are saving a space for helper text but if that is possible could we consider showing helper text dynamically - without making a reserved space for it? It preety common pattern in apps and its not considered anything bad to dynamically add helper texts when needed.
image
The counter is spaced not relatively to items on the left side - it doesn't look nice - I updated the component so now it sticks to the right side of the container - [I fixed it here >](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A115739)
image
If I once focus the title area - even if the amount of letters is in desired range - when I click anything else - the title area stays in focused state - the helper state is visible - could we collapse it - if the title meats the requirements of the helper text? Because we really dont need this information anymore when we are focusing on other elements of the form. On the other hand if we do not meet requirements - so in this case 3 chars at least - we should still display the helper text to let user know that this is not going to be accepted (also dont let user click upload button if we dont meet this requirement)

Except of that everything is fine! :)

kdembler:
> The spacing in implementation is 16px - we could make it a bit more spacey like in designs with bumping it to 32px.
I understand that this is because of the fact that we are saving a space for helper text but if that is possible could we consider showing helper text dynamically - without making a reserved space for it? It preety common pattern in apps and its not considered anything bad to dynamically add helper texts when needed.

I'm a bit unsure about this. First thing, we already preserve space for helper text for other inputs, so I think we should be consistent. Second thing, how do you imagine this working exactly? When I click to edit the video title, every other input in the form "jumps down" 20px or so?

> If I once focus the title area - even if the amount of letters is in desired range - when I click anything else - the title area stays in focused state - the helper state is visible - could we collapse it - if the title meats the requirements of the helper text? Because we really dont need this information anymore when we are focusing on other elements of the form. On the other hand if we do not meet requirements - so in this case 3 chars at least - we should still display the helper text to let user know that this is not going to be accepted (also dont let user click upload button if we dont meet this requirement)

Hmm I think it disappears for me! @drillprop any idea what could cause this?

drillprop:
> Hmm I think it disappears for me! @drillprop any idea what could cause this?

No idea, It disappears for me as well.

KubaMikolajczyk:
Alright then maybe it was just in my case :P 

If it comes to helper texts the answer is yes  - the form would expand with new space if helper text is needed. Few examples below:
- https://demo.mobiscroll.com/jquery/forms/validation#
- https://formvalidation.io/

Also gov.uk below
https://www.universal-credit.service.gov.uk/sign-in
image

Ikea follows expanding
https://www.ikea.com/pl/pl/profile/login/
image

Even google expands
https://myaccount.google.com/birthday
image

Also I believe that it wouldn't be much of a problem when there won't be too many fields expanding due to new policy of submitting forms :) 

We could expand it with instant transition for now and in future maybe we could add a smooth transition when we would have tokens :D

KubaMikolajczyk:
Great job! Looks really nice :)
image

rafalpawlow (approved):

PR #1318, opened 2021-09-02
Add ability to click on cover photo while it's loading

### LINKED ISSUE #1287: When the video thumbnail is loading user cannot go to the video by clicking its thumbnail
Take a look at the video. 

https://user-images.githubusercontent.com/51168865/131334406-438e3138-49cc-4a5a-a9ea-c848eaeab09c.mov

I think a user should be able to go to the video when the title is loaded. We should also show the play icon on the loading thumbnail.
The same problem I see on channel avatars. We cannot click on the loading avatar to go to a certain channel even when we see the channel title.

rafalpawlow:
Closes #1287


drillprop (approved):
LGTM!

PR #1319, opened 2021-09-02
Add optional prop to FormField

### LINKED ISSUE #1281: Add optional variant to FormField
image

drillprop:
Fix #1281 
Design: https://www.figma.com/file/mMGcjK0usOHzKXAwcBdwiv/[-10]-Upload-flow-improvements?node-id=3%3A7236

### DIFF
--- a/src/shared/components/FormField/FormField.style.ts
+++ b/src/shared/components/FormField/FormField.style.ts
@@ -10,15 +10,20 @@ export const FormFieldWrapper = styled.div`
   max-width: 760px;
 `
 
+export const FormFieldHeader = styled.header`
+  display: flex;
+  align-items: center;
+  width: 85%;
+  padding-bottom: ${sizes(2)};
+  word-wrap: break-word;
+`
+
 export const FormFieldTitle = styled(Text)`
   margin: 0;
-  line-height: ${sizes(5)};
-  word-wrap: break-word;
-  width: 85%;
+  margin-right: ${sizes(4)};
 `
 
 export const FormFieldDescription = styled(Text)`
-  margin-top: ${sizes(2)};
   width: 85%;
   font-size: ${typography.sizes.subtitle2};
   color: ${colors.gray[300]};
--- a/src/shared/components/FormField/FormField.style.ts
+++ b/src/shared/components/FormField/FormField.style.ts
@@ -12,14 +12,13 @@ export const FormFieldWrapper = styled.div`
 
 export const FormFieldHeader = styled.header`
   display: flex;
-  align-items: center;
+  align-items: baseline;
   width: 85%;
   padding-bottom: ${sizes(2)};
   word-wrap: break-word;
 `
 
 export const FormFieldTitle = styled(Text)`
-  margin: 0;
   margin-right: ${sizes(4)};
 `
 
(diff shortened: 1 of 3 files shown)


src/shared/components/FormField/FormField.tsx
```diff
@@ -1,17 +1,33 @@
...
+  FormFieldHeader,
+  FormFieldTitle,
+  FormFieldWrapper,
+} from './FormField.style'
+
+import { Text } from '../Text'
 
 export type FormFieldProps = {
   title: string
+  optional?: boolean
```
rafalpawlow, line 15: Maybe `isOptional` would be more readable? WDYT?

rafalpawlow (approved):

src/shared/components/FormField/FormField.tsx
drillprop: I don't know,  I kinda like `optional`. This is similar to passing `required` or `disabled` to input.

src/shared/components/FormField/FormField.style.ts
```diff
@@ -10,15 +10,20 @@ export const FormFieldWrapper = styled.div`
...
+export const FormFieldHeader = styled.header`
+  display: flex;
+  align-items: center;
+  width: 85%;
+  padding-bottom: ${sizes(2)};
+  word-wrap: break-word;
+`
+
 export const FormFieldTitle = styled(Text)`
   margin: 0;
```
kdembler: I don't think we need this, `Text` should already apply this

src/shared/components/FormField/FormField.tsx
```diff
@@ -1,17 +1,33 @@
...
     <FormFieldWrapper className={className}>
-      <FormFieldTitle variant="h6">{title}</FormFieldTitle>
+      <FormFieldHeader>
+        <FormFieldTitle variant="h6">{title}</FormFieldTitle>
+        {optional && (
+          <Text variant="body2" secondary>
+            (Optional)
+          </Text>
+        )}
+      </FormFieldHeader>
```
kdembler, line 30: This doesn't produce a great visual effect - the two text nodes aren't properly centered. It's not your fault - just issues with the font in the app and Figma. Will discuss with designers what we should do about this

Screenshot 2021-09-03 at 11 04 30
Screenshot 2021-09-03 at 11 04 42

drillprop: As a workaround, I used `align-items: baseline` instead of `center`. I think it looks slightly better now.

src/shared/components/FormField/FormField.style.ts
```diff
@@ -12,14 +12,13 @@ export const FormFieldWrapper = styled.div`
 
 export const FormFieldHeader = styled.header`
   display: flex;
-  align-items: center;
+  align-items: baseline;
```
kdembler: This will actually cause the header to have different height based on whether or not it is optional. Had a chat with designers about this, conclusion is we'll try to figure out this font issue in general. For now let's just use `center` here and accept non-perfect look. Hopefully we resolve the font issue soon and it just fixes itself

src/shared/components/FormField/FormField.style.ts
```diff
@@ -10,15 +10,19 @@ export const FormFieldWrapper = styled.div`
   max-width: 760px;
 `
 
-export const FormFieldTitle = styled(Text)`
-  margin: 0;
-  line-height: ${sizes(5)};
-  word-wrap: break-word;
+export const FormFieldHeader = styled.header`
+  display: flex;
+  align-items: baseline;
```
drillprop:
```suggestion
  align-items: center;
```

kdembler (approved):

PR #1321, opened 2021-09-02
Refactor channel follow/unfollow logic

### LINKED ISSUE #1293: Refactor channel follow/unfollow logic
Currently we have couple of different places where we handle channel follow/unfollow. I think we could improve that and use a common hook at all times. We already have `useHandleFollowChannel` but it's not used everywhere, for sure not in `ChannelView`. We should also show a confirmation dialog when unfollowing (can be found [here](https://www.figma.com/file/Vk2Z4QOiVa5bB6q3cBIG5J/Joystream-Atlas?node-id=11921%3A424905), one on the right). It should be part of the hook

rafalpawlow:
Closes #1293

### DIFF
--- a/src/hooks/useHandleFollowChannel.tsx
+++ b/src/hooks/useHandleFollowChannel.tsx
@@ -5,7 +5,6 @@ import { useDialog } from '@/providers/dialogs'
 import { usePersonalDataStore } from '@/providers/personalData'
 import { Text } from '@/shared/components/Text'
 import { SentryLogger } from '@/utils/logs'
-import { UnfollowDescriptionContainer } from '@/views/viewer/ChannelView/ChannelView.style'
 
 export const useHandleFollowChannel = (id?: string, name?: string | null) => {
   const [openUnfollowDialog, closeUnfollowDialog] = useDialog()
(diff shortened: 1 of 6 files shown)


src/hooks/useHandleFollowChannel.tsx
```diff
@@ -1,27 +1,58 @@
+import React from 'react'
+
 import { useFollowChannel, useUnfollowChannel } from '@/api/hooks'
+import { useDialog } from '@/providers/dialogs'
 import { usePersonalDataStore } from '@/providers/personalData'
+import { Text } from '@/shared/components/Text'
 import { SentryLogger } from '@/utils/logs'
+import { UnfollowDescriptionContainer } from '@/views/viewer/ChannelView/ChannelView.style'
```
kdembler: This import shouldn't be here, this component should be declared next to the hook instead

src/hooks/useHandleFollowChannel.tsx
```diff
@@ -1,27 +1,58 @@
...
     }
     try {
       if (isFollowing) {
-        updateChannelFollowing(id, false)
-        unfollowChannel(id)
+        openUnfollowDialog({
+          variant: 'error',
+          exitButton: false,
+          error: true,
+          title: 'Would you consider staying?',
```
kdembler:
```suggestion
          title: 'Do you want to unfollow?',
```

src/hooks/useHandleFollowChannel.tsx
```diff
@@ -1,27 +1,58 @@
...
+            },
+          },
+        })
       } else {
         updateChannelFollowing(id, true)
         followChannel(id)
       }
     } catch (error) {
-      SentryLogger.error('Failed to update Channel following', error, { channel: { id } })
+      SentryLogger.error('Failed to update channel following', 'ChannelView', error, { channel: { id } })
```
kdembler:
```suggestion
      SentryLogger.error('Failed to update channel following', 'useHandleFollowChannel', error, { channel: { id } })
```

kdembler (changes requested):

kdembler (approved):

PR #1325, opened 2021-09-03
Disable channel card's anchor when there is no id provided

### LINKED ISSUE #1323: User shouldn't be able to click loading channel tiles.
This applies to ChannelsView(top 10 channels section and discover new channels). If you click loading channel card it will redirect you to the non-existing channel:

image

rafalpawlow:
Closes #1323


kdembler (changes requested):
You solved the issue for `ChannelCard` but this also applies to other places. For example channels in `Discover new channels` section

kdembler (approved):

PR #1328, opened 2021-09-03
Add tooltip to upload button in EditVideoForm

### LINKED ISSUE #1275: Add tooltip to upload button in EditVideoForm
image
Also, we should update button text

rafalpawlow:
Closes #1275

### DIFF
--- a/src/shared/components/ActionBar/ActionBar.style.ts
+++ b/src/shared/components/ActionBar/ActionBar.style.ts
@@ -66,7 +66,6 @@ export const StyledButtonsContainer = styled.div`
   display: flex;
   width: 100%;
   flex-direction: row;
-  align-items: center;
   justify-content: flex-end;
 
   > *:not(:last-child) {
@@ -75,10 +74,12 @@ export const StyledButtonsContainer = styled.div`
 `
 
 export const DetailsContainer = styled.div`
+  user-select: none;
   margin-left: auto;
   display: flex;
   align-items: center;
-  padding: 13px ${sizes(4)};
+  height: 100%;
+  padding: 0 ${sizes(5)};
   transition: background-color ${transitions.timings.sharp} ${transitions.easing};
 
   :hover {
--- a/src/shared/components/ActionBar/ActionBar.style.ts
+++ b/src/shared/components/ActionBar/ActionBar.style.ts
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 
-import { colors, media, sizes, typography } from '@/shared/theme'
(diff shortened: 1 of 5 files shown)


drillprop:
made some improvements based on updated design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=798%3A90836
I think we could ignore for now checking if all required inputs are filled and deal with it in another PR.

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -78,7 +78,12 @@ export const DetailsContainer = styled.div`
   margin-left: auto;
   display: flex;
   align-items: center;
-  padding: 0 ${sizes(5)};
+  padding: 13px ${sizes(4)};
+  transition: background-color ${transitions.timings.sharp} ${transitions.easing};
+
+  :hover {
+    background-color: ${colors.transparentPrimary[18]};
+  }
```
kdembler, line 87: Now that this element has a hover state, I think we should also make the text non-selectable so you don't get a text selection cursor when hovering over text

src/shared/components/ActionBar/ActionBar.style.ts
```diff
@@ -78,7 +78,12 @@ export const DetailsContainer = styled.div`
   margin-left: auto;
   display: flex;
   align-items: center;
-  padding: 0 ${sizes(5)};
+  padding: 13px ${sizes(4)};
```
kdembler: I think we actually don't need any vertical padding as it should fill the entire available height anyway, right? For horizontal padding, design uses 20px

kdembler (changes requested):

src/shared/components/Tooltip/Tooltip.tsx
```diff
@@ -30,7 +30,7 @@ export const Tooltip: React.FC<TooltipProps> = ({
   arrowDisabled,
   className,
 }) => {
-  const [isVisible, setIsVisible] = useState(false)
+  const [isVisible, setIsVisible] = useState(true)
```
kdembler: Where is this change coming from?

drillprop:
```suggestion
  const [isVisible, setIsVisible] = useState(false)
```

drillprop: damn it 😅 . I should more carefully check what I'm committing.

kdembler (approved):

PR #1330, opened 2021-09-06
Add RadioCardButton component

### LINKED ISSUE #1273: Add radio card button component
image

drillprop:
fix #1273 
design: https://www.figma.com/file/mMGcjK0usOHzKXAwcBdwiv/[-10]-Upload-flow-improvements?node-id=3%3A7236

### DIFF
--- a/src/shared/components/RadioButton/RadioButton.style.ts
+++ b/src/shared/components/RadioButton/RadioButton.style.ts
@@ -3,7 +3,6 @@ import styled from '@emotion/styled'
 import { colors, sizes } from '@/shared/theme'
 
 import { HelperText } from '../HelperText/HelperText'
-import { Text } from '../Text'
 
 export type RadioButtonStyleProps = Partial<{
   disabled: boolean
@@ -16,27 +15,12 @@ type RadioButtonCaptionProps = {
 }
 
 export const RadioButtonLabel = styled.label<RadioButtonStyleProps>`
-  --radio-button-size: 18px;
-  --radio-button-container-size: ${sizes(6)};
-
   display: inline-grid;
   grid-template-columns: auto 1fr;
   align-items: center;
   grid-column-gap: ${sizes(3)};
   grid-row-gap: ${sizes(1)};
   cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
-`
-
-export const RadioButtonContainer = styled.div`
-  width: var(--radio-button-container-size);
-  height: var(--radio-button-container-size);
-  display: flex;
-  justify-content: center;
-  align-items: center;
-`
-
-export const StyledLabelText = styled(Text)`
-  color: ${colors.gray[50]};
 `
 
 export const StyledHelperText = styled(HelperText)<RadioButtonCaptionProps>`
--- a/src/shared/components/RadioButton/RadioButton.style.ts
+++ b/src/shared/components/RadioButton/RadioButton.style.ts
@@ -1,7 +1,6 @@
-import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 
-import { colors, sizes, transitions } from '@/shared/theme'
+import { colors, sizes } from '@/shared/theme'
 
 import { HelperText } from '../HelperText/HelperText'
 import { Text } from '../Text'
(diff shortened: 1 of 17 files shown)


kdembler:
@KubaMikolajczyk shouldn't this have some kind of hover state specified? I don't see anything in the design

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -25,6 +25,7 @@ import { Datepicker } from '@/shared/components/Datepicker'
 import { FormField } from '@/shared/components/FormField'
 import { FileErrorType, ImageInputFile, VideoInputFile } from '@/shared/components/MultiFileSelect/MultiFileSelect'
```
kdembler: 😅 

```suggestion
import { FileErrorType, ImageInputFile, VideoInputFile } from '@/shared/components/MultiFileSelect'
```

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -487,25 +480,40 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
+                        onChange={() => onChange(true)}
+                        selectedValue={value?.toString()}
+                        helperText="Visible to all"
+                      />
+                      <RadioCardButton
+                        value="false"
+                        label="Unlisted"
+                        onChange={() => onChange(false)}
+                        selectedValue={value?.toString()}
+                        helperText={'Visible with link only'}
```
kdembler: Let's use regular `"` here

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -487,25 +480,40 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
...
+                  name="isPublic"
+                  control={control}
+                  defaultValue={true}
+                  rules={{
+                    validate: (value) => value !== null,
+                  }}
+                  render={({ field: { value, onChange, ref } }) => (
+                    <RadioCardButtonsContainer>
+                      <RadioCardButton
+                        ref={ref}
```
kdembler: Why do we pass the `ref` here?

src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -16,70 +15,6 @@ type RadioButtonCaptionProps = {
...
-          background-color: ${colors.transparentPrimary[6]};
-        }
-      }
-    `
-  }
-}
-
 export const RadioButtonLabel = styled.label<RadioButtonStyleProps>`
   --radio-button-size: 18px;
   --radio-button-container-size: ${sizes(6)};
```
kdembler: `--radio-button-size` isn't used anymore, let's remove it

`--radio-button-container-size` is used for `RadioButtonContainer` but I don't think we really need it? Tbh we shouldn't need `RadioButtonContainer` at all

src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -1,43 +1,38 @@
...
-  Input,
   RadioButtonContainer,
   RadioButtonLabel,
   RadioButtonStyleProps,
   StyledHelperText,
-  StyledInput,
   StyledLabelText,
 } from './RadioButton.style'
 
+import { RadioInput } from '../RadioInput/RadioInput'
```
kdembler:
```suggestion
import { RadioInput } from '../RadioInput'
```

src/shared/components/RadioCardButton/RadioCardButton.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  return colors.gray[500]
+}
+
+export const RadioCardLabel = styled.label<RadioCardLabelProps>`
+  padding: ${sizes(4)};
+  display: flex;
+  flex-direction: column;
+  cursor: pointer;
+  border: 1px solid ${({ selected, disabled, error }) => getRadioCardBorderColor({ selected, disabled, error })};
```
kdembler: You can just pass the function, it should get called with the props:
```suggestion
  border: 1px solid ${getRadioCardBorderColor};
```

kdembler (changes requested):

drillprop:
> @KubaMikolajczyk shouldn't this have some kind of hover state specified? I don't see anything in the design

Found it: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=750%3A90714
I'll update this

drillprop:
@kdembler In the design, this component is called `option card`, but I'm not sure if we should add this. I think we could just stick to RadioCardButton for now. I don't see the usage of the checkbox with this card yet. Let me know what you think

kdembler:
I think here it makes sense to use the name from the design - our name clearly works only for radio buttons while the design allows for checkboxes inside as well. I wouldn't want us to need to change the component name later when we add checkboxes

KubaMikolajczyk:
The way those option cards work is flawless - only issues I spotted are in their responsiveness:
image
Somehow they stop filling the container after 865px breakpoint 
image
And on mobile those should use a vertical layout

Everything is explained here: https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Page-Video-workspace?node-id=314%3A6148

kdembler (commented):
All looks good but I think we still have a bit of inconsistency with the hover state. Design says that on card hover, the radio button should also have its hover state triggered. However it seems to me this may be complicated. Do you have any idea on how to achieve that?

drillprop:
> All looks good but I think we still have a bit of inconsistency with the hover state. Design says that on card hover, the radio button should also have its hover state triggered. However it seems to me this may be complicated. Do you have any idea on how to achieve that?

Check now.
I think it's relatively clean. I mean I kinda don't like that we need to import `CustomRadioInput` in `OptionCard`, but didn't find a cleaner solution.

kdembler (approved):

PR #1337, opened 2021-09-07
Fix issue with showing big play button when video is not ready to play

### LINKED ISSUE #1336: Cannot start video playback when the asset is being resolved
Like in the title - when asset resolution runs (we try to find storage provider that will provide the asset), the player shows the big, initial play button. However, until resolution finishes, clicking the play button will not have any effect. This is a bad UX, user can click the button 100 times and nothing will happen. We should show the player loading state in this case to let the user now that something is getting processed atm.

drillprop:
Fix #1336

### DIFF
--- a/src/shared/components/VideoPlayer/VideoPlayer.style.tsx
+++ b/src/shared/components/VideoPlayer/VideoPlayer.style.tsx
@@ -78,7 +78,7 @@ export const PlayControl = styled.div<PlayControlProps>`
   left: 0;
 
   /* hide PlayControl when loading on mobile */
-  ${({ isLoading }) => isLoading && `opacity: 0;`};
+  ${({ isLoading }) => isLoading && `opacity: 0;`}
   @media (hover: hover) {
     opacity: 1;
     margin-right: 0.5em;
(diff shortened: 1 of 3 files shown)


kdembler:
Thanks for jumping so quick on this! I'm wondering though, are you sure the player is the best place to fix it? I think this actually worked at some previous point, the described issue started happening after one of the refactors of the asset resolution I think. Did you look into what the `VideoPlayer` is given when the asset gets resolved? Maybe we actually should fix that on the other end?

drillprop:
> Thanks for jumping so quick on this! I'm wondering though, are you sure the player is the best place to fix it? I think this actually worked at some previous point, the described issue started happening after one of the refactors of the asset resolution I think. Did you look into what the `VideoPlayer` is given when the asset gets resolved? Maybe we actually should fix that on the other end?

Didn't think about it, but you're right. I just looked quickly at it, and I think I could make use of the `isLoadingAsset` from the `useMedia` hook. Nevertheless, I think some changes to the player are still needed. 
I'll revisit this PR a little later.

kdembler:
:+1:

kdembler (changes requested):
It seems the VideoHero will navigate you to a new video once it finishes playing, let's fix that

kdembler (approved):

Issue #1294, opened 2021-08-31
Make changing environments easier


kdembler:
I think we could have some settings overlay to more easily change environments during development. Changing should also log out the current user from studio

PR #1338, opened 2021-09-08
replace AdminView with AdminOverlay

kdembler:
You can trigger the overlay by pressing Ctrl+Shift+D. Let me know if you have any suggestions

Resolves https://github.com/Joystream/atlas/issues/1294


src/components/AdminOverlay/AdminOverlay.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+  const handleImportClick = () => {
+    const inputElement = document.createElement('input')
+    inputElement.setAttribute('type', 'file')
+    inputElement.onchange = handleFileChange
+    inputElement.click()
+  }
+
+  return (
+    <VerticalSpacedContainer>
```
drillprop: I know that's only for developers, but It would be nice to add some margin above these buttons. 
image

drillprop (approved):
Looks brilliant and works really well! Added just one suggestion

PR #1339, opened 2021-09-08
Update upload status group

### LINKED ISSUE #1277: Update UploadStatusGroup component
image

drillprop:
Fix #1277

### DIFF
--- a/src/providers/uploadsManager/types.ts
+++ b/src/providers/uploadsManager/types.ts
@@ -33,6 +33,7 @@ export type StartFileUploadOptions = {
 export type UploadStatus = {
   lastStatus?: AssetUploadStatus
   progress?: number
+  retries?: number
 }
 
 export type UploadsStatusRecord = Record<string, UploadStatus | undefined>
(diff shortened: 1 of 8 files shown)


rafalpawlow (approved):

KubaMikolajczyk:
Wow the upload asset group - it looks perfect! ❤️ ⭐ 

Two minor things maybe not even related to this particular task - 
The submit button ha no disabled state when not all required fields are filled and it shows the tooltip - it should be disabled to show the tooltip. And on the other hand when all necessary fields are filled it still shows the tooltip on hover - and now it doesn't have to do that as we already filled all required fields :)
image

Those snackbars appear when we start uploading - and while the information - 2 assets being uploaded is useful - the other one saying that something is uploaded can be still misleading and telling user that the process is over which is not true at all. We should not display the one about assets being uploaded.
image
What we can do - is to show the user a snackbar like this one below when the whole process is finished - with a handy link to open the upload in a new tab 👇 
image
Link to figma 👉 [Link](https://www.figma.com/file/mMGcjK0usOHzKXAwcBdwiv/10-Upload-flow-improvements?node-id=62%3A12271)

PR #1341, opened 2021-09-09
Fix issue with not displaying ActionBar primary button

drillprop:
I missed that in #1328. Little patch, that should fix the issue with not displaying the `Publish changes` button, when editing the channel.

### DIFF
--- a/src/shared/components/ActionBar/ActionBar.tsx
+++ b/src/shared/components/ActionBar/ActionBar.tsx
@@ -54,6 +54,13 @@ export const ActionBar: React.FC<ActionBarProps> = ({
   onCancelClick,
   primaryButtonTooltipText,
 }) => {
+  const renderButton = () =>
+    primaryButtonText ? (
+      <Button disabled={disabled} onClick={onConfirmClick} size="large" type="submit">
+        {primaryButtonText}
+      </Button>
+    ) : null
+
   return (
     <StyledActionBarContainer className={className}>
       <StyledInnerContainer>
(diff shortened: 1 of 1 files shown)


src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -54,6 +54,13 @@ export const ActionBar: React.FC<ActionBarProps> = ({
   onCancelClick,
   primaryButtonTooltipText,
 }) => {
+  const renderButton = () =>
+    primaryButtonText ? (
+      <Button disabled={disabled} onClick={onConfirmClick} size="large" type="submit">
+        {primaryButtonText}
+      </Button>
+    ) : null
```
kdembler: Do we need this to be a function? I think this could just be a variable instead?

kdembler (changes requested):

src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -54,12 +54,11 @@ export const ActionBar: React.FC<ActionBarProps> = ({
   onCancelClick,
   primaryButtonTooltipText,
 }) => {
-  const renderButton = () =>
-    primaryButtonText ? (
-      <Button disabled={disabled} onClick={onConfirmClick} size="large" type="submit">
-        {primaryButtonText}
-      </Button>
-    ) : null
+  const renderButton = primaryButtonText ? (
```
kdembler:
```suggestion
  const buttonNode = primaryButtonText ? (
```

src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -92,10 +91,10 @@ export const ActionBar: React.FC<ActionBarProps> = ({
               icon={primaryButtonTooltipText?.icon}
               placement="top-end"
             >
-              {renderButton()}
+              {renderButton}
```
kdembler:
```suggestion
              {buttonNode}
```

src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -92,10 +91,10 @@ export const ActionBar: React.FC<ActionBarProps> = ({
...
             </Tooltip>
           ) : (
-            <>{renderButton()}</>
+            <>{renderButton}</>
```
kdembler:
```suggestion
            <>{buttonNode}</>
```

kdembler (approved):

PR #1343, opened 2021-09-09
Refactor `uploadsManager` store

### LINKED ISSUE #1306: Refactor uploadManager store
Once #1239 is done,  we should use our `createStore` and immer as well

drillprop:
Fix #1306

### DIFF
--- a/src/hooks/useDeleteVideo.tsx
+++ b/src/hooks/useDeleteVideo.tsx
@@ -11,7 +11,7 @@ export const useDeleteVideo = () => {
   const { joystream } = useJoystream()
   const handleTransaction = useTransaction()
   const { activeMemberId } = useAuthorizedUser()
-  const { removeAssetsWithParent: removeAssetsFromUploads } = useUploadsStore((state) => state.actions)
+  const removeAssetsFromUploads = useUploadsStore((state) => state.actions.removeAssetsWithParent)
   const [openDeleteVideoDialog, closeDeleteVideoDialog] = useDialog()
 
   const client = useApolloClient()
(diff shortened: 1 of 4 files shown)


src/providers/uploadsManager/store.ts
```diff
@@ -10,81 +8,93 @@ type AssetFile = {
   blob: File | Blob
 }
 
-type UploadStoreState = {
+type _UploadStoreState = {
```
kdembler: Why the `_` prefix?

src/hooks/useDeleteVideo.tsx
```diff
@@ -11,7 +11,7 @@ export const useDeleteVideo = () => {
   const { joystream } = useJoystream()
   const handleTransaction = useTransaction()
   const { activeMemberId } = useAuthorizedUser()
-  const removeAssetsFromUploads = useUploadsStore((state) => state.removeAssetsWithParent)
+  const { removeAssetsWithParent: removeAssetsFromUploads } = useUploadsStore((state) => state.actions)
```
kdembler: If you need a specific action, it's better to do:
```suggestion
  const removeAssetsFromUploads = useUploadsStore((state) => state.actions.removeAssetsWithParent)
```

src/providers/uploadsManager/uploadsManager.tsx
```diff
@@ -23,20 +23,10 @@ export const UploadsManager: React.FC = () => {
...
-        pendingAssetsIds: state.pendingAssetsIds,
-        removePendingAssetId: state.removePendingAssetId,
-        setUploadStatus: state.setUploadStatus,
-      }),
-      shallow
-    )
+  const { addAsset, removeAsset, setIsSyncing, removePendingAssetId, setUploadStatus } = useUploadsStore(
+    (state) => state.actions
+  )
+  const { isSyncing, pendingAssetsIds } = useUploadsStore((state) => state)
```
kdembler: This will slightly change the behaviour. Previously we provided an object extracted from state and asked for a `shallow` comparison. That meant that only if one of the requested items changed, this component would rerender. With `useUploadsStore((state) => state)`, the component will rerender on any change in the `state`, are we sure that is what we want?

kdembler (approved):

PR #1347, opened 2021-09-13
Show more settings section

### LINKED ISSUE #1276: Add show more settings section in EditVideoForm
image

drillprop:
Fix #1276 
Design: https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/[Page]-Video-workspace?node-id=421%3A6708

### DIFF
--- a/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
+++ b/src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
@@ -122,23 +122,13 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
   } = useForm<EditVideoFormFields>({
     shouldFocusError: true,
     defaultValues: {
-      title: '',
-      isPublic: true,
-      language: 'en',
-      category: null,
-      licenseCode: 1001,
-      licenseAttribution: '',
-      licenseCustomText: '',
-      description: '',
-      hasMarketing: false,
-      publishedBeforeJoystream: null,
-      isExplicit: false,
       assets: {
         video: {
           contentId: null,
         },
         thumbnail: { cropContentId: null, originalContentId: null },
       },
+      ...tabData,
     },
   })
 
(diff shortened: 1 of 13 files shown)


src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
```diff
@@ -120,13 +125,13 @@ export const EditVideoForm: React.FC<EditVideoFormProps> = ({
       isPublic: true,
       language: 'en',
       category: null,
-      licenseCode: null,
+      licenseCode: 1001,
```
rafalpawlow: Do we need it here since it's already declared in provider? If yes, maybe we could move it to some common const? WDYT?
https://github.com/Joystream/atlas/pull/1347/files#diff-6ac1f2b87ffac0734ea21a1d15a11969a70658bce02cfde51fb3392e582e5819R73

rafalpawlow (changes requested):

src/views/studio/EditVideoSheet/EditVideoForm/EditVideoForm.tsx
drillprop: That's a good point, fixed. I'm a little worried that I could break something here though. I'm just surprised, why we have this doubled code here.
Could you also test this? Just play around - check if the draft is saved properly, check if you can upload video with correct data, etc.
I think everything should work, but it's possible that I could missed something.

eldiegod (approved):

KubaMikolajczyk:
Maybe it's just wishful thinking but is it possible to display the scrollbar as an overlay rather than a part of the layout - to avoid this prominent layout shift when expanding more settings - if before expanding screen was large enough to be displayed without scrollbar? Wdyt?

https://user-images.githubusercontent.com/31144205/133470112-25329268-2254-4466-acde-49b75877034f.mov

drillprop:
> Maybe it's just wishful thinking but is it possible to display the scrollbar as an overlay rather than a part of the layout - to avoid this prominent layout shift when expanding more settings - if before expanding screen was large enough to be displayed without scrollbar? Wdyt?

It's really hard to accomplish what you proposing with the current markup. I mean, I've tried and I didn't manage to do that 😅 . What if we leave the scrollbar always visible? Is it acceptable for you?

rafalpawlow (approved):

PR #1368, opened 2021-09-17
Add lottie animations to transaction dialog, adjust dialogs to new design requirements

### LINKED ISSUE #1279: Update transaction dialog component
image
image
image

rafalpawlow:
Closes #1279


src/providers/transactionManager/transactionManager.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 
 import { useQueryNodeStateSubscription } from '@/api/hooks'
 import { TransactionDialog } from '@/components/TransactionDialog'
+// import { ExtrinsicStatus } from '@/joystream-lib'
```
drillprop:
```suggestion
```

src/providers/transactionManager/transactionManager.tsx
```diff
@@ -13,6 +14,8 @@ export const TransactionManager: React.FC = () => {
     actions: { removeOldBlockActions, setDialogStep },
   } = useTransactionManagerStore((state) => state)
 
+  //const dialogStep = ExtrinsicStatus.ProcessingAssets
+
```
drillprop:
```suggestion
```

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -15,57 +34,139 @@ export type TransactionDialogProps = Pick<ActionDialogProps, 'className'> & {
...
+
+  const transactionSteps = Object.values(TRANSACTION_STEPS_DETAILS)
 
-  const transactionStepsWithoutProcessingAssets = Object.values(TRANSACTION_STEPS_DETAILS).filter(
-    (step) => step.title !== TRANSACTION_STEPS_DETAILS[ExtrinsicStatus.ProcessingAssets].title
-  )
   return (
     <ActionDialog
       showDialog={!!stepDetails}
       secondaryButton={{ text: 'Cancel', onClick: onClose, disabled: !canCancel }}
```
drillprop: One thing which I don't like is that when the transaction is finished we are still showing the `Cancel` button which could be wrongly interpreted by a user. 
In the prototype, we automatically close the dialog with the disabled cancel button. We could do that, but it will require adding some delay(we still want to display completed animation). 

On the other hand, I think we could just show `Close` button: 
```suggestion
      secondaryButton={{
        text: status === ExtrinsicStatus.Completed ? 'Close' : 'Cancel',
        onClick: onClose,
        disabled: !canCancel,
      }}
```
And to be honest, I'm fine with the second option, but I'll consult this with a designer and let you know.

drillprop (changes requested):

src/components/TransactionDialog/TransactionDialog.tsx
drillprop: Ok, talked to Adam, and we should automatically close the dialog after delay.

src/components/TransactionDialog/TransactionDialog.style.ts
```diff
@@ -1,55 +1,144 @@
...
+    position: absolute;
+    display: ${({ isActive, past }) => (isActive || past ? 'block' : 'none')};
+    animation-name: ${({ isActive, past }) => (isActive && !past ? 'slideAndFadeIn' : 'unset')};
+    animation-iteration-count: infinite;
+    animation-timing-function: cubic-bezier(0, 0, 0.3, 1);
+    animation-duration: 2s;
+    background-color: ${colors.gray[500]};
+    width: 100%;
+
+    @keyframes slideAndFadeIn {
```
drillprop: you should use emotion's `keyframes` here. We encountered a situation once where certain animation was applied to another element.

src/components/TransactionDialog/TransactionDialog.style.ts
```diff
@@ -1,55 +1,144 @@
...
+  align-items: center;
+  justify-content: center;
+  background-color: ${colors.blue[500]};
+  border-radius: 50%;
+  animation-iteration-count: 1;
+  animation-name: zoomOut;
+  animation-duration: 600ms;
+  animation-timing-function: ${transitions.easing};
+
+  @keyframes zoomOut {
```
drillprop: Same as above

src/components/TransactionDialog/TransactionDialog.style.ts
```diff
@@ -1,55 +1,144 @@
...
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background-color: ${colors.blue[500]};
+  border-radius: 50%;
+  animation-iteration-count: 1;
+  animation-name: zoomOut;
+  animation-duration: 600ms;
+  animation-timing-function: ${transitions.easing};
```
drillprop: btw, I think we could easily use shorthand property here.

kdembler:
It seems to me that after this change the first step (processing your assets) is always displayed for a second or two, even when no assets have been changed. This didn't happen before

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -15,58 +35,183 @@ export type TransactionDialogProps = Pick<ActionDialogProps, 'className'> & {
...
+          in={polkadotLogoVisible}
+          timeout={200}
+          classNames={transitions.names.fade}
+          mountOnEnter
+          unmountOnExit
+        >
+          <PolkadotLogoWrapper>
+            <StyledPolkadotLogo />
+            <Text secondary variant="caption">
+              Continue in Polkadot
```
kdembler:
```suggestion
              Continue in Polkadot extension
```

rafalpawlow:
> It seems to me that after this change the first step (processing your assets) is always displayed for a second or two, even when no assets have been changed. This didn't happen before

Yes, it is intentional and it was a request from @toiletgranny to force it to be shown

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -15,58 +35,183 @@ export type TransactionDialogProps = Pick<ActionDialogProps, 'className'> & {
...
+        disabled: !canCancel,
+      }}
       exitButton={false}
+      illustration
       {...actionDialogProps}
     >
       <StepsBar>
-        {transactionStepsWithoutProcessingAssets.map(({ tooltip }, idx) => (
+        {transactionSteps.map(({ tooltip }, idx) => (
           <Tooltip key={idx} text={tooltip} placement="top-end">
```
kdembler: There are currently issues with those tooltips. To start off, the displayed titles don't actually match the current step. It seems the hover trigger is not always working. I suggest we just drop it at this point since I haven't seen an indication of those still being present in the current design

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -15,58 +35,183 @@ export type TransactionDialogProps = Pick<ActionDialogProps, 'className'> & {
...
+      size: {
+        width: 216,
+        height: 216,
+      },
+      loop: true,
+    },
   },
   [ExtrinsicStatus.Unsigned]: {
-    title: 'Waiting for signature...',
+    title: 'Sign in Polkadot',
```
kdembler:
```suggestion
    title: 'Waiting for your signature'
```

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -15,58 +35,183 @@ export type TransactionDialogProps = Pick<ActionDialogProps, 'className'> & {
...
       : null
 
-  const canCancel = status === ExtrinsicStatus.ProcessingAssets || status === ExtrinsicStatus.Unsigned
+  useEffect(() => {
+    if (status !== null && initialStatusRef.current === null) {
+      initialStatusRef.current = status
+    }
+    if (status === null) {
+      initialStatusRef.current = null
+    }
```
kdembler, line 47: Why do we need this `initialStatusRef`?

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -15,58 +35,183 @@ export type TransactionDialogProps = Pick<ActionDialogProps, 'className'> & {
 
 const TRANSACTION_STEPS_DETAILS = {
```
kdembler: Maybe we could move those to a separate file to improve readability here?

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -15,58 +35,183 @@ export type TransactionDialogProps = Pick<ActionDialogProps, 'className'> & {
...
+      'An unexpected error was encountered. If this persists, our Discord community may be a good place to find some help.',
+    tooltip: 'Propagation',
+    animation: {
+      data: errorAnimation,
+      size: {
+        width: 288,
+        height: 264,
+      },
+      loop: false,
+    },
```
kdembler: Seems the details for error were provided here but I'm still seeing the old error dialog:
Screenshot 2021-09-22 at 16 13 53

Figma reference: https://www.figma.com/file/Vk2Z4QOiVa5bB6q3cBIG5J/Joystream-Atlas?node-id=9066%3A227100

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -1,11 +1,31 @@
-import React from 'react'
+import React, { useEffect, useRef, useState } from 'react'
+import { CSSTransition } from 'react-transition-group'
 
+import { ActionDialog, ActionDialogProps } from '@/components/ActionDialog'
 import { ExtrinsicStatus } from '@/joystream-lib'
+import errorAnimation from '@/shared/assets/animations/stepper/error.json'
```
kdembler: We already have this animation in the codebase, let's not duplicate it

src/components/TransactionDialog/TransactionDialog.tsx
```diff
@@ -1,11 +1,31 @@
...
+import firstStepAnimation from '@/shared/assets/animations/stepper/step1.json'
+import secondStepAnimation from '@/shared/assets/animations/stepper/step2.json'
+import thirdStepAnimation from '@/shared/assets/animations/stepper/step3.json'
+import fourthStepAnimation from '@/shared/assets/animations/stepper/step4.json'
```
kdembler: Let's improve naming on those to be explicit what this is referencing:
1. Let's put it at `@/shared/assets/animations/transaction/`
2. Let's call each file according to the step it's used for, e.g. `processing-assets.json`, etc

src/providers/transactionManager/useTransaction.ts
```diff
@@ -57,7 +55,11 @@ export const useTransaction = (): HandleTransactionFn => {
       if (preProcess) {
         setDialogStep(ExtrinsicStatus.ProcessingAssets)
         try {
-          await preProcess()
+          await new Promise((resolve) =>
+            setTimeout(() => {
+              resolve(preProcess())
+            }, 2000)
```
kdembler: If I'm not mistaken this change is supposed to make the first step displayed even if there's no processing to do, right? This is the UI/design leaking into the transaction handling logic which we should avoid. This has to be somehow handled by the dialog itself.

kdembler (changes requested):

src/components/TransactionDialog/TransactionDialog.tsx
rafalpawlow: This is hack to determine whether it's upload transaction or something else, like delete for example. If it's upload transaction, then `initialStatusRef.current === 0`

src/providers/transactionManager/useTransaction.ts
rafalpawlow: Exactly. This is possible to handle it by the dialog but this way Polkadot extension opens while you have `Processing your assets` dialog open.

kdembler: Additionally, maybe instead of predefined value of 2s we just show the animation for one loop?

src/components/TransactionDialog/TransactionDialog.tsx
kdembler: Why do we need this information?

src/providers/transactionManager/useTransaction.ts
kdembler: Oh yes I see the issue

src/components/TransactionDialog/TransactionDialog.tsx
rafalpawlow: Because we need to know how many steps we have (we show number of steps at the top of dialog). Non upload transaction has 1 step less than upload transaction.

src/providers/transactionManager/useTransaction.ts
kdembler: Based on the Rocket chat discussion we don't need this

kdembler (approved):

PR #1369, opened 2021-09-17
Update action bar

### LINKED ISSUE #1349: Update action bar component
There are some changes to the action bar especially for smaller screens, we should update this.
Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=798%3A92929

drillprop:
Fix #1349

Design https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=798%3A90836


src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -17,87 +16,131 @@ import { Button } from '../Button'
...
+  size?: ActionBarSize
   primaryText?: string
   secondaryText?: string
-  primaryButtonTooltipText?: {
-    headerText: string
+  fullWidth?: boolean
+  className?: string
+  primaryButtonText?: string
+  primaryButtonDisabled?: boolean
+  primaryButtonOnClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```
kdembler: Could we use the same API as for the `ActionDialog` here? 

```tsx
type DialogButtonProps = {
  text?: string
  disabled?: boolean
  onClick?: (e: React.MouseEvent) => void
} & Omit<ButtonProps, 'children'>

export type ActionDialogProps = {
  // ...
  primaryButton?: DialogButtonProps
  // ...
}
```

kdembler (commented):
Could we use this occasion to remove the separate `ActionBarTransaction` component? I'm pretty sure we don't need it as it seems transaction will be the only context for that component. 

Some visual feedback:
1. Seems the `Fee: ` text is using incorrect variant, the font sizes don't match
2. The height of the bar changes on mobile:

https://user-images.githubusercontent.com/12646744/134154726-17d61c05-6a2a-4b79-8ba6-79f4567b15e8.mov

3. Maybe we could fade in/out the cancel button?

src/shared/components/ActionBar/ActionBar.stories.tsx
```diff
@@ -1,14 +1,17 @@
...
-import { ActionBarTransaction, ActionBarTransactionProps } from './ActionBarTransaction'
 
 export default {
   title: 'Shared/A/ActionBar',
   component: ActionBar,
+  argTypes: {
+    size: {
+      options: ['small', 'medium', 'large'],
+      control: { type: 'radio' },
+    },
```
kdembler: Not needed anymore, right?

src/shared/components/ActionBar/ActionBar.tsx
```diff
@@ -1,103 +1,132 @@
...
-}) => {
-  const buttonNode = primaryButtonText ? (
-    <Button disabled={disabled} onClick={onConfirmClick} size="large" type="submit">
-      {primaryButtonText}
-    </Button>
-  ) : null
+export const ActionBar = React.forwardRef<HTMLDivElement, ActionBarProps>(
+  ({ primaryText, secondaryText, className, primaryButton, secondaryButton, draftBadge }, ref) => {
+    const smMatch = useMediaMatch('sm')
+    const lgMatch = useMediaMatch('lg')
```
kdembler: I see you moved the media match inside the component but now that size is handled only from within this component, I guess we could just use regular media queries in CSS? Or is there something making this hard?

kdembler (approved):

PR #1371, opened 2021-09-17
Update new video tile

### LINKED ISSUE #1364: Update new video tile
New video tile should occupy more space: https://www.figma.com/file/LF7LOMhOA46bRcYHWlw1Bq/[Page]-My-videos?node-id=215%3A4134

image

drillprop:
Fix #1364 
Design: https://www.figma.com/file/LF7LOMhOA46bRcYHWlw1Bq/[Page]-My-videos?node-id=217%3A6627

### DIFF
--- a/src/views/studio/MyVideosView/MyVideos.styles.ts
+++ b/src/views/studio/MyVideosView/MyVideos.styles.ts
@@ -2,6 +2,7 @@ import styled from '@emotion/styled'
 
 import { LimitedWidthContainer } from '@/components/LimitedWidthContainer'
 import { DismissibleBanner } from '@/shared/components/DismissibleBanner'
+import { Grid } from '@/shared/components/Grid'
 import { Pagination } from '@/shared/components/Pagination'
 import { Text } from '@/shared/components/Text'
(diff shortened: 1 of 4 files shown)


src/views/studio/MyVideosView/MyVideos.styles.ts
```diff
@@ -28,6 +29,10 @@ export const TabsContainer = styled.div`
   }
 `
 
+export const StyledGrid = styled(Grid)`
+  grid-template-rows: 1fr 1fr;
```
kdembler: What's the point of this?

drillprop: This is needed for mobile views.
without: 
image
and with this property: 
image

I guess it's a workaround, I couldn't figure out which property I need set here to not shrink this tile.

kdembler: All right. I think this will work better:
```suggestion
  grid-auto-rows: 1fr;
```

kdembler (approved):

Issue #1362, opened 2021-09-16
UI layout for all videos section in video category page


eldiegod:
Figma: https://www.figma.com/file/3uFQ2y1jlS6fbp7Ac0yIpv/%5B%2312%5D-Video-category-page-flows

estimate will probably run short

PR #1386, opened 2021-09-21
All Category videos component layout

eldiegod:
implements: #1362

Basic UI Layout for All Category videos component.

Details on the search bar to be discussed with designers as it possibly will be removed (it is removed)

Filters and popover to come in a different PR


src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+  const handleFilterClick = () => {
+    setiIsFiltersOpen((value) => !value)
+  }
+  return (
+    <Container>
+      <ControlsContainer>
+        <Text variant={mdMatch ? 'h4' : 'h5'}>All videos (441)</Text>
+        {mdMatch && <Select helperText={null} placeholder="Language" value={null} items={SORT_OPTIONS} />}
```
drillprop:
```suggestion
        {mdMatch && <Select size="small" helperText={null} placeholder="Language" value={null} items={SORT_OPTIONS} />}
```

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -0,0 +1,107 @@
...
+          ) : (
+            <Button variant="secondary" icon={<SvgActionFilters />} onClick={handleFilterClick}>
+              Filters
+            </Button>
+          )}
+        </FiltersSearchContainer>
+        {mdMatch && (
+          <SortContainer>
+            <Text variant="body2">Sort by</Text>
+            <Select helperText={null} value={sortVideosBy} items={SORT_OPTIONS} onChange={handleSorting} />
```
drillprop: It's just looks really weird right now: 
image


```suggestion
            <Select size="small" helperText={null} value={sortVideosBy} items={SORT_OPTIONS} onChange={handleSorting} />
```

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -0,0 +1,107 @@
...
+    (searchQuery: string) => {
+      setSearchQuery(searchQuery)
+      searchVideo({
+        variables: {
+          text: searchQuery,
+          whereVideo: {
+            isPublic_eq: true,
+            mediaAvailability_eq: AssetAvailability.Accepted,
+            thumbnailPhotoAvailability_eq: AssetAvailability.Accepted,
+            // channelId_eq: '0',
```
drillprop:
```suggestion
```

src/components/CategoryVideos/CategoryVideos.styles.tsx
```diff
@@ -0,0 +1,58 @@
+import styled from '@emotion/styled'
+
+import { colors, media, sizes } from '@/shared/theme'
+
+export const Container = styled.div`
+  margin: ${sizes(16)} 0 0 0;
```
drillprop:
```suggestion
  margin-top: ${sizes(16)};
```

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <Text variant={mdMatch ? 'h4' : 'h5'}>All videos (441)</Text>
+        {mdMatch && <Select helperText={null} placeholder="Language" value={null} items={SORT_OPTIONS} />}
+        <FiltersSearchContainer>
+          <Search
+            searchInputRef={searchInputRef}
+            isSearchInputOpen={isSearchInputOpen}
+            setIsSearchingInputOpen={setIsSearchingInputOpen}
+            setIsSearching={setIsSearching}
+            search={search}
+          />
```
drillprop: Search input should not expand to the full width.
Design:
image

src/components/CategoryVideos/CategoryVideos.tsx
drillprop: Also why do we have here `Language` as a placeholder and `SORT_OPTIONS` as items?

src/hooks/useToggle.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const useToggle = (initialState = false) => {
+  // Initialize the state
+  const [state, setState] = useState(initialState)
+
+  // Define and memorize toggler function in case we pass down the comopnent,
+  // This function change the boolean value to it's opposite value
+  const toggle = useCallback(() => setState((state) => !state), [])
+
+  return [state, toggle] as const
```
drillprop:
```suggestion
  return [state, toggle]
```
I don't think we need this

src/hooks/useToggle.ts
```diff
@@ -0,0 +1,12 @@
+import { useCallback, useState } from 'react'
+
+export const useToggle = (initialState = false) => {
+  // Initialize the state
```
drillprop:
```suggestion
```
I think we all know what it does 😅

src/hooks/useToggle.ts
```diff
@@ -0,0 +1,12 @@
...
+  const [state, setState] = useState(initialState)
+
+  // Define and memorize toggler function in case we pass down the comopnent,
+  // This function change the boolean value to it's opposite value
```
drillprop:
```suggestion
```
same

src/shared/components/Search/Search.tsx
```diff
@@ -0,0 +1,171 @@
...
+        setIsSearchingInputOpen(false)
+        searchInputRef.current?.blur()
+        setSearchQuery('')
+      }
+    },
+    [search, searchInputRef, searchQuery, setIsSearching, setIsSearchingInputOpen, setSearchQuery]
+  )
+
+  return (
+    <SearchContainer className={className}>
```
drillprop: I said this before in other PR and I still maintain this. We should just use `form` here, like this:
```tsx
    <SearchContainer
      as="form"
      className={className}
      onSubmit={(e) => {
        e.preventDefault()
        // all on submit logic like search(searchQuery)
      }}
    >
```
Hitting `enter` will just work out of the box(but we could leave listening on `escape` though).

drillprop (changes requested):

src/components/CategoryVideos/CategoryVideos.tsx
eldiegod: Not a real reason, it's just a visual component atm but functionality will be added later

src/hooks/useToggle.ts
eldiegod: we do, for types

drillprop: Type inference works just fine here.  Are you sure? 
Without `as const`
image
With `as const`
image
Which is basically the same.

Anyway, why do we need this hook in the first place? It seems that it's not used at all.

eldiegod: not the same with `as const` first item of the array is a boolean and second item is a fn always; without `as const` type is an array of any size can have items booleans or fns.

But yes this might get removed or not depending on another PR that builds on this one

src/components/CategoryVideos/CategoryVideos.tsx
eldiegod: I'll hold up on the searchbar changes as @kdembler mentioned it might get dropped

src/hooks/useToggle.ts
drillprop: Just tested this. My mistake, you're right about `as const` 👍🏻

drillprop: basically without `as const` it's a regular array, and with `as const` it's a **tuple**(array with a fixed number of elements)

eldiegod: Yep, that the first element is type boolean and the second of type fn which is important.

Another way to do it is to define the return type of the fn but `as const` it's just convenient

eldiegod:
Search bar was removed for now and will be added at some point in another PR

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -4,12 +4,11 @@ import { fluidRange } from 'polished'
 import { ChannelLink } from '@/components/ChannelLink'
 import { TOP_NAVBAR_HEIGHT } from '@/components/TopbarBase'
 import { Button } from '@/shared/components/Button'
-import { IconButton } from '@/shared/components/IconButton'
+import { Search } from '@/shared/components/Search/Search'
```
kdembler: Should import from `@/shared/components/Search`. If `index.ts` doesn't exist there, let's add it

src/views/viewer/CategoryView/CategoryView.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+import { CategoryVideos } from '@/components/CategoryVideos/CategoryVideos'
```
kdembler: Should import from `index.ts`

src/views/viewer/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './NewView'
 export * from './PopularView'
 export * from './SearchOverlayView'
 export * from './VideoView'
+export * from './CategoryView/CategoryView'
```
kdembler: Should re-export from `index.ts`

package.json
```diff
@@ -51,6 +51,7 @@
     "@joystream/types": "0.16.1",
     "@loadable/component": "^5.15.0",
     "@polkadot/extension-dapp": "~0.37.3-17",
+    "@popperjs/core": "^2.10.1",
```
kdembler: I don't think this and `react-popper` are needed for this PR?

src/components/CategoryVideos/CategoryVideos.styles.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const ControlsContainer = styled.div`
+  display: grid;
+  grid-template-columns: 1fr auto;
+  gap: ${sizes(4)};
+  align-items: center;
+  padding-bottom: ${sizes(4)};
+  border-bottom: 1px solid ${colors.gray[700]};
+
+  ${media.md} {
+    grid-template-columns: auto 168px 1fr 241px;
```
kdembler: Where are those values coming from?

src/shared/components/Search/Search.tsx
```diff
@@ -0,0 +1,171 @@
+import styled from '@emotion/styled'
```
kdembler: I just rebased `discovery` with `dev`. Recently we've added some additional styles to the search component so this would need to be reflected here. Although I'm a bit unsure whether we should stick that into `Search` component. Design seems to define specific visuals just for the Channel search component. Maybe for now we should keep it in the channel view since we won't be using it here anyway?

kdembler (changes requested):

package.json
eldiegod: I removed it from popover-base will wait on it to get merged to rebase discovery here

src/components/CategoryVideos/CategoryVideos.styles.tsx
eldiegod: This is the size of the dropdown sections on MD breakpoint:
https://www.figma.com/file/UdBMxrLspjcf6IVYYJS8YP/[Page]-Discover-and-Video-category?node-id=252%3A93102

kdembler: Could you help me understand what you mean by dropdown sections? Are you sure that Figma specifies this values as fixed as you currently do? I'm pretty sure the values depend on something else than a hardcoded value

eldiegod: Maybe @toiletgranny might be able to clarify here, but the language dropdown and the sort by dropdown don't seem to be elements that scale or shrink in this MD breakpoint, in any case it was the searchbar the element that used to take free available space but is not there anymore

kdembler: I still don't understand what these values are supposed to style. As far as I can tell, selects in the design just follow automatic width as all the other selects? I don't think they need to be explicitly sized like this?

src/views/viewer/ChannelView/ChannelView.style.tsx
```diff
@@ -175,7 +175,7 @@ type SearchContainerProps = {
 export const SearchContainer = styled.div<SearchContainerProps>`
   display: flex;
   grid-area: search;
-  width: 100%;
+  width: 192px;
```
kdembler: Why are we making this change to channel view? Seems outside of scope of this PR

src/hooks/useToggle.ts
```diff
@@ -0,0 +1,9 @@
+import { useCallback, useState } from 'react'
+
+export const useToggle = (initialState = false) => {
+  const [state, setState] = useState(initialState)
+
+  const toggle = useCallback(() => setState((state) => !state), [])
+
+  return [state, toggle] as const
+}
```
kdembler: Is this needed here? Let's remove it

package.json
```diff
@@ -80,6 +81,7 @@
     "react-intersection-observer": "^8.31.0",
     "react-lottie-player": "^1.4.0",
     "react-number-format": "^4.4.4",
+    "react-popper": "^2.2.5",
```
kdembler: Let's remove those deps, they're not part of this PR, so they shouldn't be here.

kdembler (changes requested):

src/views/viewer/ChannelView/ChannelView.style.tsx
eldiegod: It's a fix I added in every PR after fixing the conflicts of the Search component, not ideal but it won't matter as it will be in discovery as well.

kdembler: I don't understand. Both this PR and discovery sprint don't make any changes to the `ChannelView` at all. What conflict do you mean it solves? There shouldn't be any conflicts

src/components/CategoryVideos/CategoryVideos.styles.tsx
kdembler: @eldiegod Still waiting on response for this

eldiegod: They are the reserved space for the language dropdown and the sort by label + dropdown, I'm not sure what are you suggesting here instead? making them grow or shrink doesn't seem to fit design

src/views/viewer/ChannelView/ChannelView.style.tsx
eldiegod: Yes, I think I was confused about this, removed the change

kdembler (approved):

PR #1387, opened 2021-09-21
Check if assets is processing when launching the app

### LINKED ISSUE #1240: Double check if upload is still pending when launching the app
Currently if we launch the app and any of the channels' assets is `PENDING`, we tell the user that the upload wasn't finished and they need to start over. That may not be true though, the user may have already finished the upload but the query node haven't processed it yet. To avoid that, when starting an upload we could save the target storage provider URL in local storage. Then on refresh, if the `PENDING` asset has saved URL, we could send a HEAD request to that URL to see if it was already uploaded. If the HEAD succeeds, we will know that we only need to wait for the query node and use the same processing logic as in #1239.

HEAD strategy needs to be tested though to see if the storage provider will respond to it right after the upload was finished. If it doesn't work, we can use some other strategy, like saving upload success time in local storage and using that to determine whether the upload likely just finished.

drillprop:
Fix #1240

### DIFF
--- a/src/providers/uploadsManager/uploadsManager.tsx
+++ b/src/providers/uploadsManager/uploadsManager.tsx
@@ -23,9 +23,8 @@ export const UploadsManager: React.FC = () => {
     (state) => state.uploads.filter((asset) => asset.owner === activeChannelId),
     shallow
   )
-  const { addAsset, removeAsset, setIsSyncing, removePendingAssetId, setUploadStatus } = useUploadsStore(
-    (state) => state.actions
-  )
+  const { addAssetToUploads, removeAssetFromUploads, setIsSyncing, removePendingAssetId, setUploadStatus } =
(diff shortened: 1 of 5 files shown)


rafalpawlow (approved):

src/providers/uploadsManager/uploadsManager.tsx
```diff
@@ -37,23 +38,29 @@ export const UploadsManager: React.FC = () => {
   })
 
   useEffect(() => {
-    if (!pendingAssetsIds.length) {
+    processingAssetsIds.map((assetId) => {
+      setUploadStatus(assetId, { progress: 100, lastStatus: 'processing' })
```
kdembler, line 46: This will change status to `processing` even if it's already set

kdembler (approved):

Issue #1358, opened 2021-09-16
Create basic Popover component


eldiegod:
Create a `Popover` that is an empty panel with Figma-specified options for positioning. It doesn't define any content (footer/header/etc). Let's use this in the `ContextMenu` Later it will be used as a basis for `PopoverDialog`

Figma: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=510%3A97749

PR #1389, opened 2021-09-22
Popover base component

eldiegod:
fixes:  #1358


src/shared/components/ContextMenu/ContextMenu.tsx
```diff
@@ -27,13 +19,12 @@ export const ContextMenuItem: React.FC<MenuItemProps> = ({ icon, children, onCli
...
+export const ContextMenu: React.FC<PopoverBaseProps> = ({ children, isVisible, ...rest }) => {
   return (
-    <Portal containerRef={contextMenuContainerRef}>
-      <CSSTransition in={isActive} timeout={150} classNames="menu" unmountOnExit>
-        <StyledContainer position={position}>{children}</StyledContainer>
+    <PopoverBase isVisible={isVisible} {...rest}>
+      <CSSTransition in={isVisible} timeout={150} classNames="menu" unmountOnExit>
+        <StyledContainer>{children}</StyledContainer>
       </CSSTransition>
-    </Portal>
```
drillprop, line 37: Not sure which change causing this, but the context menu is broken now: 
image
I think it may be due to the fact that we're not using Portal here anymore.

drillprop (changes requested):

src/shared/components/ContextMenu/ContextMenu.tsx
eldiegod: z-index seems to do the job here, you think that would be enough or is the Portal imporant?

drillprop: I think it would be better to leave Portal here, but if it requires a lot of work let's just use `z-index`.

eldiegod: yeah, that was my thought as well

drillprop: I'm not sure if we understood each other here 😅 . I wanted to **keep** this portal. 
let's bring it back 😄  opened https://github.com/eldiegod/atlas/pull/3 WDYT?

eldiegod: oh lol, ok it's back now

src/hooks/useContextMenu.tsx
```diff
@@ -1,35 +1,21 @@
...
   useEffect(() => {
-    if (!isActive) {
+    if (!isVisible) {
       return
     }
-    const closeContextMenu = () => setMenuActive(false)
+    const closeContextMenu = () => setIsVisible(false)
     document.addEventListener('click', closeContextMenu, { once: true, capture: true })
-  }, [isActive])
+  }, [isVisible])
```
kdembler: Do we still need to handle this? I would expect the popover library to handle this by itself?

package.json
```diff
@@ -51,6 +51,7 @@
     "@joystream/types": "0.16.1",
     "@loadable/component": "^5.15.0",
     "@polkadot/extension-dapp": "~0.37.3-17",
+    "@popperjs/core": "^2.10.1",
```
kdembler: I've noticed we pull in `@popperjs` but we already use `tippy.js` which is a wrapper built on top of popper. From my reading of docs it seems `tippy.js` provides some more functionality on top of the base library. That kind of suggests to me that we shouldn't be using both of them at the same time. One of them should be enough I think. Wdyt?

src/providers/overlayManager.tsx
```diff
@@ -10,7 +10,7 @@ type OverlayManagerContextValue = {
   scrollLocked: boolean
   setOverlaysSet: React.Dispatch<React.SetStateAction<Set<string>>>
   dialogContainerRef: React.RefObject<HTMLDivElement>
-  contextMenuContainerRef: React.RefObject<HTMLDivElement>
+  contextContainerRef: React.RefObject<HTMLDivElement>
```
kdembler: Do we still need to go through portal and overlay manager? I think `popper` handles this automatically? After all we currently have working tooltips that don't have a separate ref in overlay manager

kdembler (changes requested):

src/providers/overlayManager.tsx
eldiegod: Switched to tippy, it has some different abstractions and handles more things on its own but animations might be a bit trickier here and that might or might not matter

src/hooks/useToggle.ts
```diff
@@ -0,0 +1,9 @@
+import { useCallback, useState } from 'react'
+
+export const useToggle = (initialState = false) => {
+  const [state, setState] = useState(initialState)
+
+  const toggle = useCallback(() => setState((state) => !state), [])
+
+  return [state, toggle] as const
+}
```
kdembler: I don't think this hook is used anywhere?

src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -343,45 +341,47 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
...
-          size="small"
-          isActive={contextMenuOpts.isActive}
-        >
-          <SvgGlyphMore />
-        </KebabMenuButtonIcon>
-        <ContextMenu contextMenuOpts={contextMenuOpts}>
-          {publisherMode ? (
+
+        <ContextMenu
+          placement={'bottom-end'}
```
kdembler: Unnecessary brackets

src/providers/overlayManager.tsx
kdembler: Seems this was removed from the `ContextMenu`, do we still need this for anything?

src/shared/components/ContextMenu/ContextMenu.tsx
```diff
@@ -1,23 +1,14 @@
 import React, { ReactNode } from 'react'
-import { CSSTransition } from 'react-transition-group'
-
-import { Portal } from '@/components/Portal'
-import { useOverlayManager } from '@/providers/overlayManager'
 
 import { StyledContainer, StyledMenuItem, StyledText } from './ContextMenu.style'
 
+import { PopoverBase, PopoverBaseProps } from '../PopoverBase/PopoverBase'
```
kdembler: Should import from `index.ts`

src/shared/components/PopoverBase/PopoverBase.tsx
```diff
@@ -0,0 +1,44 @@
...
+export type PopoverBaseProps = {
+  content: React.ReactNode
+  isVisible?: boolean
+  placement?: Placement
+  offset?: [number, number]
+  hideOnClick?: boolean
+  className?: string
+}
+
+export const PopoverBase: React.FC<PopoverBaseProps> = ({
```
kdembler: I think I'd opt for just `Popover` for this component name

kdembler (changes requested):
We're getting there :)

There are also some visual issues:
1) The kebab button VideoTile no longer has a `pointer` cursor
2) The animation we had when the `ContextMenu` was opening is gone now, we should bring it back. I think `tippy.js` has animations built-in?
3) Seems we could slightly improve the positioning of the ContextMenu, check out https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A126061
Screenshot 2021-09-29 at 14 19 26
Screenshot 2021-09-29 at 14 19 08

src/shared/components/VideoTileBase/VideoTileBase.tsx
drillprop: @kdembler Just an idea, maybe we could enable this rule? https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md 
I also tend to forget about this. Of course, not as part of this PR, wdyt?

src/shared/components/PopoverBase/PopoverBase.tsx
eldiegod: There is another component called Popover built on top coming that's why

eldiegod: see PR popover

kdembler: I think we agreed to call this one `Popover` and the one from the other PR `PopoverDialog`, no?

eldiegod: I think that might be misleading as in Figma it already is called `popover`

src/shared/components/PopoverBase/PopoverBase.tsx
```diff
@@ -0,0 +1,80 @@
...
+    opacity: 1;
+    transform: scale(1);
+  }
+
+  &.popover-exit-active {
+    opacity: 0;
+    transform: scale(0.88);
+    transition: 50ms cubic-bezier(0.25, 0.01, 0.25, 1);
+  }
+`
```
kdembler: Those transition times seems to be different than what we had previously for the context menu. Why the change?

src/shared/components/PopoverBase/PopoverBase.tsx
kdembler: We talked about this during our call last week :P The Figma component will get another name, it's temporary only. The Github issue you've linked mentions the correct names for the implementation

src/shared/components/PopoverBase/PopoverBase.tsx
eldiegod: for some reason, the exit anim was feeling very slow, not sure if it was a tippy thing so I made it faster

src/shared/components/PopoverBase/PopoverBase.tsx
eldiegod: hmm, true, since I got sick my memory got deleted lol

src/shared/components/PopoverBase/PopoverBase.tsx
kdembler: Let's use the previous timing and easings, we can update if it looks bad

src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -147,7 +147,8 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
   onDeleteVideoClick,
   isPullupDisabled,
 }) => {
-  const { openContextMenu, contextMenuOpts } = useContextMenu()
+  const contextMenuInstanceRef = useRef<TippyInstance>()
```
kdembler: We cannot require every consumer of `ContextMenu` to maintain its own instance of tippy. Let's change the `ContextMenu` to take `items` prop, for example:
```
type ContextMenuItem = {
  icon?: React.ReactNode
  title: string
  onClick?: () => void
```

Then `onClick` handlers can be managed by `ContextMenu` itself - this way we can trigger the tippy close without bothering the consumer of `ContextMenu`

kdembler (changes requested):

src/shared/components/VideoTileBase/VideoTileBase.tsx
eldiegod: I see what you mean, but in this case I think either way works as I don't expect `ContextMenu` to be used anywhere else than the videoTileBase.

PopoverDialog would be another discussion but I do think is needed in there for finetuning

kdembler: > I don't expect ContextMenu to be used anywhere else than the videoTileBase

I'm not so sure about it, but even if it was used only there, that's still a good reason to improve its API.

kdembler (approved):

PR #1392, opened 2021-09-22
Show input helper dynamically

### LINKED ISSUE #1348: Form fields should occupy more space when helper text(also warning, error) is showing.
For now, we preserve space for helper texts in our inputs, but after a short discussion in https://github.com/Joystream/atlas/pull/1317 we decided to change that. 

We should update our `InputBase`(we are setting the default value to helperText - `helperText = '\u00A0',`) and also make sure that the space between inputs in our forms(create/edit video, create member and create/edit channel) didn't change when there is no helper text. So for example, we should still keep 40px space between inputs in create/edit video form.

rafalpawlow:
Closes #1348

### DIFF
--- a/src/shared/components/InputBase/InputBase.styles.ts
+++ b/src/shared/components/InputBase/InputBase.styles.ts
@@ -19,6 +19,11 @@ export const FormGroup = styled.label<FormGroupProps>`
   }
 
   input,
+  textarea {
+    color: ${colors.gray[50]};
+  }
+
+  input,
   button,
   textarea {
     transition: border ${transitions.timings.regular} ${transitions.easing};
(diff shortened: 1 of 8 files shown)


src/shared/components/InputBase/InputBase.styles.ts
```diff
@@ -27,6 +27,11 @@ export const FormGroup = styled.label<FormGroupProps>`
     font-size: ${typography.sizes.body1};
     line-height: 28px;
 
+    input,
+    textarea {
+      color: ${colors.gray[50]};
+    }
+
```
drillprop: Text in inputs is black now. 
image
you should move this block before 
```  
  input,
  button,
  textarea {
...
```

drillprop (changes requested):
One thing, besides that looks good

src/shared/components/InputBase/InputBase.styles.ts
wyrdrender: Crap, I did it because of stylelint.

drillprop (approved):

PR #1396, opened 2021-09-23
Remove size variant in TitleArea

### LINKED ISSUE #1393: Use breakpoints instead of size prop in TitleArea

drillprop:
Fix #1393 

I also decided to install `react-textarea-autosize`. I just didn't like how we handled autosizing previously. 
it's a small library that just works out of the box https://bundlephobia.com/package/react-textarea-autosize@8.3.3

### DIFF
--- a/src/shared/components/TitleArea/TitleArea.style.tsx
+++ b/src/shared/components/TitleArea/TitleArea.style.tsx
@@ -1,4 +1,5 @@
 import styled from '@emotion/styled'
+import TextareaAutosize from 'react-textarea-autosize'
 
 import { colors, media, sizes, typography } from '@/shared/theme'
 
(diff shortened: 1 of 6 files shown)


src/shared/components/TitleArea/TitleArea.style.tsx
```diff
@@ -35,14 +29,10 @@ export const StyledTextArea = styled.textarea<StyledTextAreaProps>`
...
-    if (variant === 'large') {
-      return typography.sizes.h4
-    }
-    if (variant === 'small') {
-      return typography.sizes.h2
-    }
-  }};
+  font-size: ${typography.sizes.h4};
+  ${media.lg} {
+    font-size: ${typography.sizes.h2};
```
rafalpawlow, line 35: What about line height?

drillprop: that's a good point, fixed.

rafalpawlow (approved):

Issue #1378, opened 2021-09-21
Create PopoverDialog component


kdembler:
Create a component using a `Popover` basis that has sections for header/footer/etc

Figma: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=510%3A97749

PR #1397, opened 2021-09-23
Popover Dialog component

eldiegod:
First merge: https://github.com/Joystream/atlas/pull/1389

fixes: #1378


src/shared/components/Popover/PopoverDialog.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+
+import { PopoverDialog } from './PopoverDialog'
+
+import { Button } from '../Button'
+import { Text } from '../Text'
+
+export default {
+  title: 'Shared/P/Popover',
```
kdembler:
```suggestion
  title: 'Shared/P/PopoverDialog',
```

src/shared/components/Popover/PopoverDialog.styles.tsx
```diff
@@ -0,0 +1,51 @@
+import styled from '@emotion/styled'
+
+import { colors, sizes } from '@/shared/theme'
+
+import { Popover } from './Popover'
+
+import { Text } from '../Text'
+
+export const MAX_CONTENT_HEIGHT = 219
```
kdembler: Where is this magic number coming from?

src/shared/components/Popover/PopoverDialog.styles.tsx
```diff
@@ -0,0 +1,51 @@
...
+  isScrollable: boolean
+}
+
+export const HeaderContainer = styled(Text)<Scrollable>`
+  ${({ isScrollable }) =>
+    isScrollable &&
+    `
+    padding-bottom: ${sizes(4)};
+    border-bottom: 1px solid ${colors.gray[600]};
+  `}
```
kdembler: If we need to conditionally include a block of CSS like this, let's extract the relevant styles to a separate constant. That way we get proper CSS intellisense&linting, I don't think the tooling will be able to understand this string. 

Example:
```tsx
const headerContainerScrollableStyles = css`
  padding-bottom: ${sizes(4)};
  border-bottom: 1px solid ${colors.gray[600]};
`

export const HeaderContainer = styled(Text)<Scrollable>`
  ${({ isScrollable }) => isScrollable && headerContainerScrollableStyles};
`

**Review by kdembler — CHANGES_REQUESTED** [review]

File: src/shared/components/Popover/PopoverDialog.styles.tsx (hunk 2 above)
Comment (eldiegod) [reply src/shared/components/Popover/PopoverDialog.styles.tsx]: an attempt at replicating design: 
https://www.figma.com/file/3uFQ2y1jlS6fbp7Ac0yIpv/[-12]-Video-category-page-flows?node-id=304%3A40239

Comment (kdembler) [reply src/shared/components/Popover/PopoverDialog.styles.tsx]: The design doesn't define this value so it shouldn't be used. The design only defines the maximum height for the entire popover dialog which is 320px. But the content height will depend on whether there is a header/footer, used padding etc. We cannot hardcode it, we need to find some different approach

**Review by kdembler — CHANGES_REQUESTED** [review]

File: src/shared/components/Popover/PopoverDialog.styles.tsx [hunk 4]
```diff
@@ -0,0 +1,52 @@
...
+  grid-auto-columns: max-content;
+  justify-content: end;
+  gap: ${sizes(2)};
+
+  ${({ isScrollable }) =>
+    isScrollable &&
+    `
+    padding-top: ${sizes(4)};
+    border-top: 1px solid ${colors.gray[600]};
+  `}
```
kdembler: Same case here with conditional css block

src/shared/components/Popover/PopoverDialog.styles.tsx
eldiegod: Had to do some black magic but it is like design now

src/shared/components/Popover/Popover.tsx
```diff
@@ -53,9 +55,11 @@ export const Popover: React.FC<PopoverProps> = ({
...
       }}
       render={(attrs) => (
-        <ContentContainer {...attrs} className={className}>
-          {content}
-        </ContentContainer>
+        <div ref={contentContainerRef}>
+          <ContentContainer {...attrs} className={className}>
+            {content}
+          </ContentContainer>
+        </div>
```
kdembler: Why not apply `ref` directly to `ContentContainer`?

eldiegod: The transform css animation mess up the meassuring

kdembler (changes requested):
Like mentioned over RocketChat, let's remove the scroll measuring and replace that with static `dividers` prop

eldiegod:
> Like mentioned over RocketChat, let's remove the scroll measuring and replace that with static `dividers` prop

@kdembler  should we call the prop scrollable instead? think it makes more sense

kdembler:
Explained in the RocketChat, content should be scrollable either way, `dividers` prop is supposed to only add visual dividers, nothing else

kdembler (approved):

PR #1402, opened 2021-09-23
Add link to Pioneer member page in channel details

rafalpawlow:


src/views/viewer/ChannelView/ChannelAbout.tsx
```diff
@@ -21,6 +22,7 @@ export const ChannelAbout = () => {
   const { id } = useParams()
   const { channel } = useChannel(id)
   const { videoCount } = useChannelVideoCount(id)
+  const pioneerUrl = import.meta.env.VITE_PIONEER_URL
```
kdembler: Let's move that into `config/urls`

src/config/urls.ts
```diff
@@ -15,3 +15,5 @@ export const STORAGE_URL_PATH = 'asset/v0'
 export const COVER_VIDEO_INFO_URL = 'https://eu-central-1.linodeobjects.com/atlas-hero/cover-info.json'
 
 export const JOYSTREAM_DISCORD_URL = 'https://discord.gg/DE9UN3YpRP'
+
+export const PIONEER_URL = import.meta.env.VITE_PIONEER_URL
```
kdembler: Why aren't you using `readEnv` for that like the rest of URLs

kdembler (changes requested):

src/config/urls.ts
rafalpawlow: Because `redEnv` returns variables for `production`, `development` or `mocking` and I declared `PIONNER_URL` without environment prefix since it's independent of the environment. Maybe we should declare it the same way as we did for discord by adding string to `urls.ts` instead of `.env`?

kdembler: You're right of course. I agree, let's put the URL inside the file itself :+1:

kdembler (approved):

PR #1414, opened 2021-09-27
Fix loading tile visibility when switching tabs

### LINKED ISSUE #1411: Loading video tile should not be visible when user is switching my videos tabs
Take a look at the video

https://user-images.githubusercontent.com/51168865/134899242-e3a248dd-66d2-4df3-a1d0-48888947412f.mov

rafalpawlow:
Fixes #1411

### DIFF
--- a/src/views/studio/MyVideosView/MyVideosView.tsx
+++ b/src/views/studio/MyVideosView/MyVideosView.tsx
@@ -254,7 +254,7 @@ export const MyVideosView = () => {
         }
         return (
           <VideoTilePublisher
-            key={idx}
+            key={`video-tile-${video.id || idx}`}
             id={video.id}
             showChannel={false}
             isPullupDisabled={!!videoTabs.find((t) => t.id === video.id)}
--- a/src/views/studio/MyVideosView/MyVideosView.tsx
+++ b/src/views/studio/MyVideosView/MyVideosView.tsx
@@ -229,11 +229,11 @@ export const MyVideosView = () => {
         .slice(videosPerPage * currentPage, currentPage * videosPerPage + videosPerPage)
         .map((draft, idx) => {
           if (draft === 'new-video-tile') {
-            return <NewVideoTile loading={loading} key={idx} onClick={() => addVideoTab()} />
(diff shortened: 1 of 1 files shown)


drillprop:
This does not entirely fix the issue. If you're uploading a private video and you switch to the public tab, you'll experience the similar problem

https://user-images.githubusercontent.com/51168865/135059215-c4d9d0d9-95b7-4a84-a58e-5db3a471ef86.mov

drillprop (approved):
Looks good to me!

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -254,7 +254,7 @@ export const MyVideosView = () => {
         }
         return (
           <VideoTilePublisher
-            key={idx}
+            key={`video-tile-${video.id || idx}`}
```
kdembler: I think this may have some undesired consequences when video IDs are matching some of the indexes. Check this out:

https://user-images.githubusercontent.com/12646744/135095096-ff46b46a-dc0a-4259-838f-39836269b9fe.mov

Can you see the `banana` video flashing for a second when switching the tabs? It's because it has an ID of `1` so React tries to reuse that same tile when loading the tile with index `1`. I think this will probably help but you need to test it out:
```suggestion
            key={video.id ? `video-id-${video.id}` : `video-idx-${idx}`}
```

kdembler (changes requested):

src/views/studio/MyVideosView/MyVideosView.tsx
rafalpawlow: @kdembler Yeah, you are right, good catch 👍. By the way, we have more places in app where we use `index` as key. What do you think we fix all these places and add eslint rule for that?

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

kdembler: Not sure, I think the rule may be a bit too generic, I think there are still some places in the app where it makes sense to use index as the key, I think we just shouldn't mix those with IDs

kdembler (approved):

PR #1417, opened 2021-09-27
Change default license to CC0

### LINKED ISSUE #1409: Default license should be CC0, not PDM
as the title says. Also, the attribution text field should have the `Optional` badge

drillprop:
fix #1409

### DIFF
--- a/src/providers/editVideoSheet/hooks.tsx
+++ b/src/providers/editVideoSheet/hooks.tsx
@@ -15,6 +15,8 @@ import { EditVideoAssets, EditVideoFormFields, EditVideoSheetState, EditVideoShe
 import { channelDraftsSelector, useDraftStore } from '../drafts'
 import { useAuthorizedUser } from '../user'
 
+export const DEFAULT_LICENSE_ID = 1002
+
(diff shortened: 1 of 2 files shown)


src/providers/editVideoSheet/hooks.tsx
```diff
@@ -70,7 +70,7 @@ export const useEditVideoSheetTabData = (tab?: EditVideoSheetTab) => {
     title: tab.isDraft ? draft?.title ?? '' : video?.title ?? '',
     description: (tab.isDraft ? draft?.description : video?.description) ?? '',
     category: (tab.isDraft ? draft?.category : video?.category?.id) ?? null,
-    licenseCode: (tab.isDraft ? draft?.licenseCode : video?.license?.code) ?? 1001, // PDM as a default license
+    licenseCode: (tab.isDraft ? draft?.licenseCode : video?.license?.code) ?? 1002,
```
kdembler: Could we maybe stick that into some constant like `DEFAULT_LICENSE_ID` to avoid having this repeated?

kdembler (approved):

PR #1432, opened 2021-09-29
Implement bottom navigation for mobile devices

### LINKED ISSUE #1420: Bottom navigation
Zrzut ekranu 2021-09-28 o 14 13 32
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1341%3A94962

rafalpawlow:
Closes #1420 
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1341%3A94962

### DIFF
--- a/src/components/BottomNav/BottomNav.style.ts
+++ b/src/components/BottomNav/BottomNav.style.ts
@@ -2,6 +2,7 @@ import isPropValid from '@emotion/is-prop-valid'
 import styled from '@emotion/styled'
 import { Link, PathMatch } from 'react-router-dom'
 
+import { Text } from '@/shared/components/Text'
 import { colors, sizes, typography, zIndex } from '@/shared/theme'
 import { animation } from '@/shared/theme/tokens'
 
@@ -14,16 +15,16 @@ export const Container = styled.div`
   bottom: 0;
   z-index: ${zIndex.nearSheetOverlay};
   display: flex;
-  transition: all ${animation.medium.timing} ${animation.medium.easing};
+  transition: transform ${animation.medium.timing} ${animation.medium.easing};
 
   &.bottom-nav-active,
   &.bottom-nav-exit {
-    bottom: 0;
+    transform: translateY(0);
   }
 
   &.bottom-nav-exit-active,
   &.bottom-nav-enter {
-    bottom: -64px;
+    transform: translateY(100%);
   }
 `
 
@@ -36,17 +37,14 @@ export const NavLink = styled(Link, { shouldForwardProp: isPropValid })<{ active
   display: flex;
   flex-direction: column;
   justify-content: center;
+  align-items: center;
   padding: 0 ${sizes(2.5)};
   flex: 1;
 
   ${({ active }) => active && `background-color: ${colors.transparentPrimary[18]}`};
-
-  > svg {
-    margin: 0 auto;
-  }
 `
 
-export const NavTitle = styled.span`
+export const NavTitle = styled(Text)`
   min-width: 56px;
   margin-top: ${sizes(1)};
 `
--- a/src/components/BottomNav/BottomNav.style.ts
+++ b/src/components/BottomNav/BottomNav.style.ts
@@ -1,0 +1,52 @@
+import isPropValid from '@emotion/is-prop-valid'
+import styled from '@emotion/styled'
+import { Link, PathMatch } from 'react-router-dom'
+
+import { colors, sizes, typography, zIndex } from '@/shared/theme'
+import { animation } from '@/shared/theme/tokens'
+
+export const Container = styled.div`
+  width: 100%;
+  height: 64px;
+  background-color: ${colors.gray[800]};
+  position: fixed;
+  left: 0;
+  bottom: 0;
+  z-index: ${zIndex.nearSheetOverlay};
+  display: flex;
+  transition: all ${animation.medium.timing} ${animation.medium.easing};
+
+  &.bottom-nav-active,
+  &.bottom-nav-exit {
+    bottom: 0;
+  }
+
+  &.bottom-nav-exit-active,
+  &.bottom-nav-enter {
+    bottom: -64px;
+  }
+`
+
+export const NavLink = styled(Link, { shouldForwardProp: isPropValid })<{ active: PathMatch | null }>`
+  color: ${colors.white};
+  font-size: ${typography.sizes.button.small};
+  text-align: center;
+  text-decoration: none;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  padding: 0 ${sizes(2.5)};
+  flex: 1;
+
+  ${({ active }) => active && `background-color: ${colors.transparentPrimary[18]}`};
+
+  > svg {
+    margin: 0 auto;
+  }
+`
+
+export const NavTitle = styled.span`
+  min-width: 56px;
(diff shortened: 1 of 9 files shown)


src/components/BottomNav/BottomNav.style.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  &.bottom-nav-active,
+  &.bottom-nav-exit {
+    bottom: 0;
+  }
+
+  &.bottom-nav-exit-active,
+  &.bottom-nav-enter {
+    bottom: -64px;
+  }
```
kdembler: Let's use `transform: translateY()` instead, it has much better transition performance than `bottom`

src/components/BottomNav/BottomNav.style.ts
```diff
@@ -0,0 +1,52 @@
...
+  transition: all ${animation.medium.timing} ${animation.medium.easing};
+
+  &.bottom-nav-active,
+  &.bottom-nav-exit {
+    bottom: 0;
+  }
+
+  &.bottom-nav-exit-active,
+  &.bottom-nav-enter {
+    bottom: -64px;
```
kdembler: With `translate` you can just use `-100%`

src/components/BottomNav/BottomNav.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { CSSTransition } from 'react-transition-group'
+
+import { navItems } from '@/config/nav'
+import { transitions } from '@/shared/theme'
+
+import { Container, NavLink, NavTitle } from './BottomNav.style'
+
+export const BottomNav = () => {
+  const [open, setOpen] = useState(false)
+  const match = useMatch
```
kdembler: What's the point of this?

src/config/nav.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { absoluteRoutes } from '@/config/routes'
+import { SvgNavChannels, SvgNavHome, SvgNavNew, SvgNavPopular } from '@/shared/icons'
+
+export const navItems = [
```
kdembler: Let's keep the prefix 🙏  e.g. `viewerNavItems`

kdembler (changes requested):
I think the UX could be improved a bit:
1) First off, I think we may have the scroll direction the other way around. I'd expect that when I open the app and start scrolling down (the page content goes up), the navigation disappears. As long as I go down and scroll further content it remains hidden. Only if I go back and scroll up, it appears back. I'm pretty sure that's what Adam had in mind as well, although I see that docs in Figma can be kind of misleading.
2) Another thing, I think the navigation should be visible right after you open the app on mobile and disappear only once you start scrolling. Also, when I navigate to another page it hides after the navigation. I believe it should be visible (for example to let the user navigate further) and only hide once scrolled.

src/components/BottomNav/BottomNav.style.ts
```diff
@@ -0,0 +1,52 @@
...
+  flex-direction: column;
+  justify-content: center;
+  padding: 0 ${sizes(2.5)};
+  flex: 1;
+
+  ${({ active }) => active && `background-color: ${colors.transparentPrimary[18]}`};
+
+  > svg {
+    margin: 0 auto;
+  }
```
drillprop: I think we could remove this block and use `align-items: center` instead

src/components/BottomNav/BottomNav.tsx
```diff
@@ -0,0 +1,54 @@
...
+      timeout={parseInt(transitions.timings.routing)}
+      classNames="bottom-nav"
+      unmountOnExit
+      mountOnEnter
+    >
+      <Container>
+        {navItems.map((item) => (
+          <NavLink to={item.to} key={`bottomLink-${item.to}`} active={match(item.to)}>
+            {item.icon}
+            <NavTitle>{item.name}</NavTitle>
```
drillprop: Font is different in the design: 
image 
You should use our `Text` with `button3 ` variant here.

But... on the other hand, we could stick to `Inter` here, because it looks weird when I applied `PxGrotesk`. Let's ask designer about this, maybe?

drillprop: Ok, talked to Adam.  You should change this to `PxGrotesk`, also you should update the font in desktop navigation( you'll need to update `::after` pseudo-element probably)

toiletgranny: @drillprop knowing that soon we'll be replacing `PxGrotesk` with a completely different typeface altogether, I'd suggest sticking with `PxGrotesk` for now, to keep the implementation and design consistent when this happens.

toiletgranny:
100% agree with what @kdembler said.

> 1. First off, I think we may have the scroll direction the other way around.

That's 100% correct: the bottom navigation bar should hide upon scrolling downwards and reappear upon scrolling upwards. Sorry if the Figma documentation was misleading. 🙈 

> 2. Another thing, I think the navigation should be visible right after you open the app on mobile and disappear only once you start scrolling. Also, when I navigate to another page it hides after the navigation. I believe it should be visible (for example to let the user navigate further) and only hide once scrolled.

100% agree with this. Ideally, the bottom navigation bar is visible by default and gets hidden upon scrolling down.

Other than this, now that I've played with it myself I'm not sure if the `128px` threshold is not too generous. Perhaps we could try with something like this:
1.  Upon scrolling `64px` downwards ⬇️ , the bottom navigation bar gets hidden. 
2. Upon scrolling `24px` upwards ⬆️  , the bottom navigation bar reappears.

src/components/BottomNav/BottomNav.tsx
rafalpawlow: This is a workaround to use hook inside render

drillprop (approved):
LGTM!

src/components/BottomNav/BottomNav.tsx
kdembler: I see now. However, I think this isn't a good idea. By "workaround" I guess you mean workaround for the eslint hooks rule. It's not there without a reason though. Rules of hooks say we cannot call hooks conditionally and that's what we're doing here by triggering inside a loop. You can create a wrapper for `NavLink` instead that will call the hook and manage it's `active` state autonomously

kdembler (commented):
Looks good to me other than this one comment. However, let's also get an approval from @toiletgranny before merging

toiletgranny:
Can we have the `scrollRange` set to `24px` on both scroll directions? Other than this, looks fine by me. :)

kdembler:
After some more testing I've noticed that sometimes when I scroll to the top of the page, the navigation will disappear even though I haven't scrolled down. Here's an example video:

https://user-images.githubusercontent.com/12646744/135816841-ac6adb68-ae81-4972-bd67-844657210327.MP4


Are you able to replicate this? Maybe this is because of the scroll "bounce" off top?

rafalpawlow:
> After some more testing I've noticed that sometimes when I scroll to the top of the page, the navigation will disappear even though I haven't scrolled down. Here's an example video:
> 
>  RPReplay_Final1633335141.MP4 
> Are you able to replicate this? Maybe this is because of the scroll "bounce" off top?

@kdembler Yeah I was able to replicate this on iOS in both directions. I made a fix for this.

kdembler (approved):
Great stuff, works really well!

Issue #1382, opened 2021-09-21
Improve the VideoHero layout


kdembler:
Implement the new rules for layout and size described here: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=489%3A88635

Issue #1425, opened 2021-09-28
Ideas for community-managed featured content in Atlas


kdembler:
## The goal

The goal of this issue is to describe possible solutions for letting the Joystream community manage some featured content inside Atlas. This includes the existing video hero, but also features yet to be implemented, like the featured videos for video category. Managing all of that curated content by the dev team isn't really feasible for a long term and it's a great place for community contribution, we just need to enable it.

### Static hosted JSON files

This is the approach we use today for the main page video hero. We have a Jsgenesis-owned bucket storage where we keep a single JSON file that describes the video hero - it provides the ID of the main video, title to be displayed in the hero and also a URL for the cut video file to be played in background. We could leverage the same approach for other data, like featured videos for a specific category. There could be a single/multiple JSON file(s) describing the list of videos, in the same manner as we do with the video hero. Community members could be given access to update those files and Atlas would display the updated data. The issue with this approach is the access control. Because of the authentication of the bucket storage, it's not easy to gain/revoke access for community members. Access is done via a secret key that needs to be generated by the server's admin (Mokhtar). So the easiest approach would be to generate one secret for the community, but then this secret is valid forever until someone manually revokes it. So if there was a single role in the community responsible for this, whenever the person fulfilling this role changes, we should revoke&generate new key to ensure only one person has access. This is cumbersome because we need to ask Mokhtar.

#### Advantages
- Easy to build

#### Disadvantages 
- Problematic access control

### Gateway-like featured content server

Another way we could go about this is to build a very basic version of a gateway node that would keep info about featured content inside its database. Along with some simple authentication mechanism, the featured content could be updated via an API and the access control could be fine-tuned to the project needs. Atlas would then query this data via GraphQL, like it does with all the other data sources (query node & Orion). However, assuming some featured videos would also need to have a custom short cut of the video, we still need some kind of storage to let the community upload those files, as doing file hosting as part of this server wouldn't be really feasible.

#### Advantages
- We take the first iterative step in building the gateway node
- We use the same approach to fetch the data as with the other data sources
- We're closer to the final solution (not as temporary as static JSON files)
- Could use account-based access control

#### Disadvantages
- Needs to be built (although I think we could reuse some parts of Orion so not a giant effort, I estimate something like 1-2 days)
- We still need some kind of storage for the video files so it only solves the metadata part of the problem

### Summary

[...]

PR #1433, opened 2021-09-29
update VideoHero component

kdembler:
Resolves https://github.com/Joystream/atlas/issues/1382

Summary of changes:
- `VideoHero` is now dumb reusable component that only displays data, `HomeView` will now fetch the data
- added a starting point in `@/api/featured` for new hooks that will fetch featured content that's not hosted on chain, for more details see #1425
- improved the layout to match the latest designs


src/components/VideoHero/VideoHero.tsx
```diff
@@ -51,69 +48,71 @@ export const VideoHero: React.FC = () => {
...
+                <Link to={absoluteRoutes.viewer.video(videoHeroData.video.id)}>
+                  <Text variant={isCompact ? 'h2' : 'h4'}>{videoHeroData.heroTitle}</Text>
+                </Link>
+              ) : isCompact ? (
+                <SkeletonLoader height={48} width={360} />
               ) : (
                 <>
-                  <TitleSkeletonLoader height={34} />
-                  <TitleSkeletonLoader height={34} />
+                  <SkeletonLoader height={30} width="100%" bottomSpace={4} />
```
drillprop: I think we could add SwitchTransition here.

src/components/VideoHero/VideoHero.tsx
```diff
@@ -51,69 +48,71 @@ export const VideoHero: React.FC = () => {
...
                 </>
               )}
             </TitleContainer>
           </GridItem>
         </LayoutGrid>
-        <ButtonsSpaceKeeper>
-          {coverVideo && (
-            <ButtonsContainer>
+        <ButtonsContainer>
+          {videoHeroData ? (
```
drillprop: If we don't find `videoHeroData` or an error occurs we will show infinitely skeleton loaders, right? Do we want that? I mean, I'm not asking you to correct here something, I'm asking because I'm not sure what should we do in this scenario.

src/components/VideoHero/VideoHero.tsx
```diff
@@ -51,69 +48,71 @@ export const VideoHero: React.FC = () => {
...
-          avatarSize="small"
-        />
         <LayoutGrid>
           <GridItem colSpan={{ xxs: 12, xs: 10, sm: 6, md: 5, xl: 4, xxl: 3 }}>
+            <StyledChannelLink
+              variant="secondary"
+              id={videoHeroData?.video.channel.id}
+              overrideChannel={videoHeroData?.video.channel}
+              avatarSize="small"
+            />
```
drillprop: image
Margin is little different here

src/components/VideoHero/VideoHero.tsx
kdembler: This behaviour is currently undefined in the design. One thing we do in the logic part is that we will fallback to local JSON file if we fail to fetch so the infinite loading would happen only if something went really bad. I think we can live with that

src/components/VideoHero/VideoHero.tsx
kdembler: Added

src/components/VideoHero/VideoHero.tsx
kdembler: Fixed the margin and also the avatar size

drillprop (approved):

PR #1448, opened 2021-10-04
Video category hero

### LINKED ISSUE #1384: Create VideoCategoryHero component
Add options to `VideoHero` to:
- overwrite height
- pass top navigation ReactNode
- pass bottom ReactNode where will showcase the video video hero slider

Add logic for switching the playing video, etc

drillprop:
Fix #1384

### DIFF
--- a/src/components/VideoCategoryHero/VideoCategoryHero.tsx
+++ b/src/components/VideoCategoryHero/VideoCategoryHero.tsx
@@ -1,0 +1,79 @@
+import styled from '@emotion/styled'
+import { round } from 'lodash'
+import React, { ReactNode, useState } from 'react'
+
+import { VideoHeroData } from '@/api/featured'
+import { IconButton } from '@/shared/components/IconButton'
+import { SkeletonLoader } from '@/shared/components/SkeletonLoader'
+import { SvgGlyphChevronLeft } from '@/shared/icons'
+import { media } from '@/shared/theme'
+
+import { Divider, VideoHeroHeader, VideoHeroHeaderTitle } from './VideoCategoryHero.style'
+import { VideoHeroSlider } from './VideoHeroSlider'
+
+import { VideoHero } from '../VideoHero'
+
+export type VideoCategoryHeroProps = {
+  header: {
+    icon: ReactNode
+    title: string
+  }
+  videos?: (VideoHeroData | null)[]
+}
+
+export const VideoCategoryHero: React.FC<VideoCategoryHeroProps> = ({ header, videos }) => {
+  const [activeVideoIdx, setActiveVideoIdx] = useState(0)
+  const [videoProgress, setVideoProgress] = useState(0)
+
+  const handleVideoClick = (idx: number) => {
+    setActiveVideoIdx(idx)
+  }
+
+  const handleTimeUpdate = (e: React.SyntheticEvent<HTMLVideoElement, Event>) => {
+    const currentTime = e.currentTarget.currentTime
+    const duration = e.currentTarget.duration
+    const progressInPercentage = round((currentTime / duration) * 100, 2)
+    setVideoProgress(progressInPercentage)
+  }
+
+  return (
+    <StyledVideoHero
+      onTimeUpdate={handleTimeUpdate}
+      videoHeroData={videos ? videos[activeVideoIdx] : null}
+      headerNode={
+        <VideoHeroHeader>
+          {header ? (
+            <IconButton variant="tertiary">
+              <SvgGlyphChevronLeft />
+            </IconButton>
+          ) : (
+            <SkeletonLoader rounded height={40} width={40} />
+          )}
+          <Divider />
+          {header ? (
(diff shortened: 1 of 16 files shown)


drillprop:
Adam gave me some feedback: 
On mobile:
- active progress bar should have a different color 
- autoplay doesn't work for some reason
- height should be different
- switching videos doesn't work correctly

I'll set this to a draft for now, until I fix this those

src/components/VideoCategoryHero/VideoCategoryHero.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const [activeVideoIdx, setActiveVideoIdx] = useState(0)
+  const [videoProgress, setVideoProgress] = useState(0)
+
+  const handleVideoClick = (idx: number) => {
+    setActiveVideoIdx(idx)
+    setVideoProgress(0)
+  }
+
+  const handleEnded = () => {
+    const idx = activeVideoIdx + 1 >= 3 ? 0 : activeVideoIdx + 1
```
kdembler: We hardcode number of  videos here to `3`, let's try to make this component work with any number of videos

src/components/VideoCategoryHero/VideoHeroSlider/VideoHeroSlider.style.ts
```diff
@@ -0,0 +1,85 @@
...
+export const VideoHeroSliderWrapper = styled.div`
+  position: absolute;
+  margin: ${sizes(4)};
+  top: 0;
+  left: 0;
+  display: grid;
+  width: calc(100% - ${sizes(8)});
+  height: 100%;
+  grid-gap: ${sizes(2)};
+  grid-template-columns: repeat(3, 1fr);
```
kdembler: We're assuming 3 videos here as well, right?

kdembler (commented):
Looking nice overall. Minor comments and also those visual ones:

1. I might be imagining things but it seems to me that you can see pause between single steps of this progress indicator? I mean that it's not a constant movement but rather quick repeat of (move, wait). Maybe the progress update is not happening often enough? This is really minor and it's possible a non-issue so if you don't see the issue or the solution doesn't come to mind right away let's ignore it



https://user-images.githubusercontent.com/12646744/136266567-853b66dd-e048-4980-88a2-154f370bf7b4.mov


2. This is actually a design issue, not the implementation (ping @toiletgranny) - on mobile, since the video in the hero section changes, we can experience some layout shifts. When the title of the video changes from 1 line to 2 lines, all of the content below will shift slightly. This can be really infuriating if you're scrolling videos below. Maybe we can agree that we will show at most 2 lines of title and set the height of the title element to be fixed to that 2 lines of text?

https://user-images.githubusercontent.com/12646744/136266507-e84af0ca-7d72-4a8f-8894-d7819f2f1774.mov

drillprop:
> 1. I might be imagining things but it seems to me that you can see pause between single steps of this progress indicator? I mean that it's not a constant movement but rather quick repeat of (move, wait). Maybe the progress update is not happening often enough? This is really minor and it's possible a non-issue so if you don't see the issue or the solution doesn't come to mind right away let's ignore it

This is because I'm relying here on `timeupdate` event, which is not happening very frequently. To fix this I would have to use `setInterval` and check the video time more often(like I do in `CustomTimeline`). I mean, I could try that, but I'm a little worried that this can impact performance.

kdembler:
Let's ignore this then

toiletgranny:
> This is actually a design issue, not the implementation (ping @toiletgranny) - on mobile, since the video in the hero section changes, we can experience some layout shifts. When the title of the video changes from 1 line to 2 lines, all of the content below will shift slightly. This can be really infuriating if you're scrolling videos below. Maybe we can agree that we will show at most 2 lines of title and set the height of the title element to be fixed to that 2 lines of text?

A great catch, thanks!

Your idea is certainly a quick and straightforward solution to this problem. What I fear slightly, in this case, is having videos with long titles, possibly wrapping into more than 2 lines of text. Is my understanding correct, that these titles would be simply truncated, possibly ending with an ellipsis `...`? It's not ideal, but it's also not a necessarily terrible thing, either. I just want to make sure I understand this.

Alternatively, what I was hoping we could do is to experiment with capturing and preserving the user's scroll position, so that no layout shift is perceived from the user's perspective. Perhaps this could be approached in a similar fashion to what's been described [in this Medium article](https://medium.com/@mikolajkocieda/react-hooks-keep-scroll-position-during-change-list-views-cd86463f9629).

I realize the latter would certainly take more effort to implement, so I'd be fine with proceeding with the first (straightforward) solution. But perhaps this could be a fun challenge to tackle.

kdembler:
Yes, in the straightforward solution if some title needs more space than 2 lines it would be truncated with an ellipsis. I'd go this way, the other approach you suggested seems way too complicated for this usecase

kdembler (approved):

PR #1457, opened 2021-10-05
update topbar and sidenav with latest changes

kdembler:
## Summary of changes:
- Updated both topbar and sidenav with latest designs - adjusted spacing, sizes, colors, etc
- Fixed some issues with topbar jumping left and right when loading studio
- Added transition to sidenav when it's unavailable in studio due to not being logged in
- Renamed `SidenavStudio` (and similar) to `StudioSidenav`
- Removed export `TOP_NAVBAR_HEIGHT` constant, replaced with `--size-topbar-height` CSS variable
- Cleaned up naming of Joystream logo assets


src/components/TopbarViewer/TopbarViewer.tsx
```diff
@@ -104,7 +110,7 @@ export const ViewerTopbar: React.FC = () => {
             Start publishing
           </Button>
         )}
-        {!isFocused && !smMatch && (
+        {!isFocused && !mdMatch && (
           <IconButton>
```
drillprop: I'll look at the code later. Just noticed one thing - this button doesn't do anything :P Could you also add here `onClick={() => openInNewTab(absoluteRoutes.studio.index())}`?

kdembler: Yeah good catch, missed that. Those should actually be native links instead, will fix tomorrow

src/components/TopbarViewer/TopbarViewer.style.tsx
```diff
@@ -43,7 +43,7 @@ export const StyledSearchbar = styled(Searchbar)<FocusProps>`
   margin-left: auto;
```
drillprop, line 43: I cannot comment here, and I know that search is being reworked but I think we could change line 42 to: 
```tsx
width: ${({ hasFocus }) => (hasFocus ? '100%' : '39px')};
```
Right now there is unnecessary space between the input and hamburger button.
Design:
image
Ours: 
image

src/shared/components/HamburgerButton/HamburgerButton.tsx
```diff
@@ -7,11 +7,12 @@ export type HamburgerButtonProps = {
...
   onClick: (e: React.MouseEvent<HTMLElement>) => void
   outerStyles?: SerializedStyles
+  className?: string
 }
 
-export const HamburgerButton: React.FC<HamburgerButtonProps> = ({ active, onClick }) => {
+export const HamburgerButton: React.FC<HamburgerButtonProps> = ({ active, onClick, className }) => {
   return (
-    <Hamburger onClick={onClick} aria-label="Main menu" aria-expanded={active}>
+    <Hamburger onClick={onClick} aria-label="Main menu" aria-expanded={active} className={className}>
```
drillprop:
```suggestion
    <Hamburger tabIndex={1} onClick={onClick} aria-label="Main menu" aria-expanded={active} className={className}>
```
Slight improvement. This will make this button the first focusable element, when you navigating with `TAB` button.

drillprop (commented):
Looks good! Only two minor suggestions

src/components/SidenavBase/SidenavBase.style.ts
```diff
@@ -27,40 +22,29 @@ type SidebarNavLinkProps = {
...
-export const JoystreamLogo = styled(SvgJoystreamFullLogo)`
-  ${logoStyles}
-`
-
-export const StudioLogo = styled(SvgStudioFullLogo)`
-  ${logoStyles}
-`
-
 export const SidebarNavList = styled.ul`
   margin-top: 56px;
```
toiletgranny: According to the designs, this should be `margin-top: 28px;`

toiletgranny (commented):
Overall, this looks great. Colors are now aligned, "borders", too, we've get the new JS Studio logo on Publisher — 👌 .

Looking at designs, there're still some minor spacing and structure inconsistencies in list elements in the Side navigation or differences in the position of the logo, but I'm not sure if this is a good moment to tackle this. These are super minor details, don't really affect the UX, and also, I think this is best solved by having also a designer on board with this.

src/shared/components/HamburgerButton/HamburgerButton.tsx
kdembler: Done

src/components/TopbarViewer/TopbarViewer.style.tsx
kdembler: Done

src/components/TopbarViewer/TopbarViewer.tsx
kdembler: Replaced all the viewer/studio buttons with anchor elements

src/components/SidenavBase/SidenavBase.style.ts
kdembler: Yes, updated now. Also did some tweaks to spacing and alignment. Let me know if you see anything else

drillprop (approved):

Issue #1379, opened 2021-09-21
Create video category page filtering bar/dialog


kdembler:
Implement the filtering bar with all the necessary logic. This should output the arguments for the videos query. This also includes the badges on the filter buttons. Let's make it non-dependant on the category page

PR #1458, opened 2021-10-05
Discovery filters

eldiegod:
Fixes: #1379

It will probably have some details to be addressed 

Button badges courtesy of @wyrdrender


drillprop:
Still needs rebase :P

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -0,0 +1,480 @@
...
+  OtherFilterStyledText,
+  SortContainer,
+} from './CategoryVideos.styles'
+
+import { ActionDialog, ActionDialogProps } from '../ActionDialog'
+import { StyledTitleText } from '../MessageDialog/MessageDialog.style'
+
+type VideoLengthOptions = '0-to-4' | '4-to-10' | '10-to-9999'
+
+export const CategoryVideos = () => {
```
kdembler: The filtering component should be extracted and `CategoryVideos` shouldn't be a part of this PR

kdembler (changes requested):

src/components/CategoryVideos/CategoryVideos.tsx
eldiegod: Extracted filter component and added a playground lets keep CategoryVideos as it works very well as a demo for the playground for now and it doesn't have much out of itself

src/shared/components/Popover/PopoverDialog.styles.tsx
```diff
@@ -13,7 +13,12 @@ type Dividers = {
   dividers: boolean
 }
 
+type HasHeader = {
+  hasHeader: boolean
+}
+
 export const PopoverContainer = styled(Popover)`
+  padding: ${sizes(4)};
```
kdembler: This doesn't work well - the dividers will not span the full width of the popover

src/components/CategoryVideos/CategoryVideos.styles.tsx
```diff
@@ -0,0 +1,32 @@
+import styled from '@emotion/styled'
```
kdembler: Let's not include _anything_ related to video categories in this PR, this file included. If we want a place to test out the filters bar let's just create a playground with only the necessary logic there

src/shared/components/Button/Button.style.ts
```diff
@@ -84,7 +86,34 @@ export const ButtonIconWrapper = styled.span<ButtonIconWrapperProps>`
 
 export const StyledText = styled(Text)<TextProps>`
   /* compensate for line-height being 1 */
-  ${textPaddingStyles};
+  ${textPaddingStyles}
```
kdembler: why the removal of semicolon?

src/shared/components/Button/Button.style.ts
```diff
@@ -84,7 +86,34 @@ export const ButtonIconWrapper = styled.span<ButtonIconWrapperProps>`
...
+    css`
+      &::after {
+        ${square(8)}
+
+        content: '';
+        background-color: ${colors.white};
+        border-radius: 50%;
+        display: block;
+      }
+    `}
```
kdembler: Let's extract this block to a separate var for better readability 🙏

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,500 @@
...
+  const { id } = useParams()
+
+  // filters
+  const [selectedLanguage, setSelectedLanguage] = useState<string | null | undefined>('en')
+  const [sortVideosBy, setSortVideosBy] = useState<VideoOrderByInput>(VideoOrderByInput.CreatedAtDesc)
+  const [dateUploadedFilter, setdateUploadedFilter] = useState<number>()
+  const [licensesFilter, setLicensesFilter] = useState<number[]>()
+  const [videoLegnthFilter, setVideoLegnthFilter] = useState<VideoLengthOptions>()
+  const [paidPromotionalMaterialFilter, setPaidPromotionalMaterialFilter] = useState<boolean>()
+  const [matureContentRatingFilter, setMatureContentRatingFilter] = useState<boolean>()
```
kdembler: Why don't we put this state directly in the component? I think FiltersBar could manage its own state? if we do that, then all the filters don't have to pass through the consumer component

src/components/FiltersBar/FiltersBar.styles.tsx
```diff
@@ -0,0 +1,57 @@
...
+  align-items: center;
+`
+
+export const OtherFilterStyledIcon = styled(SvgGlyphHide)`
+  margin-right: ${sizes(2)};
+
+  & path {
+    fill: currentColor;
+    stroke: currentColor;
+  }
```
kdembler: I think those style cause the icon to not look as it should - compare with the design

kdembler (changes requested):

src/shared/components/Popover/PopoverDialog.styles.tsx
eldiegod: I see, fixed it now

src/components/CategoryVideos/CategoryVideos.styles.tsx
eldiegod: hmm I will remove manually the files then and later try a rebase without it

src/shared/components/Button/Button.style.ts
eldiegod: Because I was getting a red squiggly line in vscode for it but not sure why

src/components/FiltersBar/FiltersBar.tsx
eldiegod: Because all this is used to derive other state like the canClearFilters flags that it's needed on the parent common;ents for button badges

src/components/FiltersBar/FiltersBar.styles.tsx
eldiegod: not sure about that, looks like design to me with the same color as the text, maybe if u have a screenshot to compar

kdembler: Seems pretty clear to me:
Screenshot 2021-10-11 at 09 41 27
Screenshot 2021-10-11 at 09 41 07

src/components/FiltersBar/FiltersBar.tsx
kdembler: What other fields does the parent need other than `canClearFilters`? If it's only this, we could just pass a single field to the parent instead of all the state

src/shared/components/Button/Button.style.ts
kdembler: On the other hand, this breaks styles for WebStorm:
Screenshot 2021-10-11 at 09 44 11

If you look at emotion examples, they also add a semicolon after composition, so I think we should use them too: https://emotion.sh/docs/composition

src/components/FiltersBar/FiltersBar.styles.tsx
eldiegod: I see, it's the thing about having hard coded colors in svgs, fixed now

src/components/FiltersBar/FiltersBar.tsx
eldiegod: they need the state and setters for language and sortBy dropdown, for example, and the `canClearFilters`, then again the `canClearFilters` is computed from a combination of all the state talked above so don't see how it can be extracted

src/shared/components/Button/Button.style.ts
eldiegod: yeah I think its a vscode bug I readed it

src/components/FiltersBar/FiltersBar.tsx
kdembler: Both language and sortBy would be displayed outside of the `FiltersBar` though right? In that case I guess those could be handled by the consumer directly? In that case, only resulting filters and `canClearFilters` would need to be passed to the consumer?

src/shared/components/Button/Button.style.ts
```diff
@@ -88,3 +90,30 @@ export const StyledText = styled(Text)<TextProps>`
 
   color: inherit;
 `
+
+const BadgeDotStyles = css`
```
kdembler: Please don't use `PascalCase` for anything other than components

src/shared/components/Button/Button.style.ts
```diff
@@ -88,3 +90,30 @@ export const StyledText = styled(Text)<TextProps>`
...
+    border-radius: 50%;
+    display: block;
+  }
+`
+
+export const Badge = styled.div<{ dot?: boolean }>`
+  ${square(16)}
+
+  font-weight: ${typography.weights.bold};
+  font-size: 0.6rem;
```
kdembler: Let's specify explicit (in px) font size here, since the container size is specified explicitly as well

src/shared/components/Popover/PopoverDialog.styles.tsx
```diff
@@ -23,30 +27,33 @@ export const PopoverContainer = styled(Popover)`
...
 
 export const Header = styled(Text)<Dividers>`
-  padding: ${sizes(4)};
+  padding: ${sizes(4)} ${sizes(4)} 0 ${sizes(4)};
+
   ${({ dividers }) => dividers && headerDividersStyles};
 `
 
-export const ContentContainer = styled.div<Dividers>`
+export const ContentContainer = styled.div<HasHeader>`
```
kdembler: Doesn't seem like `HasHeader` type is needed anymore?

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,501 @@
...
+    <FilterContentContainer>
+      <RadioButton
+        onChange={() => {
+          setdateUploadedFilter(1)
+        }}
+        name="date-uploaded"
+        label="Last 24 hours"
+        value={1}
+        selectedValue={dateUploadedFilter}
+      ></RadioButton>
```
kdembler: If the JSX children are empty, let's not use both opening and closing tag. This can be a self-closing tag instead (`<RadioButton />`). This applies to a several tags in this file

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,501 @@
...
+        name="other-filters"
+        label="Mature content rating"
+        value={!!matureContentRatingFilter}
+      ></Checkbox>
+    </FilterContentContainer>
+  )
+
+  return (
+    <>
+      {betweenBaseAndSMMatch ? (
```
kdembler: I think it would be more readable if instead we did
```tsx
if (betweenBaseAndSMMatch) {
    return // mobile dialog
}

return // regular filters
```

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,501 @@
...
+                    >
+                      Clear
+                    </Button>
+                    <Button size="small" onClick={handleApplyFilter}>
+                      Apply
+                    </Button>
+                  </>
+                }
+              >
+                <Button badge={canClearDateUploadedFilter && 1} variant="secondary" onClick={() => null}>
```
kdembler: We shouldn't add `onClick={() => null}` on every Button here. I understand that you do this to add pointer styles but that's not the right way to go about this. It may be misleading to somebody not aware of why it's used. Maybe instead we can just add pointer styles to button regardless of whether it has `onClick` or not

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,501 @@
...
+                      variant="secondary"
+                      disabled={dateUploadedFilter === undefined}
+                    >
+                      Clear
+                    </Button>
+                    <Button size="small" onClick={handleApplyFilter}>
+                      Apply
+                    </Button>
+                  </>
+                }
```
kdembler: I think we should create some reusable component/function to handle this. We currently have a lot of duplication

kdembler (changes requested):

src/components/FiltersBar/FiltersBar.tsx
eldiegod: agree, a button should always have cursor pointer unless they are disabled

src/components/FiltersBar/FiltersBar.tsx
eldiegod: we should look into adding this eslint rule with autofix so hopefully we don't have to care about it no more

```
    'react/self-closing-comp': [
      'warn',
      {
        'component': true,
        'html': true,
      },
    ],
```

src/components/FiltersBar/FiltersBar.tsx
eldiegod: The thing is sortBy does appear on the mobile filters, language doesn't but I keep it in there because we still keep track of its state and reset it when clearing filters. 

Still not sure what are you proposing as a solution here to keep more logic on the filters bar component level if I'm understood that part correctly.

Then again, I don't think it's an inconvenient for the filters state to live in the hook as it all can be passed comfortably with a spread operator and the consumer can grab what it needs / compose custom logic on the fly

src/components/FiltersBar/FiltersBar.tsx
kdembler: Feel free to open a PR with that

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,386 @@
...
+import { VideoLengthOptions, useFiltersBar } from './useFiltersBar'
+
+import { ActionDialog, ActionDialogProps } from '../ActionDialog'
+import { StyledTitleText } from '../MessageDialog/MessageDialog.style'
+
+export const FiltersBar: React.FC<ReturnType<typeof useFiltersBar>> = ({
+  setVideoWhereInput,
+  videoWhereInput,
+  filters: {
+    setiIsFiltersOpen,
```
kdembler: Typo here: `seti`

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,386 @@
...
+    isFiltersOpen,
+    dateUploadedFilter,
+    setDateUploadedFilter,
+    matureContentRatingFilter,
+    setMatureContentRatingFilter,
+    paidPromotionalMaterialFilter,
+    setPaidPromotionalMaterialFilter,
+    videoLegnthFilter,
```
kdembler: should be `Length`

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,386 @@
...
+    setVideoLegnthFilter,
```
kdembler: should be `Length`

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,386 @@
...
+    licensesFilter,
+    setLicensesFilter,
+  },
+  canClearFilters: {
+    canClearDateUploadedFilter,
+    canClearVideoLegnthFilter,
```
kdembler: should be `Length`

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,386 @@
...
+    canClearAllFilters,
+    clearAllFilters,
+    clearDateUploadedFilter,
+    clearVideoLegnthFilter,
```
kdembler: should be `Length`

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -0,0 +1,386 @@
...
+    clearLicensesFilter,
+    clearOtherFilters,
+  },
+}) => {
+  const smMatch = useMediaMatch('sm')
+  const betweenBaseAndSMMatch = !smMatch
```
kdembler, line 56: Doesn't seem really needed? There's only one condition, it could use `!smMatch` directly right?

kdembler (changes requested):

src/components/FiltersBar/FiltersBar.tsx
eldiegod: Kept it for clarity, hard to know what breakpoint is !smMatch

kdembler (approved):

src/views/playground/PlaygroundLayout.tsx
```diff
@@ -42,6 +43,7 @@ const playgroundRoutes = [
   { path: 'orion-testing', element: <OrionTesting />, name: 'Orion testing' },
   { path: 'grid', element: <GridTesting />, name: 'Grid testing' },
   { path: 'breakpoints', element: <PlaygroundBreakpoints />, name: 'Breakpoints' },
+  { path: 'categoryvideos/1', element: <PlaygroundFilters />, name: 'Filters' },
```
kdembler:
```suggestion
  { path: 'filters', element: <PlaygroundFilters />, name: 'Filters' },
```

PR #1461, opened 2021-10-06
Fix issue with missing feedback when reuploading larger files

### LINKED ISSUE #1460: Missing feedback on reupload
@KubaMikolajczyk  When you interrupt the upload of a large file and try to do a reupload, there is no feedback that something is happening. After a long while the file starts uploading. I think we need some loader or some indicator that file waiting for upload.

https://user-images.githubusercontent.com/13706246/136208147-e52df16f-9f8f-46a9-a60d-370c2eb5ffbb.mov

drillprop:
Fix #1460

### DIFF
--- a/src/views/studio/MyUploadsView/UploadStatusGroup/UploadStatusGroup.tsx
+++ b/src/views/studio/MyUploadsView/UploadStatusGroup/UploadStatusGroup.tsx
@@ -86,7 +86,7 @@ export const UploadStatusGroup: React.FC<UploadStatusGroupProps> = ({ uploads, s
 
   const renderAssetsGroupInfo = () => {
     if (isWaiting) {
-      return 'Waiting for upload...'
+      return 'Starting upload...'
     }
     if (errorsCount) {
       return `${errorsCount} asset${errorsCount > 1 ? 's' : ''} upload failed`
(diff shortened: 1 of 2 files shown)


src/views/studio/MyUploadsView/UploadStatusGroup/UploadStatusGroup.tsx
```diff
@@ -85,6 +85,9 @@ export const UploadStatusGroup: React.FC<UploadStatusGroupProps> = ({ uploads, s
   }, [errorsCount, hasUploadingAsset, isCompleted, isProcessing, missingAssetsCount])
 
   const renderAssetsGroupInfo = () => {
+    if (isWaiting) {
+      return 'Waiting for upload...'
```
kdembler: This copy sounds to me like I'm supposed to do something as a user. Maybe instead we could use "Starting upload..."? Is this used only when we're computing the hash of the asset or in some other case as well?

drillprop: Yeah, I think "Starting upload..." will be better, and to be honest I'm not sure why we had this `isWaiting` before. It was shown for a split second when the upload was starting - i.e status was `inProgress` and progress was `0`.

kdembler (approved):

Issue #1427, opened 2021-09-28
Search box keyboard navigation


rafalpawlow:
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1341%3A93740

Issue #1426, opened 2021-09-28
Search box and input animations


rafalpawlow:
Open/closing animation for both, input and search box

https://www.figma.com/proto/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1271%3A90835&scaling=scale-down&page-id=1271%3A90759&starting-point-node-id=1271%3A90835&show-proto-sidebar=1

PR #1470, opened 2021-10-07
Global searchbox component

### LINKED ISSUE #1422: Global search box
- opening and closing logic
- recent searches
- videos
- channels
- keyboard navigation legend

https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1249%3A88605

rafalpawlow:
Closes #1422 
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1289%3A96544&viewport=429%2C48%2C0.07

Please note, there are separate tasks for opening animations (#1426) and keyboard shortcuts (#1427)


src/components/TopbarViewer/TopbarViewer.style.tsx
```diff
@@ -11,14 +11,15 @@ type FocusProps = {
 
 export const StyledTopbarBase = styled(TopbarBase)<FocusProps>`
   transition: background-color 0.4s ${transitions.easing};
-  background-color: ${(props) => (props.hasFocus ? colors.gray[900] : colors.black)};
+  background-color: ${({ hasFocus }) => (hasFocus ? colors.gray[900] : colors.black)};
+  ${({ hasFocus }) => hasFocus && `z-index: ${zIndex.globalOverlay}`};
```
drillprop, line 15: image

src/components/TopbarViewer/TopbarViewer.style.tsx
```diff
@@ -35,22 +36,26 @@ export const SearchbarContainer = styled.div`
...
+  justify-self: flex-end;
+  flex-shrink: 0;
+`
 
-  ${media.md} {
-    max-width: 480px;
-    width: 100%;
-    margin-left: 0;
-    height: initial;
+export const Overlay = styled.div`
```
drillprop, line 45: Something is not working with overlay. When you click on search input and then click with the mouse close to the search bar, the overlay doesn't disappear.
The same happens when you try navigating using TAB.
Related video: 

https://user-images.githubusercontent.com/51168865/136582247-d23b3ad8-bfa2-4f9c-b54a-c2a15a6543cc.mov

src/shared/components/Searchbar/SearchBox/Result.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  return (
+    <ResultWrapper to={to} onClick={handleResultClick}>
+      <ResultContent>
+        {isLoading ? (
+          <StyledSkeletonLoader width={video ? '64px' : '32px'} height={video ? '40px' : '32px'} rounded={!!channel} />
+        ) : (
+          <ResultThumbnail src={(video ? videoThumbnail : channelAvatar) || ''} rounded={!!channel} />
```
drillprop: We should show some placeholder when there is no avatar. 
image
image

src/shared/components/Searchbar/SearchBox/RecentSearchItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+  to: string
+  query?: string
+  title?: string
+}
+
+export const RecentSearchItem: React.FC<RecentSearchItemProps> = ({ title, onDelete, to, query }) => {
+  const highlightedTitle = useMemo(() => {
+    if (query) {
+      const regex = new RegExp(query, 'i')
+      return title ? title.replace(regex, (match) => `<span style="color: white">${match}</span>`) : ''
```
drillprop: We could use `mark` here I think. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark

src/shared/components/Searchbar/SearchBox/RecentSearchItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }, [query, title])
+
+  return (
+    <ResultWrapper to={to} onDelete={onDelete}>
+      <RecentSearchItemWrapper>
+        <ClockWrapper>
+          <SvgClock />
+        </ClockWrapper>
+        <Text secondary variant="button2">
+          {parse(highlightedTitle)}
```
drillprop: I don't like that we're using this library just o get this, but on the other hand, I don't know how to do it alternatively 😅 .
We could accomplish the same by using `dangerouslySetInnerHTML` but I think it's unsafe(as the name says) 

I'll just leave a comment here, maybe someone else has a better proposition of how to handle this case 😄

src/shared/components/Searchbar/SearchBox/Result.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const isLoading = video ? videoThumbnailLoading : channelAvatarLoading
+
+  const highlightedTitle = useMemo(() => {
+    const title = video ? video.title : channel?.title
+    if (query) {
+      const regex = new RegExp(query, 'i')
+      return title ? title.replace(regex, (match) => `<span style="color: white">${match}</span>`) : ''
+    }
+    return title || ''
+  }, [channel?.title, query, video])
```
drillprop: also, we're using this twice, we could create some helper function for this or a hook.

src/components/TopbarViewer/TopbarViewer.style.tsx
rafalpawlow: I fixed it but I'm not sure what do you mean by 'navigating using TAB'. Navigating using keyboard is not part of this PR, it will be accomplished in #1427

drillprop: Basically, if the search bar lost the focus, the overlay shouldn't be visible. 
So let's say you clicked on the search bar and then you hit the `TAB` on your keyboard. Focus is now on the `start publishing` button, but overlay doesn't go away. And I think it should disappear.

rafalpawlow: Not exactly. After I introduce keyboard navigation, you will jump to the closest result when hit Tab. You can try this in this draft: https://github.com/Joystream/atlas/pull/1473

rafalpawlow: Zrzut ekranu 2021-10-9 o 00 39 26

drillprop: Oh, I see, didn't check fully all the designs, sorry. I still don't like this. Other video websites like vimeo, youtube, or twitch doing this differently. I will talk to @toiletgranny on Monday about this and add a comment in the #1473 as well. 
We can resolve this conversation.

src/components/TopbarViewer/TopbarViewer.tsx
```diff
@@ -21,23 +21,24 @@ export const TopbarViewer: React.FC = () => {
...
+    if (location.pathname.includes(absoluteRoutes.viewer.search())) {
       setIsFocused(true)
       if (location.search) {
         const params = new URLSearchParams(location.search)
         const query = params.get(QUERY_PARAMS.SEARCH)
         setSearchQuery(query || '')
       }
     }
-  }, [isFocused, location.pathname, location.search])
+  }, [location.pathname, location.search])
```
kdembler: I don't think we still need this?

src/components/TopbarViewer/TopbarViewer.tsx
```diff
@@ -21,23 +21,24 @@ export const TopbarViewer: React.FC = () => {
...
 
+  // Lose focus on location change
   useEffect(() => {
-    // close the searchbar on external navigation
-    if (isFocused && !location.pathname.includes(absoluteRoutes.viewer.search())) {
-      setSearchQuery('')
+    if (location.pathname) {
       setIsFocused(false)
     }
+  }, [location.pathname])
```
kdembler: Likewise, I'm not sure if this is still needed

src/components/TopbarViewer/TopbarViewer.tsx
```diff
@@ -110,12 +104,13 @@ export const TopbarViewer: React.FC = () => {
             Start publishing
           </Button>
         )}
-        {!isFocused && !mdMatch && (
-          <IconButton to={absoluteRoutes.studio.index()} newTab>
+        {!searchQuery && !mdMatch && (
```
kdembler, line 122: It's the second time we use this condition right? Let's extract it to a var

src/providers/personalData/store.tsx
```diff
@@ -33,7 +33,8 @@ const WHITELIST = [
 export type PersonalDataStoreActions = {
   updateWatchedVideos: (__typename: WatchedVideoStatus, id: string, timestamp?: number) => void
   updateChannelFollowing: (id: string, follow: boolean) => void
-  updateRecentSearches: (id: string, type: RecentSearchType) => void
+  updateRecentSearches: (id: string, type: RecentSearchType, title?: string) => void
```
kdembler: I think this should be more like `addRecentSearch` right? Current name sound a bit misleading

kdembler: Also, do we still need `id` and `type` here? Previously we would remember what type of result was clicked but with the new UX, we only display previous search phrases right?

src/shared/components/Searchbar/SearchBox/RecentSearchItem.tsx
kdembler: I think we could just create a `HighlightedTitle` component and let it render the title? I think we should avoid parsing HTML

src/shared/icons/svgs/clock.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kdembler: This file should follow the naming scheme we have for other icons. Let's call this `glyph-clock.svg`

src/shared/components/Searchbar/Searchbar.style.tsx
```diff
@@ -39,25 +40,47 @@ export const CancelButton = styled(IconButton)`
...
+    width: ${hasQuery || hasFocus ? '100%' : '39px'};
+    padding-left: ${hasFocus || hasQuery ? sizes(2) : 0};
+    background-color: ${hasFocus ? colors.gray[800] : 'transparent'};
+    ${
+      !hasFocus &&
+      `
+      border: 1px solid ${colors.gray[700]};
+      margin-left: auto;`
+    };
+  `};
```
kdembler: This block will be not parsable by tooling because it's just a string. Let's extract this to some function for example like this:

```tsx
const getContainerFocusStyles = ({hasFocus, hasQuery}) => css`
  height: ${hasFocus ? '64px' : '39px'};
  // ...
`

const Container = styled.div`
  ${getContainerFocusStyles};
`

File: src/shared/components/Searchbar/Searchbar.style.tsx [hunk 13]
```diff
@@ -39,25 +40,47 @@ export const CancelButton = styled(IconButton)`
   }
 `
 
-export const Container = styled.div`
-  position: relative;
+export const Container = styled.div<{ hasFocus: boolean; hasQuery: boolean }>`
```
kdembler, line 44: Screenshot 2021-10-11 at 09 03 09

The searchbar has a 2px border now, 1px from `border` and 1px from `box-shadow`. I suggest we use only `box-shadow`

src/shared/components/Searchbar/Searchbar.style.tsx
```diff
@@ -39,25 +40,47 @@ export const CancelButton = styled(IconButton)`
...
+    width: 100%;
+    margin-left: 0;
     height: 48px;
     padding-left: ${sizes(4)};
   }
 `
 
 // TODO: remove override on viewer update
-export const StyledSvgOutlineSearch = styled(SvgOutlineSearch)`
+export const StyledSvgOutlineSearch = styled(SvgOutlineSearch, { shouldForwardProp: isPropValid })<{
```
kdembler: The icon used is a wrong one, compare with the design

src/shared/components/Searchbar/Searchbar.tsx
```diff
@@ -1,87 +1,111 @@
...
-  const mdMatch = useMediaMatch('md')
+export const Searchbar = React.forwardRef<HTMLDivElement, SearchbarProps>(
+  (
+    {
+      placeholder,
+      onChange,
+      onFocus,
+      onCancel,
+      controlled = false,
+      value: externalValue,
```
kdembler: Do we still need both the controlled and uncontrolled states? I think we only use one of them so maybe let's just remove the logic for the other one?

kdembler (changes requested):

src/providers/personalData/store.tsx
rafalpawlow: I know this is a super rare scenario but what if the channel name and the video name are the same?

src/components/TopbarViewer/TopbarViewer.tsx
kdembler: Oh wait we still have the old search results page as overlay so maybe we need this in this PR

src/providers/personalData/store.tsx
kdembler: Discussed on RocketChat, we had a slightly different understanding what the "Recent searches" are supposed to be in the new experience. This will need some changes

src/components/TopbarViewer/TopbarViewer.tsx
rafalpawlow: I think you're wrong. This condition is not repeated anywhere.

src/components/TopbarViewer/TopbarViewer.tsx
rafalpawlow: This condition is used to close the searchbar after selecting a specific video or channel.

src/components/TopbarViewer/TopbarViewer.tsx
kdembler: Yep, it's a bit different a above, nevermind :)

src/shared/components/Searchbar/Searchbar.tsx
```diff
@@ -1,87 +1,111 @@
...
+
+    const query = recentSearch || value
+
+    return (
+      <>
+        <Container className={className} hasFocus={hasFocus} hasQuery={!!query} ref={ref}>
+          {(mdMatch || hasFocus || !!query) && (
+            <>
+              {!mdMatch && hasFocus ? (
+                <Button onClick={onClose} iconOnly icon={<SvgGlyphChevronLeft />} variant="tertiary" />
```
drillprop: Why not `IconButton`?

src/shared/components/Button/Button.tsx
```diff
@@ -8,7 +8,7 @@ import { TextVariant } from '../Text'
 export type ButtonProps = {
   icon?: React.ReactNode
   iconPlacement?: IconPlacement
-  children: React.ReactNode
+  children?: React.ReactNode
```
drillprop: I assume that you made that change because of my last comment above, but I think we should leave this as required.

src/shared/components/Searchbar/SearchBox/ResultTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export const ResultTitle: React.FC<ResultTitleProps> = ({ title, query }) => {
+  if (!title) {
+    return null
+  }
+  if (!query) {
+    return <>{title}</>
+  }
+  const filteredQuery = query.replace(/[-[\]{}()*+?.,\\^$|#]/g, '\\$&').replace(/\s+/g, '|')
```
drillprop: let's move this `/[-[\]{}()*+?.,\\^$|#]/g` to `src/config/regex.ts` maybe?

drillprop (approved):

kdembler (changes requested):
There are some visual issues that I can see:
1. Font weights and line heights aren't matching for section headers.
2. Video/channel titles should use PxGrotesk but are using Inter, please use the associated typography styles for those (`Button 2` it seems), this will also take care of different line heights
3. The closing animation on mobile is buggy. I think it also animates `width`? It should probably just operate on `transform` instead. I'm fine with this being a followup issue
4. On mobile, even though the searchbox is a full-page overlay, I can still scroll the content underneath it.
5. Height of the search list item for videos (64px) and channels (62px) does not match the design - 56px and 48px, respectively
6. The buttons used for removing recent searches should be 32px, not 40px
7. Seems the highlighting of search term isn't always applied?
Screenshot 2021-10-13 at 15 36 30

kdembler:
Could you reply with which points from my review you've fixed? Anything you decided to do as a followup should have an issue created, please link to those as well

kdembler (approved):

PR #1473, opened 2021-10-08
Search keyboard navigation

rafalpawlow:


drillprop:
I know that it's a draft for now, but I just want to share some thoughts because I think we could improve a couple of things. 
I'll start by describing issues with your implementation. 
How it works now:
1. When the focus is on search input and you don't type any letters and you hit the TAB just nothing happens.  If you want to switch focus to the next focusable element(in this case `start publishing` button) you need to press `ESC`(overlay is disappearing) and then hit the `TAB`

https://user-images.githubusercontent.com/51168865/136742403-ded1fe03-2b5b-4662-bdef-c78c759e0840.mov

2. if you type something and you didn't find anything and then again you hit TAB nothing happens again. In this particular scenario, a lot of things work really unintuitive from my perspective: 
    - If you hit ESC the input is cleared, and the caret disappears, the overlay is still visible and you cannot type in the search box. Then, If you hit the `TAB,` the focus will be on `start publishing` button, but the overlay will be still visible.
    - If you hit ESC and you want to start again typing you need to press TAB and then press SHIFT TAB.
    - it's really hard to get rid of overlay in this scenario, I would have to press tab, then SHIFT TAB and then ESCAPE

https://user-images.githubusercontent.com/51168865/136742529-eb0258d0-ce7c-473d-b98a-57c4e6537717.mov

3. If you get results from searching, and you press TAB you can navigate through all the results. The problem is when we get to the last element we're starting over. We are basically "hijacking" the tab button and we cannot go to the next focusable element(i.e `start publishing` button). And again, it's really hard to escape from this search flow. You need to press escape, then TAB(to focus on start publishing) then shift tab(to focus again on search box) and then `escape` again (to get rid of overlay and unfocus).

https://user-images.githubusercontent.com/51168865/136742838-2f45f11c-38af-4409-823d-8b15cd50b61b.mov

How I think it should work:

1. If there is no input and we press the tab, we should focus on `start publishing` button and the overlay should disappear. `ESC` works fine in this situation.
2. If we hit `ESC` overlay should always disappear, dropdown too. If we had text in input we **shouldn't** remove text inside the search box. 
3. If we get to the last element in the dropdown and then hit TAB we should: close the dropdown, the overlay should disappear, the focus should be on `start publishing` button. If we hit `SHIFT TAB` to move back to the search box, we should show again dropdown with results and overlay
4. Maybe we could abandon the idea of navigating using the TAB button inside the dropdown? I've checked other video websites(youtube, twitch, Vimeo) and it seems that when a dropdown with autocompletion is showing and you press the tab they just close the autocompletion/suggestions dropdown and switch focus to the next focusable element.

@toiletgranny could you give your opinion on this?

toiletgranny:
Overall, this looks great, @wyrdrender! 

Thanks for taking the time to write down your suggestions, @drillprop. As per your points on how you think it should work:

1. Agree. The overlay should disappear, the dropdown should be dismissed, the text field should be no longer focused.
2. Agree. Same as above.
3. Agree.
4. Agree. I was about to suggest that myself. Not sure why, but only once I saw that implemented it occurred to me that navigating through the dropdown using `TAB` is quite unusual. I apologize for any inconvenience this imposed on you, @wyrdrender. A natural consequence of this is that we'll have to remove this from the legend at the bottom of the dropdown.

I'd like to add 2 things on top of that:
1. Can we have the content of the dropdown being scrolled a bit later, when navigating using a keyboard? Currently, the dropdown is scrolled down when going into the second element of the list, which is way to early for my taste. Instead, could we have it scrolled once we reach the first element that is out of sight on the list? 
2. I'm not sure if it's meant to be tackled later, but it seems the search field input is not affected by the currently highlighted item on the list when navigating using a keyboard. As a refresher, here's how it's supposed to look like:
CleanShot 2021-10-11 at 12 15 52


@wyrdrender @drillprop Please, let me know if you agree with dropping `TAB` navigation in the dropdown, so that I can update designs accordingly.

drillprop:
> @wyrdrender @drillprop Please, let me know if you agree with dropping `TAB` navigation in the dropdown, so that I can update designs accordingly.

yeah, I think we should drop `tab` navigation in the dropdown

src/shared/components/Searchbar/Searchbar.tsx
```diff
@@ -1,87 +1,195 @@
...
+
+    const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+      if ((event.key === 'Enter' || event.key === 'NumpadEnter') && query?.trim() && !selectedItem) {
+        const state: RoutingState = { overlaidLocation }
+        addRecentSearch(new Date().getTime(), query)
+
+        // navigate to search results
+        navigate(absoluteRoutes.viewer.search({ query: query?.trim() }), { state })
+      }
+      if (event.key === 'Escape' || event.key === 'Esc' || event.key === 'Tab') {
```
drillprop, line 89:
```suggestion
      if (event.key === 'Escape' || event.key === 'Esc') {
```
I've noticed that when the focus is on search input you need to press TAB twice to switch to `start publishing`. Is it intended?  I think we could remove Tab from this line and pass `onBlur` to SearchBar instead. Wdyt?

drillprop:
Works much better now, just added one suggestion. Will look at the code when we merge #1470

kdembler:
If you move all the way down with arrows and get to the end of the results, we properly loop and start from the beginning. However this doesn't work the same way if I move up - I don't start from the bottom. Was this intentional?

kdembler:
@toiletgranny I'd appreciate your final look at this to see if you approve ;)

rafalpawlow:
@kdembler You are right, it shouldn't work like this. I fixed this

toiletgranny:
This already feels like a great improvement over the last time I looked at it. ❤️ There are still a few minor details you might consider reviewing:
1. **AS IS:** Having typed something into the search text field, when I go down the list and then all the way back to the top, my input is lost, and the field is populated with whatever is the label of the first item on the list. 
**TO BE:** Ideally, my input is preserved, as specified in Figma documentation.
CleanShot 2021-10-18 at 13 24 12


2. **AS IS:** When I start using keyboard navigation while having a list item hovered with my cursor, the hover state remains displayed.
**TO BE:** Ideally, the hover state is dropped for the time the cursor is not visible.
CleanShot 2021-10-18 at 13 25 52

3. **AS IS:** Using keyboard navigation, when I go all the way down and then press the up arrow once, the list is scrolled up in its entirety. 
**TO BE:** Ideally, the list is scrolled up (or down) only upon reaching list items that are **not** in view.
CleanShot 2021-10-18 at 13 27 46

4. **AS IS:** Upon closing the search box after I've used keyboard navigation, the search field remains populated with whatever was the list item I happened to have highlighted in the moment of pressing the escape key. 
**TO BE:** Upon hitting escape in this case, the search field should be populated back with the user input. Ideally, when opening the search box again, highlighting list items is also reset.
CleanShot 2021-10-18 at 13 29 54


Additionally, I'm not sure if this falls under the scope of this issue, but it seems like highlighting words in list items doesn't work properly. Here's how it should work accordingly to the documentation:
1. The word "demo" should be highlighted (`Neutral / 50`) because it matches the user input,
2. The word "demo" should be highlighted because it both matches the user input, but also because the whole list item is highlighted with the keyboard navigation.
3. This shouldn't be highlighted.
CleanShot 2021-10-18 at 13 37 58

rafalpawlow:
After a short call with @toiletgranny we agreed that point 4. is not an issue and the behavior is as intended.

toiletgranny:
Feels much better now, thanks! ❤️ Following my points from the above comment:

1. ✅ Works like a charm now!
2. ✅ + 💡 This is also fixed, but with an opportunity to be further improved. Now it seems like as long as the "keyboard selection" is present, hovering with a cursor is _disabled_ altogether. Ideally (and this is an important keyword here, as none of it is a deal-breaker), moving the cursor again would overwrite the keyboard selection. I think YouTube's search dropdown has a pretty neat implementation of this, it even lets you _resume_ keyboard selection based on your cursor position. But it's a deal-breaker. 
3. ❌ This seems not to be addressed fully; the list still scrolls up instantly when I'll hit the up arrow while being at the bottom of the list. Not a deal-breaker though, no pressure.
4. ✅ Beautiful!

In terms of the "additional list" from my above comment, while the 3rd point we agreed was my mistake, I'll assume points 1 & 2 are simply a matter for another GH issue (which is 100% fine).

toiletgranny:
This is looking great! As far as the keyboard navigation goes, I think we got everything we wanted now 🥰

kdembler (approved):

PR #1472, opened 2021-10-08
Failed upload tile

drillprop:
Fix #1440 
Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1903%3A94487

PR #1477, opened 2021-10-12
Video tile buttons update

drillprop:
I forgot to add the issue for this, but the scope of this PR is to update buttons on VideoPublisherTile, and change primary on click action(previously it was opening VideoWorkspace, now it's opening video in the new browser tab). 
 
Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A125742
Loom video https://www.loom.com/share/e748fd67207f4f5fb7ca8005a4d32f69 from 1:00

~~Please note, that it will be ready to review after #1472 merge~~

### DIFF
--- a/src/shared/components/VideoTileBase/PullUp.tsx
+++ b/src/shared/components/VideoTileBase/PullUp.tsx
@@ -2,7 +2,7 @@ import styled from '@emotion/styled'
 import React from 'react'
 
 import { SvgGlyphEdit } from '@/shared/icons'
-import { sizes } from '@/shared/theme'
+import { colors, sizes } from '@/shared/theme'
 
 import { IconButton } from '../IconButton'
 import { Tooltip } from '../Tooltip'
@@ -29,6 +29,6 @@ const StyledIconButton = styled(IconButton)`
   padding: ${sizes(2)};
 
   :hover {
-    background-color: black;
+    background-color: ${colors.gray[600]};
   }
 `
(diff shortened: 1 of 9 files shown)


src/shared/components/VideoTileBase/PullUp.tsx
```diff
@@ -1,27 +1,34 @@
...
 )
+
+const StyledIconButton = styled(IconButton)`
+  background-color: black;
+  border-radius: unset;
+  padding: ${sizes(2)};
+
+  :hover {
+    background-color: black;
+  }
```
kdembler: That doesn't seem correct according to the design:
Screenshot 2021-10-13 at 09 42 48

src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -260,11 +260,11 @@ export const MyVideosView = () => {
             key={video.id ? `video-id-${video.id}` : `video-idx-${idx}`}
             id={video.id}
             showChannel={false}
-            isPullupDisabled={!!videoTabs.find((t) => t.id === video.id)}
-            onClick={() => handleVideoClick(video.id)}
+            onClick={() => openInNewTab(absoluteRoutes.viewer.video(video.id), true)}
```
kdembler: I think we could use a regular navigation for the primary click now that we can link to a specific video? Currently when I hover over video, it interprets it as a link to `studio/videos` which can be a bit misleading

kdembler (changes requested):

kdembler (approved):

PR #1489, opened 2021-10-14
Introduce color tokens

### LINKED ISSUE #1480: Introduce color design tokens to our codebase
https://github.com/Joystream/atlas-resources/tree/main/design_tokens
We'll need to write some script or find some tool that will generate code based on the JSON file provided in the repository above. 
When creating this, let's also try to make use of CSS variables.  We'll need to probably find a solution how to make them type-safe(read: https://twitter.com/pveyes/status/1249729653591293952)

This task doesn't require replacing all of the color values in our codebase. Let's do this gradually

drillprop:
Fix #1480 

### Quickguide

- `fetchToken.js` - this  script will copy all token files located in https://github.com/Joystream/atlas-resources/tree/main/design_tokens to `src/styles/tokens`. To run the script you need to type `yarn fetch-tokens`. 
- to translate design tokens I'm using [style-dictionary](https://amzn.github.io/style-dictionary/#/README).  All the needed configuration is located in `sd.config.js`. To generate new variables type `codegen:styles`, generated files will be placed inside `src/styles/generated`

### What needs to be done
- as you may see we're getting `[object Object]` when generating some of the variables. I'm not sure why this is happening and how to properly translate them. I assume that maybe reading about [custom transform](https://amzn.github.io/style-dictionary/#/transforms) could be a good place to start. 
- ~~In the style-dictionary config I've created a custom parser that is adding `.value` to every aliases. For example `"value": "{core.neutral.default.900}",` will become  `"value": "{core.neutral.default.900.value}"`.  I think we could avoid this by using custom transform too.~~
- ~~We need to find a way how to introduce CSS variables to our codebase and make them type-safe. I think we could create a follow-up issue just for that. Francesco wrote here how we could accomplish this https://github.com/Joystream/atlas/issues/34~~ Done!


drillprop:
> ### What needs to be done
> * as you may see we're getting `[object Object]` when generating some of the variables. I'm not sure why this is happening and how to properly translate them. I assume that maybe reading about [custom transform](https://amzn.github.io/style-dictionary/#/transforms) could be a good place to start.

This applies only to `animation` tokens. I'll deal with them in another PR

kdembler (approved):
Looks good, great job!

Issue #1380, opened 2021-09-21
Implement an infinite grid for the video category page 


kdembler:
Should follow the Figma design if possible (infinite scroll is "triggered" by a button). Will need to take filter arguments into account.

PR #1490, opened 2021-10-14
Category infinite grid

eldiegod:
Useful figma links:
https://www.figma.com/file/3uFQ2y1jlS6fbp7Ac0yIpv/%5B%2312%5D-Video-category-page-flows?node-id=304%3A39876
https://www.figma.com/file/UdBMxrLspjcf6IVYYJS8YP/%5BPage%5D-Discover-%26-Video-category?node-id=252%3A93224

Fixes: #1380

Extends the infinite video grid component with a new "infinite on-demand" mode to be used in the category page

### DIFF
--- a/src/components/CategoryVideos/CategoryVideos.styles.tsx
+++ b/src/components/CategoryVideos/CategoryVideos.styles.tsx
@@ -1,11 +1,12 @@
 import styled from '@emotion/styled'
 
-import { Text } from '@/shared/components/Text'
-import { SvgGlyphHide } from '@/shared/icons'
 import { colors, media, sizes } from '@/shared/theme'
 
 export const Container = styled.div`
-  margin-top: ${sizes(16)};
+  /* margin-top: ${sizes(16)}; */
+
+  /* todo remove line below and uncomment above when layout is ready */
+  margin-top: ${sizes(32)};
 `
 
 export const ControlsContainer = styled.div`
@@ -31,45 +32,4 @@ export const SortContainer = styled.div`
     grid-template-columns: auto 1fr;
     grid-area: initial;
   }
-`
-
-export const FiltersContainer = styled.div<{ open: boolean }>`
-  display: ${({ open }) => (open ? 'flex' : 'none')};
-  justify-content: space-between;
-  padding: ${sizes(4)};
-  background-color: ${colors.gray[900]};
-`
-
-export const FiltersInnerContainer = styled.div`
-  display: grid;
-  gap: ${sizes(4)};
-  grid-auto-flow: column;
-  grid-auto-columns: max-content;
-`
-
-export const OtherFilterStyledText = styled(Text)`
-  margin: 0 0 ${sizes(2)} ${sizes(2)};
-  display: flex;
-  align-items: center;
-`
-
-export const OtherFilterStyledIcon = styled(SvgGlyphHide)`
-  margin-right: ${sizes(2)};
-
-  & path {
-    fill: currentColor;
-    stroke: currentColor;
-  }
-`
-
-export const DateUploadFilterContainer = styled.div`
-  padding-bottom: ${sizes(1)};
-  display: grid;
-  gap: ${sizes(3)};
-`
-
-export const MobileFilterContainer = styled.div`
-  display: grid;
-  padding-bottom: ${sizes(6)};
-  gap: ${sizes(3)};
 `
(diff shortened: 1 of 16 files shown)


src/components/CategoryVideos/CategoryVideos.styles.tsx
```diff
@@ -2,9 +2,13 @@ import styled from '@emotion/styled'
...
 
+import { InfiniteVideoGrid } from '../InfiniteGrids'
+
 export const Container = styled.div`
-  margin-top: ${sizes(16)};
-  position: relative;
+  /* margin-top: ${sizes(16)}; */
+
+  /* todo remove line below and uncomment above when layout is ready */
+  margin-top: ${sizes(32)};
```
kdembler, line 12: What's going on here?

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -1,53 +1,88 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
```
kdembler: I think this file should be inside `views/` not inside `components/`?

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -1,53 +1,88 @@
...
+      ...value,
+      languageId_eq: selectedLanguage,
+    }))
+  }, [selectedLanguage, setVideoWhereInput])
+
   return (
     <Container>
       <ControlsContainer>
         <GridItem colSpan={{ base: 2, md: 1 }}>
           <Text variant={mdMatch ? 'h4' : 'h5'}>All videos (441)</Text>
```
kdembler: What about number in this header?

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -1,53 +1,88 @@
...
-  FiltersContainer,
-  FiltersInnerContainer,
-  SortContainer,
-} from './CategoryVideos.styles'
+import { Container, ControlsContainer, SortContainer, StyledVideoGrid } from './CategoryVideos.styles'
+
+import { FiltersBar, useFiltersBar } from '../FiltersBar'
 
 export const CategoryVideos = () => {
+  const { id } = useParams()
```
kdembler: Let's make the parent view pass this here, without direct access

src/components/CategoryVideos/CategoryVideos.tsx
```diff
@@ -1,53 +1,88 @@
...
+  const {
+    setVideoWhereInput,
+    filters: { setSelectedCategoryIdFilter, setIsFiltersOpen },
+    canClearFilters: { canClearAllFilters },
+    videoWhereInput,
+  } = filtersBarLogic
+
   const [sortVideosBy, setSortVideosBy] = useState<VideoOrderByInput>(VideoOrderByInput.CreatedAtDesc)
-  const [isFiltersOpen, setiIsFiltersOpen] = useState(false)
+  const [selectedLanguage, setSelectedLanguage] = useState<string | null | undefined>('en')
```
kdembler: Let's keep the default language in a single variable so we can change it if needed

kdembler (dismissed):
We need to improve z indexes:
Screenshot 2021-10-15 at 13 26 45
Screenshot 2021-10-15 at 13 26 48

src/components/CategoryVideos/CategoryVideos.tsx
kdembler: Actually after giving it second thought, I'm pretty sure that by default no language should be selected

eldiegod: problem is once a language is selected one cannot unselect languages, but this is a problem in every language selector that we have

src/components/CategoryVideos/CategoryVideos.styles.tsx
eldiegod: These are temp margins in the meanwhile the other components missing are added otherwise it doesn't look good with the hero

src/views/viewer/CategoryView/CategoryVideos.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <Container>
+      <ControlsContainer>
+        <GridItem colSpan={{ base: 2, md: 1 }}>
+          <Text variant={mdMatch ? 'h4' : 'h5'}>All videos ({videoCount})</Text>
+        </GridItem>
+        <StyledSelect
+          placeholder="Select language"
+          onChange={setSelectedLanguage}
+          size="small"
+          helperText={null}
```
rafalpawlow: As far as I remember, few weeks ago we allowed `helperText` to be undefined, so I think we can safely get rid of this.

rafalpawlow (changes requested):

rafalpawlow (approved):

PR #1492, opened 2021-10-15
Search animations

### LINKED ISSUE #1426: Search box and input animations

rafalpawlow:
Closes #1426
https://www.figma.com/proto/Pf31tuYpozYmpq163U2ho8/Web-Components?page-id=1271%3A90759&node-id=1305%3A95843&viewport=429%2C48%2C0.4&scaling=scale-down&starting-point-node-id=1271%3A90835

### DIFF
--- a/src/components/TopbarBase/TopbarBase.style.tsx
+++ b/src/components/TopbarBase/TopbarBase.style.tsx
@@ -6,7 +6,7 @@ import { colors, media, sizes, zIndex } from '@/shared/theme'
 export const Header = styled.header`
   position: fixed;
   top: 0;
-  left: 0;
+  left: var(--size-sidenav-width-collapsed);
   right: 0;
   z-index: ${zIndex.header};
   min-height: var(--size-topbar-height);
@@ -22,8 +22,7 @@ export const Header = styled.header`
     calc(${sizes(4)} + 48px + ${sizes(2)});
 
   ${media.md} {
-    padding: ${sizes(4)} calc(${sizes(8)} + var(--size-scrollbar-width)) ${sizes(4)}
-      calc(var(--size-sidenav-width-collapsed) + ${sizes(8)});
+    padding: ${sizes(4)} calc(${sizes(8)} + var(--size-scrollbar-width)) ${sizes(4)} ${sizes(8)};
   }
 `
 
--- a/src/components/TopbarBase/TopbarBase.style.tsx
+++ b/src/components/TopbarBase/TopbarBase.style.tsx
@@ -6,7 +6,7 @@ import { colors, media, sizes, zIndex } from '@/shared/theme'
 export const Header = styled.header`
   position: fixed;
   top: 0;
-  left: var(--size-sidenav-width-collapsed);
+  left: 0;
   right: 0;
   z-index: ${zIndex.header};
   min-height: var(--size-topbar-height);
@@ -22,7 +22,8 @@ export const Header = styled.header`
     calc(${sizes(4)} + 48px + ${sizes(2)});
 
   ${media.md} {
-    padding: ${sizes(4)} calc(${sizes(8)} + var(--size-scrollbar-width)) ${sizes(4)} ${sizes(8)};
+    padding: ${sizes(4)} calc(${sizes(8)} + var(--size-scrollbar-width)) ${sizes(4)}
+      calc(var(--size-sidenav-width-collapsed) + ${sizes(8)});
   }
 `
 
--- a/src/components/TopbarViewer/TopbarViewer.style.tsx
+++ b/src/components/TopbarViewer/TopbarViewer.style.tsx
@@ -15,6 +15,12 @@ export const StyledTopbarBase = styled(TopbarBase)<FocusProps>`
   background-color: ${({ hasFocus }) => (hasFocus ? colors.gray[900] : colors.black)};
   ${({ hasFocus }) => hasFocus && `z-index: ${zIndex.globalOverlay}`};
 
+  ${media.md} {
+    left: ${({ hasFocus }) => (hasFocus ? 'var(--size-sidenav-width-collapsed)' : 0)};
+    padding: ${({ hasFocus }) => `${sizes(4)} calc(${sizes(8)} + var(--size-scrollbar-width)) ${sizes(4)}
+    ${hasFocus ? sizes(8) : `calc(var(--size-sidenav-width-collapsed) + ${sizes(8)})`}`};
+  }
+
   &.topbar-exit {
     z-index: ${zIndex.globalOverlay};
   }
--- a/src/components/TopbarViewer/TopbarViewer.style.tsx
+++ b/src/components/TopbarViewer/TopbarViewer.style.tsx
@@ -13,6 +13,10 @@ export const StyledTopbarBase = styled(TopbarBase)<FocusProps>`
   transition: background-color 0.4s ${transitions.easing};
   background-color: ${({ hasFocus }) => (hasFocus ? colors.gray[900] : colors.black)};
   ${({ hasFocus }) => hasFocus && `z-index: ${zIndex.globalOverlay}`};
+
+  &.topbar-exit {
+    z-index: ${zIndex.globalOverlay};
+  }
 `
 
 export const SearchbarContainer = styled.div`
--- a/src/components/TopbarViewer/TopbarViewer.style.tsx
+++ b/src/components/TopbarViewer/TopbarViewer.style.tsx
@@ -55,7 +55,7 @@ export const Overlay = styled.div`
   left: 0;
   background-color: ${colors.transparentBlack[54]};
   display: none;
-  transition: opacity ${animation.medium.timing}ms ${animation.medium.easing};
+  transition: opacity ${animation.medium.timing} ${animation.medium.easing};
 
   ${media.sm} {
     display: block;
--- a/src/components/TopbarViewer/TopbarViewer.style.tsx
+++ b/src/components/TopbarViewer/TopbarViewer.style.tsx
@@ -2,6 +2,7 @@ import styled from '@emotion/styled'
 
 import { IconButton } from '@/shared/components/IconButton'
 import { colors, media, sizes, square, transitions, zIndex } from '@/shared/theme'
+import { animation } from '@/shared/theme/tokens'
 
 import { TopbarBase } from '../TopbarBase'
 
@@ -54,9 +55,20 @@ export const Overlay = styled.div`
   left: 0;
   background-color: ${colors.transparentBlack[54]};
   display: none;
+  transition: opacity ${animation.medium.timing}ms ${animation.medium.easing};
 
   ${media.sm} {
     display: block;
+
+    &.searchbar-overlay-enter-active,
+    &.searchbar-overlay-exit {
+      opacity: 1;
+    }
+
+    &.searchbar-overlay-enter,
+    &.searchbar-overlay-exit-active {
+      opacity: 0;
+    }
   }
 `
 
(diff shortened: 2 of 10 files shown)


src/shared/components/Searchbar/SearchBox/SearchBox.style.ts
```diff
@@ -7,17 +7,65 @@ import { SkeletonLoader } from '@/shared/components/SkeletonLoader'
...
+  background-color: ${colors.gray[800]};
   box-shadow: inset 0 1px 0 ${colors.gray[700]};
-  ${({ visible }) => !visible && 'display: none'};
+  ${({ visible }) => !visible && 'height: 0'};
+
+  &.searchbox-enter {
+    opacity: 0;
+    height: 0;
+
+    ${media.sm} {
```
eldiegod: Is this media query here relevant?

src/shared/components/Searchbar/SearchBox/SearchBox.style.ts
```diff
@@ -7,17 +7,65 @@ import { SkeletonLoader } from '@/shared/components/SkeletonLoader'
...
+    }
+  }
+
+  &.searchbox-exit {
+    opacity: 1;
+    height: 100vh;
+    overflow-y: hidden;
+
+    ${media.sm} {
+      opacity: 1;
```
eldiegod: this line can be removed

src/shared/components/Searchbar/SearchBox/SearchBox.style.ts
```diff
@@ -7,17 +7,65 @@ import { SkeletonLoader } from '@/shared/components/SkeletonLoader'
...
+    }
+  }
+
+  &.searchbox-exit-active {
+    opacity: 0;
+    height: 0;
+    transition: all ${animation.slow.timing} ${animation.medium.easing};
+
+    ${media.sm} {
+      opacity: 0;
```
eldiegod: same as above

eldiegod (changes requested):
Did you have issues with the enter animation? the height part specifically looks a bit different from figma

Might be related but on mobile the animations look staggered first the input animates in and after the recent searches box

Sounds tricky to do with CSS animation, if that's what is causing the trouble you can consider using `react-spring` instead as it's more reliable for complex stuff

kdembler:
From my quick look I've noticed some animation differences as well, however I tested on desktop. Also on mobile I see that we are hiding the logo during the animation but the design doesn't specify that

src/shared/theme/tokens/animation.ts
```diff
@@ -1,18 +1,18 @@
...
   slow: {
-    timing: '500ms',
+    timing: 500,
     easing: 'cubic-bezier(0.3, 1.5, 0.6, 0.95)',
   },
   callout: {
-    timing: '250ms',
+    timing: 250,
     easing: 'cubic-bezier(0.3, 1.5, 0.6, 0.95)',
   },
```
eldiegod: What was the reason for this change?

eldiegod (approved):
Much better,`react-spring` always reliable 🔥

src/shared/theme/tokens/animation.ts
rafalpawlow: I wanted to use this in `react-spring` and I think it can take only `number` as timing parameter
https://github.com/Joystream/atlas/pull/1492/files#diff-25d1af4cbde354120bccdbdb57f53ae039964b0fad3cb91d53a94f2e182fee3fR65

src/components/TopbarBase/TopbarBase.style.tsx
```diff
@@ -6,7 +6,7 @@ import { colors, media, sizes, zIndex } from '@/shared/theme'
 export const Header = styled.header`
   position: fixed;
   top: 0;
-  left: 0;
+  left: var(--size-sidenav-width-collapsed);
```
kdembler: Do we have to make this change? It was like this for a long time but I recently changed it to `0` so that it looks nice when studio loads. With this change it looks like this:
Screenshot 2021-10-26 at 11 26 25

src/components/TopbarViewer/TopbarViewer.tsx
```diff
@@ -70,18 +71,20 @@ export const TopbarViewer: React.FC = () => {
...
-          value={searchQuery}
-          onFocus={handleFocus}
-          onCancel={handleCancel}
-          showCancelButton={!!searchQuery}
-          onClose={onClose}
-          controlled
-          hasFocus={isFocused}
-          onClick={handleFocus}
-        />
+        <CSSTransition classNames="searchbar" in={isFocused} timeout={{ enter: 250, exit: 500 }}>
```
kdembler: Shouldn't those `enter` and `exit` values come from animation tokens?

src/shared/components/Searchbar/SearchBox/SearchBox.style.ts
```diff
@@ -1,27 +1,73 @@
...
 import { IconButton } from '@/shared/components/IconButton'
 import { ShortcutIndicator } from '@/shared/components/ShortcutIndicator'
 import { SkeletonLoader } from '@/shared/components/SkeletonLoader'
 import { Text } from '@/shared/components/Text'
 import { SvgAvatarSilhouette } from '@/shared/illustrations'
 import { colors, media, sizes, square, typography } from '@/shared/theme'
+import { animation } from '@/shared/theme/tokens'
 
-export const Container = styled.div<{ visible: boolean }>`
+export const Container = styled(animated.div)<{ isVisible: boolean }>`
```
kdembler: I don't think we need this to be `animated.div` anymore?

src/shared/components/Searchbar/SearchBox/SearchBox.tsx
```diff
@@ -19,6 +20,11 @@ import {
 type SearchBoxProps = {
   searchQuery: string
   onSelectRecentSearch: (title?: string) => void
+  className?: string
+  styles?: {
+    opacity: SpringValue<number>
+    maxHeight: SpringValue<string>
+  }
```
kdembler: Don't think we need these anymore?

src/components/TopbarViewer/TopbarViewer.tsx
rafalpawlow: I removed timeout completely, it is not needed here

src/components/TopbarViewer/TopbarViewer.style.tsx
```diff
@@ -13,6 +14,16 @@ export const StyledTopbarBase = styled(TopbarBase)<FocusProps>`
   transition: background-color 0.4s ${transitions.easing};
   background-color: ${({ hasFocus }) => (hasFocus ? colors.gray[900] : colors.black)};
   ${({ hasFocus }) => hasFocus && `z-index: ${zIndex.globalOverlay}`};
+
+  ${media.md} {
+    left: ${({ hasFocus }) => (hasFocus ? 'var(--size-sidenav-width-collapsed)' : 0)};
+    padding: ${({ hasFocus }) => `${sizes(4)} calc(${sizes(8)} + var(--size-scrollbar-width)) ${sizes(4)}
+    ${hasFocus ? sizes(8) : `calc(var(--size-sidenav-width-collapsed) + ${sizes(8)})`}`};
+  }
```
kdembler: Could you add a comment here to display why we need to change `left`?

kdembler (approved):
Looks good to me! Let's get an approval from Adam as well and we can merge

toiletgranny:
Oh wow, this already matches our Figma prototype very closely, I'm super impressed! ❤️ There are a few details I wanted to point out:

1. Upon opening and closing the dropdown, I can see the `opacity` CSS property is being animated. This creates a rather unpleasant result when for a brief moment you can see things blending through each other. Could we drop the `opacity` animation altogether?
2. When the content of the dropdown changes, sometimes you can see the scrollbar trying its best to really come into play, appearing and disappearing randomly, causing a horizontal layout shift happening for a split of a second. At this point, I'm fairly confident that _it is_ a custom-styled scrollbar (not a browser native one), but I'm not sure what can we do about that. 
Some ideas, right off top of my head are: getting rid of the custom styled scrollbar over here or forcing the scrollbar to be always visible (or at least giving it some fixed space, so that upon reappearing it doesn't move the content on the left). Ideally, we could discuss this together.
3. When the dropdown changes its height, unlike in Figma, these height differences are currently not being animated. Ideally, we could have these transitions animated using the same timing and easing as with everything else in here (`transition.medium`). Is this possible? 
4. It seems like the colors of the skeleton loader blend with the background color of the dropdown and I'm wondering what we can do about that. 
    - In Figma, I (kind of) cheater by applying a `Screen` blending mode on the skeleton shapes, which would make them appear brighter. But I know that doing this in the browser is quite risky. 
    - Fortunately, now — thanks to color design tokens — we have `Lighten` and `Darken` color versions of our Neutral palette, which come with different opacity levels. I think this is best approached when a designer is paired with a developer for a 15-20 minutes session so that we can quickly tweak the colors to adjust the end result. I'd be happy to help with that.

kdembler:
@toiletgranny You've made this point a couple of times already but I'm not sure what you mean by `I'm fairly confident that it is a custom-styled scrollbar (not a browser native one)`. We are using native browser scrollbars thought the whole application, we only apply some CSS rules to change their colors, so they match the app style better. Other than color and width of the scrollbar, as far as I can tell, they function exactly like regular, unstyled scrollbars. I'm not sure which behaviour gives you an impression of scrollbars being custom.

toiletgranny:
As discussed yesterday, here are updated color values for the skeleton loader, based on our new [color tokens](https://github.com/Joystream/atlas-resources/blob/main/design_tokens/colors.token.json):

**Base:**
```
background-color: rgba(187, 217, 246, 0.13); // core.neutral.800.lighten
```

**Gradient wave:** 
```
background: linear-gradient(
  104deg,
  rgba(183, 200, 250, 0) 15%,        // core.neutral.900.lighten, with no opacity
  rgba(183, 200, 250, 0.06) 30%,     // core.neutral.900.lighten
  rgba(187, 217, 246, 0.13) 48%,     // core.neutral.800.lighten
  rgba(187, 217, 246, 0.13) 52%,     // core.neutral.800.lighten
  rgba(183, 200, 250, 0.06) 70%,     // core.neutral.900.lighten
  rgba(183, 200, 250, 0) 85%         // core.neutral.900.lighten, with no opacity
);
```

In case you plan on using color tokens, rather than pasting raw values, please note that in the `15%` and `85%` steps of the gradient wave I'm referencing the `core.neutral.900.lighten` color, the original value of which is `rgba(183, 200, 250, 0.06)`. Note the `0.06` alpha channel, which is being "overridden" to `0` in the gradient wave. Let me know if this is going to cause any trouble.

toiletgranny:
We've just jumped off a quick call with @wyrdrender, below is a recap of what we talked about:
1. The dropdown's `max-height` is currently set to `90vh`, which comes with a risk of the search box being clipped on smaller screens. Ideally, the `max-height` is set to: `calc(100vh - 16px - 48px - 16px)`, where `16px` is the distance between the input and the edge of the viewport and `48px` is the height of the search text field. The last `16px` could be a nice touch of ensuring that the distance between the dropdown and the edge of the viewport is consistently the same as the distance between the input and the top of the viewport.
2. This might belong in a separate issue, I believe: 
**AS IS:** I've just noticed that the "Recent searches" section is not being filtered by the user input. 
**TO BE:** Upon typing anything into the search box, the "Recent searches" section should be treated just like the "Videos" and "Channels" sections right now, meaning: being filtered by the user input and displaying the top 3 results. [More on this in Figma.](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1289%3A96197)

kdembler:
Let's tackle the second point as a separate issue. @wyrdrender can you add that to the milestone?

rafalpawlow:
@kdembler  https://github.com/Joystream/atlas/issues/1520

kdembler (approved):

PR #1494, opened 2021-10-15
Open failed upload drawer when navigate from snackbar

### LINKED ISSUE #1475: Allow linking to specific upload in My uploads
Sometimes we will navigate user to `My uploads` to show a specific upload. More specifically, we do that when one of the uploads failed - this navigation will be possible from the snackbar or the video tile. In those cases, it'd be really nice if after the navigation, the app would automatically expand the specific upload item. Currently, all the items are collapsed by default

rafalpawlow:
Closes #1475 

@kdembler  I'm not sure if you'll like this solution, but it seems the simplest to me. WDYT?


kdembler (commented):
I'm fine with the general solution but would like to avoid polluting the path with this info. Maybe instead of a query param we could use routing state? This way we can follow a similar approach but not present this in the URL

src/config/routes.ts
```diff
@@ -68,4 +68,5 @@ export const absoluteRoutes = Object.entries(BASE_PATHS).reduce((absoluteRoutesA
 export const QUERY_PARAMS = {
   SEARCH: 'query',
   JOIN: 'step',
+  UPLOADS: 'highlightFailed',
```
kdembler: Is this used anywhere now?

kdembler (approved):

PR #1501, opened 2021-10-18
Player finish state adjustments

rafalpawlow:


src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.style.ts
```diff
@@ -110,12 +129,18 @@ export const StyledChannelLink = styled(ChannelLink)<StyledChannelLinkProps>`
...
   flex-shrink: 0;
-  margin: ${sizes(6)} ${sizes(4)};
+  margin-right: ${sizes(8)};
+  margin-left: ${sizes(4)};
   position: relative;
   display: flex;
-  height: ${sizes(14)};
   justify-content: center;
   align-items: center;
+  height: calc(100% - 4.25em);
```
eldiegod, line 141: Any reason to use em here?

rafalpawlow: TBH it was here, I just moved it to another component. But you are right it's worth to change it

eldiegod (approved):
Didn't find any issues 👍

src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.style.ts
kdembler: I think there was some reason to use it, I remember that Bartosz used `em` in a couple of places in the video player. But I don't remember what this was about

kdembler:
I think we should also remove all the references to `posterUrl` in `src/shared/component/VideoPlayer/videoJsPlayer.ts`

src/shared/components/VideoPlayer/VideoOverlays/EndingOverlay.style.ts
rafalpawlow: Maybe let's keep it then? 😉

kdembler: yeah

toiletgranny:
The change alone looks great and is 100% as specified in Figma.

However, I've just noticed one thing that apparently broke some time ago and we just didn't catch it (it's even on production now). This might deserve its own issue.
1. **AS IS:** Clicking on the pause countdown button redirects the user to the next video.
2. **TO BE:** Clicking on the pause countdown button should just stop the countdown. The button itself should be replaced with a play button, allowing the user to go into the next video (described above the button). Only upon clicking on the play button, the user should be redirected to the next video.
CleanShot 2021-10-19 at 13 32 25

rafalpawlow:
@toiletgranny fixed, please confirm it's working properly now

toiletgranny:
Confirmed, thank you!

kdembler:
Found a very tiny issue, otherwise good to go!

When no next video is presented, channel name is centred horizontally in the design but not in the app: Screenshot 2021-10-21 at 16 59 03

kdembler (approved):

Issue #1359, opened 2021-09-16
Category Page Layout


eldiegod:
- Make sure everything works together nicely
- Add category cards shortcuts
- Add CTAs

PR #1502, opened 2021-10-18
Category page layout

eldiegod:
fixes:  #1359

### DIFF
--- a/src/views/viewer/CategoryView/CategoryVideos.tsx
+++ b/src/views/viewer/CategoryView/CategoryVideos.tsx
@@ -7,7 +7,6 @@ import { languages } from '@/config/languages'
 import { SORT_OPTIONS } from '@/config/sorting'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
 import { Button } from '@/shared/components/Button'
-import { IconButton } from '@/shared/components/IconButton'
 import { GridItem } from '@/shared/components/LayoutGrid'
 import { Text } from '@/shared/components/Text'
 import { SvgActionFilters } from '@/shared/icons'
@@ -16,8 +15,6 @@ import { Container, ControlsContainer, SortContainer, StyledSelect, StyledVideoG
 
 export const CategoryVideos: React.FC<{ categoryId: string }> = ({ categoryId }) => {
   const mdMatch = useMediaMatch('md')
-  const lgMatch = useMediaMatch('lg')
-  const betweenMdAndLgMatch = mdMatch && !lgMatch
 
   const filtersBarLogic = useFiltersBar()
   const {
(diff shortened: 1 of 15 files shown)


src/views/viewer/CategoryView/CategoryView.tsx
```diff
@@ -27,9 +39,24 @@ const dummyHeroVideos = [
 ]
 
 export const CategoryView = () => {
-  const { videos } = useVideos({ limit: 3 })
+  const MdBreakpointMatch = useMediaMatch('md')
```
kdembler: shouldn't be upper-cased

kdembler:
we also need some empty state when filters produce no result (also the header is broken in that case):
Screenshot 2021-10-21 at 16 51 42

src/views/viewer/CategoryView/CategoryView.style.ts
```diff
@@ -0,0 +1,25 @@
...
+  padding-bottom: ${sizes(6)};
+  border-bottom: 1px solid ${colors.gray[700]};
+`
+
+export const CategoriesContainer = styled(LayoutGrid)`
+  margin-top: ${sizes(12)};
+`
+
+export const FallbackWrapper = styled.div`
+  /* margin: ${sizes(46)} auto ${sizes(31)} auto; */
```
kdembler: What's the purpose of this?

src/views/viewer/CategoryView/CategoryView.style.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  ${media.sm} {
+    margin: ${sizes(46)} auto ${sizes(31)} auto;
```
kdembler: These shouldn't be expressed with `sizes`

src/views/viewer/CategoryView/CategoryVideos.tsx
```diff
@@ -64,8 +62,8 @@ export const CategoryVideos: React.FC<{ categoryId: string }> = ({ categoryId })
   return (
     <Container>
       <ControlsContainer>
-        <GridItem colSpan={{ base: 2, md: 1 }}>
-          <Text variant={mdMatch ? 'h4' : 'h5'}>All videos ({videoCount})</Text>
+        <GridItem colSpan={{ base: 2, sm: 1 }}>
+          <Text variant={mdMatch ? 'h4' : 'h5'}>All videos {videoCount !== undefined && `(${videoCount})`}</Text>
         </GridItem>
         <StyledSelect
           placeholder="Select language"
```
kdembler: This should be `Any language`

src/views/viewer/CategoryView/CategoryView.style.ts
eldiegod: is to adjust the margins around the Fallback component for desktop as styling the component directly didn't work out too well

kdembler: I was asking about the comment but I see it's been removed since

kdembler (approved):

PR #1509, opened 2021-10-21
introduce featured content fetched from Orion

kdembler:
This PR adds ability for Atlas to fetch featured content from Orion. This currently includes video hero and featured videos for video categories. Raw featured data is fetched from Orion and later additional data (title, etc) is fetched from query node. Data is also cached in indexeddb so that cached values will always be available without a need to wait for Orion to respond. Once fresh data is fetched, it will replace the persisted cache, but will only be used from the next app visit.

### DIFF
--- a/src/api/featured/categoriesFeaturedVideos.ts
+++ b/src/api/featured/categoriesFeaturedVideos.ts
@@ -1,0 +1,36 @@
+import { useApolloClient } from '@apollo/client'
+import { useMemo } from 'react'
+
+import { useVideos } from '@/api/hooks'
+import { VideoFieldsFragment } from '@/api/queries'
+import {
+  GetCategoriesFeaturedVideosDocument,
+  GetCategoriesFeaturedVideosQuery,
+} from '@/api/queries/__generated__/featured.generated'
+import { createLookup } from '@/utils/data'
+
+import { useGenericFeaturedData } from './helpers'
+
+type CategoriesFeaturedVideos = Record<string, (VideoFieldsFragment & { videoCutUrl?: string | null })[]>
+
+export const useCategoriesFeaturedVideos = (): CategoriesFeaturedVideos | null => {
+  const client = useApolloClient()
+  const fetchCategoriesFeaturedVideos = useMemo(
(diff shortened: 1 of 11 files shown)


src/api/featured/categoriesFeaturedVideos.ts
```diff
@@ -0,0 +1,36 @@
...
+type CategoriesFeaturedVideos = Record<string, (VideoFieldsFragment & { videoCutUrl?: string | null })[]>
+
+export const useCategoriesFeaturedVideos = (): CategoriesFeaturedVideos | null => {
+  const client = useApolloClient()
+  const fetchCategoriesFeaturedVideos = useMemo(
+    () => async () =>
+      (await client.query<GetCategoriesFeaturedVideosQuery>({ query: GetCategoriesFeaturedVideosDocument })).data
+        .allCategoriesFeaturedVideos,
+    [client]
+  )
```
eldiegod, line 23: Can you add some context on why we have to use the apollo client directly like this?

eldiegod (commented):
I see there is the `useVideoHeroData` hook that works for the home video hero but what are we going to use for i.e. the category page video heroes

src/api/featured/categoriesFeaturedVideos.ts
kdembler: I wanted the `useGenericFeaturedData` to be able to manually trigger data fetching. Originally I only fetched when cache was X hours old, but decided to drop it. I don't think we can have a function that returns Promise with data without using client directly.

kdembler:
> I see there is the useVideoHeroData hook that works for the home video hero but what are we going to use for i.e. the category page video heroes

That could be handled by the hook that fetched featured videos for categories, you would just extract first few results for a specific category. I guess thumbnails would need to be loaded separately because video hero excepts the URL. So possibly a wrapper hook could be created that extracts videos for video hero and resolves thumbnails

eldiegod:
Ok fair enough, I think we can merge and I'll see if I hit any roadblocks

eldiegod (approved):

PR #1513, opened 2021-10-25
Update search page

rafalpawlow:


src/views/viewer/SearchView/SearchResults/SearchResults.style.ts
```diff
@@ -0,0 +1,69 @@
...
+`
+
+export const Filters = styled.div`
+  align-items: center;
+  background-color: ${colors.gray[700]};
+  overflow: hidden;
+  will-change: height;
+  transition: height ${transitions.timings.routing} ${transitions.easing};
+  margin: 0 var(--size-global-horizontal-padding);
+  height: calc(100vh - 145px);
```
kdembler: Where does 145px come from?

src/views/viewer/SearchView/SearchResults/SearchResults.style.ts
```diff
@@ -0,0 +1,69 @@
...
+  ${media.sm} {
+    background-color: ${colors.gray[800]};
+    margin: 0;
+    display: flex;
+    height: 72px;
+  }
+
+  &.filters-active,
+  &.filters-exit {
+    height: calc(100vh - 145px);
```
kdembler: Same here

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -1,32 +1,47 @@
...
 } from './FiltersBar.styles'
 import { VideoLengthOptions, useFiltersBar } from './useFiltersBar'
 
-import { ActionDialog, ActionDialogProps } from '../ActionDialog'
-import { StyledTitleText } from '../MessageDialog/MessageDialog.style'
+import { ActionDialogProps } from '../ActionDialog'
 
-export const FiltersBar: React.FC<ReturnType<typeof useFiltersBar>> = ({
+type FiltersBarProps = {
+  variant?: 'default' | 'secondary'
```
kdembler: What is this variant referring to? Do we have differences between filter bar designs?

kdembler (commented):
1. Those channel cards definitely shouldn't look like this :P
Screenshot 2021-10-27 at 10 07 28
2. I think we could improve the filters bar enter animation. Currently seems the content below isn't pushed in sync with the bar appearing. Also you can see an overlap between the filters bar and the tabs. I think the filters should slide from under the tabs, so without overlap.

https://user-images.githubusercontent.com/12646744/139065749-8c9cc15e-5d98-4abf-85d8-c87a657a09b0.mov

src/components/FiltersBar/FiltersBar.tsx
rafalpawlow: I've found some differences in filters on discovery branch but tbh I didn't compare it to figma. As far as I can see, the filter bar in figma looks the same everywhere so I can get rid of that.

kdembler: If so, we should have a single variant. Please coordinate any differences and necessary changes with @eldiegod so we end up with single version of this component

src/views/viewer/SearchView/SearchResults/EmptyFallback.tsx
```diff
@@ -36,7 +38,14 @@ export const EmptyFallback: React.FC = () => (
     <SvgEmptyStateIllustration />
     <Message>
       <Title variant="h3">Sorry, we couldn&apos;t find any matches.</Title>
-      <Subtitle>Please try a different keyword.</Subtitle>
+      <Subtitle>Please try a different keyword or change filtering criteria.</Subtitle>
+      <StartNewSearch
+        variant="secondary"
+        size="large"
+        onClick={() => window.dispatchEvent(new KeyboardEvent('keydown', { key: '/' }))}
```
kdembler: This looks really risky to me, it will be easy to make this break in the future, couldn't we use some other approach to trigger the input?

toiletgranny:
### Filter popovers
1. The "Categories" filters popover seems to be missing a divider between its content and the buttons at the bottom. The "License" popover has it set up properly.
2. Ideally, filter popovers should have the "Apply" button disabled when no change is made in the filters selection. Currently, when you open up eg. a "Category" filter, the "Apply" button is active right from the beginning.
3. In the designs, options for the video length are presented using checkboxes, whereas in the implementation there are radio buttons. I'm 99% sure there's a reason for this, but in case there's not — ideally, we have checkboxes to control that.

### Filters bar
1. I think we could use an `overflow: hidden` on the filters bar, so that during the opening and closing animation its content doesn't escape out of it. To top it off, ideally, we could also have the `opacity` CSS property being animated on the _content_ of the filters bar (same timing, same easing).
2. Although this wasn't specified in the Figma file, it would be great if we could use the `transition.medium` animation token to drive the speed and the easing of the filters bar in and out transitions.
3. When I open up the filters bar, having scrolled down the page, the content of the page is being shifted down. This is as expected, so it's not an issue, however ideally the user's scroll position is preserved. If there's a time and space for such a tweak — I'd love it.

### "No results" state
1. This might be case for a separate issue: the "Empty fallback" component implementation doesn't match the designs in terms of spacing between items. Please, let me know if further guidance is needed here.
2. It seems like the copy in the implementation doesn't match the designs.
3. Clicking on the "Start new search" button should open up the search box, which is happening right now, but it was also supposed to wipe out the user's input from the search text field.

### Mobile (iPhone, iOS 15, Safari)
1. The filters dialog seems to be displayed using a different component, than the one we have in Figma. This creates a couple of issues:
  - When I scroll down in the dialog, I'm not able to scroll back up again. Scrolling doesn't work. 
  - The dialog header, containing the "X" button, is not fixed when scrolling. As a consequence, when I scroll down in the dialog, I'm not able to close it. I'm stuck here and I need to refresh the page.
  - Ideally, I should be able to close the dialog by tapping anywhere outside of it.
  - Hitting the "Apply" should close the dialog and apply selected filters. I'm not exactly sure what is happening right now.
  - When I'm tapping on the filters icon button to open up the filters dialog, it seems like there's something else weird happening in the background, underneath the dialog. Could it be the filters bar?

src/views/viewer/SearchView/SearchResults/EmptyFallback.tsx
rafalpawlow: The only thing which comes to my mind is to use context, but I was affraid it would be overkill. WDYT?

kdembler:
Found couple more issues on mobile:
1. I'd expect that after triggering the search, the input would be focused right away. Currently I need to tap the search button and then search the input to start typing.
2. Let's try using `type="search"` on the input, this should make the mobile keyboard adapt to this type of field. However, we need to test out styling doesn't break on mobile as I think this type may introduce some additional visual elements.
3. Scrolling inside the searchbox on mobile is broken a bit and I think scrolls the page underneath?

https://user-images.githubusercontent.com/12646744/139228677-69e2cd70-60ab-4f1e-b712-1b98ac27b772.MP4

src/views/viewer/SearchView/SearchResults/EmptyFallback.tsx
kdembler: Yeah it's tricky. Maybe instead of context we could create a search store that would hold reference to the input ref and expose focus function? It's not exactly better than context but I think we should avoid what we have currently

src/components/TopbarViewer/TopbarViewer.tsx
```diff
@@ -16,24 +17,27 @@ export const TopbarViewer: React.FC = () => {
...
+    searchOpen,
+    actions: { setSearchOpen },
+  } = useSearchStore()
 
   const [searchQuery, setSearchQuery] = useState('')
-  const [isFocused, setIsFocused] = useState(false)
 
-  useEffect(() => {
-    if (isFocused) {
+  useLayoutEffect(() => {
```
kdembler: Does this still needs to be `layoutEffect` instead of the regular?

src/shared/components/Searchbar/SearchBox/SearchBox.style.ts
```diff
@@ -10,36 +10,48 @@ import { SvgAvatarSilhouette } from '@/shared/illustrations'
 import { colors, media, sizes, square, typography } from '@/shared/theme'
 import { animation } from '@/shared/theme/tokens'
 
-const containerDesktopHeight = `calc(90vh - var(--size-topbar-height) + ${sizes(4)})`
+const CONTAINER_DESKTOP_HEIGHT = `calc(90vh - var(--size-topbar-height) + ${sizes(4)})`
 
-export const Container = styled.div<{ isVisible: boolean; hasQuery?: string }>`
-  position: absolute;
+const getContainerMobileHeight = (visualViewportHeight: number, hasFocus: boolean) =>
+  `calc(${hasFocus ? `${visualViewportHeight}px` : '100vh'})`
```
kdembler: Let's clean this up after changes:
1. `calc` is no longer needed
2. To avoid some of the repetition let's do something like this:
```tsx
type ContainerProps = {
  isVisible: boolean
  hasQuery?: string
  visualViewportHeight: number
  hasFocus: boolean
}

const getContainerMobileHeight = ({ visualViewportHeight, hasFocus }: ContainerProps) => // ...

  // ...
  height: ${getContainerMobileHeight};
`
```

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -95,15 +108,29 @@ export const Tabs: React.FC<TabsProps> = ({
...
+              icon={<SvgActionFilters />}
+              iconPlacement="left"
+              variant="secondary"
+              badge={filtersActive}
+              onClick={onFiltersClick}
+            >
+              {smMatch && 'Filters'}
+            </Button>
+          )}
+        </FiltersWrapper>
```
kdembler: I don't think this should be done here. We're sticking very specific language select into a generic Tabs component. We already have similar layouts done outside of the Tabs component, for example in My Videos.

src/shared/components/Tabs/Tabs.styles.tsx
```diff
@@ -36,7 +36,7 @@ export const Tab = styled.div<TabProps>`
   transition: box-shadow 0.125s ease, color 0.125s ease;
   width: ${TAB_WIDTH}px;
   min-width: ${TAB_WIDTH}px;
-  padding: ${sizes(6)} 0;
+  padding: ${sizes(7)} 0;
```
kdembler: Where is this change coming from? If it's only related to the select, let's revert it, as part of the other comment

src/views/viewer/SearchView/SearchResults/SearchResults.style.ts
```diff
@@ -0,0 +1,80 @@
...
+  height: 100vh;
+  width: 100%;
+  padding: 0 var(--size-global-horizontal-padding);
+  transition: height ${transitions.timings.routing} ${transitions.easing};
+
+  ${media.sm} {
+    background-color: ${colors.gray[800]};
+    margin: 0;
+    display: flex;
+    height: 72px;
```
kdembler: I guess we need an explicit height, but let's at least stick it in some variable (preferably inside FilterBar component iteself)

src/views/viewer/SearchView/SearchResults/SearchResults.style.ts
```diff
@@ -0,0 +1,80 @@
...
+import { colors, media, sizes, zIndex } from '@/shared/theme'
+import { transitions } from '@/shared/theme'
+
+export const PaddingWrapper = styled.div`
+  padding: 0 var(--size-global-horizontal-padding);
+`
+
+export const Filters = styled.div`
+  align-items: center;
+  background-color: ${colors.gray[700]};
```
kdembler: Why do we need to change `background-color` here? I'd expect this to be handled by the component directly

src/shared/components/Tabs/Tabs.styles.tsx
rafalpawlow: Because in search results tabs have 72px height, I had to add `variant` prop to it.

src/components/FiltersBar/FiltersBar.tsx
kdembler: @wyrdrender We still have a `variant` prop here, I thought we'd drop it

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -16,6 +16,11 @@ export type TabsProps = {
   onSelectTab: (idx: number) => void
   selected?: number
   className?: string
+  onFiltersClick?: () => void
+  onSelectedLanguage?: (language: unknown) => void
+  selectedLanguage?: unknown
+  filtersActive?: boolean
```
kdembler: Doesn't seem like those props are needed here?

kdembler:
Also please take a look at https://github.com/Joystream/atlas/pull/1513#discussion_r739644244 It's up above so don't want it to get lost

src/shared/components/Tabs/Tabs.tsx
```diff
@@ -16,6 +16,10 @@ export type TabsProps = {
   onSelectTab: (idx: number) => void
   selected?: number
   className?: string
+  onFiltersClick?: () => void
+  onSelectedLanguage?: (language: unknown) => void
+  selectedLanguage?: unknown
```
kdembler: You removed one of the props but these remaining 3 are still unnecessary

src/components/FiltersBar/useFiltersBar.ts
```diff
@@ -4,16 +4,18 @@ import { VideoWhereInput } from '@/api/queries'
...
+export const useFiltersBar = (initiallyFiltersOpen = true) => {
   // filters
   const [selectedCategoryIdFilter, setSelectedCategoryIdFilter] = useState<string>()
   const [dateUploadedFilter, setDateUploadedFilter] = useState<number>()
   const [licensesFilter, setLicensesFilter] = useState<number[]>()
   const [videoLengthFilter, setVideoLengthFilter] = useState<VideoLengthOptions>()
   const [paidPromotionalMaterialFilter, setPaidPromotionalMaterialFilter] = useState<boolean>()
   const [matureContentRatingFilter, setMatureContentRatingFilter] = useState<boolean>()
+  const [categoriesFilter, setCategoriesFilter] = useState<string[]>()
+  const [language, setLanguage] = useState<unknown>()
```
eldiegod, line 16: Don't think we should add language in here as we dont use it for clear all or can clear it's more of an independent filter and can be misleading

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -281,11 +399,14 @@ export const FiltersBar: React.FC<ReturnType<typeof useFiltersBar>> = ({
                   disabled: licensesFilter === undefined || licensesFilter?.length === 0,
                 }}
                 applyButtonProps={{
-                  onClick: () =>
+                  disabled: !licensesFilter || !licensesFilter.length,
```
eldiegod: Better not to disable the apply buttons because if you select a filter then apply it then open the filter and unselect it you cannot apply nor clear that filter

same for the other apply buttons

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -168,6 +213,28 @@ export const FiltersBar: React.FC<ReturnType<typeof useFiltersBar>> = ({
         title="Filters"
         content={
           <>
+            {mobileLanguageSelector && (
```
eldiegod, line 217: Language has since been removed from the mobile filter dialog and should now live on the layout outside if the change is not yet on figma please talk to adam he probably forgot to update it in here

src/components/FiltersBar/FiltersBar.styles.tsx
```diff
@@ -16,21 +19,52 @@ export const MobileFilterContainer = styled.div`
...
+  margin: 0 var(--size-global-horizontal-padding);
+  background-color: ${colors.gray[700]};
   display: ${({ open }) => (open ? 'flex' : 'none')};
   justify-content: space-between;
-  padding: ${sizes(4)};
-  background-color: ${colors.gray[900]};
-  will-change: opacity;
-  transition: opacity ${transitions.timings.sharp} ${transitions.easing};
+  padding: ${sizes(4)} var(--size-global-horizontal-padding);
+  will-change: opacity, height;
```
eldiegod: the open and close anim is looking a bit rough because of the padding doesn't allow to fully have 0 height this happened to me as well when I tried to implement so I'm not sure what to do.

I would use react-spring but Im sure @kdembler wouldnt like that

eldiegod (changes requested):
Looks very good overall, left a couple comments relevant to filterbar

src/components/FiltersBar/FiltersBar.styles.tsx
kdembler: I agree that the current animation doesn't look great but I don't think `react-spring` is the way to go here. I think the issue is that we're animating `height` and `opacity` but we could just `transform` the element instead. It could hide underneath the tabs bar and slide from there, right? This should fix issues with `height` transition

kdembler: But let's do any work on this in a separate PR, to restrict the scope here

src/components/FiltersBar/FiltersBar.tsx
rafalpawlow: I don't know anything about this, in Figma it's still inside in dialog. @toiletgranny?

src/components/FiltersBar/FiltersBar.tsx
rafalpawlow: Good, point. @toiletgranny you asked me to do this but this actually may be the issue. I'm thinking, maybe we could disable this button only before first apply?

src/components/FiltersBar/FiltersBar.styles.tsx
eldiegod: alternatively, a div wrapper could be added and the animation could be done there as is, I think as long as the wrapper doesn't have padding it should work tho I haven't tried it, the advantage being not having to touch z-index

src/components/FiltersBar/FiltersBar.tsx
toiletgranny: My initial idea was to have the "Apply" button enabled upon a _change_ in the selection, and by "change" I meant both _selecting_ and — as @eldiegod mentioned — _unselecting_. But if this requires creating and maintaining a troublesome logic then sure, I'm okay with proceeding with your idea @wyrdrender! :)

src/components/FiltersBar/FiltersBar.tsx
toiletgranny: @eldiegod In my mind, the change you're referring to was supposed to take place on the Video category page only. 

The context in which the language filter is presented on the Search results page is a little bit different, hence I think it makes sense to keep it as part of the mobile filter dialog.

src/components/FiltersBar/FiltersBar.styles.tsx
kdembler: That's also a possible solution but animating `height` will always be more rough than animating `transform`

src/components/FiltersBar/FiltersBar.tsx
eldiegod: @toiletgranny the problem with that is that we use the same component/logic for both, @wyrdrender any solution ideas?

rafalpawlow: Why can't just keep it like this? The solution I made does not conflict with your component.

src/components/FiltersBar/useFiltersBar.ts
rafalpawlow: After a moment of reflection, I come to the conclusion that it's not really because we are showing language selector in filters modal. WDYT?

src/components/FiltersBar/FiltersBar.styles.tsx
toiletgranny: @kdembler @eldiegod I did a quick experiment with animating the `height` and it turns out that if we make the animation quick enough, we might actually get away with that rough and sudden change in the height not being animated. Here's the result: https://jsfiddle.net/toiletgranny/1snkh2bL/93/ (try changing the speed to 1 second and you'll see the difference :D)

If you decide on going with the above, then I'd suggest using the `transition.fast` animation token.

Another, "bonus" thing I discovered today is that if we'll use `position: sticky`, then the change in the overall height of the filters bar doesn't seem to affect the user's scroll position. That's pretty cool! <3

rafalpawlow: @toiletgranny I've done transform transition, take a look. Unfortunately `position: sticky` didn't work well in this case.

toiletgranny: @wyrdrender Oh man, it looks so pretty right now! It's not 100% what we had in that Figma prototype but I don't mind at all, I might even like it more than the original! 

In case you're still on it, then I'd suggest dropping the opacity animation (it's not needed right now).

eldiegod:
really great looking anim now 🥳

eldiegod (approved):

PR #1516, opened 2021-10-26
Dynamic featured content

eldiegod:
Add dynamic content from orion to:

- Featured category cards in /discover
- Featured videos section in /category/id
- Category video hero in /category/id


src/views/viewer/CategoryView/CategoryView.tsx
```diff
@@ -111,3 +104,33 @@ export const CategoryView = () => {
...
+  })
+
+  if (!videoHeroVideos) return [null, null, null]
+
+  videoHeroVideos[0].thumbnailPhotoUrl = thumbnailPhotoUrl1 ?? ''
+  videoHeroVideos[1].thumbnailPhotoUrl = thumbnailPhotoUrl2 ?? ''
+  videoHeroVideos[2].thumbnailPhotoUrl = thumbnailPhotoUrl3 ?? ''
+
+  return videoHeroVideos
+}
```
kdembler: Couple of notes:
1. I think we should filter `featuredVideos` first to make sure we operate only on videos that have `videoCutUrl` defined.
2. We must remember that this component doesn't have to work with 3 videos at all times. We should try displaying 3 but we must handle the case when for example only 2 or 1 video has `videoCutUrl` defined. I'm not saying the current code doesn't work in this case because I didn't test it yet, but a note to you.
3. I think we may need to update video hero in category view to work without full-size thumbnails. As discussed in some other places, soon thumbnails will be only lowres so will not work well in full width context. I think we will require custom image only for main video hero on home page. For those I think we could figure out something else. For example, we can force the video player to show the first frame when the video is still loading. Can you make sure this will work fine with lowres images?

src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -14,9 +15,23 @@ import {
   FeaturedCategoriesContainer,
   StyledLimitedWidthContainer,
 } from './DiscoverView.style'
-import { featuredVideoCategories, videoCategories } from './data'
+import { selectedFeaturedVideoCategories, videoCategories } from './data'
 
 export const DiscoverView: React.FC = () => {
+  const categoriesFeaturedVideos = useCategoriesFeaturedVideos()
+  const featuredVideoCategoryCardsData = selectedFeaturedVideoCategories.map((category) => {
+    categoriesFeaturedVideos?.[category.id]
```
kdembler: We don't need this statement

kdembler (commented):
Additionally, like discussed over RocketChat, let's make sure the new views work properly on all deployments, even when the category list is different

src/views/viewer/CategoryView/CategoryView.tsx
```diff
@@ -54,26 +45,33 @@ export const CategoryView = () => {
...
         header={{
-          title: currentCategory?.title,
+          title: currentCategory?.name,
           icon: currentCategory?.icon,
         }}
-        videos={dummyVideos}
+        videos={videoHeroVideos}
       />
+
+      {(featuredVideos?.length ?? 0) > 0 && (
```
kdembler: Let's just do `data?.[id] ?? []` in line 30 and then we can get rid of all the optional chaining

src/views/viewer/CategoryView/CategoryView.tsx
```diff
@@ -54,26 +45,33 @@ export const CategoryView = () => {
...
+        <>
+          <TitleContainer>
+            <Text variant="h4">Featured Videos</Text>
+          </TitleContainer>
+          <Grid>
+            {[...(featuredVideos ?? [])]?.map((video, idx) => (
```
kdembler: Why do we need to expand this array here?

src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -14,9 +15,29 @@ import {
   FeaturedCategoriesContainer,
   StyledLimitedWidthContainer,
 } from './DiscoverView.style'
-import { featuredVideoCategories, videoCategories } from './data'
+import { selectedFeaturedVideoCategories, videoCategories } from './data'
 
 export const DiscoverView: React.FC = () => {
+  const { categories } = useCategories()
+  const videoCategoriesArray = Object.values(videoCategories)
```
kdembler: Is the key in the `videoCategories` object used in any way now? If not, maybe let's just the `id` be the key instead and then we don't need `Object.values` and later `find`, we could just do `videoCategories[category.id]`

kdembler (changes requested):
Opening https://deploy-preview-1516--atlas-dev.netlify.app/discover with the mainnet produces quite a broken screen

src/views/viewer/DiscoverView/DiscoverView.tsx
eldiegod: true

kdembler (approved):

PR #1518, opened 2021-10-27
Discovery sprint

eldiegod:


toiletgranny:
I jumped on a call with @eldiegod earlier this morning to discuss the Video category page, and here's what we agreed on:
1. Ideally, the expanded filters bar (in the "All videos" section) looks and works the same way as on the recently implemented Search results page by @wyrdrender, in terms of the opening/closing animation, borders, colors, etc.
2. Because having both the language filter and the "Sort by" select displayed together takes an enormous amount of vertical space **on mobile**, ideally:
   - the language filter is moved back to the Filters dialog box (this way we're also being consistent with how the language filter is being treated on the Search results page),
   - the "Sort by" select takes place of that language filter on mobile, while dropping the "Sort by" label at the top and having it displayed in the select component label.
   
    image


3. Currently, the language filter makes it impossible to reset the selection once a specific language has been selected. Thus, my suggestion is to add an "All languages" option, have it displayed at the top of the list and have it being selected by default. 
image

4. I appreciate your effort in affecting the user's scroll position upon changing filters, this is very thoughtful. One suggestion I have is to perhaps try to make that scroll position change animated, so that the user is not concerned or worried about being taken to a completely different page or view.

src/config/nav.tsx
```diff
@@ -19,6 +19,11 @@ export const viewerNavItems = [
     name: 'New',
     to: absoluteRoutes.viewer.new(),
   },
+  {
+    icon: <SvgNavDiscover />,
+    name: 'Discover',
+    to: absoluteRoutes.viewer.discover(),
+  },
```
kdembler: Let's hide this for now so we can deploy `dev` if needed before the featured content is ready

kdembler (approved):

PR #1519, opened 2021-10-27
fix colors, add comments with references to design tokens

kdembler:


toiletgranny (approved):
What an exceptional commit, great work!

PR #1527, opened 2021-10-27
filter recent search when typing query

### LINKED ISSUE #1520: Filter recent searches by user input

rafalpawlow:
Closes #1520

### DIFF
--- a/src/providers/personalData/store.tsx
+++ b/src/providers/personalData/store.tsx
@@ -80,6 +80,7 @@ export const usePersonalDataStore = createStore<PersonalDataStoreState, Personal
             state.recentSearches.splice(index, 1)
           }
           state.recentSearches.unshift({ title, id })
+          state.recentSearches.splice(6, 1)
         })
       },
       deleteRecentSearch: (id) => {
--- a/src/providers/personalData/store.tsx
+++ b/src/providers/personalData/store.tsx
@@ -26,8 +26,8 @@ const WHITELIST = [
 export type PersonalDataStoreActions = {
   updateWatchedVideos: (__typename: WatchedVideoStatus, id: string, timestamp?: number) => void
   updateChannelFollowing: (id: string, follow: boolean) => void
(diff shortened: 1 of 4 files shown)


src/shared/components/Searchbar/SearchBox/SearchBox.tsx
```diff
@@ -83,15 +84,23 @@ export const SearchBox: React.FC<SearchBoxProps> = ({
...
 
-  const slicedReccentSearches = recentSearches.slice(0, 6)
+  const filteredRecentSearches = searchQuery.length
+    ? recentSearches
+        .filter((item) =>
+          new RegExp(`${searchQuery.replace(SPECIAL_CHARACTERS, '\\$&').replace(/\s+/g, '|')}`, 'i').test(
+            item.title || ''
+          )
+        )
+        .slice(0, 3)
```
kdembler, line 94: Why do we use only 3 elements here?

src/providers/personalData/store.tsx
```diff
@@ -80,6 +80,7 @@ export const usePersonalDataStore = createStore<PersonalDataStoreState, Personal
             state.recentSearches.splice(index, 1)
```
kdembler: Cannot provide a comment for the full function but I think it's a bit hard to read what's happening here. What do you think of this instead?
```tsx
addRecentSearch: (title) => {
  set((state) => {
    const filteredCurrentSearches = state.recentSearches.filter((item) => item.title !== title)
    const newSearches = [{ title }, ...filteredCurrentSearches]
    state.recentSearches = newSearches.slice(0, 6)
  }
},
```

This takes into account the discussion we had over rocket chat on unnecessary id field

src/shared/components/Searchbar/SearchBox/SearchBox.tsx
rafalpawlow: This was requested by @toiletgranny

kdembler (approved):

PR #1533, opened 2021-11-02
Always show video context menu on mobile

### LINKED ISSUE #1521: Mobile version: "Dots button" not visible till hover
**Issue:**
On mobile devices there is no mouse hover effect, as result we can see three dots for each video only when touch on scroll: [http://joxi.ru/RmzGZ7ZUjaJ1K2](http://joxi.ru/RmzGZ7ZUjaJ1K2)

**Browser:**
All mobile browsers. 
Tested on iPhone 11 Pro, iOS 15.0.2, Safari browser.

**Suggestion:**
Always show this button for mobile version

rafalpawlow:
Fixes #1521

### DIFF
--- a/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
+++ b/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
@@ -2,14 +2,12 @@ import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 import { Link } from 'react-router-dom'
 
-import { colors, media, sizes, square, typography } from '@/shared/theme'
+import { colors, sizes, square, typography } from '@/shared/theme'
 
 import { Avatar } from '../Avatar'
 import { IconButton } from '../IconButton'
 import { SkeletonLoader } from '../SkeletonLoader'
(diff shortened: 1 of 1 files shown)


src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -62,9 +62,13 @@ export const TextContainer = styled.div`
...
   ${square(32)};
 
+  opacity: 1;
   margin-left: ${sizes(2)};
-  opacity: ${({ isActive }) => (isActive ? 1 : 0)};
   pointer-events: ${({ isActive }) => (isActive ? 'auto' : 'none')};
+
+  ${media.md} {
+    opacity: ${({ isActive }) => (isActive ? 1 : 0)};
+  }
```
kdembler: Let's not base this on screen width but instead of media query which tells us whether the device has a pointer. `@media (any-pointer: none)` should do the trick

kdembler (changes requested):

kdembler (approved):

Issue #1381, opened 2021-09-21
Wrap up the all videos section


kdembler:
- Sticky header
- Filers applied to the grid
- Anything else

PR #1534, opened 2021-11-02
Sticky header for the filters bar in category vids

eldiegod:
fixes: #1381

Uses react-stickynode to make the filtersbar sticky in the category videos section of the category page

https://github.com/yahoo/react-stickynode


drillprop (commented):
The code generally looks good, although I'm not sure if we should use another library. Is it required to write a lot of code to accomplish that? If yes, I'm fine with it.
Found some visual glitches: 
1. When you hover on video which is behind sticky header

https://user-images.githubusercontent.com/51168865/139869726-38ccb7fe-82b1-4863-9488-14faa6b6eb0e.mov

2. When you update filters and there is no result you'll not see information about no results until you move your mouse wheel up. Could we improve that? This applies to mobile view:

https://user-images.githubusercontent.com/51168865/139871982-38834121-9596-45a5-a154-54f25b4041fc.mov

eldiegod:
@drillprop fixed `1.` for `2.` it shows `All videos (0)` but not sure what else we could do there, what did you have in mind?

drillprop:
> @drillprop fixed `1.` for `2.` it shows `All videos (0)` but not sure what else we could do there, what did you have in mind?

I would expect that when there are no videos, the header should not stick anymore(immediately). Right now this header just covers up the `No videos` illustration. You need to scroll up to see the illustration, which is a weird experience(take a look at the second video: 0:03  - 0:07).

eldiegod:
> I would expect that when there are no videos, the header should not stick anymore(immediately). Right now this header just covers up the `No videos` illustration. You need to scroll up to see the illustration, which is a weird experience(take a look at the second video: 0:03 - 0:07).

@drillprop  I see, I made some adjustments I think it should be better now, lmk

kdembler:
I see some issue that makes the screen scroll down when I open category view. I should be on top of the page after navigation

https://user-images.githubusercontent.com/12646744/140068230-bf465eab-7731-423c-8912-626b68a905d4.mov

eldiegod:
@kdembler fixed!

src/views/viewer/CategoryView/CategoryVideos.tsx
```diff
@@ -52,48 +62,65 @@ export const CategoryVideos: React.FC<{ categoryId: string }> = ({ categoryId })
...
+      containerRef.current?.scrollIntoView()
+    }
+
+    // account for videoWhereInput initialization
+    if (!isEqual(videoWhereInput, {})) {
+      setScrollWhenFilterChange(true)
+    }
+    // we dont want to rerun effect when scrollWhenFilterChange changes but only videoWhereInput
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [videoWhereInput])
```
drillprop: If I understand this code correctly, we could just use `ref` instead of `useState`. 
```suggestion
  const scrollWhenFilterChange = useRef(false)
  useEffect(() => {
    if (scrollWhenFilterChange.current) {
      containerRef.current?.scrollIntoView()
    }

    // account for videoWhereInput initialization
    if (!isEqual(videoWhereInput, {})) {
      scrollWhenFilterChange.current = true
    }
  }, [videoWhereInput])
```

drillprop (approved):
Looks good to me!

PR #1537, opened 2021-11-03
Design docs

toiletgranny:
Moving design docs from the `atlas-resources` to the main `atlas` repo.


kdembler (commented):
2 small point already brought up on RocketChat:
1) Let's call the directory with images `assets` instead of `src`
2) Let's try to remove whitespaces from the image names 🙏

toiletgranny:
@kdembler done!

kdembler (approved):

PR #1544, opened 2021-11-04
update dialogs/modals with new designs

kdembler:
The long-coming update for dialogs/modals is here!

Summary of changes:
- `BaseDialog` renamed to `Modal` - it doesn't contain any body now, just a logic to display an overlay over the page
- `ActionDialog` and `MessageDialog` merged as `Dialog` - it doesn't contain any modal-related logic now, it only concerns itself with the content. It was also updated to follow the new design - header/body/footer sections.
- `DialogModal` created for ease of use (just `Dialog` inside `Modal`)
- `ImageCropDialog` renamed to `ImageCropModal`
- `TransactionDialog` renamed to `TransactionModal`
- `Multistepper` renamed to `StepperModal`


drillprop (dismissed):
Tested all the modal scenarios and looked at the code, didn't find any issues. Looks great overall 👍🏻

kdembler:
Rebased but also had to make some additional changes. Because of that please review again. Summary of changes:
- Updated `PopoverDialog` - changed name to `DialogPopover` and made it use the new `Dialog` component
- Updated `FiltersBar` - removed some unnecessary styles and moved the playground to storybook instead, also made the list of categories come from the consumer so that component doesn't need to fetch by itself and rely on Apollo
- Made slight changes to the `Popover` API - previously it accepted `content` for content and `children` for trigger element, now it accepts `trigger` for trigger and `children` for content, should be more inline with Dialog API
- Updated styles of `Dialog` component so that content always has top&bottom padding to avoid scroll issues

src/shared/components/Dialog/Dialog.styles.ts
```diff
@@ -52,21 +47,17 @@ export const DialogContainer = styled.div<SizeProps>`
...
-`
-
-export const Header = styled.div<HeaderProps>`
+export const Header = styled.div<DividersProps>`
   display: flex;
   flex-direction: row;
   justify-content: space-between;
   align-items: flex-start;
-  padding: var(--local-size-dialog-padding);
+  padding: var(--local-size-dialog-padding) var(--local-size-dialog-padding) 0;
```
drillprop, line 58: according to the design on mobile padding is a little different when the title is present. It's 24px top and bottom not 16px. 
Yours: 
image
Design: 
image

drillprop: Also, I think the header should have bottom padding if it's possible, not the dialog content. Look at the screen:
image

(I know that it's an extreme situation, but maybe we could consider adding it if it's not problematic?)

kdembler: So regarding the first issue you brought up, I think the padding actually is 16px, you need to look at the biggest element in the header which is the close button:
Screenshot 2021-11-05 at 10 23 21

Regarding the second case for when we're scrolling I'm afraid you are right but I just don't want to fix it :P Adding the padding to the content instead of the header fixed the scrollbar issue I've mentioned. Also, in theory, whenever scroll will be needed inside the container, the designers should use the `dividers` variant so we shouldn't run into this case at all

drillprop: Regarding the first one - my mistake, you're right. Regarding the second point - we can ignore this.

drillprop (approved):

PR #1548, opened 2021-11-04
Apply generated variables to color theme

### LINKED ISSUE #1547: Apply generated css variables to `src/shared/theme/colors`

drillprop:
Fix #1547

### DIFF
--- a/src/shared/components/SkeletonLoader/SkeletonLoader.tsx
+++ b/src/shared/components/SkeletonLoader/SkeletonLoader.tsx
@@ -2,6 +2,8 @@ import { keyframes } from '@emotion/react'
 import styled from '@emotion/styled'
 import React from 'react'
 
+import { cVar } from '@/styles/generated/variables'
+
 export type SkeletonLoaderProps = {
   width?: string | number
   height?: string | number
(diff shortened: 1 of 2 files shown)


src/shared/components/SkeletonLoader/SkeletonLoader.tsx
```diff
@@ -65,12 +49,12 @@ const SkeletonLoaderAnimated = styled.div<SkeletonLoaderProps>`
   transform: translateX(0%);
   background: linear-gradient(
     104deg,
-    rgba(183, 200, 250, 0) 15%,
-    rgba(183, 200, 250, 0.06) 30%,
-    rgba(187, 217, 246, 0.13) 48%,
-    rgba(187, 217, 246, 0.13) 52%,
-    rgba(183, 200, 250, 0.06) 70%,
-    rgba(183, 200, 250, 0) 85%
+    rgba(0, 0, 0, 0) 15%,
```
rafalpawlow, line 52: Why have you changes this? According to this comment: https://github.com/Joystream/atlas/pull/1492#issuecomment-952601955
It should be `rgba(183, 200, 250, 0) 15%`

rafalpawlow (approved):
I only have one concern which I included in a comment. Other than that it looks good.

src/shared/components/SkeletonLoader/SkeletonLoader.tsx
kdembler: So the starting color has 0 opacity which makes it fully transparent. Bartosz did testing and turns out it doesn't matter what the base color is, if the opacity is 0, the transition will look the same. Here are the test results (one is transparent black, the other `rgba(183, 200, 250, 0)` :
Clipboard - 4 listopada 2021 12_47

kdembler (approved):

PR #1550, opened 2021-11-04
Handle search store migration

rafalpawlow:


src/providers/personalData/store.tsx
```diff
@@ -107,13 +107,15 @@ export const usePersonalDataStore = createStore<PersonalDataStoreState, Personal
     persist: {
       key: 'personalData',
       whitelist: WHITELIST,
-      version: 0,
+      version: 0.1,
```
kdembler: So far we have been using only integer versions for store

```suggestion
      version: 1,
```

src/providers/personalData/store.tsx
```diff
@@ -107,13 +107,15 @@ export const usePersonalDataStore = createStore<PersonalDataStoreState, Personal
...
       onRehydrateStorage: () => {
         WHITELIST.forEach((item) => {
           window.localStorage.removeItem(item)
         })
       },
-      migrate: () => null,
+      migrate: (oldState) => ({
+        recentSearches: oldState.filter((item: RecentSearch) => !Object.prototype.hasOwnProperty.call(item, 'type')),
```
kdembler: Old state here isn't an array but an object. If this migration is run, an exception will be thrown: `Failed to migrate store "personalData": e.filter is not a function`. You can test it out yourself by getting some recent searches on deployment with old search, use export state function from developer overlay and then import into app running those changes and refresh

kdembler (changes requested):

src/providers/personalData/store.tsx
```diff
@@ -107,13 +107,17 @@ export const usePersonalDataStore = createStore<PersonalDataStoreState, Personal
...
         WHITELIST.forEach((item) => {
           window.localStorage.removeItem(item)
         })
       },
-      migrate: () => null,
+      migrate: (oldState) => ({
+        recentSearches: oldState.recentSearches.filter(
+          (item: RecentSearch) => !Object.prototype.hasOwnProperty.call(item, 'type')
+        ),
+      }),
```
kdembler: This removed all the other data except `recentSearches` from `personalData` store

kdembler (changes requested):

kdembler (approved):

PR #1551, opened 2021-11-04
Import svg icons using figma API

### LINKED ISSUE #1041: Look into importing icons using Figma API
It's possible to extract icons from Figma programatically, without a need to go and manually download the files. We could have a script that fetches files from Figma and places them inside the codebase.

rafalpawlow:
Closes #1041

### DIFF
--- a/figma-import.config.js.example
+++ b/figma-import.config.js.example
@@ -1,0 +1,6 @@
+module.exports = {
+  FIGMA_TOKEN: '',
+  FILE_KEY: '',
+  FRAME_WITH_ICONS_ID: '',
+  CONCURRENCY: 1,
+}
--- a/figma-import.config.js.example
+++ b/figma-import.config.js.example
@@ -1,6 +1,12 @@
+/**
+ * FIGMA_TOKEN - your personal API token, generated in figma profile settings
+ * FILE_KEY - unique file id
+ * FRAME_WITH_ICONS_ID - frame id with assets to export
+ * e.g. https://www.figma.com/file/[FILE_KEY]/Joystream-Atlas?node-id=[FRAME_WITH_ICONS_ID]
+ */
+
 module.exports = {
   FIGMA_TOKEN: '',
   FILE_KEY: '',
   FRAME_WITH_ICONS_ID: '',
-  CONCURRENCY: 1,
 }
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "svgs:illustrations": "svgr --config-file svgr.config.js -d src/shared/illustrations src/shared/illustrations/svgs",
     "svgs": "yarn svgs:icons && yarn svgs:illustrations",
     "graphql": "graphql-codegen --config codegen.config.yml",
-    "figma-import": "node ./scripts/figma-import"
+    "svgs:import": "node ./scripts/figma-import"
   },
   "lint-staged": {
     "*.{ts,tsx,js,jsx,json}": [
@@ -64,7 +64,6 @@
     "bezier-easing": "^2.1.0",
     "bn.js": "^4.12.0",
     "buffer": "^5.7.1",
-    "change-case": "^4.1.2",
     "cropperjs": "^1.5.12",
     "date-fns": "^2.25.0",
     "downshift": "^6.1.0",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "svgs:icons": "svgr --config-file svgr.config.js -d src/shared/icons src/shared/icons/svgs",
     "svgs:illustrations": "svgr --config-file svgr.config.js -d src/shared/illustrations src/shared/illustrations/svgs",
     "svgs": "yarn svgs:icons && yarn svgs:illustrations",
-    "graphql": "graphql-codegen --config codegen.config.yml"
+    "graphql": "graphql-codegen --config codegen.config.yml",
+    "figma-import": "node ./scripts/figma-import"
   },
   "lint-staged": {
     "*.{ts,tsx,js,jsx,json}": [
@@ -61,8 +62,10 @@
     "awesome-debounce-promise": "^2.1.0",
     "axios": "^0.21.1",
     "bezier-easing": "^2.1.0",
+    "bluebird": "^3.7.2",
     "bn.js": "^4.12.0",
     "buffer": "^5.7.1",
+    "change-case": "^4.1.2",
     "cropperjs": "^1.5.12",
     "date-fns": "^2.25.0",
     "downshift": "^6.1.0",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "awesome-debounce-promise": "^2.1.0",
     "axios": "^0.21.1",
     "bezier-easing": "^2.1.0",
-    "bluebird": "^3.7.2",
     "bn.js": "^4.12.0",
     "buffer": "^5.7.1",
     "change-case": "^4.1.2",
--- a/scripts/figma-import/icons.js
+++ b/scripts/figma-import/icons.js
@@ -1,8 +1,8 @@
-const config = require('../../figma-import.config')
+const config = require('./figma-import.config')
 
 const path = require('path')
 const fs = require('fs').promises
-const { paramCase } = require('change-case')
+const { kebabCase } = require('lodash')
 
 const { getImageContent, getNodeChildren, getSvgImageUrl } = require('./utils/api')
 
@@ -34,7 +34,7 @@ const clearIconsDir = async () => {
 const generateIcon = async (iconNode, total) => {
   const iconUrl = await getSvgImageUrl(iconNode.id)
 
-  const iconName = paramCase(iconNode.name)
+  const iconName = kebabCase(iconNode.name)
 
   try {
     await fs.access(iconsDir)
--- a/scripts/figma-import/icons.js
+++ b/scripts/figma-import/icons.js
@@ -18,7 +18,7 @@ let counter = 0
  */
 const clearIconsDir = async () => {
   try {
-    await fs.rmdir(IconsDir, { recursive: true })
+    await fs.rm(IconsDir, { recursive: true })
     console.log(`${IconsDir} successfully deleted!`)
   } catch (err) {
     console.error(`Error while deleting ${IconsDir}.`)
--- a/scripts/figma-import/icons.js
+++ b/scripts/figma-import/icons.js
@@ -1,10 +1,8 @@
 const config = require('../../figma-import.config')
 
 const path = require('path')
-const fs = require('fs')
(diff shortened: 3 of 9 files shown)


figma-import.config.js.example
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  FIGMA_TOKEN: '',
+  FILE_KEY: '',
+  FRAME_WITH_ICONS_ID: '',
+  CONCURRENCY: 1,
+}
```
kdembler: This file needs some documentation, I have no idea what this options are supposed to be supplied with

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,81 @@
+const config = require('../../figma-import.config')
+
+const path = require('path')
+const fs = require('fs')
+const { exec } = require('child_process')
+const { pascalCase } = require('change-case')
+const Promise = require('bluebird')
```
kdembler: What is this, why not use built-in promises?

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,81 @@
...
+const writeFile = Promise.promisify(fs.writeFile)
+let counter = 0
+
+/**
+ * clear icons dir
+ *
+ */
+const clearIconsDir = () => {
+  exec(`rm -rf ${IconsDir}`)
+}
```
kdembler: This could be done via node API, without invoking shell directly, right?

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,81 @@
...
+
+let counter = 0
+
+/**
+ * clear icons dir
+ *
+ */
+const clearIconsDir = async () => {
+  try {
+    await fs.rmdir(IconsDir, { recursive: true })
```
drillprop: Im getting this in the console: 
```
(node:22653) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
```
This is probably due the fact that, Im using node 16, but I think ` fs.rm(IconsDir, { recursive: true })` should also works on node 14 and we can safely replace it.

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,81 @@
+const config = require('../../figma-import.config')
+
+const path = require('path')
+const fs = require('fs').promises
+const { pascalCase } = require('change-case')
+
+const { getImageContent, getNodeChildren, getSvgImageUrl } = require('./utils/api')
+
+const IconsDir = path.resolve(__dirname, '../../src/shared/icons/figmaSvgs')
```
drillprop: let's write this variable with camelCase, maybe? `iconsDir`

drillprop:
```suggestion
const iconsDir = path.resolve('src/shared/icons/figmaSvgs')
```

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,81 @@
...
+ * [iconName].svg files
+ *
+ * @param {object} iconNode
+ * @param {number} total
+ * @return {Promise<void>}
+ */
+const generateIcon = async (iconNode, total) => {
+  const iconUrl = await getSvgImageUrl(iconNode.id)
+
+  const iconName = pascalCase(iconNode.name)
```
drillprop: I think we could stick to `kebabCase` names like we have in `src/shared/icons/svgs`. 
We also don't need `change-case` library, we could just use `kebabCase` from `lodash`

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,81 @@
...
+ * clear icons dir
+ *
+ */
+const clearIconsDir = async () => {
+  try {
+    await fs.rm(IconsDir, { recursive: true })
+    console.log(`${IconsDir} successfully deleted!`)
+  } catch (err) {
+    console.error(`Error while deleting ${IconsDir}.`)
+  }
```
drillprop:
```suggestion
    await fs.rm(iconsDir, { recursive: true })
    console.log(`${iconsDir} successfully deleted!`)
  } catch (err) {
    console.error(`Error while deleting ${iconsDir}.`)
  }
```

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,81 @@
...
+
+const path = require('path')
+const fs = require('fs').promises
+const { pascalCase } = require('change-case')
+
+const { getImageContent, getNodeChildren, getSvgImageUrl } = require('./utils/api')
+
+const IconsDir = path.resolve(__dirname, '../../src/shared/icons/figmaSvgs')
+
+const getIconFolderPath = () => path.resolve(IconsDir)
```
drillprop: Why do we need this function? Is it `iconsDir` enough?

drillprop (commented):
Missed some things.
Also, small suggestion - maybe we could throw some error, which notifies that you didn't provide values for `figma-import.config.js`?
In general works really great 👍🏻

scripts/figma-import/icons.js
rafalpawlow: Unfortunately I wasn't able to use `lodash-es` because of CommonJS usage in this script `require() of ES modules is not supported.`. Do you think we can live with that? I think it much less effort than changing everything to ES modules.

package.json
```diff
@@ -31,7 +31,8 @@
     "svgs:icons": "svgr --config-file svgr.config.js -d src/shared/icons src/shared/icons/svgs",
     "svgs:illustrations": "svgr --config-file svgr.config.js -d src/shared/illustrations src/shared/illustrations/svgs",
     "svgs": "yarn svgs:icons && yarn svgs:illustrations",
-    "graphql": "graphql-codegen --config codegen.config.yml"
+    "graphql": "graphql-codegen --config codegen.config.yml",
+    "figma-import": "node ./scripts/figma-import"
```
kdembler: Let's call this `svgs:import` maybe?

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,78 @@
+const config = require('../../figma-import.config')
```
kdembler: I'd suggest to keep this config in the same directory as the script to not pollute the top-level directory

figma-import.config.js.example
```diff
@@ -0,0 +1,12 @@
...
+ * FILE_KEY - unique file id
+ * FRAME_WITH_ICONS_ID - frame id with assets to export
+ * e.g. https://www.figma.com/file/[FILE_KEY]/Joystream-Atlas?node-id=[FRAME_WITH_ICONS_ID]
+ */
+
+module.exports = {
+  FIGMA_TOKEN: '',
+  FILE_KEY: '',
+  FRAME_WITH_ICONS_ID: '',
+}
```
kdembler: I wanted to test it out but I still didn't know what to provide for the last two fields of config. I think we should either provide detailed description on what to select in Figma or just provide those values supplied here already

scripts/figma-import/icons.js
drillprop: hmm, I guess I'm fine with using it

drillprop: Oh, sorry, consulted this with Klaudiusz. Let's just use `lodash`( the one without `-es`). I'm using it as well in `sd.config.js` and it works.

scripts/figma-import/icons.js
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * generate icons components
+ *
+ * @param {[Object]} iconNodesArr - array of icon nodes from frame
+ * @return {Promise<void>}
+ */
+const generateIcons = async (iconNodesArr) => {
+  await Promise.all(
+    iconNodesArr.map((item, index) => {
+      setTimeout(() => generateIcon(item, iconNodesArr.length), 1000 * index)
```
kdembler, line 65: What's happening here? Seems we're forcing this process to take at least 3 minutes (we have ~180 icons)

kdembler:
I see this PR successfully fetches the SVGs but we're still using the old ones in the codebase to generate icon components. Do you plan to work on this as well?

kdembler:
Also, there are the following problems:
- Some of the icon names overlap between icon types and we end up overwriting the results. For example the icons file defines `image-file` as 16x16 but also as 24x24. Because we use a single name for this, our code will end up writing only one of those. You can also see this with the social platforms icons, with the first variant being called `on-light` for all the different platforms.
- Previously we used prefixes to more easily distinct between different icon types (action, glyph, etc.). The design currently uses slightly different names but our implementation doesn't use any prefixes. I would like us to keep using them (we can use the new names).

So, above points both boil down to the names used in Figma. As far as I understand we won't be able to figure out the icon type ourselves since the export file doesn't contain any information on that. However, I think we could ask Adam to tweak the names of nodes used in the export file so we keep the type information. Wdyt?

scripts/figma-import/icons.js
rafalpawlow: We don't have choice due to figma api limitations. Without this or with smaller timeout value this process is failing.

kdembler: Ah, so this is the solution to the too many requests problem you've mentioned?

figma-import.config.js.example
rafalpawlow: I don't know how to write this without sounding passive aggressive 😛, but TBH I'm not sure what is unclear in the description above, and I think providing values here is not a good idea because, we might commit `figma-import.config.js` as well. Everything you need is in the figma url with icons, as I showed in the example here https://github.com/Joystream/atlas/blob/835367d89084df82fd095c82e7bb76c2608709a3/figma-import.config.js.example#L5
Anyway I improved description a little bit

scripts/figma-import/icons.js
rafalpawlow: Exactly

figma-import.config.js.example
kdembler: So I see the following problems:
1. It's not immediately clear which file should be used to extract icons.
2. I have no idea what `FRAME_WITH_ICONS_ID` should be, i.e. how I get this value. We should make sure that someone that isn't able to ask this questions is able to figure it out by themselves so I think we need to improve the instructions a bit.

> I think providing values here is not a good idea because, we might commit figma-import.config.js as well

I don't really understand what you mean by that.

scripts/figma-import/icons.js
kdembler: All right, sounds reasonable

figma-import.config.js.example
rafalpawlow: Please see if it's better now

kdembler: Looks good, thanks for improving this

kdembler (approved):

PR #1559, opened 2021-11-05
Fix other filters logic

### LINKED ISSUE #1556: Other filters in search results page don't act like described
Those filters should exclude the given types of content, not filter by them

rafalpawlow:
Fixes #1556

### DIFF
--- a/src/api/client/cache.ts
+++ b/src/api/client/cache.ts
@@ -54,8 +54,10 @@ const getSearchKeyArgs = (args: Record<string, SearchQueryVariables['whereVideo'
   const languageIdEq = args?.whereVideo?.languageId_eq || ''
   const categoryIdIn = args?.whereVideo?.categoryId_in || []
   const createdAtGte = args?.whereVideo?.createdAt_gte ? JSON.stringify(args.whereVideo.createdAt_gte) : ''
+  const durationGte = args?.whereVideo?.duration_gte || null
+  const durationLte = args?.whereVideo?.duration_lte || null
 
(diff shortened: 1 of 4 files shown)


src/api/client/cache.ts
```diff
@@ -49,13 +49,15 @@ const getChannelKeyArgs = (args: Record<string, GetChannelsConnectionQueryVariab
 const getSearchKeyArgs = (args: Record<string, SearchQueryVariables['whereVideo']> | null) => {
   const text = args?.text || ''
   const licenseIdIn = args?.whereVideo?.licenseId_in || []
-  const hasMarketingEq = args?.whereVideo?.hasMarketing_eq || ''
-  const isExplicitEq = args?.whereVideo?.isExplicit_eq || ''
+  const hasMarketingEq = args?.whereVideo?.hasMarketing_eq || true
+  const isExplicitEq = args?.whereVideo?.isExplicit_eq || true
```
kdembler: Why are we assuming `true` for those if they're missing? This can lead to wrong results

src/hooks/useSearchResults.ts
```diff
@@ -42,6 +42,7 @@ export const useSearchResults = ({ searchQuery, limit = 50, videoWhereInput }: S
       },
     },
     {
+      fetchPolicy: 'no-cache',
```
kdembler: Why `no-cache`? This will force us to make network request on every filter change even if we already know the result, right?

kdembler (approved):

PR #1560, opened 2021-11-05
Fix colors in tertiary button and in transaction dialog

drillprop:


src/components/TransactionModal/TransactionModal.style.ts
```diff
@@ -123,7 +123,8 @@ export const Success = styled.div`
   display: flex;
   align-items: center;
   justify-content: center;
-  background: ${rgba(colors.blue[500], 0.2)};
+  /* TODO replace with variable: blue 500 */
+  background: ${rgba('#4038ff', 0.2)};
```
kdembler: I'm not really sure what the actionable item here is. Could you do a simple structure change here so that the background is a separate element and we can apply `opacity` to it?

src/shared/components/ButtonBase/ButtonBase.style.ts
```diff
@@ -84,11 +84,11 @@ const variantStyles = ({ variant, textOnly, iconOnly }: ButtonBaseStyleProps): S
 
         &:hover,
         &:focus {
-          background-color: ${colors.transparentPrimary[12]};
+          background-color: ${colors.transparentPrimary[18]};
         }
 
         &:active {
-          background-color: ${colors.transparentPrimary[6]};
+          background-color: ${colors.transparentPrimary[10]};
```
kdembler, line 91: We should do the same updates in `HamburgerButton.style.ts`. Unfortunately those 2 components share their styles but aren't connected in the implementation

kdembler (approved):

PR #1563, opened 2021-11-08
Replace urls in video description

rafalpawlow:


src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -104,6 +104,24 @@ export const VideoView: React.FC = () => {
     }
   }, [video?.id, handleTimeUpdate, updateWatchedVideos])
 
+  const replaceUrls = (text: string) => {
+    const urlRegex = /(https?:\/\/[^\s]+)/g
```
drillprop, line 108: not sure, but maybe we could support also addresses that start with `www`?

src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -104,6 +104,24 @@ export const VideoView: React.FC = () => {
...
+    for (let i = 1; i < parts.length; i += 2) {
+      const part = parts[i] as string
+      parts[i] = (
+        <Button
+          textOnly
+          key={`description-link-${i}`}
+          to={part.startsWith(window.origin) ? part.replace(window.origin, '') : part}
+        >
+          {part}
+        </Button>
```
drillprop, line 115: @toiletgranny could you look at this? Rafał said to me that you wanted to use our button component here. The problem is we used different font-family and different font-size for content. It looks a little bit odd to me.
image

drillprop: BTW you can test it here https://deploy-preview-1563--atlas-dev.netlify.app/video/7175 (you need to switch to main testnet)

toiletgranny: @drillprop In terms of the size, I would recommend simply using a `Large` variant of the Button component, which comes with a `16px` text label.

In terms of the font family, I wouldn't worry about that right now, because we'll be introducing new typography tokens soon anyway, which will eventually replace all `PxGrotesk` instances in buttons and other component labels with `Inter`.

drillprop (approved):

src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -104,6 +104,25 @@ export const VideoView: React.FC = () => {
...
+    const urlRegex = /(https?:\/\/[^\s]+|www.[^\s]+)/g
+    const parts: string[] | React.ReactNodeArray = text.split(urlRegex) || []
+    for (let i = 1; i < parts.length; i += 2) {
+      const part = parts[i] as string
+      parts[i] = (
+        <Button
+          size="large"
+          textOnly
+          key={`description-link-${i}`}
+          to={part.startsWith(window.origin) ? part.replace(window.origin, '') : part}
```
kdembler: Why do we need this logic vs just using the link directly?

Issue #1508, opened 2021-10-21
Discovery milestone implementation feedback


kdembler:
Couple of things we should fix:

- [x] No `Discovery` navigation item in sidenav
- [x] It's not possible to navigate via category cards
- [x] On category view, title of the section changes from `All videos ()` to `All videos(xxx)`. Let's improve the loading state and use just `All videos` when the number isn't available
- [x] The `Other filters` section is supposed to exclude selected items, not filter by them as it does currently
- [x] The loading state should be triggered when filters are changes to show the user something is happening. Currently there is no indication that new filters were applied

https://user-images.githubusercontent.com/12646744/138249368-e476668c-c234-4509-95d4-ad7792b9e1a5.mov

kdembler:
- [x] There's no spacing between the last category card and the end of the page (vertically). The design specifies 64px there
- [x] The hover interaction with the category card can be broken by placing cursor on the edge of the card. Something like this has been fixed recently for VideoTiles by @drillprop, I suggest to reach out to him for quick fix

https://user-images.githubusercontent.com/12646744/140294205-e9cf2370-e6a8-47ba-8666-a30bdbafca1e.mov

- [x] Navigation via the category card doesn't work properly. It seems it's not handled by react router and navigation causes the full page to reload
- [x] Changing a filter causes the page to scroll down

https://user-images.githubusercontent.com/12646744/140294852-0d6d36cd-6035-4e41-9ec4-70799b141373.mov

PR #1570, opened 2021-11-09
Category wrapup

eldiegod:
Address last feedback on category pages both from:
https://github.com/Joystream/atlas/issues/1508
https://github.com/Joystream/atlas/pull/1518

fixes: #1381

### DIFF
--- a/src/components/FiltersBar/FiltersBar.tsx
+++ b/src/components/FiltersBar/FiltersBar.tsx
@@ -34,14 +34,12 @@ type FilterCategory = {
 
 export type FiltersBarProps = {
   categories?: FilterCategory[]
-  mobileLanguageSelector?: boolean
 }
 
 export const FiltersBar: React.FC<ReturnType<typeof useFiltersBar> & FiltersBarProps> = ({
   setVideoWhereInput,
   videoWhereInput,
   categories,
-  mobileLanguageSelector,
   filters: {
     setIsFiltersOpen,
     isFiltersOpen,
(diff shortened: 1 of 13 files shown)


rafalpawlow (changes requested):
Let's add padding to results when filters are open, to avoid covering video grid. Like on the search result page. Other than this, LGTM.
Zrzut ekranu 2021-11-9 o 10 34 06
.

src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -277,7 +273,7 @@ export const FiltersBar: React.FC<ReturnType<typeof useFiltersBar> & FiltersBarP
               hasMarketing_eq: excludePaidPromotionalMaterialFilter ? !excludePaidPromotionalMaterialFilter : undefined,
               isExplicit_eq: excludeMatureContentRatingFilter ? !excludeMatureContentRatingFilter : undefined,
               categoryId_in: categoriesFilter,
-              languageId_eq: language as string,
+              languageId_eq: language === 'undefined' ? undefined : language,
```
rafalpawlow, line 276: Why can't we use just `languageId_eq: language`?

eldiegod: because Adam changed the language select a little bit and now has an `All languages` default option but this option can't be undefined in the `select` but it has to be undefined in the `languageId_eq`

rafalpawlow (approved):

PR #1572, opened 2021-11-09
Avoid rendering video tile context menu while uploading

### LINKED ISSUE #1566: VideoTIle in uploading variant has invisible context menu
Even though there are is no visible trigger button, you can still open the context menu when the video is uploading. We should disable it as the design doesn't specify context menu for this variant

https://user-images.githubusercontent.com/12646744/140779188-bddfa8f2-08c1-4584-b031-4c70d1dd2695.mov

rafalpawlow:
Fixes #1566

### DIFF
--- a/src/shared/components/ContextMenu/ContextMenu.tsx
+++ b/src/shared/components/ContextMenu/ContextMenu.tsx
@@ -8,6 +8,7 @@ type MenuItemProps = {
   icon: ReactNode
   title: string
   onClick?: () => void
+  disabled?: boolean
 }
 
 export const ContextMenuItem: React.FC<MenuItemProps> = ({ icon, onClick, title }) => {
@@ -21,10 +22,10 @@ export const ContextMenuItem: React.FC<MenuItemProps> = ({ icon, onClick, title 
 
 type ContextMenuProps = { items: MenuItemProps[] } & Omit<PopoverProps, 'content' | 'instanceRef'>
 
-export const ContextMenu: React.FC<ContextMenuProps> = ({ children, items, ...rest }) => {
+export const ContextMenu: React.FC<ContextMenuProps> = ({ children, items, disabled, ...rest }) => {
   const contextMenuInstanceRef = useRef<PopoverImperativeHandle>(null)
   return (
-    <Popover hideOnClick ref={contextMenuInstanceRef} {...rest}>
+    <Popover hideOnClick ref={contextMenuInstanceRef} disabled={disabled} {...rest}>
       <StyledContainer>
         {items.map((item, index) => (
           <ContextMenuItem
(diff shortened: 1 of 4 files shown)


src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -272,25 +272,27 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
...
+                      icon: <SvgGlyphCopy />,
+                      onClick: onCopyVideoURLClick,
+                      title: 'Copy video URL',
+                    },
+                  ]
+            }
+            trigger={
+              <KebabMenuButtonIcon onClick={() => null} variant="tertiary" size="small" disabled>
+                <SvgGlyphMore />
+              </KebabMenuButtonIcon>
```
kdembler: Why have you added the `disabled` prop here? This makes all the buttons visible right away, without a hover.

src/shared/components/VideoTileBase/VideoTileBase.tsx
```diff
@@ -272,25 +272,27 @@ export const VideoTileBase: React.FC<VideoTileBaseProps> = ({
...
-                  },
-                ]
-          }
-          trigger={
-            <KebabMenuButtonIcon onClick={() => null} variant="tertiary" size="small">
-              <SvgGlyphMore />
-            </KebabMenuButtonIcon>
-          }
-        />
+        {!isUploading && (
```
kdembler: I'm not sure fully hiding the trigger is the right way to go. This will make the available space change between uploading and regular tile and can cause some layout shifts. Instead we could add a prop to `KebabMenuButtonIcon` to make the button present in the layout but disabled

src/shared/components/ContextMenu/ContextMenu.tsx
```diff
@@ -21,10 +22,10 @@ export const ContextMenuItem: React.FC<MenuItemProps> = ({ icon, onClick, title
 
 type ContextMenuProps = { items: MenuItemProps[] } & Omit<PopoverProps, 'content' | 'instanceRef'>
 
-export const ContextMenu: React.FC<ContextMenuProps> = ({ children, items, ...rest }) => {
+export const ContextMenu: React.FC<ContextMenuProps> = ({ children, items, disabled, ...rest }) => {
   const contextMenuInstanceRef = useRef<PopoverImperativeHandle>(null)
   return (
-    <Popover hideOnClick ref={contextMenuInstanceRef} {...rest}>
+    <Popover hideOnClick ref={contextMenuInstanceRef} disabled={disabled} {...rest}>
```
kdembler: Doesn't seem we need to extract this specifically, `{...rest}` should already do the job right?

kdembler (approved):

PR #1573, opened 2021-11-09
remove temporary hardcoded tokens & rename actions

kdembler:
changes:
- renamed scripts to be more consistent
- moved tokens script to top-level `scripts` directory
- removed temporary `@/shared/theme/tokens`


package.json
```diff
@@ -25,14 +25,14 @@
...
-    "codegen:illustrations": "svgr --config-file svgr.config.js -d src/shared/illustrations src/shared/illustrations/svgs",
-    "codegen:svgs": "yarn codegen:illustrations && yarn codegen:icons",
-    "codegen:styles": "yarn fetch-tokens && yarn build-tokens"
+    "tokens:fetch": "node ./scripts/fetchTokens && prettier --write src/styles/tokens",
+    "tokens:build": "style-dictionary build --config ./sd.config.js && prettier --write src/styles/generated/variables.ts",
+    "tokens": "yarn tokens:fetch && yarn tokens:build",
+    "svgs:icons": "svgr --config-file svgr.config.js -d src/shared/icons src/shared/icons/svgs",
+    "svgs:illustrations": "svgr --config-file svgr.config.js -d src/shared/illustrations src/shared/illustrations/svgs",
+    "svgs": "yarn svgs:icons && yarn svgs:illustrations",
+    "graphql:watch": "yarn codegen:graphql --watch",
```
drillprop: This is wrong. Should be 
```suggestion
    "graphql:watch": "yarn graphql --watch",
```
But on the second thought - do we actually need this script? I never used this and If someone needs that he could just run: 
```
yarn run graphql --watch
```

drillprop (changes requested):

package.json
kdembler: True, will remove

drillprop (approved):

PR #1580, opened 2021-11-10
Fix no empty fallback for channels

### LINKED ISSUE #1576: No empty state for Channels tab in search results
Screenshot 2021-11-09 at 17 17 15

rafalpawlow:
Fix #1576

### DIFF
--- a/src/views/viewer/SearchView/SearchResults/SearchResults.tsx
+++ b/src/views/viewer/SearchView/SearchResults/SearchResults.tsx
@@ -99,7 +99,9 @@ export const SearchResults: React.FC<SearchResultsProps> = React.memo(({ query }
       </SearchControls>
       <Results filtersOpen={isFiltersOpen}>
         <LimitedWidthContainer big>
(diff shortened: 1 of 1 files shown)


src/views/viewer/SearchView/SearchResults/SearchResults.tsx
```diff
@@ -99,7 +99,9 @@ export const SearchResults: React.FC<SearchResultsProps> = React.memo(({ query }
       </SearchControls>
       <Results filtersOpen={isFiltersOpen}>
         <LimitedWidthContainer big>
-          {!loading && channels.length === 0 && videos.length === 0 && !!query ? (
+          {!loading &&
+          ((videos.length === 0 && selectedIndex === 0) || (channels.length === 0 && selectedIndex === 1)) &&
+          !!query ? (
```
drillprop: What's is selectedIndex in this context? Is it a selected tab index?
If yes, could we change this to `selectedTabIndex`?
Also, could we store this long expression in a variable?  I think it should be more readable.

drillprop (approved):

PR #1582, opened 2021-11-10
Stretch create channel form when creating first channel

### LINKED ISSUE #1485: Make the channel edit view stretch full width when creating first channel
Seems like there is a margin on the left hand side when trying to create your first channel during onboarding, where the consumer side menu would normally be.

Screenshot 2021-10-13 at 17 21 41

rafalpawlow:
Fix #1485

### DIFF
--- a/src/views/studio/StudioLayout.tsx
+++ b/src/views/studio/StudioLayout.tsx
@@ -76,7 +76,7 @@ const StudioLayout = () => {
         nodeConnectionStatus={nodeConnectionStatus}
         isConnectedToInternet={internetConnectionStatus === 'connected'}
       />
-      <TopbarStudio hideChannelInfo={!memberSet} />
+      <TopbarStudio noSidebar={!channelSet} hideChannelInfo={!memberSet} />
       <CSSTransition
         in={channelSet}
         timeout={parseInt(transitions.timings.regular)}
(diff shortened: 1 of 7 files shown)


eldiegod (dismissed):
Looks good.

Only one detail, I would say for clarity, name `noSidebar` prop `hasSidebar` instead and negate that when needed.
Having negated props by default makes it harder to think about what the result is i.e noSidebar=false I have to stop for a second to understand if this will have sidebar or not

kdembler:
I think we could improve this solution. One thing I'd like to avoid is the studio logo jumping left and right, some time ago we introduced a change that makes the logo consistent in place and I'd like to keep it this way. This also introduces visual issues when the sidenav is disappearing, for example when logging out.

Another issue I see is the amount of components that we need to pass the `hasSidebar` prop to. Maybe in `StudioLayout` we could define some CSS variable, like `--size-sidenav-current-width` that would be 0 if channel is not set. Then channel edit view could use that to style action wrapper?

src/views/studio/StudioLayout.tsx
```diff
@@ -147,7 +147,9 @@ const StudioLayout = () => {
   )
 }
 
-const MainContainer = styled.main`
+const MainContainer = styled.main<{ hasSidebar: boolean }>`
+  --size-sidenav-width-collapsed: ${({ hasSidebar }) => (hasSidebar ? '72px' : 0)};
```
kdembler: This will break the mobile views because you set this to constant `72px` although it changes to 0 on mobile. We should use the original `--size-sidenav-width-collapsed` value here that changes automatically. However that means we need to introduce a separate variable

kdembler (approved):

PR #1583, opened 2021-11-10
Fix ranking video gallery shift

### LINKED ISSUE #1575: Layout shift in numbered carousel
Once the loading is finished, you can see the numbers moving down:

https://user-images.githubusercontent.com/12646744/140962104-2d04a777-0832-4dfd-909d-64d79f5135bd.mov

rafalpawlow:
Fix #1575

### DIFF
--- a/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
+++ b/src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
@@ -2,7 +2,7 @@ import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 import { Link } from 'react-router-dom'
 
-import { colors, sizes, square, typography } from '@/shared/theme'
+import { colors, media, sizes, square, typography } from '@/shared/theme'
 
 import { Avatar } from '../Avatar'
 import { IconButton } from '../IconButton'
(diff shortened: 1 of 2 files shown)


eldiegod (approved):
Does the trick

src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
```diff
@@ -85,6 +85,15 @@ export const Container = styled.article<LoadingProps>`
...
   flex-direction: column;
+  min-height: 250px;
+
+  ${media.sm} {
+    min-height: 260px;
+  }
+
+  ${media.xxl} {
+    min-height: 270px;
+  }
```
kdembler: Is the video tile height really the problem here? I would expect that this is more of an layout/alignment issue with the big number.

kdembler (changes requested):
Problem still persists on some screen sizes:

https://user-images.githubusercontent.com/12646744/141782979-da4437ac-e0e1-4865-9145-dae0387886d4.mov

src/shared/components/VideoTileBase/VideoTileBase.styles.tsx
rafalpawlow: Unfortunately it uses `align-items: center;`, problem appears when loaded video has more lines than skeleton. I will try figure out some other way to position it.

kdembler (approved):

PR #1585, opened 2021-11-12
Fix duplicated play button on mobile

### LINKED ISSUE #1471: Duplicated play icon on iOS
IMG_2300 2

rafalpawlow:
Fixes #1471

### DIFF
--- a/src/shared/components/VideoPlayer/VideoPlayer.tsx
+++ b/src/shared/components/VideoPlayer/VideoPlayer.tsx
@@ -499,20 +499,22 @@ const VideoPlayerComponent: React.ForwardRefRenderFunction<HTMLVideoElement, Vid
               />
               <CustomControls isFullScreen={isFullScreen} isEnded={playerState === 'ended'}>
                 <PlayControl isLoading={playerState === 'loading'}>
-                  <PlayButton
-                    isEnded={playerState === 'ended'}
-                    onClick={handlePlayPause}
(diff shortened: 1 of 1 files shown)


src/shared/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -499,20 +499,22 @@ const VideoPlayerComponent: React.ForwardRefRenderFunction<HTMLVideoElement, Vid
...
-                  >
-                    {playerState === 'ended' ? (
-                      <SvgPlayerRestart />
-                    ) : isPlaying ? (
-                      <SvgPlayerPause />
-                    ) : (
-                      <SvgPlayerPlay />
-                    )}
-                  </PlayButton>
+                  {!showBigPlayButton && (
```
drillprop: This will remove play button on desktop view.
Zrzut ekranu 2021-11-15 o 12 47 17


@toiletgranny what do you think about it?

toiletgranny: Thanks for including me in this! 

Frankly, I would recommend having the small play button visible and active in the above case scenario. This might be useful for users that have already developed a "muscle memory" for playing/pausing the video in the bottom left corner of the video player.

To have the big play button in the middle, as an addition, is ok for me, by the way.

drillprop (approved):

PR #1592, opened 2021-11-16
Storybook housekeep

### LINKED ISSUE #1333: Storybook controls not working properly for lots of components
If you take a look at some storybook components, lots of controls are disabled or not working properly. For instance, this is from Snackbar: 
image
Also, we are getting lots of console warnings about deprecated properties 
image
This is the effect of updating the storybook(the API changed a bit) which happened together with the vite migration.

We should go through all of these storybook components, update them, and make sure that everything works as before.

eldiegod:
fixes: #1333


kdembler (changes requested):
Okay, we will need to make changes here. It's hard for me to tell what exactly is the issue (I guess the new docgen plugin), but the build process is broken. Some stats to compare:
- Build time: ~40s -> ~2min
- RAM usage: 2.7GB -> 7GB 🤯
- `dist` size after build: 6MB -> 30MB 🤯

This is not acceptable just to make Storybook props work. So we either need to figure out what the issue is and fix it (there should be no build regression) or just use something different.

drillprop:
I don't think it works as it should. For instance, we still need to provide some default text, when the prop type is a `string`. You could try it yourself by removing `arg.title` from `src/shared/components/ChannelCardBase/ChannelCardBase.stories.tsx`.  You'll get something like this:
image



Ideally, it should work just out of the box i.e. you shouldn't need to set args by yourself, and you should see here just text input.
I mean, It worked like that before when we had `create-react-app`.  To prove my words I just move back to the commit when we had CRA and this is how `TextArea` controls looked: 
image
And we didn't need to provide `args` and `argTypes`, every control input was just automatically added. 

Not sure what to do with this. I'm a little worried that there is no solution to resolve this for now.

kdembler:
I just asked Diego to skip this part of the PR and we'll try to solve it at some later time

src/shared/components/OptionCard/OptionCard.stories.tsx
```diff
@@ -4,7 +4,7 @@ import React, { useState } from 'react'
 import { OptionCard, OptionCardProps } from './OptionCard'
 
 export default {
-  title: 'Shared/O/OptionCard',
+  title: 'other/OptionCard',
```
drillprop: let's move this to `inputs`

src/shared/components/HelperText/HelperText.stories.tsx
```diff
@@ -4,7 +4,7 @@ import React from 'react'
 import { HelperText, HelperTextProps } from './HelperText'
 
 export default {
-  title: 'Shared/H/HelperText',
+  title: 'other/HelperText',
```
drillprop: I think this one is used only for inputs. We could move it there

src/components/NoConnectionIndicator/NoConnectionIndicator.stories.tsx
```diff
@@ -7,7 +7,7 @@ import { Snackbars } from '@/providers/snackbars'
 import { NoConnectionIndicator, NoConnectionIndicatorProps } from './NoConnectionIndicator'
 
 export default {
-  title: 'General/NoConnectionIndicator',
+  title: 'other/NoConnectionIndicator',
```
drillprop: Maybe we should move it to overlays? I mean, it basically behaves like a modal

.storybook/main.ts
```diff
@@ -23,10 +23,16 @@ const config: CustomizedStorybookConfig = {
...
+      (plugin) =>
+        // @ts-ignore totally justified ignore
+        plugin.name !== 'vite:react-babel' &&
+        // @ts-ignore totally justified ignore
+        plugin.name !== 'vite:react-refresh' &&
+        // @ts-ignore totally justified ignore
+        plugin.name !== 'vite:react-jsx' &&
+        // @ts-ignore totally justified ignore
+        plugin.name !== 'vite-plugin-checker'
     )
```
drillprop, line 35: Just a suggestion: 
```tsx
    const ignoredPlugins = ['vite:react-babel', 'vite:react-refresh', 'vite:react-jsx', 'vite-plugin-checker']
    const filteredVitePlugins = vitePlugins.filter(
      (plugin) =>
        // @ts-ignore totally justified ignore
        !ignoredPlugins.includes(plugin.name)
    )
```

drillprop (approved):
Looks good to me, too bad we didn't bring back inferring arguments. I hope that future releases of storybook/storybook-vite-builder will resolve this.

kdembler:
Yeah I think it's just too much trouble to fix it ourselves at this point, hopefully support comes soon

kdembler (approved):

PR #1598, opened 2021-11-16
Handle quota storage error

### LINKED ISSUE #701: No meaningful error when storage quota exceeded
A user hit the storage quota for their channel and recieved the following error when trying to upload:
image

Their storage quota was full:
```size_limit: 10,000,000,000,
  objects_limit: 500,
  size_used: 9,853,065,660,
  objects_used: 30```

It would be better if Joystream Studio could:
* Show the storage quota status
* Tell a user when they are exceeding the limit

### PR DESCRIPTION
**Pull request by rafalpawlow**
Fixes #701

### DIFF
--- a/src/components/TransactionModal/TransactionModal.constants.ts
+++ b/src/components/TransactionModal/TransactionModal.constants.ts
@@ -70,4 +70,17 @@ export const TRANSACTION_STEPS_DETAILS = {
       loop: false,
     },
   },
+  [ExtrinsicStatus.VoucherSizeLimitExceeded]: {
+    title: 'Storage limit exceeded',
+    description:
+      'Each publisher has a storage limit. If your file exceed yours you won’t be able to upload it. You can aquire additional storage space by asking on our discord.',
+    animation: {
+      data: errorAnimation,
+      size: {
+        width: 288,
+        height: 264,
+      },
+      loop: false,
+    },
+  },
 }
--- a/src/components/TransactionModal/TransactionModal.constants.ts
+++ b/src/components/TransactionModal/TransactionModal.constants.ts
@@ -73,7 +73,7 @@ export const TRANSACTION_STEPS_DETAILS = {
   [ExtrinsicStatus.VoucherSizeLimitExceeded]: {
     title: 'Storage limit exceeded',
     description:
-      'Each publisher has a storage limit. If your file exceed yours you won’t be able to upload it. You can aquire additional storage space by asking on our discord.',
+      "Your transaction failed because publishing associated assets would exceed your storage quota. Each channel has a dedicated storage limit that's controlled by the DAO storage working group. You can ask for additional storage space in #storage-provider channel on Joystream Discord and then try again.",
     animation: {
       data: errorAnimation,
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/components/TransactionModal/TransactionModal.tsx [hunk 1]
```diff
@@ -124,8 +128,22 @@ export const TransactionModal: React.FC<TransactionModalProps> = ({ status, onCl
       </StyledTransactionIllustration>
       <Dialog
         title={stepDetails?.title}
+        primaryButton={
+          status === ExtrinsicStatus.VoucherSizeLimitExceeded
+            ? {
+                text: 'Go to discord',
+                to: 'https://discord.com/invite/DE9UN3YpRP',
```
kdembler: Let's extract this to a constant in `config/urls`. Note that we already have `JOYSTREAM_DISCORD_URL` in there - this is a general Discord link. For this usage I'd like to direct users directly to the right channel so they don't need to ask around. I'm coordinating with the community on which channel this should be, will provide you with a link at a later time

kdembler: Use this: https://discord.gg/WUb7XwW72a

src/providers/transactionManager/useTransaction.ts
```diff
@@ -86,13 +86,16 @@ export const useTransaction = (): HandleTransactionFn => {
         })
         return false
       }
-
       if (e instanceof ExtrinsicFailedError) {
         SentryLogger.error('Extrinsic failed', 'TransactionManager', e)
```
kdembler: I don't think we need to send error to sentry in case it's an exceeded voucher

src/components/TransactionModal/TransactionModal.constants.ts
```diff
@@ -70,4 +70,17 @@ export const TRANSACTION_STEPS_DETAILS = {
       loop: false,
     },
   },
+  [ExtrinsicStatus.VoucherSizeLimitExceeded]: {
+    title: 'Storage limit exceeded',
+    description:
+      'Each publisher has a storage limit. If your file exceed yours you won’t be able to upload it. You can aquire additional storage space by asking on our discord.',
```
kdembler:
```suggestion
      'Your transaction failed because publishing associated assets wold exceed your storage quota. Each channel has a dedicated storage limit that's controlled by the DAO storage working group. You can ask for additional storage space in #storage-provider channel on Joystream Discord and then try again.',
```

src/components/TransactionModal/TransactionModal.tsx
```diff
@@ -124,8 +129,22 @@ export const TransactionModal: React.FC<TransactionModalProps> = ({ status, onCl
       </StyledTransactionIllustration>
       <Dialog
         title={stepDetails?.title}
+        primaryButton={
+          status === ExtrinsicStatus.VoucherSizeLimitExceeded
+            ? {
+                text: 'Go to discord',
```
kdembler:
```suggestion
                text: 'Open Discord',
```

kdembler (approved):

Issue #1591, opened 2021-11-15
Create view with embeddable video player


kdembler:
Until we can get proper social previews and embeddable video player done, we could introduce a simple workaround to allow embedding videos in some regard. We could create a view that display nothing but the video player. So no navigation, title, etc, just the video player spanning the full width. This view could later be embedded inside an `iframe` to allow displaying Joystream videos in other apps. Disclaimer: there may be something wrong with my reasoning above but let's try!

PR #1606, opened 2021-11-19
Embedded view

eldiegod:
An attempt at a workaround for embeddable video

the route is `embedded/video/${videoId}`

fixes: #1591


src/views/viewer/EmbeddedView/EmbeddedView.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const { url: mediaUrl, isLoadingAsset: isMediaLoading } = useAsset({ entity: video, assetType: AssetType.MEDIA })
+
+  const [startTimestamp, setStartTimestamp] = useState<number>()
+  useEffect(() => {
+    const duration = video?.duration ?? 0
+    if (!timestampFromQuery || timestampFromQuery > duration) {
+      return
+    }
+    setStartTimestamp(timestampFromQuery)
+  }, [video?.duration, timestampFromQuery])
```
kdembler, line 36: I think we can skip this part in the embedded view

kdembler (commented):
Looks pretty awesome in general, good work! One thing, it seems to me the player slightly overflows the view? i.e. I have a vertical scroll on this view, can you see that as well?

src/views/viewer/EmbeddedView/EmbeddedView.tsx
eldiegod: I think they mentioned in the call today they wanted the timestamp so this does it already if I got that part right

kdembler: Yeah let's leave it

kdembler:
Looks good! Tested it in an `iframe` and works pretty nice actually! There's one thing we need to fix though and that is that after the video finishes playing, you will get a suggestion for next video and that will navigate you to the regular `video` view, breaking the embed experience. I'd suggest we display this for embedded videos (for now at least) and just don't do anything once the playback is finished

eldiegod:
I think the end screen now works pretty good check it out

jodator:
FYI: Out of curiosity I've checked this in [Pioneer 2](https://deploy-preview-1818--pioneer-2.netlify.app/#/proposals/current):



Works nicely :rocket:

The only thing I've noticed is that the body has `overflow-y: scroll` gives this scroll bar on the right side. Changing to `auto` or `hidden` solves that issue:

kdembler:
Awesome news @jodator, glad it works for you! @eldiegod I think we should be able to add global style on embedded view to force `overflow: hidden`?

eldiegod:
Great to hear, yeah I will implement it right away

src/components/_video/VideoPlayer/VideoOverlay.tsx
```diff
@@ -10,21 +10,23 @@ import { getRandomIntInclusive } from '@/utils/number'
...
 
-type VideoOverlaProps = {
+type VideoOverlayProps = {
   playerState: PlayerState
   onPlay: () => void
   channelId?: string
   currentThumbnailUrl?: string | null
   videoId?: string
   isFullScreen?: boolean
+  playRandomVideoOnEnded?: boolean
```
kdembler, line 20: You haven't provided a default for that so now regular playback will also not have this functionality, right? Let's change props so that this is enabled by default. Actually, I think we could just introduce an `embedded` prop  to the video player instead, we may want to change some other things based on that as well in the future.

kdembler (approved):

PR #1615, opened 2021-11-19
Optimize VideoTile rerender amount

### LINKED ISSUE #1607: Optimize VideoTile performance

rafalpawlow:
Closes #1607

### DIFF
--- a/src/components/_video/VideoTile/VideoTile.tsx
+++ b/src/components/_video/VideoTile/VideoTile.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useCallback } from 'react'
 
 import { useVideo } from '@/api/hooks'
 import { absoluteRoutes } from '@/config/routes'
@@ -14,13 +14,17 @@ export type VideoTileProps = {
 } & VideoTileBaseMetaProps &
   Pick<VideoTileBaseProps, 'progress' | 'className'>
 
-export const VideoTile: React.FC<VideoTileProps> = ({ id, onNotFound, ...metaProps }) => {
+export const VideoTile: React.FC<VideoTileProps> = React.memo(({ id, onNotFound, ...metaProps }) => {
   const { video, loading, videoHref, thumbnailPhotoUrl, avatarPhotoUrl, isLoadingThumbnail, isLoadingAvatar } =
     useVideoSharedLogic({
       id,
       isDraft: false,
       onNotFound,
     })
+
+  const onCopyVideoURLClick = useCallback(() => {
+    copyToClipboard(videoHref ? location.origin + videoHref : '')
+  }, [videoHref])
 
   return (
     <VideoTileBase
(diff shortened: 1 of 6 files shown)


drillprop (approved):
Didn't find anything, looks good to me 👍🏻

src/components/_video/VideoTile/VideoTile.tsx
```diff
@@ -14,14 +14,18 @@ export type VideoTileProps = {
...
-export const VideoTile: React.FC<VideoTileProps> = ({ id, onNotFound, ...metaProps }) => {
+export const VideoTile: React.FC<VideoTileProps> = React.memo(({ id, onNotFound, ...metaProps }) => {
   const { video, loading, videoHref, thumbnailPhotoUrl, avatarPhotoUrl, isLoadingThumbnail, isLoadingAvatar } =
     useVideoSharedLogic({
       id,
       isDraft: false,
       onNotFound,
     })
 
+  const onCopyVideoURLClick = useCallback(() => {
```
kdembler: Small detail but according to our styleguide (the mental one, not sure if we have this documented), this should be called `handleCopyVideoURLClick`

src/components/_video/VideoTileBase/VideoTileCover.tsx
```diff
@@ -67,144 +67,154 @@ type VideoTileCoverProps = {
...
+    isLoadingThumbnail,
+    isDraft,
+    isUnlisted,
+    publisherMode,
+    duration,
+    onPullupClick,
+    removeButton,
+    thumbnailAlt,
+  }) => {
+    const isUploading = useMemo(() => uploadStatus && uploadStatus.lastStatus !== 'completed', [uploadStatus])
```
kdembler: Do we really need `useMemo` here? Seems like a very simple calculation and it also outputs `boolean` which is primitive so shouldn't trigger props change for children?

src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
```diff
@@ -11,54 +11,61 @@ import { VideoTileBase, VideoTilePublisherProps } from '../VideoTileBase'
...
+      (state) => state.uploadsStatus[video?.mediaDataObject?.joystreamContentId || '']
+    )
 
-  const hasThumbnailUploadFailed = video?.thumbnailPhotoAvailability === AssetAvailability.Pending
-  const hasVideoUploadFailed = video?.mediaAvailability === AssetAvailability.Pending
-  const hasAssetUploadFailed = hasThumbnailUploadFailed || hasVideoUploadFailed
+    const hasThumbnailUploadFailed = video?.thumbnailPhotoAvailability === AssetAvailability.Pending
+    const hasVideoUploadFailed = video?.mediaAvailability === AssetAvailability.Pending
+    const hasAssetUploadFailed = hasThumbnailUploadFailed || hasVideoUploadFailed
+    const onCopyVideoURLClick = useCallback(() => {
```
kdembler: Likewise, should update the handler name

src/components/_video/VideoTileBase/VideoTileCover.tsx
rafalpawlow: I made this change on the basis of `Why Did You Render` output, it showed that this variable was causing unnecessary rerenders.

src/components/_video/VideoTile/VideoTile.tsx
rafalpawlow: btw, we could collect our mental styleguides one day and make some kind of document out of it 😉. I think this would be useful for new devs.

kdembler: Yes that's for sure! We actually have an old (a bit outdated at this point) piece of styleguide [here](https://github.com/Joystream/atlas/blob/master/docs/styleguide.md). I'd love to get this updated but there just always are some other, more important things to do :(

kdembler (approved):

PR #1617, opened 2021-11-20
made notes of sorting in featured-content.md

mochet:
(would've liked to have tested this to make 100% sure, but the test deployment isn't working with categories at the moment)


kdembler (approved):
True, and thanks for adding this!

Issue #1616, opened 2021-11-20
Category Featured Videos do not work on categories above 10


mochet:
**Describe the bug**

When accessing the test deployment of Atlas (pointed at the current active testnet `sumer`) any video category above `10` will not work.

Examples:
https://atlas-dev.netlify.app/category/10 - works as expected

https://atlas-dev.netlify.app/category/11 (or any category above this that has the featured videos set results in the below)

![image](https://user-images.githubusercontent.com/50830161/142720729-d5691da6-f461-4b7e-8716-0ffab3f7f0dc.png)

Note: If the category above `10` does not have featured videos set, then it works, but the hero section of that category is obviously blank as no data is present from Orion

PR #1627, opened 2021-11-22
Discovery community content fixes

eldiegod:
fixes: #1616, #1637, #1636 and  some other details found along the way

### DIFF
--- a/src/api/featured/categoriesFeaturedVideos.ts
+++ b/src/api/featured/categoriesFeaturedVideos.ts
@@ -34,9 +34,7 @@ export const useCategoriesFeaturedVideos = (): CategoriesFeaturedVideos | null =
   )
   const videosLookup = createLookup(videos || [])
   const categoriesLookup = rawData?.reduce((acc, cur) => {
-    acc[cur.categoryId] = cur.videos.map((v) => {
-      return { ...videosLookup[v.videoId], videoCutUrl: v.videoCutUrl }
-    })
+    acc[cur.categoryId] = cur.videos.map((v) => ({ ...videosLookup[v.videoId], videoCutUrl: v.videoCutUrl }))
     return acc
   }, {} as CategoriesFeaturedVideos)
 
--- a/src/api/featured/categoriesFeaturedVideos.ts
+++ b/src/api/featured/categoriesFeaturedVideos.ts
@@ -16,6 +16,7 @@ export type CategoriesFeaturedVideos = Record<
   Array<VideoFieldsFragment & { videoCutUrl?: string | null }> | undefined
 >
 
+// TODO: hook to only fetch a single category featured videos
 export const useCategoriesFeaturedVideos = (): CategoriesFeaturedVideos | null => {
   const client = useApolloClient()
   const fetchCategoriesFeaturedVideos = useMemo(
@@ -27,11 +28,15 @@ export const useCategoriesFeaturedVideos = (): CategoriesFeaturedVideos | null =
   const { data: rawData } = useGenericFeaturedData('categories-featured-videos', fetchCategoriesFeaturedVideos)
 
   const allVideosIds = rawData?.reduce((acc, cur) => [...acc, ...cur.videos.map((v) => v.videoId)], [] as string[])
-  const { videos } = useVideos({ where: { id_in: allVideosIds } }, { skip: !allVideosIds || !allVideosIds.length })
+  const { videos } = useVideos(
+    { limit: allVideosIds?.length, where: { id_in: allVideosIds } },
+    { skip: !allVideosIds || !allVideosIds.length }
+  )
   const videosLookup = createLookup(videos || [])
-
   const categoriesLookup = rawData?.reduce((acc, cur) => {
-    acc[cur.categoryId] = cur.videos.map((v) => ({ ...videosLookup[v.videoId], videoCutUrl: v.videoCutUrl }))
+    acc[cur.categoryId] = cur.videos.map((v) => {
+      return { ...videosLookup[v.videoId], videoCutUrl: v.videoCutUrl }
+    })
     return acc
   }, {} as CategoriesFeaturedVideos)
 
--- a/src/components/_video/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
+++ b/src/components/_video/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
@@ -57,7 +57,7 @@ export const FeaturedVideoCategoryCard: React.FC<FeaturedVideoCategoryCardProps>
           color={color}
         >
           <PlayerContainer>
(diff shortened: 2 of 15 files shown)


src/components/_video/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
```diff
@@ -48,21 +50,14 @@ export const FeaturedVideoCategoryCard: React.FC<FeaturedVideoCategoryCardProps>
...
           <PlayerContainer>
-            {
-              <BackgroundVideoPlayer
-                src={isLoading ? undefined : videoUrl}
-                loop
-                muted
-                playing={variant === 'default' ? isVideoHovering : true}
-              />
-            }
+            {<BackgroundVideoPlayer src={isLoading ? undefined : videoUrl} loop muted playing={isPlaying} />}
```
drillprop:
```suggestion
            <BackgroundVideoPlayer src={isLoading ? undefined : videoUrl} loop muted playing={isPlaying} />
```
We don't need curly braces here

src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -25,22 +26,28 @@ export const DiscoverView: React.FC = () => {
...
+            categoriesFeaturedVideos?.[category.id]?.find((video) => !!video.videoCutUrl) ??
+            categoriesFeaturedVideos?.[category.id]?.[0] ??
+            undefined
+          return video === undefined
+            ? undefined
+            : {
+                videoTitle: video?.title ?? '',
+                videoUrl: video?.videoCutUrl ?? '',
+                ...mappedVideoCategories?.find((cat) => cat.id === category.id),
+              }
```
drillprop:
```suggestion
          if (video) {
            return {
              videoTitle: video?.title ?? '',
              videoUrl: video?.videoCutUrl ?? '',
              ...mappedVideoCategories?.find((cat) => cat.id === category.id),
            }
          }
```

src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -25,22 +26,28 @@ export const DiscoverView: React.FC = () => {
...
-    })
-    .filter((cat) => !!cat) ?? [null, null, null]
+  const featuredVideoCategoryCardsData = React.useMemo(() => {
+    const _featuredVideoCategoryCardsData = sampleSize(
+      categories
+        ?.map((category) => {
+          const video =
+            categoriesFeaturedVideos?.[category.id]?.find((video) => !!video.videoCutUrl) ??
+            categoriesFeaturedVideos?.[category.id]?.[0] ??
+            undefined
```
drillprop:
```suggestion
          const video =
            categoriesFeaturedVideos?.[category.id]?.find((video) => !!video.videoCutUrl) ??
            categoriesFeaturedVideos?.[category.id]?.[0]
```
Could be wrong, but I don't think we need this `?? undefined` here.

src/api/featured/categoriesFeaturedVideos.ts
```diff
@@ -27,11 +28,15 @@ export const useCategoriesFeaturedVideos = (): CategoriesFeaturedVideos | null =
...
+    { limit: allVideosIds?.length, where: { id_in: allVideosIds } },
+    { skip: !allVideosIds || !allVideosIds.length }
+  )
   const videosLookup = createLookup(videos || [])
-
   const categoriesLookup = rawData?.reduce((acc, cur) => {
-    acc[cur.categoryId] = cur.videos.map((v) => ({ ...videosLookup[v.videoId], videoCutUrl: v.videoCutUrl }))
+    acc[cur.categoryId] = cur.videos.map((v) => {
+      return { ...videosLookup[v.videoId], videoCutUrl: v.videoCutUrl }
+    })
```
drillprop:
```suggestion
    acc[cur.categoryId] = cur.videos.map((v) => ({ ...videosLookup[v.videoId], videoCutUrl: v.videoCutUrl }))
```

drillprop (commented):
I see also two warnings in the console: 
image
image
This is probably not related to your changes, I think it was there before. Could you fix them as part of this PR as well?

src/views/viewer/DiscoverView/DiscoverView.tsx
drillprop: Actually, we can just do: 
```tsx
const video = categoriesFeaturedVideos?.[category.id]?.find((video) => !!video.videoCutUrl)
```
Right🤔 ?

eldiegod: yes this works

src/views/viewer/DiscoverView/DiscoverView.tsx
eldiegod: I rather keep it explicit that we are possibly returning undefined here

drillprop: Hmm, I don't agree with this. I mean, it's quite clear for me, that if we don't return from `Array.map` we will just return `undefined`. Your current code gives me the impression that something else going on here, and I need to decode this every time I'm looking at these lines 😅 . 
Of course, this is only my opinion. Not sure if everyone on the team agrees with me. Let's leave that unresolved for now, maybe?

kdembler: I think we actually shouldn't rely on `undefined` that much, `null` seems to be a better fit here as it explicitly tells you the value is missing. How about this, should be both readable and explicit:
```tsx
const video =
  categoriesFeaturedVideos?.[category.id]?.find((video) => !!video.videoCutUrl) ??
  categoriesFeaturedVideos?.[category.id]?.[0]

if (!video) return null

return {
  // ...
}
```

kdembler (changes requested):
Found some more issues:
1. The thumbnails in Video Hero don't have a proper loading state:

https://user-images.githubusercontent.com/12646744/143020714-93082044-6008-4445-a2ed-6900b51d51fb.mov

2. We're missing bottom space in `Discover` view
3. How do we pick featured categories? They seem to be different on every load, also if I go to specific category and navigate back, the list will change. We should have those at least semi-consistent. I'm not sure what the correct long-term solution is, for the time being maybe we just hardcode IDs or just show the first 3 categories? Open to suggestions on this one

kdembler:
I think we just improve the loading of thumbnails inside video hero. You've added placeholders for images but even if those are loading, we should still show the progress bar indicating which video is currently playing.

kdembler:
The video hero thumbnails still don't work properly when loading:

https://user-images.githubusercontent.com/12646744/143212152-40be34cc-bef3-40e2-8f56-f7bd6c162ebe.mov

The progress bar should be spaced from the skeleton and also seems like the progress bar has a different color when the thumbnail is loading? Maybe an overlap?

kdembler (approved):

PR #1629, opened 2021-11-23
Fix storybook build

drillprop:
We were getting `setImmediate is not defined` on the storybook build. I'm not sure what's `setImmediate`  and why storybook needs it, but it should resolve the issue. 

To test it locally, you can run 
```
yarn build-storybook
npx http-server storybook-static
```

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,5 +1,8 @@
 <link rel="stylesheet" href="https://eu-central-1.linodeobjects.com/atlas-assets/fonts/PxGrotesk/PxGrotesk.css" />
 <link rel="stylesheet" href="https://eu-central-1.linodeobjects.com/atlas-assets/fonts/Inter/inter.css" />
+<script>
+  window.setImmediate = window.setTimeout
+</script>
 <style>
   body {
     margin: 0;


.storybook/preview-head.html
```diff
@@ -1,5 +1,8 @@
 <link rel="stylesheet" href="https://eu-central-1.linodeobjects.com/atlas-assets/fonts/PxGrotesk/PxGrotesk.css" />
 <link rel="stylesheet" href="https://eu-central-1.linodeobjects.com/atlas-assets/fonts/Inter/inter.css" />
+<script>
+  window.setImmediate = window.setTimeout
+</script>
```
drillprop, line 5: Basically, when you run `build-storybook` this script tag will be added to `iframe.html` inside `storybook-static`

kdembler: Not sure if you're aware but we actually do the same thing in our main `index.html`:
```html
    <!--  manual polyfills  -->
    <script>
      window.setImmediate = setTimeout
      window.global = window
      window.__DEV__ = false
    </script>
```

kdembler (approved):

Issue #1622, opened 2021-11-22
VideoWorkspace new tab bug


rafalpawlow:
**Describe the bug**

When you try to add new tab in video workspace, it's in edit mode and there is no way to add video file.

**To Reproduce**

1. go to studio
2. click "Upload video"
3. click "+" button on tabs bar

**Expected behavior**

Clicking "+" button should open new tab with possibility of adding a video file.

**Screenshots**

![Zrzut ekranu 2021-11-22 o 14 16 17](https://user-images.githubusercontent.com/13706246/142868333-46046ea9-f673-4597-a6ca-36793b7cf0c7.png)

**Device (please complete the following information):**

- OS: macOS Monterey
- Browser Chrome 96

drillprop:
Cannot reproduce on dev. Read https://github.com/Joystream/atlas/pull/1652#issuecomment-982781661

PR #1652, opened 2021-11-24
Optimize VideoWorkspace performance

rafalpawlow:
Closes #1652


kdembler (changes requested):
I think we need to focus on this component a little bit more than the rest of the app, after all, this is the place that we noticed the problems with. I've ran a simple test before and after your changes. The test consisted of going to Studio, opening the video workspace and closing it. I've counted the render times of specific components, take a look.

**Before changes:**
`VideoWorkspace` renders: 12
`VideoWorkspaceForm` renders: 18
`VideoWorkspaceTabsBar` renders: 12

**After changes:**
`VideoWorkspace` renders: 11 (-1)
`VideoWorkspaceForm` renders: 17 (-1)
`VideoWorkspaceTabsBar` renders: 8 (-4)

As you can see, our savings are quite low and both the workspace and the form component still rerender too often. If you trigger the opening/closing animation with devtools open, you can see the animation lagging. We should try to do more drastic improvements here, to make sure the animation is smooth.

drillprop:
Just want to say that the video workspace feels much more performant, but the animation seems to me too fast and too "aggressive" so to speak. Could we tweak this a bit? Maybe we could change timeouts and transition delays to`500ms`? Maybe designers have something to say here? 
Correct me if I'm wrong, but I think we used default values from react-spring before, and we actually never discuss this with them.  The animation just felt right, and no one objects to that.

kdembler:
Agreed with @drillprop here, I suggest to ask @toiletgranny for his input on this.

Also, when minimizing the VideoWorkspace, the backdrop doesn't disappear:

Screenshot 2021-11-30 at 08 49 56

toiletgranny:
@wyrdrender I'd suggest using our [animation design tokens](https://github.com/Joystream/atlas-resources/blob/main/design_tokens/animation.token.json) to drive both timing and easing of this animation. Over here my suggestion would be to go with `{transition.callout}`, or `{transition.slow}` if the first one looks weird.

drillprop (approved):
Looks much better now from my perspective. I don't have anything to add, very needed changes 👍🏻

rafalpawlow:
I've made a mistake and created #1622 based on this branch. It needs to be fixed under this PR.
CC @drillprop

kdembler (changes requested):
The transition doesn't seem to work properly. It animates when I minimize, but it just disappears when I close the workspace

drillprop:
Looks like it also fixes #1689

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
```diff
@@ -2,8 +2,33 @@ import styled from '@emotion/styled'
 import { animated } from 'react-spring'
 
 import { oldColors, zIndex } from '@/styles'
+import { cVar } from '@/styles'
+import { VIDEO_WORKSPACE_TABS_BAR_HEIGHT } from '@/views/studio/VideoWorkspace/VideoWorkspaceTabsBar'
```
kdembler: This should be a relative import

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
```diff
@@ -2,8 +2,33 @@ import styled from '@emotion/styled'
...
 
-export const Container = styled(animated.div)`
+export const DrawerOverlay = styled(animated.div)`
```
kdembler: Does tis still need to be `animated.div`? We're no longer using `react-spring`, right?

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
```diff
@@ -14,14 +39,27 @@ export const Container = styled(animated.div)`
...
+    }
+
+    &-enter,
+    &-exit-active {
+      transform: translateY(100%);
+      opacity: 0;
+    }
+
+    &--minimized {
+      top: calc(100% - ${VIDEO_WORKSPACE_TABS_BAR_HEIGHT}px - 1px);
```
kdembler: Where does that `1px` come from? From my quick testing it seems it works better without it

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
```diff
@@ -14,14 +39,27 @@ export const Container = styled(animated.div)`
...
+      transition: all ${cVar('animationTransitionSlow')};
+    }
+
+    &-enter,
+    &-exit-active {
+      transform: translateY(100%);
+      opacity: 0;
+    }
+
+    &--minimized {
```
kdembler: I had no idea you can compose selectors like that :o

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
kdembler: Also, why are we using `top` here instead of `transform`? This will make the animation less performant

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
```diff
@@ -14,14 +39,27 @@ export const Container = styled(animated.div)`
...
-  bottom: 0;
-  left: 0;
-  background-color: ${oldColors.transparentBlack[54]};
+  &.video-workspace {
+    &-exit,
+    &-enter-active,
+    &-enter-done {
+      opacity: 1;
+      transform: translateY(0);
+      transition: all ${cVar('animationTransitionSlow')};
```
kdembler: This should only need `opacity` and `transform`, right?

kdembler (commented):
I've tested the functionality and all seems to work properly, have some notes for the code though

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
rafalpawlow: Yeah, this is the way SASS supports BEM

src/views/studio/VideoWorkspace/VideoWorkspace.style.ts
rafalpawlow: TBH, I took this `1px` from previous implementation https://github.com/Joystream/atlas/pull/1652/files#diff-169352e72b00dd3da5f910131bf7694dd34eaa84124ea8a38c49b8594a6c1cd9L13

src/views/studio/VideoWorkspace/VideoWorkspaceForm/VideoWorkspaceForm.tsx
```diff
@@ -64,629 +76,877 @@ const knownLicensesOptions: SelectItem<License['code']>[] = knownLicenses.map((l
...
+      forceReset,
+      reset,
+      tabDataLoading,
+      tabData,
+      updateSelectedVideoTab,
+      selectedVideoTabCachedDirtyFormData,
+      setValue,
+    ])
+
+    const onSubmit = useCallback(
```
drillprop, line 261: I can't comment on these lines , but I think we could safely refactor lines 156-159 to: 
```tsx
    const addAsset = useAssetStore((state) => state.actions.addAsset)
    const mediaAsset = useRawAsset(watch('assets.video.contentId'))
    const thumbnailAsset = useRawAsset(watch('assets.thumbnail.cropContentId'))
    const originalThumbnailAsset = useRawAsset(watch('assets.thumbnail.originalContentId'))
```
and if we do that we could remove `defaultValues` from line 145 as well. It seems that we don't need it at all for some reason. Did some testing and I think we should be fine with these changes

kdembler:
I think the problem with new tab being in edit mode still persists, it's just hidden a bit. If you use the primary buttons in MyVideos (icon in topnav and the one next to sort select) to open a new tab, all works great. But if you use the new video tile that's in the grid, the tab still opens in edit mode.

Other than above everything seems to work correctly so I'm fine to merge as long as we fix the above in the followup

rafalpawlow:
@kdembler Please take a look now

kdembler (approved):

Issue #1633, opened 2021-11-23
Apply generated typography css variables to `Text` component


drillprop:

dmtrjsg:
⚠️  need to sync up with Adam and make the changes requried.

dmtrjsg:
@toiletgranny FYI we've added you to this task to be worked on this week :)

PR #1667, opened 2021-11-26
Typography adjustments

eldiegod:
fixes: #1633


.storybook/preview-head.html
```diff
@@ -1,5 +1,8 @@
-<link rel="stylesheet" href="https://eu-central-1.linodeobjects.com/atlas-assets/fonts/PxGrotesk/PxGrotesk.css" />
 <link rel="stylesheet" href="https://eu-central-1.linodeobjects.com/atlas-assets/fonts/Inter/inter.css" />
```
kdembler: We don't need this one anymore, right?

src/components/Text/Text.tsx
```diff
@@ -12,7 +12,7 @@ export type TextProps = {
 } & React.DetailedHTMLProps<React.HTMLAttributes<HTMLHeadingElement>, HTMLHeadingElement>
 
 export const Text = React.forwardRef<HTMLHeadingElement, TextProps>(
-  ({ variant = 'body2', secondary, as, ...otherProps }, ref) => {
+  ({ variant = 't200', secondary, as, ...otherProps }, ref) => {
```
kdembler: I suggest we remove the default value here and require all usages to be explicit

src/components/Text/Text.styles.ts
```diff
@@ -1,139 +1,121 @@
...
-  button3: styled.span<TextProps>`
-    ${baseStyles};
-
-    font-size: ${oldTypography.sizes.button.small};
-    line-height: ${oldTypography.lineHeights.button};
-    font-weight: ${oldTypography.weights.bold};
-    font-family: ${oldTypography.fonts.headers};
+    font: ${cVar('typographyDesktopT100Strong')};
+    letter-spacing: ${cVar('typographyDesktopT100StrongLetterSpacing')};
+    text-transform: ${cVar('typographyDesktopT100StrongTextTransform')};
```
kdembler: Any specific reason to make `t300` and `t200` render as `p` but `t300-strong`, `t200-strong`, `t100` and `t100-strong` as `span`?

src/components/Text/Text.styles.ts
```diff
@@ -1,139 +1,121 @@
...
 
-    font-size: ${oldTypography.sizes.subtitle1};
-    line-height: ${oldTypography.lineHeights.subtitle1};
-    font-weight: ${oldTypography.weights.regular};
+    font: ${cVar('typographyDesktopH200')};
+    letter-spacing: ${cVar('typographyDesktopH200LetterSpacing')};
+    text-transform: ${cVar('typographyDesktopH200TextTransform')};
   `,
-  subtitle2: styled.span<TextProps>`
+  h100: styled.span<TextProps>`
```
kdembler: All the header styles render as `h*` but these 2 as `span`s, let's use `h6` for those

kdembler (commented):
I love the new look and really glad to get this done and close this chapter! Here's a list of feedback/questions to the new look:

1. I'm not sure about the change of secondary text color from neutral-300 to neutral-400. It seems it creates low contrast issues in the app. Chrome devtools report the contrast as too low according to accessibility guidelines. My personal opinion is that those blend just too much with the dark background:
Screenshot 2021-11-26 at 19 59 10
Screenshot 2021-11-26 at 19 48 57
Screenshot 2021-11-26 at 20 06 31

2. Items inside the sidenav seem to be more squished together than they previously were (spacing between icon and label):

Before:
Screenshot 2021-11-26 at 19 54 04
After:
Screenshot 2021-11-26 at 19 53 57

It seems we're missing a 4px spacing between those that's defined in the designs.

3. The `Sort by` label in `My videos` has now a bold font that seems to direct a bit too much attention to it. At the same time, the same label, but in Channel view, isn't bolded. So we should pick one of those styles and use it consistently.

Screenshot 2021-11-26 at 20 12 05

Screenshot 2021-11-26 at 20 11 50

4. The license info inside video view is now uppercased but wasn't previously. Can you confirm this was intentional?

Before:
Screenshot 2021-11-26 at 20 14 05

After:
Screenshot 2021-11-26 at 20 14 18

cc @toiletgranny

toiletgranny:
Klaudiusz, thanks so much for taking the time to review this! Below you'll find my notes:

As per your comments @kdembler:
1. Actually, none of the color changes were intentional and my best guess is that it must be a side effect of some bulk "find and replace" action. Nevertheless, 🚨 @eldiegod: we must take all color changes back, as it's not only the matter of insufficient contrast ratios (as pointed out by Klaudiusz), but also eg. broken hover effects in the "Global search box" component.
2. In this case, it's the `margin-bottom` property of the `::after` element holding a text label being set to `10px`. 
🚨 @eldiegod: this should be changed to `8px`.
3. Thanks for catching that! 🚨 @eldiegod: The "Select" component on the "My videos" view is completely off. I'd suggest for us two to catch up on Monday and inspect that together, as I fear it might be a more serious issue. Let me know when you're available! :)
4. No, not intentional at all. Actually, there are more things that need our attention on the Video view. 🚨 @eldiegod:
image
    a. Views & date uploaded should be displayed using `T300 Regular` text style. Now it seems to be random,
    b. This should be `T300 Strong`,
    c. Both license and attribution information should be displayed using `T100 Regular`.


Other comments:
1. While we have adjusted top & bottom padding of the "Select" component in its "Big" size variant, we forgot about two other sizes it comes with: "Medium" & "Small". For those the top & bottom paddings should be, respectively: `10px` and `6px`.
     - ⚠️ Please note that the same padding values should be applied to the "Select list item" components in their corresponding sizes.
     - ⚠️ Please note that the same changes might be needed for "Text field" & "Text area" components.
2. We've mistakenly set the  "Add cover image" label to be `T300 Strong`, but it should be `T200 Stron` instead
image

src/components/Text/Text.styles.ts
eldiegod: no reason really it was all there already and then I reordered them, should we make it all `p`'s?

kdembler: Maybe let's try `span`, we don't really have that much content that should go into paragraphs

kdembler (approved):

drillprop:
@kdembler one thing left, you need to remove `export * from './deprecated/typography'` from `src/styles/index.ts` :P

kdembler:
Yes, just pushed :P

drillprop (changes requested):
One thing is unfortunately broken after these changes. Look at VideoWorkspace:
image

kdembler:
oh so it worked before only because the details were `p` which is a block element

eldiegod:
Made that one specific instance a p now

drillprop:
Sorry, found one more thing 😅 
Tooltips headers don't feel right 
image
image

@toiletgranny could you guide us, what we should apply here?

toiletgranny:
@drillprop Actually, the implementation of this component matched the designs so everything is technically correct, at least looking from the typography adjustments perspective.

I agree this component could use some other visual adjustments, but I would leave this for later right now, knowing what other things are ahead.

drillprop (approved):
Looks good to me! Fantastic work guys! 🥳

PR #1671, opened 2021-11-27
update deps, fix commit hooks, fix stylelint

kdembler:
Summary of changes:
- updated some dependencies
- fixed `husky` and `lint-staged` setup to run ESLint, Prettier and StyleLint on all staged files during commit
- updated stylelint which seems to have added support for much more rules for our code
- fixed all the stylelint issues
- added CI fail on stylelint warnings
- updated our Prettier plugin for sorting imports, the new version supports ordering import members as well, so we don't need eslint rule for that anymore


src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -106,16 +106,18 @@ export const VideoView: React.FC = () => {
...
           {part}
         </Button>
+      ) : (
+        part
       )
-    }
-    return parts
+
+      return [...acc, node]
+    }, [] as React.ReactNode[])
```
kdembler, line 120: I've updated this code because it was producing TS error after some dependency updated. (It was complaining about overwriting the array entries).

However, I also think there was a bug here because of the `i += 2`, this would likely fail if there were two links next to each other so this should also fix it

rafalpawlow (changes requested):
LGTM. However I think it still does not fully solves the problem with `husky`. I got an error I mentioned on rocket chat. According to config it should fix the problems and commit them, shouldn't it?
Zrzut ekranu 2021-11-27 o 14 23 38

rafalpawlow (approved):

PR #1686, opened 2021-11-30
Restore scroll on going back

### LINKED ISSUE #379: Restore scroll position when coming back from details view
When a user browse a channels list and scrolls pretty far down and then opens one of the channels for details, it would be nice to restore their scroll if they decide to go back to the channels list. Right now they will always be back to the top of the list which isn't great.

We could do the same for videos view and also maybe home page?

rafalpawlow:
Closes #379

### DIFF
--- a/src/views/viewer/ViewerLayout.tsx
+++ b/src/views/viewer/ViewerLayout.tsx
@@ -26,6 +26,8 @@ import {
 
 import { DiscoverView } from './DiscoverView/DiscoverView'
 
+const ROUTING_ANIMATION_OFFSET = 100
+
 const viewerRoutes = [
   { path: relativeRoutes.viewer.search(), element: <SearchView /> },
   { path: relativeRoutes.viewer.index(), element: <HomeView /> },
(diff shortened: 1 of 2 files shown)


src/views/viewer/ViewerLayout.tsx
```diff
@@ -64,9 +67,9 @@ export const ViewerLayout: React.FC = () => {
 
     // delay scroll to allow transition to finish first
     setTimeout(() => {
-      window.scrollTo(0, 0)
+      window.scrollTo(0, navigationType !== 'POP' ? 0 : scrollPosition.current)
```
kdembler: This doesn't seem to work properly, my scroll position isn't correctly restored:

https://user-images.githubusercontent.com/12646744/144088538-3816019e-582f-435c-9c38-8e13d7f77bde.mov

Couldn't we use native browser scroll restoration instead?

kdembler (changes requested):

kdembler (approved):

PR #1697, opened 2021-12-01
Remove SignInJoin and SignInProcessView

### LINKED ISSUE #857: Remove / hide the second page in sign in flow
In the process of analysing the current implementation of Joystream studio we decided that the second step of sign in should be removed/ hidden for the user for now
image

drillprop:
Fix #857


kdembler:
I think we can also remove the coin illustration now, right?

drillprop:
> I think we can also remove the coin illustration now, right?

removed

kdembler (approved):

PR #1700, opened 2021-12-02
Textfield start and end nodes

eldiegod:
fixes: #1634

https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=110%3A18288


src/components/_inputs/TextField/TextField.tsx
```diff
@@ -32,27 +35,41 @@ const TextFieldComponent: React.ForwardRefRenderFunction<HTMLInputElement, TextF
...
     placeholder,
     defaultValue,
+    nodeStart,
+    nodeEnd,
     ...inputBaseProps
   },
   ref
 ) => {
+  const [nodeLeftRef, nodeLeftBounds] = useMeasure()
+  const [nodeRightRef, nodeRightBounds] = useMeasure()
```
kdembler, line 45: Why do we need to measure anything here? I would suspect we can just make the nodes part of the layout and the input will get sized automatically?

src/components/_inputs/TextField/TextField.styles.ts
```diff
@@ -1,5 +1,24 @@
...
+
+export const NodeContainer = styled.div<{ left?: boolean }>`
+  position: absolute;
+  display: flex;
+  align-items: center;
+  height: 50px;
+  z-index: 2;
+  top: 0;
+  padding: 0 ${sizes(4)};
+  ${({ left }) => (left ? 'left: 0px' : 'right: 0px')};
```
kdembler: Let's try to avoid having any styles as a string as this will not be properly parsed by our tooling, including Stylelint. You can do it like this for example:

```tsx
type PlacementProps = {
  placement: 'left' | 'right'
}

const nodePlacementStyles = ({ placement }: PlacementProps) => placement === 'left' ? css`left: 0;` : css`right: 0;`

export const NodeContainer = styled.div<PlacementProps>`
  // ...
  ${nodePlacementStyles};
`
```

src/components/_inputs/TextField/TextField.tsx
kdembler: I see now that this is because the nodes have `position: absolute;` but do they need to?

toiletgranny:
Looks good to me! 🎉 Just a quick question: is my understanding correct that it is _the_ `Button` component that's nested inside the `TextField` in Storybook? And that we could replace it with anything we want, without affecting the overall height of the `TextField` component?

src/components/_inputs/TextField/TextField.tsx
eldiegod: I didn't find a good way to do it on the first try so I used chakra UI for inspiration and they had it using absolute.

For ref: https://chakra-ui.com/docs/form/input

src/components/_inputs/TextField/TextField.stories.tsx
```diff
@@ -43,3 +44,9 @@ Default.argTypes = {
 export const WithControlledInput = TemplateWithControlledInput.bind({})
 
 export const WithUncontrolledInput = TemplateWithUncontrolledInput.bind({})
+
+export const withIcons = Template.bind({})
+withIcons.args = {
```
drillprop:
```suggestion
export const WithIcons = Template.bind({})
WithIcons.args = {
```
Just spotted a small typo. Let's keep these names consistent

.env
```diff
@@ -16,3 +16,5 @@ VITE_PRODUCTION_ORION_URL=https://orion.joystream.org/graphql
 VITE_PRODUCTION_NODE_URL=wss://rome-rpc-endpoint.joystream.org:9944
 VITE_PRODUCTION_FAUCET_URL=https://member-faucet.joystream.org/register
 VITE_PRODUCTION_OFFICIAL_JOYSTREAM_CHANNEL_ID=3
+VITE_PRODUCTION_HOTJAR_ID=2692611
+VITE_PRODUCTION_LIVESESSION_ID=029d5963.a21f20e4
```
kdembler: This shouldn't be here

src/components/_inputs/TextField/TextField.tsx
kdembler: What's stopping us from using a flex row layout? The nodes could be placed at the ends and the middle text input would grow to fill the available space

src/components/_inputs/TextField/TextField.styles.ts
```diff
@@ -1,5 +1,37 @@
...
+        left: 0;
+      `
+    : css`
+        right: 0;
+      `
+export const NodeContainer = styled.div<PlacementProps>`
+  position: absolute;
+  display: flex;
+  align-items: center;
+  height: 50px;
```
kdembler: Where does this come from?

eldiegod: I guess you mean the 50px height, it's just what the input height is including the border I just have it fixed for the absolute nodes to work properly, and because this component should not be bigger or smaller than 50px

eldiegod:
> What's stopping us from using a flex row layout? The nodes could be placed at the ends and the middle text input would grow to fill the available space

@kdembler  I'm really struggling to picture in my head what you propose, care to make a sandbox or something like that?

kdembler:
@eldiegod Very quick demo here: https://codesandbox.io/s/nostalgic-engelbart-h0oy0?file=/src/App.js

There are basically 2 approaches we can take:
1. What you did here, make the start/end node reside inside the actual input and add dynamic padding. This is what Chakra does as well.
2. Make the input start only after the start node, shrinking it. That's what I did in the demo and also what MUI does: https://mui.com/components/text-fields/#icons

So this is a tradeoff I think, 1. solution requires us to do the measurement which I don't feel very comfortable about. I had a brief chat with Adam and he said he's fine with taking the second approach, so I'd suggest we use this

eldiegod:
> 2\. ake the input start only after the start node, shrinking it. That's what I did in the demo and also wh

@kdembler 

The reason I didn't go with approach number 2 is that we do all our form styling on `InputBase` -> `FormGroup` and it targets directly inputs, buttons and textareas so this approach would mean we have to remove the styling from the input detach it and duplicate it into the div container which would move away from that convention and be an extra thing to maintain.

What would be the problem with the measurements, performance shouldn't be a thing because they use ResizeObserver which is fairly optimized + a big debounce can be added as this is a static component.

On another note, not sure why but this build is failing quite misteriously

src/components/_inputs/TextField/TextField.styles.ts
kdembler: We shouldn't need this height stated here explicitly, this leaves out room for problems if this height changes somehow and we miss it. If this is positioned absolutely, we can just use `bottom: 0` instead

kdembler (approved):

Issue #1635, opened 2021-11-23
Create new ListItem components


kdembler:

eldiegod:
same issue as #1589

PR #1701, opened 2021-12-02
List item component

eldiegod:
fixes: #1635

https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3152%3A103472

### DIFF
--- a/src/components/ListItem/ListItem.stories.tsx
+++ b/src/components/ListItem/ListItem.stories.tsx
@@ -1,0 +1,58 @@
+import styled from '@emotion/styled'
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+
+import { ListItem, ListItemProps } from './ListItem'
+
+import { SvgActionChevronR, SvgActionExclamation } from '../_icons'
+
+export default {
+  title: 'Other/ListItem',
+  component: ListItem,
+  argTypes: {
+    size: {
+      control: { type: 'select', options: ['small', 'medium', 'large'] },
+      defaultValue: 'medium',
+    },
+    captionPosition: {
+      control: { type: 'select', options: ['bottom', 'right'] },
+      defaultValue: 'bottom',
+    },
+  },
+  args: {
+    label: 'ListItem label',
+    caption: 'This is a caption',
+    selected: false,
+    disabled: false,
+    destructive: false,
+  },
+} as Meta
+
+const Template: Story<ListItemProps> = (args) => (
+  <Container>
+    <ListItem {...args} />
+    <ListItem {...args} />
+    <ListItem {...args} />
+  </Container>
+)
+
+export const Default = Template.bind({})
+
+export const IconStart = Template.bind({})
+IconStart.args = {
+  nodeStart: <SvgActionExclamation />,
+}
+export const IconEnd = Template.bind({})
+IconEnd.args = {
+  nodeEnd: <SvgActionChevronR />,
+}
+export const IconStartEnd = Template.bind({})
+IconStartEnd.args = {
+  nodeStart: <SvgActionExclamation />,
+  nodeEnd: <SvgActionChevronR />,
+}
+
+const Container = styled.div`
+  background-color: #181c20;
+  width: 280px;
+`
--- a/src/components/ListItem/ListItem.stories.tsx
+++ b/src/components/ListItem/ListItem.stories.tsx
@@ -36,7 +36,15 @@ const Template: Story<ListItemProps> = (args) => (
   </Container>
 )
 
+const TemplateSingle: Story<ListItemProps> = (args) => (
+  <Container>
+    <ListItem {...args} />
+  </Container>
+)
+
 export const Default = Template.bind({})
+
+export const Single = TemplateSingle.bind({})
 
 export const IconStart = Template.bind({})
 IconStart.args = {
--- a/src/components/ListItem/ListItem.styles.ts
+++ b/src/components/ListItem/ListItem.styles.ts
@@ -1,6 +1,7 @@
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 
+import { Text } from '@/components/Text'
 import { cVar, sizes } from '@/styles'
 
 import { ListItemSizes } from '.'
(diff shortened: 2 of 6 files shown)


src/components/ListItem/ListItem.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+    disabled: false,
+    destructive: false,
+  },
+} as Meta
+
+const Template: Story<ListItemProps> = (args) => (
+  <Container>
+    <ListItem {...args} />
+    <ListItem {...args} />
+    <ListItem {...args} />
```
drillprop, line 35: Could you also add a template with a single `ListItem` to see more clearly how they look in isolation?

src/components/ListItem/ListItem.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+      return css`
+        padding: ${sizes(2)} ${sizes(4)};
+      `
+    case 'small':
+      return css`
+        padding: ${sizes(1)} ${sizes(4)};
+      `
+  }
+}
+type ContainerProps = { size: ListItemSizes; hasNodeStart: boolean; disabled?: boolean }
```
drillprop: Let's move this close to the Container
```suggestion
```

src/components/ListItem/ListItem.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+  &:focus,
+  &:hover {
+    background-color: ${cVar('colorBackgroundAlpha')};
+  }
+
+  &:active {
+    background-color: ${cVar('colorBackgroundMutedAlpha')};
+  }
+`
+export const Container = styled.div<ContainerProps>`
```
drillprop, line 41:
```suggestion
type ContainerProps = { size: ListItemSizes; hasNodeStart: boolean; disabled?: boolean }
export const Container = styled.div<ContainerProps>`
```

src/components/ListItem/ListItem.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+      return css`
+        padding: ${sizes(1)} ${sizes(4)};
+      `
+  }
+}
+type ContainerProps = { size: ListItemSizes; hasNodeStart: boolean; disabled?: boolean }
+const disabledStyles = css`
+  opacity: 0.25;
+  cursor: not-allowed;
+`
```
drillprop, line 29: hmm I think disabled variants looks different in the designs: 
image

src/components/ListItem/ListItem.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+    background-color: ${cVar('colorBackgroundMutedAlpha')};
+  }
+`
+export const Container = styled.div<ContainerProps>`
+  display: grid;
+  grid-gap: ${sizes(3)};
+  grid-auto-flow: column;
+  grid-template-columns: ${({ hasNodeStart }) => (hasNodeStart ? 'auto 1fr' : '1fr')};
+  align-items: center;
+  user-select: none;
```
drillprop, line 47: This component should have also different default background if Im not mistaken. 
```
background: var(--color-background-muted);
```
Design: 
image

src/components/Text/Text.styles.ts
```diff
@@ -4,12 +4,13 @@ import styled from '@emotion/styled'
...
 
 type TextProps = {
+  color?: string
   isSecondary?: boolean
 }
 
 const baseStyles = (props: TextProps) => css`
   margin: 0;
-  color: ${props.isSecondary ? cVar('colorCoreNeutral300') : cVar('colorCoreNeutral50')};
+  color: ${props.color ? props.color : props.isSecondary ? cVar('colorCoreNeutral300') : cVar('colorCoreNeutral50')};
```
drillprop: Let's refactor this a bit maybe? 
```js
const baseStyles = ({ color, isSecondary }: TextProps) => css`
  margin: 0;
  color: ${color ? color : isSecondary ? cVar('colorCoreNeutral300') : cVar('colorCoreNeutral50')};
`
```

drillprop (commented):
Looks pretty solid in general

src/components/ListItem/ListItem.styles.ts
drillprop: oh, wait, my mistake. I didn't understand the design. It's transparent by default, right?

src/components/ListItem/ListItem.styles.ts
drillprop: https://github.com/Joystream/atlas/pull/1701#discussion_r761041470

src/components/ListItem/ListItem.styles.ts
eldiegod: yep

kdembler:
@eldiegod why are all the comments resolved? AFAICT there were no changes to the PR made after the comments were made so they shouldn't be resolved

eldiegod:
> @eldiegod why are all the comments resolved? AFAICT there were no changes to the PR made after the comments were made so they shouldn't be resolved

@kdembler my bad, forgot to push them

toiletgranny:
Omg, I can't believe this! This is so spot on! How did you do that?! You made my day, perhaps even week! There are some tiny-winy little details that I've found:

Here's all I found:
1. When the `destructive` property is set to `true`, the icon should inherit the red color as well. This is an important thing to have, I'm afraid.
2. When the `captionPosition` property is set to `right`, ideally the CSS `text-alignment` is also set to `right`, so that in rare scenarios when the text wraps, it stays aligned to the right side.
3. As I can't test this in Storybook (which is understandable), I just want to make sure: is it still possible to replace the icon with any other component we want in the future? Eg. an avatar?

kdembler:
Just wanted to drop in and mention: great job on the story, I love that you took time to include all the arguments and all the different types of stories. I think that we could use this as an example on how the stories should look like 🥇

.env
```diff
@@ -16,3 +16,5 @@ VITE_PRODUCTION_ORION_URL=https://orion.joystream.org/graphql
 VITE_PRODUCTION_NODE_URL=wss://rome-rpc-endpoint.joystream.org:9944
 VITE_PRODUCTION_FAUCET_URL=https://member-faucet.joystream.org/register
 VITE_PRODUCTION_OFFICIAL_JOYSTREAM_CHANNEL_ID=3
+VITE_PRODUCTION_HOTJAR_ID=2692611
+VITE_PRODUCTION_LIVESESSION_ID=029d5963.a21f20e4
```
drillprop: These changes shouldn't be included

drillprop (changes requested):
Looks good to me, but your PR still includes some weird commit. You need to rebase
image

drillprop (approved):

kdembler (approved):

PR #1703, opened 2021-12-02
Remove polished library

### LINKED ISSUE #290: Clean up font sizing w.r.t. responsiveness
We should figure out some common solution for how we do font-sizing w.r.t. to viewport width changes. Currently there is no one clear way to do it. We use combination of media queries, `fluidRange` and nothing. Ideally, this shouldn't require any effort.

Let's remove `polished` library altogether

drillprop:
Fix #290

FYI, I allowed myself to choose the values for font sizes. Didn't want to set hardcoded values, I wanted to use values that were generated from design tokens. They differ from the designs, but designs for the ChannelView and the VideoView are quite old now and I think they're using different fonts anyway. 

VideoView: https://www.figma.com/file/Vk2Z4QOiVa5bB6q3cBIG5J/Joystream-Atlas?node-id=6345%3A149653
ChannelView: https://www.figma.com/file/Vk2Z4QOiVa5bB6q3cBIG5J/Joystream-Atlas?node-id=6277%3A8778
CategoryCards: https://www.figma.com/file/UdBMxrLspjcf6IVYYJS8YP/%5BPage%5D-Discover-%26-Video-category?node-id=244%3A97798

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "ipfs-only-hash": "^4.0.0",
     "localforage": "^1.10.0",
     "lodash-es": "^4.17.21",
-    "polished": "^4.1.3",
     "rc-slider": "^9.7.4",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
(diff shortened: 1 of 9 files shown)


eldiegod (approved):
Removing polished parts looks good, and I dont mind the arbitrary font sizes added coz those views design is going to change soon and it doesn't break anything

kdembler (approved):

PR #1705, opened 2021-12-02
Add consistent padding-bottom to LimitedWidhtContainer and update views

### LINKED ISSUE #1594: Ensure consistent bottom spacing in Studio views
Currently almost all of the Studio views have different bottom spacing (space we leave out so page doesn't end immediately with the content). We should possibly create a template and use it for all the views. Problems we should solve:
- Log in screen (with membership cards) doesn't have any bottom spacing
- Channel edit view has too much bottom spacing
- My videos view has spacing only if the pagination is visible

drillprop:
Fix #1594

### DIFF
--- a/src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
+++ b/src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
@@ -6,6 +6,8 @@ import { TitleArea } from '@/components/_inputs/TitleArea'
 import { media, sizes, transitions, zIndex } from '@/styles'
 import { SubTitle, TitleSection } from '@/views/viewer/ChannelView/ChannelView.styles'
 
+import { VIDEO_WORKSPACE_TABS_BAR_HEIGHT } from '../VideoWorkspace/VideoWorkspaceTabsBar'
+
 export const StyledTitleSection = styled(TitleSection)`
   display: inline-flex;
   width: auto;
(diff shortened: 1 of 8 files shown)


src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
```diff
@@ -38,7 +40,7 @@ export const TitleContainer = styled.div`
 export const InnerFormContainer = styled.div`
   width: 100%;
   margin-top: 50px;
-  padding-bottom: 250px;
+  padding-bottom: ${VIDEO_WORKSPACE_TABS_BAR_HEIGHT}px;
```
kdembler: Why?

drillprop: I need some additional space for ActionBar. Without this padding I can't see `select` element fully, because it's hidden:
image

With: 
image

kdembler: What does it have to do with the video workspace though? It seems completely unrelated to me

drillprop: I don't know why I picked this, for some reason I thought that this variable contains the height of the action bar 😅 
Anyway, should be good now. Basically did the same as we have in the VideoWorkspace form.

kdembler (approved):

PR #1706, opened 2021-12-03
Fix wrong button order in dialogs

### LINKED ISSUE #1601: Wrong button order in Dialogs
If one of the buttons in the Dialog footer is an `a` link and the other is a regular button, the order of the buttons will be wrong. This is because I've incorrectly used `:first-of-type` to do the order styling

Screenshot 2021-11-17 at 11 51 19

(primary button should be on the right)

drillprop:
Fix #1601 

I'm not sure, what's was the reason for using `order` here. Please double-check if removing this property breaks anything

### DIFF
--- a/src/components/_overlays/Dialog/Dialog.styles.ts
+++ b/src/components/_overlays/Dialog/Dialog.styles.ts
@@ -115,9 +115,5 @@ export const FooterButtonsContainer = styled.div`
     grid-auto-flow: column;
     grid-auto-columns: auto;
     justify-content: end;
-
-    > :first-of-type {
-      order: 2;
-    }
   }
 `
--- a/src/components/_overlays/Dialog/Dialog.styles.ts
+++ b/src/components/_overlays/Dialog/Dialog.styles.ts
@@ -1,6 +1,7 @@
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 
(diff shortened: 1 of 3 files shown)


kdembler (changes requested):
The problem I was trying to solve with this `order` style was this: on mobile the order of the buttons should be primary on top, but on desktop the primary button should be on the right. If you open this PR in the storybook and take a look at dialog modal on small screen, you will see that the order of the buttons is incorrect

drillprop:
This is a little hacky, but it works. I used [direction](https://developer.mozilla.org/pl/docs/Web/CSS/direction) property, which is [well-supported](https://caniuse.com/?search=direction)

src/components/_overlays/Dialog/Dialog.styles.ts
```diff
@@ -114,10 +114,9 @@ export const FooterButtonsContainer = styled.div`
   ${media.sm} {
     grid-auto-flow: column;
     grid-auto-columns: auto;
-    justify-content: end;
 
-    > :first-of-type {
-      order: 2;
-    }
+    /* those two properties below will reverse order of button */
+    direction: rtl;
```
kdembler: `direction` has a quite specific meaning and it supposed to be used only for different languages that use` rtl`. Let's find something else for styling

kdembler (approved):

PR #1707, opened 2021-12-03
Improve Avatar styles

### LINKED ISSUE #1649: Avatar component adjustments
### Description

When an avatar with a black image is placed against a pure black background, the visual boundaries of this component get blurred. 

### Figma link

_No response_

### How urgent this is?

It's blocking me right now

rafalpawlow:
Closes #1649 
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=104%3A76

### DIFF
--- a/src/components/Avatar/Avatar.styles.ts
+++ b/src/components/Avatar/Avatar.styles.ts
@@ -1,16 +1,16 @@
 import { SerializedStyles, css } from '@emotion/react'
 import styled from '@emotion/styled'
-import { TransitionGroup } from 'react-transition-group'
 
 import { SvgIllustrativeFileFailed } from '@/components/_icons'
 import { SvgAvatarSilhouette } from '@/components/_illustrations'
 import { SkeletonLoader } from '@/components/_loaders/SkeletonLoader'
-import { cVar, media, oldColors, transitions } from '@/styles'
+import { cVar, media, oldColors, square, transitions, zIndex } from '@/styles'
 
-export type AvatarSize = 'preview' | 'cover' | 'view' | 'default' | 'fill' | 'small' | 'channel' | 'channel-card'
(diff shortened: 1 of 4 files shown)


eldiegod (commented):
Looks good overall, would be nice to have a story with an editable avatar tho

eldiegod (approved):

src/components/Avatar/Avatar.styles.ts
```diff
@@ -90,21 +85,30 @@ const getAvatarSizeCss = (size: AvatarSize): SerializedStyles => {
...
 }
 
+export const Outline = styled.div`
+  ${square('100%')};
+
+  border: 1px solid ${cVar('colorBorderAlpha')};
+  border-radius: 50%;
+  z-index: ${zIndex.overlay};
+  pointer-events: none;
+`
```
kdembler: Why do we need a separate element for that? Couldn't we just use `box-shadow` directly on the `Container`?

kdembler:
It seems we are also not fully following the design here - the design specifies the hover/active border also for non-editable avatars. Let's update that

src/components/Avatar/Avatar.styles.ts
```diff
@@ -113,6 +107,29 @@ export const Container = styled.div<ContainerProps>`
...
+
+  ::after {
+    ${square('100%')};
+
+    content: '';
+    display: block;
+    border-radius: 50%;
+    z-index: ${zIndex.overlay};
+    pointer-events: none;
+    box-shadow: inset 0 0 0 1px ${cVar('colorBorderAlpha')};
```
kdembler: This uses a wrong color, should be `colorBorderAlphaMuted`

kdembler (approved):

PR #1708, opened 2021-12-03
Fix issue with showing connection indicator when refreshing

### LINKED ISSUE #761: Page refresh in studio shows connection closed alert
The alert slides down when someone refreshes, which it should not if the user is actively refreshing.

drillprop:
Fix #761


rafalpawlow (approved):

kdembler (changes requested):
With this change if node looses connection, we will run `joystream?.destroy()` which hasn't happened before. I'm not sure if this is desired and I think testing it would be outside of scope of this issue. Can we instead just add a visual delay of the banner showing of like `1s`?

drillprop:
I don't think one second is enough, I set this to two and a half seconds. I reverted previous changes as well.

kdembler (approved):

PR #1713, opened 2021-12-03
Hide fullscreen player button if fullscreen isn't available

### LINKED ISSUE #1630: Hide fullscreen player button if fullscreen isn't available
We currently always show the fullscreen button in the player component. However, fullscreen is available in all context. For example, if you embed Atlas video in an `iframe` and don't set proper permissions, it will not be able to trigger fullscreen. We should hide the button in that case. I found this for checking whether fullscreen is available: https://stackoverflow.com/questions/49584537/how-to-detect-if-a-page-is-rendered-in-an-iframe-that-allows-fullscreen We should test it out

drillprop:
Fix #1630


src/views/viewer/EmbeddedView/EmbeddedView.tsx
```diff
@@ -54,6 +54,8 @@ export const EmbeddedView: React.FC = () => {
     })
   }, [addVideoView, videoId, channelId, categoryId])
 
+  const allowFullscreen = (window.frameElement as HTMLIFrameElement)?.allowFullscreen
```
drillprop: I thought that it will be working just out of the box. We just pass `allowFullScreen`, and every element inside the iframe has fullscreen disabled.  Unfortunately, it won't work like that in our case. We need to pass the attribute and read it in the component. I could be wrong, but I think it's because we have several HTML elements in EmbededdedView. 

I've noticed that it only works if we had only a native video inside this iframe like in this snippet: 
```js
      <iframe
        src="http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"
        frameBorder="2"
        width={600}
        height={400}
        allowFullScreen={false}
      ></iframe>
```

kdembler (approved):

PR #1714, opened 2021-12-03
Fix video player error overlay

### LINKED ISSUE #1640: Video player error overlay allows starting playback
**Describe the bug**

If there's some issue and the error overlay appears in the video player, it's possible to click the background of it and force to player to start playback.

**To Reproduce**

1. Go to https://play.joystream.org/video/7815
2. Click the greyed out background
3. Player loading is triggered

**Expected behavior**

It shouldn't be possible to tigger the playing state, error should be persistent

**Screenshots**

https://user-images.githubusercontent.com/12646744/143215640-303650cf-dd8e-43e5-96a0-da0fc74be93f.mov

rafalpawlow:
Fixes #1640

### DIFF
--- a/src/components/_video/VideoPlayer/VideoPlayer.tsx
+++ b/src/components/_video/VideoPlayer/VideoPlayer.tsx
@@ -416,13 +416,16 @@ const VideoPlayerComponent: React.ForwardRefRenderFunction<HTMLVideoElement, Vid
   }, [currentVolume, volumeToSave, player, setCachedVolume])
 
   // button/input handlers
-  const handlePlayPause = () => {
+  const handlePlayPause = useCallback(() => {
+    if (playerState === 'error') {
+      return
+    }
     if (isPlaying) {
       pauseVideo(player, true, () => setIsPlaying(false))
     } else {
       playVideo(player, true, () => setIsPlaying(true))
     }
-  }
+  }, [isPlaying, pauseVideo, playVideo, player, playerState])
 
   const handleChangeVolume = (event: React.ChangeEvent<HTMLInputElement>) => {
     setCurrentVolume(Number(event.target.value))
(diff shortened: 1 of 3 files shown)


kdembler:
I think @drillprop may be a better fit for reviewer here

src/components/_video/VideoPlayer/VideoPlayer.tsx
```diff
@@ -416,13 +416,16 @@ const VideoPlayerComponent: React.ForwardRefRenderFunction<HTMLVideoElement, Vid
   }, [currentVolume, volumeToSave, player, setCachedVolume])
 
   // button/input handlers
-  const handlePlayPause = () => {
+  const handlePlayPause = useCallback(() => {
+    if (playerState === 'error') {
```
drillprop, line 420:
```suggestion
    if (playerState === 'error' || playerState === 'ended') {
```
This is happens also when a video is ended. I think it makes sense to refactor this line

drillprop (changes requested):
Also, I found that the text on the ending screen is not centered. Could we also fix this in one go? Should be relatively easy.
image
 
I think this is because the paragraph beneath the title is a `span` now and should be `paragraph`. You could just add `as="p"` on the Text component if I'm not mistaken.

src/components/_video/VideoPlayer/VideoPlayer.tsx
drillprop: Rafał just point me out that applying this suggestion will break the replay button on controls.

Anyway, I got confirmation from Adam that we shouldn't be able to replay a video by clicking on the background when a video is ended. We need to fix this at some point as well. 
I'll create a new issue just for this particular bug because it will require more changes to the code.

drillprop (approved):
Looks good to me 👍🏻

PR #1720, opened 2021-12-04
handle asset downloads in the new storage system

### LINKED ISSUE #1715: Handle downloading assets in new storage system

kdembler:
Resolves #1715 

First part of changes for Atlas to work with Giza. This PR breaks Studio but we will fix that when we handle uploads. At this point all of the viewer experience should work. 

**Notes:**
- **The Netlify deployment will most likely not work since it will override env variables to use `sumer-dev-2` instead of the Giza deployment. For now let's test locally, we will figure something out soon**
- Will reenable asset logger at a later point
- `@joystream/types` currently uses a local reference, let's keep it that way until we do uploads
- We're getting rid of the data object availability field (query node can now do nested filtering). We also get rid of URL assets
- Based on above, simplified `useAsset` API, it now just take a single argument - the data object
- Changed all the filtering inputs to accept nested arguments instead of shortcuts. So instead of doing `where: {channelId_eq: "xxx"}` we will do `where: {channel: {id_eq: "xxx"} }` should be more clear and future-proof

Happy to answer any questions on Giza


src/components/FiltersBar/useFiltersBar.ts
```diff
@@ -20,8 +20,7 @@ export const useFiltersBar = () => {
   const canClearVideoLengthFilter =
     videoWhereInput?.duration_lte !== undefined || videoWhereInput?.duration_gte !== undefined
   const canClearOtherFilters = videoWhereInput?.hasMarketing_eq === false || videoWhereInput?.isExplicit_eq === false
-  const canClearCategoriesFilter =
-    videoWhereInput?.categoryId_in !== undefined && videoWhereInput?.categoryId_in?.length !== 0
+  const canClearCategoriesFilter = videoWhereInput?.category !== undefined
```
rafalpawlow: without this, when you select and unselect category, clear button is not disabled
```suggestion
  const canClearCategoriesFilter = videoWhereInput?.category !== undefined || videoWhereInput?.category.id_in.length !== 0
```

src/providers/assets/assetsManager.tsx
```diff
@@ -1,117 +1,142 @@
...
+  dataObject: StorageDataObjectFieldsFragment
+): DistributorInfo[] => {
+  const dataObjectIdBn = new BN(dataObject.id)
+  const distributorsCountBn = new BN(distributors.length)
+  const firstDistributorIndex = dataObjectIdBn.mod(distributorsCountBn).toNumber()
+  return [...distributors.slice(firstDistributorIndex), ...distributors.slice(0, firstDistributorIndex)]
+}
+
+const createDistributorDataObjectUrl = (distributor: DistributorInfo, dataObject: StorageDataObjectFieldsFragment) => {
+  return joinUrlFragments(distributor.endpoint, 'api/v1/assets', dataObject.id)
```
rafalpawlow: Let's move this url to `config/urls.ts`. WDYT?

src/providers/distributors.tsx
```diff
@@ -0,0 +1,98 @@
+import { useApolloClient } from '@apollo/client'
+import React, { useCallback, useContext, useEffect, useRef, useState } from 'react'
+
+import {
+  GetDistributionBucketsWithOperatorsDocument,
+  GetDistributionBucketsWithOperatorsQuery,
+  GetDistributionBucketsWithOperatorsQueryVariables,
+} from '@/api/queries/__generated__/storage.generated'
```
kdembler: Should import from `@/api/queries`

rafalpawlow (changes requested):

src/providers/distributors.tsx
```diff
@@ -0,0 +1,98 @@
...
+DistributorsContext.displayName = 'DistributorsContext'
+
+export const DistributorsContextProvider: React.FC = ({ children }) => {
+  const distributorsMappingPromiseRef = useRef<Promise<DistributorsMapping>>()
+  const [distributorsError, setDistributorsError] = useState<unknown>(null)
+
+  const client = useApolloClient()
+
+  useEffect(() => {
+    const promise = client.query<
```
drillprop: Could we use a more meaningful name here?

src/providers/distributors.tsx
```diff
@@ -0,0 +1,98 @@
...
+      GetDistributionBucketsWithOperatorsQueryVariables
+    >({
+      query: GetDistributionBucketsWithOperatorsDocument,
+      fetchPolicy: 'network-only',
+    })
+    distributorsMappingPromiseRef.current = promise.then((result) => {
+      const mapping: DistributorsMapping = {}
+      const buckets = result.data.distributionBuckets
+      buckets.forEach((bucket) => {
+        const bags = bucket.bags.map((bag) => bag.id)
```
drillprop: Let's name it `bagIds` maybe? Bags suggest that we have something else here, not only strings. 

Also, what's a bag?

src/providers/distributors.tsx
```diff
@@ -0,0 +1,98 @@
...
+        const endpoints: DistributorInfo[] = bucket.operators
+          .filter((operator) => operator && operator.metadata)
+          .map((operator) => ({ id: operator.id, endpoint: operator.metadata?.nodeEndpoint || '' }))
+
+        bags.forEach((bag) => {
```
drillprop: if we agree on the comment above, let's rename `bag` to `bagId`

src/views/viewer/CategoryView/CategoryVideos.tsx
```diff
@@ -55,7 +55,9 @@ export const CategoryVideos: React.FC<{ categoryId: string }> = ({ categoryId })
 
   useEffect(() => {
     setVideoWhereInput({
-      categoryId_eq: categoryId,
+      category: {
+        id_eq: categoryId,
+      },
```
drillprop, line 60: Can't comment on this, but you possibly also need to update line 84. I'm getting this warning when I'm changing language: 
image

drillprop (changes requested):

src/providers/distributors.tsx
```diff
@@ -0,0 +1,98 @@
...
+const DistributorsContext = React.createContext<DistributorsContextValue | undefined>(undefined)
+DistributorsContext.displayName = 'DistributorsContext'
+
+export const DistributorsContextProvider: React.FC = ({ children }) => {
+  const distributorsMappingPromiseRef = useRef<Promise<DistributorsMapping>>()
+  const [distributorsError, setDistributorsError] = useState<unknown>(null)
+
+  const client = useApolloClient()
+
+  useEffect(() => {
```
eldiegod, line 28: Im guessing this effect fetches storage providers but a short comment description above the effect wouldn't hurt

src/components/FiltersBar/useFiltersBar.ts
kdembler: Adjusted your suggestion slightly, please check

src/providers/assets/assetsManager.tsx
kdembler: Moved `'api/v1/assets'` to `config/urls`

src/providers/distributors.tsx
kdembler: Changed to `distributorsPromise`, lmk if you'd like something with more meaning

src/providers/distributors.tsx
kdembler: Yes, good catch, changed.

A bag is a named storage "bucket". For example, each channel will have its own storage bag for its asset. The bag id would be `dynamic-channel-${id}`

src/providers/distributors.tsx
kdembler: Changed

src/views/viewer/CategoryView/CategoryVideos.tsx
kdembler: Updated 👍

src/providers/distributors.tsx
kdembler: added a comment

rafalpawlow (approved):

eldiegod (approved):

drillprop (approved):

PR #1758, opened 2021-12-06
Video category auto playback based on type of screen

### LINKED ISSUE #1690: Use a cursor media query for automatic featured category card playback
Our current featured category card has a background video that will play on desktop when you hover over the card. In card's `compact` variant, the playback will be automatic instead. However, we pick the `default` and `compact` variant by the device's screen size. Now you can imagine a big tablet used in a horizontal mode that satisfies minimum screen width for the `default` variant. However, on the tablet you don't have a pointer so you won't be able to hover over the card to trigger playback.

To fix this problem, we could base the automatic playback not on the variant of the card, but on a media query that lets us know whether the device supports a pointer (`@media (any-pointer: coarse)` can be used for that)

rafalpawlow:
Closes #1690

### DIFF
--- a/src/components/_video/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
+++ b/src/components/_video/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
@@ -6,6 +6,7 @@ import { SkeletonLoader } from '@/components/_loaders/SkeletonLoader'
 import { BackgroundVideoPlayer } from '@/components/_video/BackgroundVideoPlayer'
 import { absoluteRoutes } from '@/config/routes'
 import { useHover } from '@/hooks/useHover'
+import { useTouchDevice } from '@/hooks/useTouchDevice'
 import { sizes, transitions } from '@/styles'
 
 import {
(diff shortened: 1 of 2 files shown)


src/components/_video/VideoCategoryCard/FeaturedVideoCategoryCard.tsx
```diff
@@ -39,7 +39,8 @@ export const FeaturedVideoCategoryCard: React.FC<FeaturedVideoCategoryCardProps>
 }) => {
   const [hoverRef, isVideoHovering] = useHover<HTMLAnchorElement>()
   const isLoading = !title && !videoUrl && !icon
-  const isPlaying = variant === 'default' ? isVideoHovering : true
+  const isTouchDevice = !!('ontouchstart' in window || navigator.maxTouchPoints)
```
kdembler: Could we use a media query instead? This could work similar way that our `useMediaMatch` does. This should do the trick: `@media (any-pointer: coarse)`

kdembler (approved):

PR #1759, opened 2021-12-07
add note about mutation overriding previous vids

mochet:
Added note clarifying that a mutation submitted via Orion will override whatever was previously set.


kdembler (approved):
Thanks a bunch!

PR #1775, opened 2021-12-09
Indirect sign in dialog

### LINKED ISSUE #1653: Unify sign in flow
Unify sign-in experience for both experiences publisher and viewer

Please note, that the new Sign in flow requires having implemented other recently changed/introduced components:

- https://github.com/Joystream/atlas/issues/1645
- https://github.com/Joystream/atlas/issues/1588
- https://github.com/Joystream/atlas-resources/issues/2#issuecomment-979026239

drillprop:
Partly resolves #1653 
The scope of this PR is to create just simple indirect dialog. You can find the design here: 
image
https://www.figma.com/file/ymHYOFItrmFaLeBC5Z4tpQ/Sign-in-(draft-once-again)?node-id=513%3A18747

To see this hook in action you can go to the playground - `/playground/indirect-signin-dialog`. Buttons will do nothing and this is intended because we can't buy, bid or comment yet.


eldiegod (approved):
Looks legit, the buttons switch one breakpoint too early from stacked to columns but that's how our confirmation dialog works so I would say that should be more of a design inconsistency probably

Issue #1645, opened 2021-11-24
Create a new member/channel dropdown to allow for having memberships in the viewing experience


toiletgranny:
### Description

Based on information gathered from @Lenorette, as part of the NFT MVP scope of work.

### How urgent this is?

It's blocking me right now

toiletgranny:
The new Member dropdown component, along with a full documentation designers & engineers, can be found over here: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3201%3A111989

Please note, that the new Member dropdown requires having implemented other recently changed/introduced components:
- https://github.com/Joystream/atlas/issues/1649
- https://github.com/Joystream/atlas/issues/1656
- https://github.com/Joystream/atlas/issues/1589

toiletgranny:
**⚠️ Attention**
_I've mistakenly started a discussion regarding this component in the wrong GitHub issue. Discussion relevant to this component can be found over here: https://github.com/Joystream/atlas/issues/1588_

kdembler:
I LOVE IT

kdembler:
Displaying number of tJOYs will be handled in #1762

We will do member switch animation in #1763

PR #1788, opened 2021-12-14
Unify sign in flow

drillprop:
Fix #1653 
Summary of changes
- removed old `create member` view and replace it with `create membership` dialog
- fixed validation for changing Avatar
- moved all auth components to `_auth`
- various changes to AccountStep(introduce Identicon for Polkadot, margin and padding tweaks)
- added loader to Viewer

There are some things which still needs to be done: 
- Didn't remove the studio sign-in page yet, because it's impossible to change the member without a new member dropdown. For the same reason, you can't change the member on the viewer side.
- I also talked to Adam about the scrollbar in `AccountStep` and we probably should have scroll on the dialog container not the accounts wrapper like we have now. Unfortunately, It will require a lot of changes to our `StepperDialog`, so I just left this for another time. 
How the scrollbar is positioned right now: 
image
How it should be positioned: 
image

PR #1789, opened 2021-12-14
Member dropdown

eldiegod:
Fixes: #1645

Testing path: '/playground/member-dropdown'

Figma: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=110%3A18281

Missing things to come later at some point:
- Real Tjoy balance
- Learn about tJOY link
- my profile link
- switch member animation


src/components/_overlays/MemberDropdown/MemberDropdown.tsx
```diff
@@ -0,0 +1,204 @@
...
+    setIsSwitchingMember(false)
+  }
+
+  return (
+    <Container>
+      {isSwitchingMember ? (
+        <div>
+          <SwitchMemberItemListContainer>
+            <ListItem
+              onClick={() => setIsSwitchingMember(false)}
```
drillprop: We had already `handleSwitchMemberMode` which we could use here as well(you'd have to add a parameter to the function of course)

src/components/_overlays/MemberDropdown/MemberDropdown.tsx
```diff
@@ -0,0 +1,204 @@
...
+                onClick={() => handleMemberChange(member.id)}
+                nodeStart={<Avatar assetUrl={member.avatarUri} />}
+                label={member.handle ?? ''}
+                selected={member.id === activeMembership?.id}
+              />
+            ))}
+            <ListItem
+              nodeStart={<IconWrapper icon={<SvgActionNewChannel />} />}
+              onClick={() => handleAddNewMember()}
+              label={'Add new member...'}
```
drillprop:
```suggestion
              label="Add new member..."
```
Ok, we should have some eslint/prettier plugin already for changing that 😅  I think [this one](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md) should work, but let's do this in another PR.

src/components/_overlays/MemberDropdown/MemberDropdown.tsx
```diff
@@ -0,0 +1,204 @@
...
+            />
+          </SectionContainer>
+        </div>
+      ) : (
+        <div>
+          <BlurredBG url={activeMembership?.avatarUri}>
+            <Filter />
+            <MemberInfoContainer>
+              <StyledAvatar size="fill" assetUrl={activeMembership?.avatarUri}></StyledAvatar>
```
drillprop: Let's collapse this empty tag

src/components/_overlays/MemberDropdown/MemberDropdown.tsx
```diff
@@ -0,0 +1,204 @@
...
+                  <Text
+                    variant="t100"
+                    as="a"
+                    // @ts-ignore our types don't allow this but its fine here
+                    href="https://www.joystream.org/"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    secondary
+                    color={cVar('colorCoreNeutral200Lighten')}
+                  >
```
drillprop: In the design `Learn about tJOY` isn't underlined. 
image
But to be honest I like underlined text here. It's suggesting that it's a clickable link. Could you ask @toiletgranny if we could leave this without change as you have here?

drillprop: btw. I'm fine with @ts-ignore

src/components/_overlays/MemberDropdown/MemberDropdown.tsx
```diff
@@ -0,0 +1,204 @@
...
+            ))}
+            <ListItem
+              nodeStart={<IconWrapper icon={<SvgActionNewChannel />} />}
+              onClick={() => handleAddNewMember()}
+              label={'Add new member...'}
+            />
+          </SectionContainer>
+        </div>
+      ) : (
+        <div>
```
drillprop: I think we don't need a div here, we could use Fragment.

drillprop (changes requested):
https://user-images.githubusercontent.com/51168865/146333258-022665a1-8bc4-412d-a9f5-656cb24d9f16.mov

I've noticed some layout shifts when switching members. Could we do something about it?
Besides that, found some minor stuff. Looks pretty good in general 👍🏻

kdembler:
Is there any specific reason to not use it in Studio in place of the old one right away?

src/components/_overlays/MemberDropdown/MemberDropdown.tsx
eldiegod: @toiletgranny what do you think?

src/components/_overlays/MemberDropdown/MemberDropdown.tsx
eldiegod: I think since the func is just the setter, it's not needed, will remove and use the setter instead for now

eldiegod:
> Is there any specific reason to not use it in Studio in place of the old one right away?

@kdembler added it to studio topbar but now we are lacking the logout action

src/components/_overlays/MemberDropdown/MemberDropdown.tsx
drillprop: Talked to Adam about this, and he decided that it should be underlined only when you're hovering over this link.

toiletgranny:
@eldiegod, @kdembler, @drillprop Great work, looks stunning! Can't wait to see that transition animation when going into "Switch member" mode! For all other comments and questions, I recorded a quick loom (4 min). https://www.loom.com/share/435e049ab7ce476e8510c2051af73578

eldiegod:
@toiletgranny I think I implemented everything you mentioned besides the `Select first channel when switching members to a member with channels` part because this is probably more suitable as an issue on its own since it has more to do with our internal user logic than with the member dropdown itself

drillprop:
> besides the `Select first channel when switching members to a member with channels` part because this is probably more suitable as an issue on its own since it has more to do with our internal user logic than with the member dropdown itself 

I'll deal with it in https://github.com/Joystream/atlas/pull/1788 or in follow-up PR once this member dropdown PR is merged. I also think that it will be better to merge your commits into `Joystream:signin`, not `Joystream:dev`, wdyt?

src/components/_navigation/TopbarStudio/TopbarStudio.tsx
```diff
@@ -50,90 +38,53 @@ type ChannelInfoProps = {
...
+  //     resetActiveUser()
+  //     navigate(absoluteRoutes.studio.index())
+  //   }
+  //   setIsMemberDropdownActive(false)
+  //   if (anyVideoTabsCachedAssets) {
+  //     openWarningDialog({ onConfirm: () => logout() })
+  //   } else {
+  //     logout()
+  //   }
+  // }
```
drillprop: Let's just remove this

src/components/ListItem/ListItem.tsx
```diff
@@ -40,16 +45,46 @@ export const ListItem = React.forwardRef<HTMLDivElement, ListItemProps>(
...
     return (
-      <Container onClick={onClick} disabled={disabled} hasNodeStart={!!nodeStart} size={size} ref={ref}>
-        {!!nodeStart && <NodeContainer destructive={destructive}>{nodeStart}</NodeContainer>}
+      <Container
+        className={className}
+        onClick={onClick}
+        disabled={disabled}
+        hasNodeStart={!!nodeStart}
+        size={size}
+        ref={mergeRefs([hoverRef, ref])}
```
drillprop, line 63: hmm, that's a cool library!

eldiegod: yea right

toiletgranny:
> @toiletgranny I think I implemented everything you mentioned besides the `Select first channel when switching members to a member with channels` part because this is probably more suitable as an issue on its own since it has more to do with our internal user logic than with the member dropdown itself

Ok, thanks! Visually, everything looks good now! Based on your reply, I'd assume that closing the member dropdown upon selecting a different member, which I also mentioned in the loom video, would also be tackled in that separate issue, right?
CleanShot 2021-12-20 at 12 25 55

Other than this — it's a go from my side!

eldiegod:
> > besides the `Select first channel when switching members to a member with channels` part because this is probably more suitable as an issue on its own since it has more to do with our internal user logic than with the member dropdown itself
> 
> I'll deal with it in #1788 or in follow-up PR once this member dropdown PR is merged. I also think that it will be better to merge your commits into `Joystream:signin`, not `Joystream:dev`, wdyt?

Nice!

~~About merging; I'm not sure, I would merge to dev to not make that release bigger but is not clear, thoughts @kdembler ?~~

nvm we are merging to signin

eldiegod:
> > @toiletgranny I think I implemented everything you mentioned besides the `Select first channel when switching members to a member with channels` part because this is probably more suitable as an issue on its own since it has more to do with our internal user logic than with the member dropdown itself
> 
> Ok, thanks! Visually, everything looks good now! Based on your reply, I'd assume that closing the member dropdown upon selecting a different member, which I also mentioned in the loom video, would also be tackled in that separate issue, right? CleanShot 2021-12-20 at 12 25 55
> 
> Other than this — it's a go from my side!

@toiletgranny  No we don't have to wait for that I just added it so we should be feature complete here

src/components/_overlays/MemberDropdown/MemberDropdown.styles.ts
```diff
@@ -0,0 +1,113 @@
...
+const paddingStyles = css`
+  padding: ${sizes(6)} ${sizes(4)};
+`
+
+export const Container = styled.div<{ isActive: boolean }>`
+  position: fixed;
+  right: ${sizes(4)};
+  top: 0;
+  width: 280px;
+  max-height: calc(${window.innerHeight - sizes(4, true)}px - var(--size-topbar-height));
```
kdembler: Couldn't we use just `100vh` instead? We have a guarantee this will always be responsive, `window.innerHeight` is iffy

kdembler (commented):
Just a small comment from myself, other than that looks great to me!

I will still wait for final approval from @drillprop

drillprop (approved):
Looks good to me! 🎉

src/components/_overlays/MemberDropdown/MemberDropdown.styles.ts
eldiegod: yeah this is better just pushed it

kdembler (approved):

Issue #1776, opened 2021-12-09
Node connection indicator is not centered


rafalpawlow:
**Describe the bug**

Connection indicator has unnecessary left margin

**Expected behavior**

Connection indicator should be centered relative to the content

**Screenshots**

![Zrzut ekranu 2021-12-9 o 13 53 28](https://user-images.githubusercontent.com/13706246/145400427-75b2d4eb-35bf-4a4e-bcb0-0e51a1d635e4.png)

PR #1795, opened 2021-12-15
Center node connection indicator

eldiegod:
fixes: #1776

### DIFF
--- a/src/components/NoConnectionIndicator/NoConnectionIndicator.styles.ts
+++ b/src/components/NoConnectionIndicator/NoConnectionIndicator.styles.ts
@@ -35,7 +35,7 @@ export const IndicatorWrapper = styled.div`
   padding: ${sizes(3)} ${sizes(5)};
 
   ${media.md} {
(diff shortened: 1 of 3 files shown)


rafalpawlow (changes requested):
It's still not centered. Shouldn't we get rid of this margin?
Zrzut ekranu 2021-12-17 o 14 17 19
Zrzut ekranu 2021-12-17 o 14 17 56

eldiegod:
> It's still not centered. Shouldn't we get rid of this margin? Zrzut ekranu 2021-12-17 o 14 17 19 Zrzut ekranu 2021-12-17 o 14 17 56

No, it just needed to be conditionally applied

rafalpawlow (changes requested):
Now content has no padding for side nav
Zrzut ekranu 2021-12-20 o 18 55 12

eldiegod:
That didnt work out too well so tried a diff approach

rafalpawlow (approved):
Looks good now!

Issue #1768, opened 2021-12-08
Video player ending overlay allows starting playback 


drillprop:
**Describe the bug**

This is very similar to #1640. We shouldn't be able to replay a video by clicking on the background when a video is ended. 

**To Reproduce**

1. Go to video,
2. Wait until the end 
3. Click anywhere on the background
4. Video is playing again

**Expected behavior**

When you click on the ending overlay background it shouldn't be possible to play a video again.

**Device (please complete the following information):**

- OS: [e.g. macOS Monterey]
- Browser [e.g. Chrome]

PR #1814, opened 2021-12-17
Fix: clicking video end overlay makes it playback

eldiegod:
Fixes: #1768

Additional quality of life change:
Made it so clicking on the next video card on the video end overlay
takes you to that video

### DIFF
--- a/src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
+++ b/src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
@@ -32,7 +32,7 @@ type EndingOverlayProps = {
   isEnded: boolean
 }
 // 10 seconds
-const NEXT_VIDEO_TIMEOUT = 10000
+const NEXT_VIDEO_TIMEOUT = 10 * 1000
 
 export const EndingOverlay: React.FC<EndingOverlayProps> = ({
   onPlayAgain,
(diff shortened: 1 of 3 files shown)


src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -103,17 +103,32 @@ export const EndingOverlay: React.FC<EndingOverlayProps> = ({
...
+      onClick={(e) => {
+        e.stopPropagation()
+      }}
+    >
       {randomNextVideo ? (
         <Container isFullScreen={isFullScreen}>
-          <InnerContainer>
+          <InnerContainer
+            onClick={() => {
+              navigate(absoluteRoutes.viewer.video(randomNextVideo.id))
```
rafalpawlow, line 115: As above, move inline function to const

src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -103,17 +103,32 @@ export const EndingOverlay: React.FC<EndingOverlayProps> = ({
...
             <VideoThumbnail src={thumbnailUrl} />
             <VideoInfo>
               <Text variant={mdMatch ? 't300' : 't200'} secondary>
                 Up next
               </Text>
               <Heading variant={mdMatch ? 'h500' : 'h400'}>{randomNextVideo.title}</Heading>
-              <StyledChannelLink id={channelId} avatarSize="default" textVariant={mdMatch ? 't300' : 't200'} />
+              <StyledChannelLink
+                onClick={(e) => {
+                  e.stopPropagation()
```
rafalpawlow: As above, move inline function to const

src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -103,17 +103,32 @@ export const EndingOverlay: React.FC<EndingOverlayProps> = ({
   }, [currentThumbnailUrl, randomNextVideo, randomNextVideoThumbnailUrl])
 
   return (
-    <OverlayBackground>
+    <OverlayBackground
+      onClick={(e) => {
```
rafalpawlow: Let's move this function to const. Inline functions creates new reference on each render which cause unnecessary rerender.

src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
```diff
@@ -32,7 +32,7 @@ type EndingOverlayProps = {
   isEnded: boolean
 }
 // 10 seconds
-const NEXT_VIDEO_TIMEOUT = 10000
+const NEXT_VIDEO_TIMEOUT = 10 * 1000
```
rafalpawlow, line 35: Why did you make this change?

rafalpawlow (changes requested):

src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
kdembler: Moving it to const doesn't really change anything, `const` will have a new reference on each render as well

rafalpawlow: Yeah I was thinking about `useCallback`. But either way, we have 2 inline functions here which does the same, so maybe it's worth to declare it above?

kdembler: Sure let's move it to `const` if we duplicate but let's not overuse `useCallback` if not necessary

src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.tsx
eldiegod: just to make it clearer that it is 10 secs and easier to change later

rafalpawlow (approved):

PR #1815, opened 2021-12-17
2 Needed eslint rules and its --fix result

eldiegod:
will self close empty tags and will remove
curly braces around string props or children when not needed

'react/jsx-curly-brace-presence':
'react/self-closing-comp'


drillprop (approved):
Thanks 🙏🏻

kdembler (approved):
I love it!

PR #1816, opened 2021-12-17
joystream api as web worker

### LINKED ISSUE #1764: Extract Polkadot logic to a worker thread

rafalpawlow:
Closes #1764

### DIFF
--- a/src/joystream-lib/api.ts
+++ b/src/joystream-lib/api.ts
@@ -491,8 +491,8 @@ export class JoystreamJs {
 
   private _prepareVideoAssets(inputAssets: InputAssets) {
     return {
-      videoContent: inputAssets.video && this._createFileAsset(inputAssets.video),
-      thumbnailContent: inputAssets.thumbnail && this._createFileAsset(inputAssets.thumbnail),
+      videoContent: inputAssets.video ? this._createFileAsset(inputAssets.video) : [],
+      thumbnailContent: inputAssets.thumbnail ? this._createFileAsset(inputAssets.thumbnail) : [],
     }
   }
 
@@ -507,7 +507,6 @@ export class JoystreamJs {
       videoContent: [video, videoId],
       thumbnailContent: [thumbnail, thumbnailId],
     } = this._prepareVideoAssets(inputAssets)
-    await this.ensureApi()
 
     return this._createOrUpdateVideo(
       null,
@@ -529,9 +528,9 @@ export class JoystreamJs {
     cb?: ExtrinsicStatusCallbackFn
   ): Promise<ExtrinsicResult<VideoId>> {
     const { videoContent, thumbnailContent } = this._prepareVideoAssets(inputAssets)
-    const assets: { [field: string]: [NewAsset, string] | undefined } = {
-      videoAsset: videoContent || undefined,
-      thumbnailAsset: thumbnailContent || undefined,
+    const assets: { [field: string]: [NewAsset, string] | never[] } = {
+      videoAsset: videoContent || [],
+      thumbnailAsset: thumbnailContent || [],
     }
     const contentIds = {
       videoId: assets.videoAsset ? assets.videoAsset[1] : '',
--- a/src/joystream-lib/api.ts
+++ b/src/joystream-lib/api.ts
@@ -31,6 +31,7 @@ import {
 } from '@polkadot/types'
 import { DispatchError } from '@polkadot/types/interfaces/system'
 import BN from 'bn.js'
+import { transfer } from 'comlink'
 
 import { ConsoleLogger, SentryLogger } from '@/utils/logs'
 
@@ -391,7 +392,6 @@ export class JoystreamJs {
   }
 
   /* Public */
-  // @ts-ignore test
   async setActiveAccount(accountId: AccountId | null, signer?: Signer) {
     if (!accountId) {
       this._selectedAccountId = null
@@ -413,7 +413,7 @@ export class JoystreamJs {
     return new BN(balance.freeBalance).toNumber()
   }
 
-  createFileAsset({ ipfsContentId, size }: AssetMetadata): [NewAsset, string] {
+  private _createFileAsset({ ipfsContentId, size }: AssetMetadata): [NewAsset, string] {
     const contentId = ContentId.generate(this.api.registry)
     const b = new Bytes(this.api.registry, ipfsContentId)
     const content = new ContentParameters(this.api.registry, {
@@ -421,44 +421,68 @@ export class JoystreamJs {
       // hardcoded type_id - it's not used but needs to be one of the allowed values
       type_id: new U64(this.api.registry, 1),
       size: new U64(this.api.registry, size),
-      ipfs_content_id: b,
+      ipfs_content_id: transfer(b, [b.buffer]),
     })
     return [new NewAsset(this.api.registry, { upload: content }), contentId.encode()]
+  }
+
+  private _prepareChannelAssets(inputAssets: InputAssets) {
+    return {
+      avatarContent: inputAssets.avatar && this._createFileAsset(inputAssets.avatar),
+      coverContent: inputAssets.cover && this._createFileAsset(inputAssets.cover),
+    }
   }
 
   async createChannel(
     memberId: MemberId,
     inputMetadata: CreateChannelMetadata,
-    inputAssets: ChannelAssets,
-    cb?: ExtrinsicStatusCallbackFn
+    inputAssets: InputAssets,
+    cb?: ExtrinsicStatusCallbackFn,
+    contentIdCb?: (contentId: ContentIdCbArgs) => void
   ): Promise<ExtrinsicResult<ChannelId>> {
+    const {
+      avatarContent: [avatar, avatarId],
+      coverContent: [cover, coverId],
+    } = this._prepareChannelAssets(inputAssets)
+    await contentIdCb?.({ avatarId, coverId })
     await this.ensureApi()
 
-    return this._createOrUpdateChannel(null, memberId, inputMetadata, inputAssets, cb)
+    return this._createOrUpdateChannel(null, memberId, inputMetadata, { avatar, cover }, cb)
   }
 
   async updateChannel(
     channelId: ChannelId,
     memberId: MemberId,
     inputMetadata: CreateChannelMetadata,
-    inputAssets: ChannelAssets,
-    cb?: ExtrinsicStatusCallbackFn
+    inputAssets: InputAssets,
+    cb?: ExtrinsicStatusCallbackFn,
+    contentIdCb?: (contentId: ContentIdCbArgs) => void
   ): Promise<ExtrinsicResult<ChannelId>> {
+    const { avatarContent, coverContent } = this._prepareChannelAssets(inputAssets)
+    const assets: { [field: string]: [NewAsset, string] | undefined } = {
+      avatarAsset: avatarContent || undefined,
+      coverAsset: coverContent || undefined,
+    }
+    await contentIdCb?.({
+      avatarId: assets.avatarAsset ? assets.avatarAsset[1] : '',
+      coverId: assets.avatarAsset ? assets.avatarAsset[1] : '',
+    })
     await this.ensureApi()
 
-    return this._createOrUpdateChannel(channelId, memberId, inputMetadata, inputAssets, cb)
+    return this._createOrUpdateChannel(
+      channelId,
+      memberId,
+      inputMetadata,
+      { avatar: assets.avatarAsset && assets.avatarAsset[0], cover: assets.coverAssets && assets.coverAssets[0] },
+      cb
+    )
   }
 
-  private _prepareAssets(inputAssets: InputAssets) {
-    // @ts-ignore test
-    const [video, videoId] = this.createFileAsset(inputAssets.video)
-    // @ts-ignore test
-    const [thumbnail, thumbnailId] = this.createFileAsset(inputAssets.thumbnail)
-    const assets: VideoAssets = {
-      video,
-      thumbnail,
+  private _prepareVideoAssets(inputAssets: InputAssets) {
+    return {
+      videoContent: inputAssets.video && this._createFileAsset(inputAssets.video),
+      thumbnailContent: inputAssets.thumbnail && this._createFileAsset(inputAssets.thumbnail),
     }
-    return { assets, thumbnailId, videoId }
   }
 
   async createVideo(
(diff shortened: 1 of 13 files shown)


src/joystream-lib/api.ts
```diff
@@ -411,65 +413,127 @@ export class JoystreamJs {
...
+  private _createFileAsset({ ipfsContentId, size }: AssetMetadata): [NewAsset, string] {
     const contentId = ContentId.generate(this.api.registry)
     const b = new Bytes(this.api.registry, ipfsContentId)
     const content = new ContentParameters(this.api.registry, {
       content_id: contentId,
       // hardcoded type_id - it's not used but needs to be one of the allowed values
       type_id: new U64(this.api.registry, 1),
       size: new U64(this.api.registry, size),
-      ipfs_content_id: b,
+      ipfs_content_id: transfer(b, [b.buffer]),
```
kdembler: Why do we need this? I would like to avoid having any `comlink` specific code in here

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -319,21 +316,37 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
...
-          ? 'Your channel was created and saved on the blockchain. Feel free to start using it!'
-          : 'Changes to your channel were saved on the blockchain.',
-      },
-    })
+    const setContentIds = async ({ avatarId, coverId }: ContentIdCbArgs) => {
+      avatarContentId = avatarId
+      coverContentId = coverId
+    }
+
+    const instance = await joystream
```
kdembler: I don't fully understand why we need this. Couldn't this be awaited in the provider so that consumers don't need to run this `await`?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -319,21 +316,37 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
...
+        txFactory: async (updateStatus) =>
+          newChannel
+            ? await instance.createChannel(activeMemberId, metadata, assets, proxy(updateStatus), proxy(setContentIds))
+            : await instance.updateChannel(
+                activeChannelId ?? '',
+                activeMemberId,
+                metadata,
+                assets,
+                proxy(updateStatus),
+                proxy(setContentIds)
```
kdembler: I think we could do this without actually proxying a callback. Instead, we could just return the content IDs from the `updateChannel` function, then `onTxFinalize` would get access to the content IDs. Makes sense?

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -319,21 +316,37 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
...
+        preProcess: processAssets,
+        txFactory: async (updateStatus) =>
+          newChannel
+            ? await instance.createChannel(activeMemberId, metadata, assets, proxy(updateStatus), proxy(setContentIds))
+            : await instance.updateChannel(
+                activeChannelId ?? '',
+                activeMemberId,
+                metadata,
+                assets,
+                proxy(updateStatus),
```
kdembler: I'm wondering about those `proxy` calls, I'm not sure if they should be inside our components, seems like internal workings are leaking. Although I guess they are necessary when callbacks are involved. I'm considering potentially having wrappers for all the joystream functions in one place (provider?) and we only use `comlink` there so it doesn't affect the components. WDYT?

src/providers/joystream/provider.tsx
```diff
@@ -1,77 +1,100 @@
...
     (connected: boolean) => {
       setNodeConnection(connected ? 'connected' : 'disconnected')
     },
     [setNodeConnection]
   )
+  const joystream = useRef<Remote<JoystreamJs> | undefined>()
 
-  useEffect(() => {
-    let joystream: JoystreamJs
+  const proxyCallback = <T extends object>(callback: T) => proxy(callback)
```
kdembler: We pass this as hook dependency, let's move it into `useCallback`

src/providers/joystream/provider.tsx
```diff
@@ -1,77 +1,100 @@
...
-        joystream = new JoystreamJs(nodeOverride || NODE_URL)
-        setJoystream(joystream)
-
-        joystream.onNodeConnectionUpdate = handleNodeConnectionUpdate
+        joystream.current = await new api(nodeOverride ?? NODE_URL, proxy(handleNodeConnectionUpdate))
+        setInitialized(true)
       } catch (e) {
         handleNodeConnectionUpdate(false)
-        SentryLogger.error('Failed to create JoystreamJS instance', 'JoystreamProvider', e)
+        SentryLogger.error('Failed to create JoystreamJS joystream', 'JoystreamProvider', e)
```
kdembler: I suspect this is a mistake of find&replace

src/providers/joystream/provider.tsx
```diff
@@ -1,77 +1,100 @@
...
   )
+  const joystream = useRef<Remote<JoystreamJs> | undefined>()
 
-  useEffect(() => {
-    let joystream: JoystreamJs
+  const proxyCallback = <T extends object>(callback: T) => proxy(callback)
 
-    const init = async () => {
+  useEffect(() => {
+    const getjoystream = async () => {
```
kdembler: could we make this camelCase?

src/providers/joystream/provider.tsx
```diff
@@ -1,77 +1,100 @@
...
-
+    getjoystream()
     return () => {
-      joystream?.destroy()
+      const destroy = async () => {
+        if (joystream.current) {
+          await joystream.current.destroy()
+        }
+      }
+      destroy()
```
kdembler: What's a difference between this and:

```suggestion
      joystream.current?.destroy()
```

src/providers/joystream/provider.tsx
```diff
@@ -1,77 +1,100 @@
...
-      return
-    }
-
-    if (joystream.selectedAccountId === activeAccountId) {
+    if (!initialized) {
       return
     }
+    const init = async () => {
+      const instance = joystream.current
+      const accountId = await instance?.selectedAccountId
```
kdembler: Why do we need this `await`? Also I think we could move this after the first `if`, like in the first version?

src/providers/joystream/provider.tsx
```diff
@@ -1,77 +1,100 @@
...
-        const accountInjector = await web3FromAddress(activeAccountId)
-        joystream.setActiveAccount(activeAccountId, accountInjector.signer)
-      } else {
-        joystream.setActiveAccount(activeAccountId)
+      const setActiveAccount = async () => {
+        if (activeAccountId) {
+          const { signer } = await web3FromAddress(activeAccountId)
+          if (!signer) {
+            return
+          }
```
kdembler: We should probably log something here, it would be bad if this happened

src/providers/joystream/provider.tsx
```diff
@@ -1,77 +1,100 @@
...
+          const { signRaw, signPayload } = signer
+          await instance.setActiveAccount(
+            activeAccountId,
+            proxy({ signRaw: proxy(signRaw), signPayload: proxy(signPayload) })
```
kdembler: Doesn't just `proxy(signer)` work here?

src/joystream-lib/types.ts
```diff
@@ -29,6 +29,10 @@ export type VideoAssets = {
   video?: NewAsset
 }
 
+export type ContentIdArgs = {
+  [fieldName: string]: string
+}
```
kdembler: We should:
1. Update the name since this is no longer args as I understand. Maybe `ExtrinsicContentIds`? The name will be a better fit after Giza but that should come soon
2. Introduce some typesafety, currently this is prone to typos. This could be:
```tsx

type ExtrinsicChannelContentIds = {
  avatarPhotoId?: string
  coverPhotoId?: string
}

type ExtrinsicVideoContentIds = {
  thumbnailPhotoId?: string
  mediaId?: string
}

export type ExtrinsicContentIds = ExtrinsicChannelContentIds | ExtrinsicVideoContentIds
```

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -319,21 +313,29 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
...
-      onTxSync: refetchDataAndCacheAssets,
-      successMessage: {
-        title: newChannel ? 'Channel successfully created!' : 'Channel successfully updated!',
-        description: newChannel
-          ? 'Your channel was created and saved on the blockchain. Feel free to start using it!'
-          : 'Changes to your channel were saved on the blockchain.',
-      },
-    })
+    const completed =
+      joystream &&
```
kdembler: We already check this in L234, let's remove this

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -319,21 +313,29 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
...
+            ? await joystream.createChannel(activeMemberId, metadata, assets, proxyCallback(updateStatus))
+            : await joystream.updateChannel(
+                activeChannelId ?? '',
+                activeMemberId,
+                metadata,
+                assets,
+                proxyCallback(updateStatus)
+              ),
+        onTxFinalize: await uploadAssets,
+        onTxSync: await refetchDataAndCacheAssets,
```
kdembler: Why do we have all these additional `await`s here? I don't think they add anything

src/views/studio/VideoWorkspace/VideoWorkspaceForm/VideoWorkspaceForm.tsx
```diff
@@ -406,28 +400,39 @@ export const VideoWorkspaceForm: React.FC<VideoWorkspaceFormProps> = React.memo(
...
         }
 
-        const completed = await handleTransaction({
-          preProcess: processAssets,
-          txFactory: (updateStatus) =>
-            isNew
-              ? joystream.createVideo(activeMemberId, activeChannelId, metadata, assets, updateStatus)
-              : joystream.updateVideo(
+        const completed =
+          joystream &&
```
kdembler: Same comment as in channel edit - we already check this

src/views/studio/VideoWorkspace/VideoWorkspaceForm/VideoWorkspaceForm.tsx
```diff
@@ -406,28 +400,39 @@ export const VideoWorkspaceForm: React.FC<VideoWorkspaceFormProps> = React.memo(
...
-              ? 'Your video was created and saved on the blockchain. Upload of video assets may still be in progress.'
-              : 'Changes to your video were saved on the blockchain.',
-          },
-        })
+                  proxyCallback(updateStatus)
+                )
+              }
+            },
+            onTxFinalize: await uploadAssets,
+            onTxSync: await refetchDataAndCacheAssets,
```
kdembler: Same as in channel edit - why do we need these `await`s?

src/joystream-lib/api.ts
```diff
@@ -65,20 +67,20 @@ export class JoystreamJs {
   }
 
   // if needed these could become some kind of event emitter
-  public onNodeConnectionUpdate?: (connected: boolean) => unknown
+  public onNodeConnectionUpdate?: (connected: boolean) => void
```
kdembler: Do we need this as a class field anymore now that constructor gets it from the argument?

kdembler (changes requested):
Some code details in the comments.

Other than those, I did quite extensive testing and haven't found any major problems, everything seems to be working fine on localhost.

One thing I did find, is that we have a regression in handling of cancelled signature. Previously, when you get a popup from Polkadot and you cancel the signature, the transaction modal would disappear and we would display a snackbar saying that the signature was cancelled. Now you will see an `Unexpected error` state of the transaction modal.

src/providers/joystream/provider.tsx
rafalpawlow: `instance?.selectedAccountId` is a promise, so we need it here

src/joystream-lib/types.ts
```diff
@@ -42,4 +54,9 @@ export type ExtrinsicStatusCallbackFn = (status: ExtrinsicStatus.Unsigned | Extr
 export type ExtrinsicResult<T> = {
   block: number
   data: T
+  contentIds?: ExtrinsicContentIds
+}
+
+export type InputAssets = {
```
kdembler: Let's improve this type, similar to what we did with `ExtrinsicContentIds`, to not allow arbitrary fields, but rather have options predefined

src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -321,10 +315,16 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
 
     const completed = await handleTransaction({
       preProcess: processAssets,
-      txFactory: (updateStatus) =>
+      txFactory: async (updateStatus) =>
         newChannel
-          ? joystream.createChannel(activeMemberId, metadata, assets, updateStatus)
-          : joystream.updateChannel(activeChannelId ?? '', activeMemberId, metadata, assets, updateStatus),
+          ? await joystream.createChannel(activeMemberId, metadata, assets, proxyCallback(updateStatus))
+          : await joystream.updateChannel(
```
kdembler: Do we need those `async` and `await`s?

src/views/studio/VideoWorkspace/VideoWorkspaceForm/VideoWorkspaceForm.tsx
```diff
@@ -408,17 +402,26 @@ export const VideoWorkspaceForm: React.FC<VideoWorkspaceFormProps> = React.memo(
...
+            if (isNew) {
+              return await joystream.createVideo(
+                activeMemberId,
+                activeChannelId,
+                metadata,
+                assets,
+                proxyCallback(updateStatus)
+              )
+            } else {
+              return await joystream.updateVideo(
```
kdembler: Do we need those `async` and `await`s?

src/joystream-lib/api.ts
```diff
@@ -424,52 +427,130 @@ export class JoystreamJs {
...
+      thumbnailAsset: thumbnailContent || undefined,
+    }
+    const contentIds = {
+      videoId: assets.videoAsset ? assets.videoAsset[1] : '',
+      thumbnailId: assets.thumbnailAsset ? assets.thumbnailAsset[1] : '',
+    }
     await this.ensureApi()
 
-    return this._createOrUpdateVideo(videoId, memberId, channelId, inputMetadata, inputAssets, cb)
+    return this._createOrUpdateVideo(
```
kdembler, line 490: There isn't a good place for this comment so just leaving it here - the error with cancelled signature is still there

kdembler (changes requested):

kdembler (approved):

PR #1823, opened 2021-12-21
Design documentation cleanup

toiletgranny:
@kdembler As the whole design documentation has been transferred over to Notion this morning, I created this branch to delete all markdown documents with design docs from our GH repo. My proposal is to leave the "design" folder and the updated "README" file, which now just mentions that the documentation can be found in Notion.


kdembler (approved):
Thanks!

Issue #1765, opened 2021-12-07
POC: Introduce basic unit tests


kdembler:
Let's introduce some simple test as a proof of concept to discuss with the team

PR #1832, opened 2021-12-21
Vitest testing set up

eldiegod:
POC: Introduce basic unit tests

 Fixes: #1765

There are some improvements to be done on the test watching and
some annoying eslint warnings appear too but the rest looks good; 
should be a matter of tweaking the vite config


kdembler (changes requested):
Main app build is failing

rafalpawlow:
Good job, works really good! However I think we should add this to github actions. @kdembler WDYT?

rafalpawlow (approved):

package.json
```diff
@@ -35,7 +35,8 @@
     "svgs:icons-generate": "svgr --config-file svgr.config.js -d src/components/_icons src/components/_icons/svgs && prettier --write src/components/_icons",
     "svgs:icons": "yarn svgs:icons-import && yarn svgs:icons-generate",
     "graphql": "graphql-codegen --config codegen.config.yml",
-    "postinstall": "husky install"
+    "postinstall": "husky install",
+    "test": "vitest -w"
```
rafalpawlow: I wonder if we should add also non-watching script with `--run` flag. For the purposes of CI/CD pipeplines. But not sure if this is needed.

kdembler: Yes, we should have something like that as well. I propose `test:run` and `test:watch`

kdembler (commented):
Looks all good, really cool that you found a solution based on Vite. I think we should go with Rafał's suggestion and add it as part of Github action checks

eldiegod:
Apparently we don't need to specify watch or run for CI environment as it is handled automatically:

imagen

So `test` command should work for CI and as `--watch` locally, I did add dev mode command as `test:dev` thought I'm not 100% sure what the difference is from `--watch`

.github/workflows/checks.yml
```diff
@@ -20,3 +20,21 @@ jobs:
         run: yarn install --frozen-lockfile
       - name: Run ESLint
         run: yarn lint
+  test:
```
kdembler: I think we could just add a single step to the job we already have? We would save quite some time because we don't have to run `yarn install` 2 times

kdembler (approved):

Issue #1553, opened 2021-11-05
Refactor `VideoTile` component


toiletgranny:
The scope of this task is to clean up & address common issues known with the set of "Video tile" design components, but also video tile row components. We'll use this opportunity to also add the NFT functionalities to "Video tile" components.

toiletgranny:
![image](https://user-images.githubusercontent.com/35307309/146149541-3a2c0616-752f-4aa1-b060-580c70e84aee.png)


Designs for this component can be found here: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3630%3A116415

I highly recommend catching up on Tandem with the assigned developer(s) _before_ the implementation of NFT & Video tile components begins.

---

Related components:
- https://github.com/Joystream/atlas/issues/1790
- https://github.com/Joystream/atlas/issues/1793
- https://github.com/Joystream/atlas/issues/1553
- https://github.com/Joystream/atlas/issues/1792
- https://github.com/Joystream/atlas/issues/1791

PR #1838, opened 2021-12-27
Video thumbnail

### LINKED ISSUE #1790: Create a `VideoThumbnail` component

drillprop:
Fix #1790 
Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3629%3A119375
Just FYI didn't apply this component in VideoTile yet, I'll do it when I start working on https://github.com/Joystream/atlas/issues/1553

### DIFF
--- a/src/components/_video/VideoThumbnail/VideoThumbnail.styles.ts
+++ b/src/components/_video/VideoThumbnail/VideoThumbnail.styles.ts
@@ -1,3 +1,4 @@
+import isPropValid from '@emotion/is-prop-valid'
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 
@@ -8,10 +9,8 @@ const sharedOverlayStyles = css`
   position: absolute;
   top: 0;
   left: 0;
-  bottom: 0;
-  right: 0;
   width: 100%;
-  height: 0;
+  height: 100%;
   padding-top: 56.25%;
   transition: ${cVar('animationTransitionFast')};
 `
@@ -52,6 +51,7 @@ const getSlotPosition = (slotPosition: SlotPosition) => {
 export const SlotContainer = styled.div<{ position: SlotPosition }>`
   position: absolute;
   ${({ position }) => getSlotPosition(position)};
+  user-select: none;
 `
 
 export const ContentOverlay = styled.div`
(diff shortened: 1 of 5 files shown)


rafalpawlow (changes requested):
When I click on thumbnail, sometimes NFT pill is being highlighted. Maybe we could set `user-select: none` to Pill component?
https://user-images.githubusercontent.com/13706246/147603282-6894b8d6-54cb-4e5f-b3ca-fabdbe4ae7b9.mov

src/components/_video/VideoThumbnail/VideoThumbnail.styles.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const sharedOverlayStyles = css`
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
+  width: 100%;
+  height: 0;
+  padding-top: 56.25%;
```
rafalpawlow, line 14: Wouldn't `height 100%` be enough since we have position absolute here?

rafalpawlow (changes requested):

src/components/_video/VideoThumbnail/VideoThumbnail.styles.ts
drillprop: I refactored this a bit, but I don't think I could just get rid of this padding-top if that's what you mean

drillprop:
> When I click on thumbnail, sometimes NFT pill is being highlighted. Maybe we could set `user-select: none` to Pill component? https://user-images.githubusercontent.com/13706246/147603282-6894b8d6-54cb-4e5f-b3ca-fabdbe4ae7b9.mov

Should be fixed. I didn't set `user-select` on Pill, I set this to its container(SlotContainer) in the video thumbnail.

I added some improvements as well. Now if we don't want to make the slot clickable, we can pass this as an option.  
This is will prevent situations when you click for example the pill and lifting animation don't start. Check out the video below: 



https://user-images.githubusercontent.com/51168865/147649976-81666408-292e-4ee9-9c24-c2b15f85e3cb.mov

rafalpawlow:
@drillprop  Now the whole thumbnail is getting highlighted 😛

https://user-images.githubusercontent.com/13706246/147748525-bc8a1221-4e00-4b3c-99a4-2b351204b701.mov

rafalpawlow (approved):
After a discussion with @drillprop on rocket chat we decided that highlighting on click is not really an issue because user will not be able to click multiple times on the same tile.

PR #1843, opened 2022-01-03
Refactor video tile

### LINKED ISSUE #1553: Refactor `VideoTile` component

drillprop:
Fix #1553


drillprop:
Adam gave me some small feedback(I'm writing this down, to not lose this, will fix this later)
- ~~We should use Pills with size `medium` in current VideoTiles. In fact, medium size should be the default size for Pills(currently it's `small`, which is wrong).~~ 
- ~~When you hover on VideoThumbnail, the kebab menu should show up~~
- ~~the edit button on the publisher tile should be different, we can get rid of the tooltip.~~


Edit: all fixed

toiletgranny:
Overall, looks stunning, I love it! Here are the things I found:

1. I appreciate you letting me play with the number of views displayed in the VideoTile component on Storybook. This led me into discovering that we need to update how large numbers are formatted based on the newly accepted guidelines, which you can find [here, on Notion.](https://www.notion.so/Date-time-number-formats-in-Atlas-c5e7c7aec24e4ddfa83104ca4c2e924e#b0996679a7224eadbeacb5258389ee5b)
2. I don't know if it matters, but it seems like there's an incorrect size of the Pill component used in the "Default" VideoTile component on Storybook. Currently, it's "Small", it should be "Medium".
3. Again, not sure if this changes anything in the actual component, but there's a video length Pill component missing in the "Default" VideoTile component on Storybook.
4. **It seems like the list items used in the ContextMenu component are not the newly introduced ListItem component (@eldiegod). It should be. This one might deserve its own issue, as I believe this would require changing the ContextMenu component itself, not the VideoTile.**
5. Not sure if this is because our recent updates to the Text component and the text colors specified inside are not available in this branch, but just in case: we're using wrong text colors in the video details section. Eg. `--color-core-neutral-300` instead of `--color-text` for the number of views, the channel name, or when the video has been uploaded.
6. Can't seem to be able to get the Video tile component to switch into the "Small" size variant (upon going below 300px of width) and verify if it's ok. Is it because of the Storybook?

Additionally, some questions out of my curiosity:
1. Thanks to Storybook I have found out that the VideoTile component has this `15px` of the left margin. It's def not something you found in Figma, which makes me wonder: why is that?

toiletgranny:
@drillprop There's one more thing I found, which is about the behavior of the VideoTile component when the uploading process fails. Here's all about it: https://www.loom.com/share/f075ea9012b14eb8a118f0df864684d1

drillprop:
> Overall, looks stunning, I love it! Here are the things I found:
> 
> 1. I appreciate you letting me play with the number of views displayed in the VideoTile component on Storybook. This led me into discovering that we need to update how large numbers are formatted based on the newly accepted guidelines, which you can find [here, on Notion.](https://www.notion.so/Date-time-number-formats-in-Atlas-c5e7c7aec24e4ddfa83104ca4c2e924e#b0996679a7224eadbeacb5258389ee5b)
> 2. I don't know if it matters, but it seems like there's an incorrect size of the Pill component used in the "Default" VideoTile component on Storybook. Currently, it's "Small", it should be "Medium".
> 3. Again, not sure if this changes anything in the actual component, but there's a video length Pill component missing in the "Default" VideoTile component on Storybook.
> 4. **It seems like the list items used in the ContextMenu component are not the newly introduced ListItem component (@eldiegod). It should be. This one might deserve its own issue, as I believe this would require changing the ContextMenu component itself, not the VideoTile.**
> 5. Not sure if this is because our recent updates to the Text component and the text colors specified inside are not available in this branch, but just in case: we're using wrong text colors in the video details section. Eg. `--color-core-neutral-300` instead of `--color-text` for the number of views, the channel name, or when the video has been uploaded.
> 6. Can't seem to be able to get the Video tile component to switch into the "Small" size variant (upon going below 300px of width) and verify if it's ok. Is it because of the Storybook?
> 
> Additionally, some questions out of my curiosity:
> 
> 1. Thanks to Storybook I have found out that the VideoTile component has this `15px` of the left margin. It's def not something you found in Figma, which makes me wonder: why is that?

AD. 1. I think we should deal with this in another PR, wdyt @kdembler 
AD. 2. Fixed, used medium size as a default for PillGroup
AD. 3. Fixed
AD. 4. I'll create another issue just for that
AD. 5. Already fixed
AD. 6. will be fixed

Questions: 
Add. 1.
This margin is not applied to the video tile itself, so I don't think it should be a problem. This margin comes from our storybook template: 
```css
  & > * + * {
    margin-left: 15px;
  }
```
I'm not sure if I'm should remove this, It's there for some reason I think, so I'll just leave it as it is 😅 

> @drillprop There's one more thing I found, which is about the behavior of the VideoTile component when the uploading process fails. Here's all about it: https://www.loom.com/share/f075ea9012b14eb8a118f0df864684d1

Thank you for this video! ❤️   Totally missed that, will be fixed!

drillprop:
@toiletgranny all points should be fixed

src/components/_video/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -12,88 +12,103 @@ import {
...
-    clickable?: boolean
-  }
-}
+export type SlotsObject =
+  | {
+      [Property in SlotPosition]?: {
+        element: React.ReactNode
+        type?: 'default' | 'hover'
+        clickable?: boolean
+      } | null
```
rafalpawlow: Do we really need `null` here? Undefined wouldn't be enough?

rafalpawlow (approved):

kdembler (approved):
Looks amazing, great job! Let's get final approval from @toiletgranny before merging

toiletgranny (approved):
Looks great! Thank you!

PR #1858, opened 2022-01-04
move atlas app to monorepo package

rafalpawlow:


package.json
```diff
@@ -16,25 +14,11 @@
...
-    "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
-    "lint:ts": "eslint --ext .js,.jsx,.ts,.tsx --max-warnings 0 .",
-    "lint:prettier": "prettier --check .",
-    "lint:css": "stylelint --max-warnings 0 './src/**/*.{tsx,ts}'",
-    "lint": "yarn lint:css && yarn lint:ts && yarn lint:prettier",
+    "dev:atlas": "yarn workspace @joystream/atlas dev",
+    "lint:atlas": "yarn workspace @joystream/atlas lint",
+    "build:atlas": "yarn workspace @joystream/atlas build",
+    "storybook:atlas": "yarn workspace @joystream/atlas storybook",
```
kdembler: I think it may make sense to flip those, we use convention of `category:action`, so maybe `atlas:dev`, etc?

.github/workflows/checks.yml
```diff
@@ -19,4 +19,4 @@ jobs:
       - name: Install modules
         run: yarn install --frozen-lockfile
       - name: Run ESLint
```
kdembler:
```suggestion
      - name: Run Atlas linting
```

packages/atlas/package.json
```diff
@@ -0,0 +1,154 @@
+{
+  "name": "@joystream/atlas",
+  "description": "UI for consuming Joystream - a user governed video platform",
+  "license": "GPL-3.0",
+  "workspaces": {
+    "packages": [
+      "packages/*"
+    ]
+  },
```
kdembler: Do we need this here?

packages/atlas/package.json
```diff
@@ -0,0 +1,154 @@
...
+  "directories": {
+    "src": "src"
+  },
+  "files": [
+    "src"
+  ],
```
kdembler: Let's remove those, they don't seem needed

packages/atlas/.eslintignore
```diff
@@ -0,0 +1,7 @@
+node_modules
```
kdembler: Let's cleanup this file and `.prettierignore`. Now we have exact same copies in the root and in atlas package. Opened PR to your branch with my suggestions

packages/atlas/package.json
```diff
@@ -0,0 +1,154 @@
...
+    "eslint-plugin-react": "^7.27.1",
+    "eslint-plugin-react-hooks": "^4.3.0",
+    "postcss-syntax": "^0.36.2",
+    "rollup-plugin-visualizer": "^5.5.2",
+    "storybook-builder-vite": "^0.1.10",
+    "style-dictionary": "^3.0.3",
+    "vite": "^2.6.14",
+    "vite-plugin-checker": "^0.3.4"
+  },
+  "resolutions": {
```
kdembler: Turns out `resolutions` will work only if placed in the root

packages/atlas/.eslintignore
kdembler: https://github.com/rafalpawlow/atlas/pull/4

packages/atlas/package.json
kdembler: Included in https://github.com/rafalpawlow/atlas/pull/4

packages/atlas/package.json
rafalpawlow: Nah, just a mistake

kdembler (approved):

PR #1869, opened 2022-01-06
add nft card component

### LINKED ISSUE #1678: Create a "NFT card" component
### Description

Based on information gathered from @KubaMikolajczyk and @TCzechowski, as part of the NFT MVP scope of work:
- https://www.figma.com/file/uY98FOmQPgozpNOtAiklBg/Issuing-NFT?node-id=1063%3A67422
- https://www.figma.com/file/yVS7nkKjUtCeBcjUDqBbCF/Transaction?node-id=660%3A33264

### Figma link

_No response_

### How urgent this is?

This will block me soon

rafalpawlow:
Closes #1678

### DIFF
--- a/src/components/NftCard/Members.tsx
+++ b/src/components/NftCard/Members.tsx
@@ -1,5 +1,9 @@
 import React from 'react'
 
+import { Text } from '@/components/Text'
+import { useMediaMatch } from '@/hooks/useMediaMatch'
+import { cVar } from '@/styles'
+
 import {
   AvatarGroupWrapper,
   AvatarWrapper,
@@ -7,10 +11,7 @@ import {
   MembersWrapper,
   StyledAvatar,
   StyledAvatarGroup,
-} from '@/components/NftCard/NftCard.styles'
-import { Text } from '@/components/Text'
-import { useMediaMatch } from '@/hooks/useMediaMatch'
-import { cVar } from '@/styles'
+} from './NftCard.styles'
 
 export type Member = {
   assetUrl?: string | null
--- a/src/components/NftCard/Members.tsx
+++ b/src/components/NftCard/Members.tsx
@@ -1,0 +1,59 @@
+import React from 'react'
+
+import {
+  AvatarGroupWrapper,
+  AvatarWrapper,
+  Caption,
+  MembersWrapper,
+  StyledAvatar,
+  StyledAvatarGroup,
+} from '@/components/NftCard/NftCard.styles'
+import { Text } from '@/components/Text'
+import { useMediaMatch } from '@/hooks/useMediaMatch'
+import { cVar } from '@/styles'
+
+export type Member = {
+  assetUrl?: string | null
+  name: string
+}
+
+type MembersProps =
+  | {
+      caption: 'Creator' | 'Owner'
+      members: Member
+    }
+  | { caption: 'Supporters'; members: Member[] }
+
+export const Members: React.FC<MembersProps> = ({ caption, members }) => {
+  const avatars =
+    members && Array.isArray(members)
+      ? members.map((member) => ({ assetUrl: member.assetUrl, tooltipText: member.name }))
+      : null
+  const smMatch = useMediaMatch('sm')
+  return (
+    <MembersWrapper>
+      <Caption variant="t200" secondary>
+        {caption}
+      </Caption>
+      {Array.isArray(members) && avatars && (
+        <AvatarGroupWrapper>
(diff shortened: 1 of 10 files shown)


src/components/NftCard/NftCard.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+      { assetUrl: 'https://placedog.net/100/100?random=5', tooltipText: 'Jeff' },
+      { assetUrl: 'https://placedog.net/100/100?random=6', tooltipText: 'Alan' },
+      { assetUrl: 'https://placedog.net/100/100?random=7', tooltipText: 'Marietta' },
+    ],
+    owner: { assetUrl: 'https://placedog.net/100/100?random=8', tooltipText: 'John' },
+  },
+} as Meta
+
+const Template: Story<NftCardProps> = (args) => (
+  <div style={{ width: '360px' }}>
```
drillprop: shouldn't we apply this width to NFTCard itself?

src/components/NftCard/Members.tsx
```diff
@@ -0,0 +1,59 @@
...
+type MembersProps =
+  | {
+      caption: 'Creator' | 'Owner'
+      members: Member
+    }
+  | { caption: 'Supporters'; members: Member[] }
+
+export const Members: React.FC<MembersProps> = ({ caption, members }) => {
+  const avatars =
+    members && Array.isArray(members)
```
drillprop, line 30: we used `Array.isArray(members)` two times. Let's store this boolean in a variable maybe?

src/components/NftCard/NftCard.styles.ts
```diff
@@ -0,0 +1,95 @@
...
+      text-align: right;
+      justify-self: flex-end;
+    }
+  }
+`
+
+export const Separator = styled.hr`
+  grid-column: 1 / span 2;
+  background-color: ${cVar('colorBorderMutedAlpha')};
+  border: 0;
```
drillprop, line 65: Why? Does the hr tag have a default border?

src/components/NftCard/NftCard.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <NftCard {...args} />
+  </div>
+)
+
+export const Default = Template.bind({})
```
drillprop, line 41: Could you also add a template with no supporters as we have in Figma?
image

drillprop (changes requested):

src/components/NftCard/NftCard.stories.tsx
rafalpawlow: @toiletgranny does Nft Card should have fixed width?

kdembler: Screenshot 2022-01-10 at 15 04 16

Seems so

toiletgranny: @rafalpawlow @kdembler Exactly! There's nothing inside of the NFT card component that could determine its width. And so the width should be fixed, with an option of being adjusted depending on the use case (eg. on one page we might want to set the width of this component to be of 3 columns, whereas on another page we might want it to span across 2 columns only.)

src/components/NftCard/NftCard.styles.ts
rafalpawlow: Yes, `hr` has default border

drillprop (approved):
~Not sure if we actually need the `fullWidth` variant, I'd use just `max-width` with `360px`(instead of `width`). This needs to be consulted with Adam(maybe I didn't catch his intentions correctly).~
edit: ignore the above ☝🏻 , we resolved this on rocket chat.
Besides that, looks good to me 👍🏻

src/components/NftCard/Members.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+
+import {
+  AvatarGroupWrapper,
+  AvatarWrapper,
+  Caption,
+  MembersWrapper,
+  StyledAvatar,
+  StyledAvatarGroup,
+} from '@/components/NftCard/NftCard.styles'
```
kdembler: Should use relative import

kdembler (commented):
The component in the Storybook is missing labels for Creator/Owner?

rafalpawlow:
@kdembler my bad, please check again

kdembler (approved):
Looks good to me but let's get final approval from @toiletgranny

toiletgranny (approved):

Issue #1718, opened 2021-12-03
Atlas Giza testing and tweaking


kdembler:
Scope:

- [ ] Remove any TODO items from initial Giza PRs
- [ ] Fix re-uploading
- [ ] Fix failed video tile when updating the video thumbnail
- [ ] Estimate fee of uploading
- [ ] Use WASM for blake3 hashing
- [ ] Update dimensions of uploaded images
- [ ] Handle refetching bucket assignements when creating a new channel

PR #1882, opened 2022-01-07
Giza: update content extrinsics, enable uploads

kdembler:
Working uploads on Giza network. Will likely not work on Vercel deployment yet, check locally.

Summary of changes:
- Refactored Joystream lib to better split the code in multiple files, (hopefully) greatly improved readability
- Improved type safety of some parts of Joystream lib, to make it easier to write new extrinsics
- Made the changes to uploading introduced in Giza


src/providers/assets/operatorsProvider.tsx
```diff
@@ -0,0 +1,209 @@
...
+        const operatorsInfos: OperatorInfo[] = bucket.operators
+          .filter((operator) => operator.metadata?.nodeEndpoint?.includes('http') && operator.status === 'ACTIVE')
+          .map((operator) => ({ id: operator.id, endpoint: operator.metadata?.nodeEndpoint || '' }))
+
+        bagIds.forEach((bagId) => {
+          if (!mapping[bagId]) {
+            mapping[bagId] = operatorsInfos
+          } else {
+            // TODO: make sure endpoints are unique
+            mapping[bagId] = [...mapping[bagId], ...operatorsInfos]
```
eldiegod, line 63: Is this still needed before merge?

src/providers/assets/operatorsProvider.tsx
```diff
@@ -0,0 +1,209 @@
...
+          id: bucket.id,
+          endpoint,
+        }
+
+        bagIds.forEach((bagId) => {
+          if (!mapping[bagId]) {
+            mapping[bagId] = [operatorInfo]
+          } else {
+            // TODO: make sure endpoints are unique
+            mapping[bagId] = [...mapping[bagId], operatorInfo]
```
eldiegod, line 101: same as above

eldiegod (commented):
Checked the code first but I'm not sure how to test on the app yet, could you share some brief info on that?

src/providers/assets/operatorsProvider.tsx
kdembler: No, we can leave that for cleanup as part of #1718

src/providers/assets/operatorsProvider.tsx
kdembler: same as above

kdembler:
To test out, you should just run it in dev mode as you would always do, create a membership, create a channel and do some uploads. One blocker is that currently the faucet will not send you any tokens and those are need now to do an upload. You can send me your account address and I will send you some

drillprop (changes requested):
The code generally looks good but I spot two issues
1. I cannot re-upload a video when the upload is failed. Tried a few times on different files.
2. When I changed the thumbnail to already uploaded video I'm getting `asset thumbnail failed` on the video tile until uploading/processing is done
image

kdembler:
@drillprop Good finds, I've included those in scope of a followup task here: https://github.com/Joystream/atlas/issues/1718

rafalpawlow (changes requested):
Something is wrong with upload error handling. I got this error in the console but at the same time success modal appeared.
Zrzut ekranu 2022-01-11 o 09 46 14

rafalpawlow:
I'm not sure if this is related to this PR, but right after I delete the video, there is a request to graphql which is cancelled 
Zrzut ekranu 2022-01-11 o 10 02 16
Zrzut ekranu 2022-01-11 o 10 02 33

rafalpawlow (approved):

drillprop (approved):

eldiegod:
@kdembler this is my test acc address `5CJzt8u1KYRZFTn7xd7MLuvQ1q5eybSwtyxYdHvNH75358c7`

Issue #1676, opened 2021-11-30
Update "Option card" components


toiletgranny:
### Description

Based on information gathered from @KubaMikolajczyk, as part of the NFT MVP scope of work.

### Figma link

https://www.figma.com/file/uY98FOmQPgozpNOtAiklBg/Issuing-NFT?node-id=796%3A48187

### How urgent this is?

This will block me soon

toiletgranny:
![image](https://user-images.githubusercontent.com/35307309/146954693-7572cc7a-b52f-4962-b2b2-e1a6f7906b73.png)

The updated "Option card" component, along with a full documentation can be found here: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4118%3A144975

While the full list of changes can be found in the "Changelog" part of the documentation in Figma, the key changes are:
- Adding an icon variant to be used in the issuing NFT flows,
- Bringing Figma component and its implementation counterpart closer together, eg. by adding "disabled" or "error" states which were missing in Figma and present in the implementation, but also taking this opportunity to adjust the styling of the above states.

---

![image](https://user-images.githubusercontent.com/35307309/146955329-f645988a-8194-4255-b72a-d563ca10a6db.png)

Supporting the standalone "Option card", there's now a new "Option card group" component combining two other components that were previously used in Figma: "Option row horizontal" & "Option row vertical".

This component and its documentation can be found here: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4118%3A155697

---

If this is helpful, here's the original component design in Figma, which served as a base for the current implementation: https://www.figma.com/file/yaw2k1Fw5sezLoYLLRU10v/Web-Components-(PX-Grotesk)?node-id=750%3A90472

kdembler:
Dev AC:

- [x] Add any missing variants
- [x] Make sure implementation follows all the colors/states from Figma

PR #1896, opened 2022-01-11
Update OptionCard component

eldiegod:
fixes: #1676

figma: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4118%3A144975

### DIFF
--- a/src/components/_inputs/OptionCard/OptionCard.stories.tsx
+++ b/src/components/_inputs/OptionCard/OptionCard.stories.tsx
@@ -1,6 +1,8 @@
 import { Meta, Story } from '@storybook/react'
 import React, { useState } from 'react'
 
+import { SvgActionTrash } from '@/components/_icons'
+
 import { OptionCard, OptionCardProps } from './OptionCard'
 
 export default {
@@ -21,9 +23,10 @@ const Template: Story<OptionCardProps> = (args) => {
     setSelected(element.value)
   }
   return (
-    <div style={{ display: 'grid', gridTemplateColumns: 'auto auto', gap: '16px' }}>
-      <OptionCard {...args} value="1" selectedValue={selected} onClick={handleClick} />
+    <div style={{ display: 'grid', gridTemplateColumns: 'auto auto auto', gap: '16px' }}>
+      <OptionCard {...args} icon={<SvgActionTrash />} value="1" selectedValue={selected} onClick={handleClick} />
       <OptionCard {...args} value="2" selectedValue={selected} onClick={handleClick} />
+      <OptionCard {...args} value="3" selectedValue={selected} onClick={handleClick} />
     </div>
   )
 }
--- a/src/components/_inputs/OptionCard/OptionCard.stories.tsx
+++ b/src/components/_inputs/OptionCard/OptionCard.stories.tsx
@@ -3,11 +3,13 @@ import React, { useState } from 'react'
 
 import { SvgActionTrash } from '@/components/_icons'
 
-import { OptionCard, OptionCardProps } from './OptionCard'
+import { OptionCardBase, OptionCardProps, OptionCardRadio } from './OptionCard'
+
+import { OptionCardCheckbox } from '.'
 
 export default {
   title: 'inputs/OptionCard',
-  component: OptionCard,
+  component: OptionCardBase,
   args: {
     label: 'Radio input label',
     helperText: 'Radio helper text',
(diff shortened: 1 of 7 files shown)


rafalpawlow (changes requested):
In figma, radio component doesn't inherit error state
Zrzut ekranu 2022-01-13 o 18 49 39

Another thing is, in figma we have two types of option card `Radio` and `Checkbox`. Second one is missing in your PR.

src/components/_inputs/OptionCard/OptionCard.stories.tsx
```diff
@@ -21,9 +23,10 @@ const Template: Story<OptionCardProps> = (args) => {
     setSelected(element.value)
   }
   return (
-    <div style={{ display: 'grid', gridTemplateColumns: 'auto auto', gap: '16px' }}>
-      <OptionCard {...args} value="1" selectedValue={selected} onClick={handleClick} />
+    <div style={{ display: 'grid', gridTemplateColumns: 'auto auto auto', gap: '16px' }}>
```
rafalpawlow: In figma we have 2 gaps depending on the breakpoint `16px` and `24px`. Also I think we should move `Option card group` as a separate component.

rafalpawlow:
Figma says: `In the Option card component, the display limit for both the main label is set to 1 line of text. Upon exceeding this limit, the text should be trimmed and ended with an ellipsis.`.

But current implementation allows 2 lines of text
Zrzut ekranu 2022-01-13 o 19 00 47

rafalpawlow (changes requested):

rafalpawlow:
I'm not sure about spacing between text and hovered input. @toiletgranny is that fine?
Zrzut ekranu 2022-01-13 o 19 09 48

rafalpawlow:
Border colors are wrong
we should use the following:

**white**
default: `colorBorderAlpha`
hover: `colorBorderStrongAlpha`
pressed: `colorBorderMutedAlpha`

**error**
hover: `colorCoreRed300`
default: `colorCoreRed500`
pressed: `colorCoreRed700`

currently we use `colorCoreRed400` for all variants

**blue**
hover: `colorCoreBlue300`
pressed: `colorCoreBlue700`

toiletgranny:
I agree with everything @rafalpawlow pointed out here, and to answer/comment on some of the points he brought up:
1. Yes, the spacing between the label and the Radio button/Checkbox should be `16px`, awesome catch! This can be inspected in Figma, though it might be tricky. 
CleanShot 2022-01-14 at 11 02 41

2. It's great to see text & icon colors using our new design tokens, but @rafalpawlow is right that we forgot about borders. Especially in "the error state". In case that's helpful, @eldiegod I put up this quick prototype for you as a quick reference on how border colors are changing upon hovering and pressing the Option card component.


https://www.figma.com/proto/Pf31tuYpozYmpq163U2ho8/Web-Components?page-id=4645%3A137109&node-id=4645%3A137110&viewport=396%2C48%2C1&scaling=min-zoom&starting-point-node-id=4645%3A137110

eldiegod:
@rafalpawlow @toiletgranny I think I have addressed all the comments except for the optionCardGroup components as I'm not sure it should be done here or it's a matter for another ticket.

Saying it because the initial complexity scope for this ticket was a 2 but it has gotten a bit intricate so not sure if we just missed that or we are doing more than planned

rafalpawlow:
@eldiegod component still allows for 2 lines of title

rafalpawlow:
Something is wrong with input when there is a lot of text ant card is in `active` state
Zrzut ekranu 2022-01-17 o 15 53 11

eldiegod:
> Something is wrong with input when there is a lot of text ant card is in `active` state

Can you tell me exactly what you mean with wrong? @rafalpawlow

rafalpawlow:
@eldiegod Look at the screenshot, radio button is not round, it is egg-shaped

eldiegod:
> @eldiegod Look at the screenshot, radio button is not round, it is egg-shaped

wanted to make sure, I think I fixed it now @rafalpawlow

rafalpawlow (approved):
Looks good! However I found one more thing which we probably should tackle in another issue. Pressed radio component looks a little bit different in figma:

Zrzut ekranu 2022-01-18 o 13 04 19
Zrzut ekranu 2022-01-18 o 13 04 57

eldiegod:
> Looks good! However I found one more thing which we probably should tackle in another issue. Pressed radio component looks a little bit different in figma:
> 
> Zrzut ekranu 2022-01-18 o 13 04 19 Zrzut ekranu 2022-01-18 o 13 04 57

@rafalpawlow Awesome! Yep that's right, that would be more of an existing issue with the `RadioInput` component itself

kdembler:
@eldiegod Could you please create an issue for this difference? We may have missed that

PR #1897, opened 2022-01-11
Refactor channel card

### LINKED ISSUE #1767: Add a new variant of the "Channel card" component + update the component to use color design tokens inside
### Description

I need a version of channel card without a button
Maybe a smaller size of the card could be also created

### Figma link

https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A25135

### How urgent this is?

This will block me soon

drillprop:
Fix #1767 
Designs: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4094%3A144631


src/components/_channel/ChannelCardBase/ChannelCardBase.tsx
```diff
@@ -37,36 +38,51 @@ export const ChannelCardBase: React.FC<ChannelCardBaseProps> = ({
...
                   </ChannelFollows>
-                  <FollowButton variant="secondary" size="small" onClick={onFollow}>
-                    {isFollowing ? 'Unfollow' : 'Follow'}
-                  </FollowButton>
+                  {onFollow && (
+                    <FollowButton
+                      variant="secondary"
+                      size="small"
+                      onClick={handleFollowButtonClick}
+                      onMouseOut={() => setActiveDisabled(false)}
```
rafalpawlow: Does `stopPropagation` wouldn't do the job?

src/components/_channel/ChannelCardBase/ChannelCardBase.stories.tsx
```diff
@@ -22,12 +26,23 @@ export default {
...
-      <ChannelCardBase {...args} />
-    </>
-  )
-}
+const Template: Story<ChannelCardBaseProps> = (args) => <ChannelCardBase {...args} />
 
 export const Default = Template.bind({})
+export const WithButton = Template.bind({})
+WithButton.args = {
+  onFollow: () => null,
```
rafalpawlow: Let's hide it from table
Zrzut ekranu 2022-01-12 o 11 21 38

rafalpawlow (changes requested):

src/components/_channel/ChannelCardBase/ChannelCardBase.tsx
drillprop: Unfortunately, it wouldn't - clicking the follow button triggers pulldown animation and my intention was to not do that. There is a CSS-only solution for this, but it's not supported on Safari(requires using `focus-within`). We discussed a similar case on RocketChat when I was working on VideoTile.

drillprop:
Btw. I was wondering - right now we have `ChannelCardBase`, which is basically a dumb component, and `ChannelCard`, which has some logic for following and displaying avatar(this is like 3 or 4 lines of code). Maybe I could rename `ChannelCardBase` to just `ChannelCard`, move all the logic there, and then remove the old `ChannelCard`?
Or maybe we could keep the logic outside of the component and pass onFollow and avatar URL from its parent? Wdyt?

drillprop:
> Btw. I was wondering - right now we have `ChannelCardBase`, which is basically a dumb component, and `ChannelCard`, which has some logic for following and displaying avatar(this is like 3 or 4 lines of code). Maybe I could rename `ChannelCardBase` to just `ChannelCard`, move all the logic there, and then remove the old `ChannelCard`? Or maybe we could keep the logic outside of the component and pass onFollow and avatar URL from its parent? Wdyt?

After quick discussion with Klaudiusz on RocketChat removed `ChannelCardBase`. All logic that is responsible for displaying avatar, channel information, and for following is inside `ChannelCard`

src/components/_channel/ChannelCard/ChannelCard.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { StorageProvidersProvider } from '@/providers/storageProviders'
+
+import { ChannelCard, ChannelCardProps } from './ChannelCard'
+
+export default {
+  title: 'channel/ChannelCard',
+  component: ChannelCard,
+  args: {
+    id: '3',
+    withFollowButton: true,
```
rafalpawlow, line 20: Let's add `loading` here

rafalpawlow (changes requested):

rafalpawlow (approved):

PR #1931, opened 2022-01-14
add edit membership view

### LINKED ISSUE #1755: Editing member details
AC:
- [ ] Member details are editable 

Design:
[Link to figma](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=890%3A130337)

drillprop:
Fix #1755

### DIFF
--- a/packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
+++ b/packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
@@ -23,13 +23,23 @@ import { SkeletonLoader } from '../_loaders/SkeletonLoader'
 
 export type MembershipInfoProps = {
   avatarUrl?: string | null
+  hasAvatarUploadFailed?: boolean
+  onAvatarEditClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
   handle?: string | null
   address?: string | null
   loading?: boolean
   isOwner?: boolean
 }
 
-export const MembershipInfo: React.FC<MembershipInfoProps> = ({ address, avatarUrl, handle, loading, isOwner }) => {
+export const MembershipInfo: React.FC<MembershipInfoProps> = ({
+  address,
+  avatarUrl,
+  hasAvatarUploadFailed,
+  onAvatarEditClick,
+  handle,
+  loading,
+  isOwner,
+}) => {
   const smMatch = useMediaMatch('sm')
   return (
     <SwitchTransition>
@@ -40,7 +50,14 @@ export const MembershipInfo: React.FC<MembershipInfoProps> = ({ address, avatarU
       >
         <MembershipHeader>
           <MembershipInfoContainer>
-            <Avatar size={smMatch ? 'preview' : 'channel-card'} assetUrl={avatarUrl} loading={loading} />
+            <Avatar
+              size={smMatch ? 'preview' : 'channel-card'}
+              editable
+              onEditClick={onAvatarEditClick}
+              assetUrl={avatarUrl}
+              loading={loading}
+              hasAvatarUploadFailed={hasAvatarUploadFailed}
+            />
             <MembershipDetails>
               {loading || !handle ? (
                 <SkeletonLoader width={200} height={smMatch ? 56 : 40} bottomSpace={8} />
--- a/packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
+++ b/packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
@@ -22,9 +22,9 @@ import { SvgActionEdit } from '../_icons'
 import { SkeletonLoader } from '../_loaders/SkeletonLoader'
 
 export type MembershipInfoProps = {
-  avatarUrl?: string
-  handle?: string
-  address?: string
+  avatarUrl?: string | null
+  handle?: string | null
+  address?: string | null
   loading?: boolean
   isOwner?: boolean
 }
--- a/packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
+++ b/packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { CSSTransition, SwitchTransition } from 'react-transition-group'
 
 import { useMediaMatch } from '@/hooks/useMediaMatch'
@@ -11,6 +11,7 @@ import {
   MembershipHeader,
   MembershipInfoContainer,
   StyledHandle,
+  StyledSvgActionCheck,
   StyledSvgActionCopy,
   StyledText,
 } from './MembershipInfo.style'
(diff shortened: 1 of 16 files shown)


src/views/viewer/ViewerLayout.tsx
```diff
@@ -36,11 +39,22 @@ const viewerRoutes = [
...
   { path: relativeRoutes.viewer.channel(), element: <ChannelView /> },
   { path: relativeRoutes.viewer.category(), element: <CategoryView /> },
+  {
+    path: relativeRoutes.viewer.editMembership(),
+    element: (
+      <PrivateRoute
+        isAuth={isActiveMember}
+        element={<EditMembershipView />}
+        redirectTo={absoluteRoutes.viewer.index()}
+      />
```
eldiegod: in firefox I get this error if I go to the URL without login in `Unhandled exception was thrown: The operation is insecure. 
Object { cause: React ErrorBoundary Error }`

on Chrome it works fine.

Not sure if worth fixing or what is causing it

src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -0,0 +1,192 @@
...
+      image.onload = () => {
+        resolve(true)
+      }
+      image.onerror = () => resolve(false)
+      image.src = value
+      if (!value) {
+        resolve(true)
+      }
+    })
+
```
eldiegod: This can live outside the component and in fact, utils would probably be a good place for it

src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const handleUniqueValidation = async (value: string, prevValue?: string) => {
```
eldiegod: Maybe call it `handleUniqueMembershipHandleValidation` for clarity

src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -0,0 +1,192 @@
...
+    if (!activeMembershipLoading && activeMembership) {
+      resetForm()
+    }
+  }, [activeMembership, activeMembershipLoading, resetForm])
+
+  const handleEditMember = handleSubmit(async () => {
+    // TODO
+    // handle submiting edited member
+  })
+
```
eldiegod: We dont have the tech for this yet?

eldiegod (changes requested):
Looks very nice already! added some comments for improvements 

In my eyes, it might be confusing at first sight if I want to change the avatar how to do it, I might start clicking on the avatar right away trying to change it but that does nothing; I think it would be could that if we click on the avatar it would focus the `Avatar URL` input if possible.

src/views/viewer/EditMembershipView/EditMembershipView.tsx
drillprop: I looked at: https://github.com/Joystream/membership-faucet/tree/master/src and I don't think we have. 
@kdembler should I do something about it? If yes, I'll probably need some help

drillprop:
> Looks very nice already! added some comments for improvements
> 
> In my eyes, it might be confusing at first sight if I want to change the avatar how to do it, I might start clicking on the avatar right away trying to change it but that does nothing; I think it would be could that if we click on the avatar it would focus the `Avatar URL` input if possible.

Hmm, that's a really good idea but I'll ask the designer what should I do here just to be sure!

src/views/viewer/EditMembershipView/EditMembershipView.tsx
kdembler: This may be like a 3rd place in the app where we use something similar, definitely agree

src/views/viewer/EditMembershipView/EditMembershipView.tsx
kdembler: This will not be faucet's job. Faucet just creates memberships and supplies them with some tokens. This will be a regular extrinsic to update the membership. We can sync in chat about this

src/views/viewer/ViewerLayout.tsx
drillprop: hmm, I didn't do anything to fix this but I think it works now for some reason 🤔  Let me know if you manage to reproduce this!

src/views/viewer/EditMembershipView/EditMembershipView.tsx
drillprop: ok, fixed this. It's possible now

drillprop:
Added avatar input focus and updating membership info on the fly on membership header. 
Just FYI both these changes were discussed with Nina.

src/views/viewer/ViewerLayout.tsx
eldiegod: Is still the same for me but maybe is just my machine idk what it is

eldiegod (approved):
Nice, awesome work!

src/views/viewer/ViewerLayout.tsx
drillprop: Ok, I managed to reproduce this, but only once. Then I tried to do this with or without extension and with cleared local storage.  Everything seems to be fine. 
I assume that it's caused by changes inside ViewerLayout - I used PrivateRoute there and Navigate. 
Maybe I should revert that and we should do all the private navigation in another PR?

kdembler:
I'm seeing the same thing @eldiegod mentioned:

Screenshot 2022-01-19 at 11 52 37

The first message suggest we are calling something too many times?

src/joystream-lib/extrinsics.ts
```diff
@@ -38,6 +38,8 @@ import {
   VideoInputMetadata,
 } from './types'
 
+import { MemberExtrisincResult } from '.'
```
kdembler: This should be imported from `./types`

src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -0,0 +1,211 @@
...
+      if (membershipByUniqueInput) {
+        return false
+      } else {
+        return true
+      }
+    }, 500)
+  )
+  const avatarInputRef = useRef<HTMLInputElement | null>(null)
+
+  const schema = z.object({
```
kdembler: This and the form itself (including its state) already exists in the app, right? In the sign up flow. Could we use a single component in both places?

kdembler (commented):
Minor visual points:
1. The spacing between the avatar and header is wrong in the app:
Screenshot 2022-01-24 at 17 01 57
 2. We're missing the interaction with the copy address button: https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=1020%3A69829

packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
```diff
@@ -10,27 +10,48 @@ import {
...
+}) => {
+  const [copyButtonClicked, setCopyButtonClicked] = useState(false)
   const smMatch = useMediaMatch('sm')
+
+  const handleCopyAddress = () => {
+    if (!address) {
+      return
+    }
+    copyToClipboard(address)
+    setCopyButtonClicked(true)
```
kdembler, line 52:
```suggestion
    setCopyButtonClicked(true)
    setTimeout(() => {
      setCopyButtonClicked(false)
    }, 3000)
```

PR #1952, opened 2022-01-18
update docs to reflect the latest Orion and Giza changes

### LINKED ISSUE #1828: Deploy updated Orion to production instance & update docs

kdembler:
Resolves #1828


traumschule:
Have `Bandwidth` nodes/providers been renamed to `Distributors`? They seem to be equivalent but could cause confusion.

kdembler:
These names are used interchangeably but I think distributor is more precise, that's what they are called in the repo https://github.com/Joystream/joystream/tree/giza/distributor-node

kdembler:
Just checked and [the Handbook](https://joystream.gitbook.io/joystream-handbook/subsystems/storage) seems to use `Bandwidth` exclusively. It's a bit messy since like you can see the code in the monorepo seems to use `Distributor` exclusively. I think we will need to accept that both names are used

eldiegod (approved):

Issue #1628, opened 2021-11-23
Create information component


KubaMikolajczyk:
### Description

A unified information component is needed - a hoverable element triggering tooltip.

### How urgent this is?

Not a blocker, but still worth considering

toiletgranny:
@kdembler I think the `design-system` label should be added automatically to all issues created with the issue template I provided you with. Is there anything that changed when you uploaded that issue template?

kdembler:
@toiletgranny I didn't change anything in the template, the issue is that the template assumed a label called `design-system` but in our repo it was called `design system` (with a space). I changed the label name yesterday so it should be properly applied going forward

toiletgranny:
![image](https://user-images.githubusercontent.com/35307309/150374703-8b9a5bb8-f7a8-424e-8256-f54643af56a5.png)

Designs and the documentation for this component [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4819%3A169964).

For the implementation of this component, it's critical to support keyboard & touch-screen devices usage. All the details can be found in the documentation.

PR #1956, opened 2022-01-20
Switchable form field

### LINKED ISSUE #1926: Switchable Field Component
**Context**

Supporting Components for issue NFT flow

**AC**
- [ ] add component for a switchable field
- [ ] add a toggle component
- [ ] make the component accept a tooltip

**Designs**

[Reference in the flow> ](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=996%3A28264)
[Local Components Page on Figma](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=919%3A21501)

drillprop:
Fix #1926 and #1628
Designs: https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=919%3A21501
Switch component: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=1859%3A94121

Summary of changes: 
- added Switch component
- ~added InformationButton component. I think it makes sense to create it, I saw this button in a few places, for example https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=916%3A21147~ 
- refactor FormField component in general, make it switchable.


kdembler:
@drillprop I think you also accidentally worked on this task: https://github.com/Joystream/atlas/issues/1628 Since you already started I suggest to tackle this here as well, but please take all the information in Figma into consideration to make sure we have feature parity

drillprop:
> @drillprop I think you also accidentally worked on this task: #1628 Since you already started I suggest to tackle this here as well, but please take all the information in Figma into consideration to make sure we have feature parity

Okay, no problem. I also notified Kuba about this, because he is still using the Button component there: 
image

src/components/Tooltip/Tooltip.tsx
```diff
@@ -37,6 +37,8 @@ export const Tooltip: React.FC<TooltipProps> = ({
   return (
     <Tippy
       onMount={() => setIsVisible(true)}
+      hideOnClick={false}
```
rafalpawlow: I'm not sure it's a good idea to hardcode this. Can we pass it from prop?

src/components/_buttons/Information/Information.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+
+import { Information, InformationProps } from '.'
+
+export default {
+  title: 'button/Information',
```
rafalpawlow: This shouldn't be in buttons group

rafalpawlow (changes requested):

toiletgranny:
### Information component
1. Looks stunning: colors, spacing: 👌
2. Now that I got a chance to review this live, in implementation, I'm no longer a fan of the behavior I proposed to be used on touch-screen devices, which is displaying a tooltip upon tapping and holding the finger over the touch target. Can I propose to switch this back to a default behavior, which is displaying a tooltip upon tapping on the Information component once, and then dismissing it upon tapping anywhere else on the screen? I will update the Figma documentation accordingly today.
3. When activating a tooltip using my keyboard, by pressing down the `tab` key, the "Information" icon is not lit up, like in the `hover` state. Ideally, the `active` state looks exactly the same as the `hover` state.

### Switch component
Again, to me, it's ready to go as-is, but there are some little adjustments that you can consider:
1. That's a clever move with the `isLabelTitle` property, but I'm not sure if this is something I would keep on the "Switch" component level. I think the custom label styling (bold, bright) is sort of a one-time custom situation that takes place only in that "Switchable Form Field" component. But that's just my perspective on this, ultimately it's up to you on how you decide to proceed with this.
2. The `border-radius` CSS property on the dot is currently set to `50%`, can I ask to switch this to something like `999px`?
3. I think the styling of the disabled state does not match the designs when you look at the color of the text label.

### Switchable form field
@KubaMikolajczyk would be best suitable to review this component.

toiletgranny (changes requested):
### Information component
1. Looks stunning: colors, spacing: 👌
2. Now that I got a chance to review this live, in implementation, I'm no longer a fan of the behavior I proposed to be used on touch-screen devices, which is displaying a tooltip upon tapping and holding the finger over the touch target. Can I propose to switch this back to a default behavior, which is displaying a tooltip upon tapping on the Information component once, and then dismissing it upon tapping anywhere else on the screen? I will update the Figma documentation accordingly today.
3. When activating a tooltip using my keyboard, by pressing down the `tab` key, the "Information" icon is not lit up, like in the `hover` state. Ideally, the `active` state looks exactly the same as the `hover` state.

### Switch component
Again, to me, it's ready to go as-is, but there are some little adjustments that you can consider:
1. That's a clever move with the `isLabelTitle` property, but I'm not sure if this is something I would keep on the "Switch" component level. I think the custom label styling (bold, bright) is sort of a one-time custom situation that takes place only in that "Switchable Form Field" component. But that's just my perspective on this, ultimately it's up to you on how you decide to proceed with this.
2. The `border-radius` CSS property on the dot is currently set to `50%`, can I ask to switch this to something like `999px`?
3. I think the styling of the disabled state does not match the designs when you look at the color of the text label.

### Switchable form field
@KubaMikolajczyk would be best suitable to review this component.

KubaMikolajczyk:
Checked the switchable form field and it looks great ✅

src/components/_inputs/FormField/FormField.tsx
```diff
@@ -29,7 +31,14 @@ export const FormField: React.FC<FormFieldProps> = React.memo(
     return (
       <FormFieldWrapper className={className} dense={dense}>
         <FormFieldHeader>
-          {switchProps ? <Switch {...switchProps} isLabelTitle label={title} /> : <Text variant="h300">{title}</Text>}
+          {switchProps ? (
+            <SwitchLabel>
+              <Switch {...switchProps} /> <SwitchTitle variant="h300">{title}</SwitchTitle>
+            </SwitchLabel>
+          ) : (
+            // <Switch {...switchProps} isLabelTitle label={title} />
```
rafalpawlow: I guess we don't need that

rafalpawlow (changes requested):

rafalpawlow (approved):

toiletgranny (changes requested):
Both the Switch and Information components look very decent now, thank you!

I'm going for "Request changes", but please consider this to be a nice-to-have and disregard if there are more important items on your to-do list. 

Ok, here it is: on touch-screen devices when I tap on the "i" icon, the tooltip does indeed show up and remains open until I tap anywhere outside, which is great, just what we wanted! But the "i" icon itself is not lit up for the time of the tooltip being visible. 

The reason this might be problematic is eg. when we have a couple of information components stacked really closely next to each other. In such a case, not having the "i" icon lit up, there's no way of telling which information component does a given tooltip belongs to, right?

And so I was wondering if there's a way for us to have that "i" icon remain to lighten up for the time of tooltip being visible, on touch-screen devices?

toiletgranny (approved):
giphy

Issue #1856, opened 2022-01-04
NFT (Channel view) Layout Update and Responsive


dmtrjsg:
**Context**

General layout and Responsive updates ticket

**AC**
- [x] Match designs for all break points
- [x] Make sure the typography styles are correct

**Designs**
[ Link to Figma >](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=510%3A24710)

PR #1948, opened 2022-01-18
Collectors NFT panel

drillprop:
Fix #1852 
Designs: 
https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=515%3A35388
https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=507%3A31258

PR #1958, opened 2022-01-20
Channel layout Update and Responsive + NFT

eldiegod:
fixes: #1856

### DIFF
--- a/packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
+++ b/packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
@@ -7,7 +7,7 @@ import { media, sizes, transitions, zIndex } from '@/styles'
 import { SubTitle, TitleSection } from '@/views/viewer/ChannelView/ChannelView.styles'
 
 export const StyledTitleSection = styled(TitleSection)`
-  margin-top: 0 !important;
+  margin-top: 0;
   padding-top: ${sizes(8)};
   display: inline-flex;
   width: auto;
--- a/packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
+++ b/packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
@@ -7,13 +7,11 @@ import { media, sizes, transitions, zIndex } from '@/styles'
 import { SubTitle, TitleSection } from '@/views/viewer/ChannelView/ChannelView.styles'
 
 export const StyledTitleSection = styled(TitleSection)`
+  margin-top: 0 !important;
+  padding-top: ${sizes(8)};
   display: inline-flex;
   width: auto;
   flex-direction: row;
-  padding-top: ${sizes(8)};
-  ${media.sm} {
-    padding-top: 0;
-  }
 
   /* Hidden visibility on container to not block hover state on Channel cover. 
   TitleArea, SubTitle and Avatar must be visible  */
@@ -22,7 +20,6 @@ export const StyledTitleSection = styled(TitleSection)`
 
 export const StyledTitleArea = styled(TitleArea)`
   visibility: visible;
-  margin-left: ${sizes(2)};
 `
 
 export const StyledSubTitle = styled(SubTitle)`
--- a/packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
+++ b/packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
@@ -144,7 +144,7 @@ export const TabsContainer = styled.div`
     align-items: center;
     box-shadow: ${cVar('effectDividersBottom')};
     gap: ${sizes(8)};
-    grid-template: 1fr / auto 1fr 193.333px;
+    grid-template: 1fr / auto 1fr 160px;
   }
 `
 
--- a/packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
+++ b/packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
@@ -28,31 +28,36 @@ const activeUnderline = css`
 `
 
 export const TitleSection = styled.div`
+  margin-top: -${sizes(6)};
   display: grid;
-  grid-template-columns: auto 1fr;
-  grid-row-gap: ${sizes(4)};
-  align-items: center;
+  grid-template-rows: repeat(auto, 3);
+  grid-template-columns: 1fr;
+  grid-row-gap: ${sizes(6)};
+  justify-items: center;
   width: 100%;
-  margin: ${sizes(8)} 0 ${sizes(14)} 0;
 
-  ${media.xs} {
-    grid-template-columns: auto 1fr auto;
+  ${media.sm} {
+    margin-top: 0;
+    grid-template-columns: min-content max-content auto;
   }
 `
+
 export const TitleContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
   max-width: 100%;
   overflow: hidden;
-
-  ${media.md} {
-    max-width: 60%;
-  }
-
   z-index: 2;
+
+  ${media.sm} {
+    align-items: normal;
+    justify-content: center;
+  }
 `
 
 export const Title = styled(Text)`
   margin-bottom: 0;
-  padding: ${sizes(1)} ${sizes(2)} 5px;
   white-space: nowrap;
   text-overflow: ellipsis;
   overflow: hidden;
@@ -69,8 +74,7 @@ export const SortContainer = styled.div`
 `
 
 export const SubTitle = styled(Text)`
-  padding: ${sizes(1)} ${sizes(2)};
-  margin-top: ${sizes(1)};
+  margin: ${sizes(2)} 0;
   color: ${oldColors.gray[300]};
   display: inline-block;
 `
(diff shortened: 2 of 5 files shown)


src/components/Avatar/Avatar.styles.ts
```diff
@@ -40,7 +40,7 @@ const channelAvatarCss = css`
   width: 88px;
   min-width: 88px;
   height: 88px;
-  ${media.md} {
+  ${media.sm} {
```
drillprop: Hmm, I'm a little worried that this change could break avatar somewhere else. Are you sure that we need this? Maybe we could use `useMediaMatch`  inside component instead.

src/views/viewer/ChannelView/ChannelView.styles.ts
```diff
@@ -28,31 +29,64 @@ const activeUnderline = css`
...
+  justify-items: center;
+  width: 100%;
+
+  ${media.sm} {
+    margin-top: 0;
+    grid-template-columns: min-content max-content auto;
+  }
+`
+
+export const CollectorsBox = styled.div`
```
drillprop: I don't see the point of creating this, let's just use [Collectors panel](https://github.com/Joystream/atlas/pull/1948) once it's merged. I would just remove this and not include this in this PR. I know that my PR is blocking you, but I think we could include finished CollectorsBox in a separate PR.

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -214,6 +220,28 @@ export const ChannelView: React.FC = () => {
     }
   }, [currentTabName])
 
+  const collectorBoxNode = (
```
drillprop: Let's remove this as well

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -237,12 +265,13 @@ export const ChannelView: React.FC = () => {
     <ViewWrapper>
       <ChannelCover assetUrl={coverPhotoUrl} />
       <LimitedWidthContainer>
+        {smMatch ? <CollectorsBoxContainer>{collectorBoxNode}</CollectorsBoxContainer> : null}
```
drillprop:
```suggestion
```
Let's just delete this. Also, CollectorsBoxContainer doesn't work well, If you remove `CollectorsBoxContainer`(If I'm not mistaken, some channels will not have collectorsBox), the layout will be broken.
Look: 
image

src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -252,6 +281,7 @@ export const ChannelView: React.FC = () => {
               </>
             )}
           </TitleContainer>
+          {smMatch ? null : collectorBoxNode}
```
drillprop: let's just delete this as well for now.

drillprop (changes requested):

src/components/Avatar/Avatar.styles.ts
eldiegod: I think the api decision might be inconsistent with the other avatar sizes, but as the variable name suggests, this is only to be used for the `channelAvatar`. I also would like this to be separate implementation but at the moment is a variant. 

Not too sure what to do here.

drillprop: Did you check studio views as well? if you're sure that it doesn't break anything, I'm fine with that change 👍🏻

drillprop:
Just took a glance at the studio. Your PR is breaking spacing on `studio/channel`. 
image
should be: 
image

src/components/Avatar/Avatar.styles.ts
eldiegod: It does break sadly.

Should we just keep it consistent with the variant or what do we do here @toiletgranny ?

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -236,12 +240,33 @@ export const ChannelView: React.FC = () => {
     <ViewWrapper>
       <ChannelCover assetUrl={coverPhotoUrl} />
       <LimitedWidthContainer>
+        {smMatch ? (
+          <CollectorsBoxContainer>
+            <CollectorsBox
+              collectors={[
```
drillprop: If there are no collectors we will see: 
image

And if I remove CollectorsBox there will be a problem with spacing: 
image

drillprop (changes requested):

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.styles.ts
```diff
@@ -7,13 +7,11 @@ import { media, sizes, transitions, zIndex } from '@/styles'
 import { SubTitle, TitleSection } from '@/views/viewer/ChannelView/ChannelView.styles'
 
 export const StyledTitleSection = styled(TitleSection)`
+  margin-top: 0 !important;
```
drillprop: I'm pretty sure, that we don't need to add `!important` here.

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
```diff
@@ -131,20 +134,17 @@ export const PaginationContainer = styled.div`
...
-    padding-top: ${sizes(8)};
-  }
-
   ${media.sm} {
     align-items: center;
-    border-bottom: solid 1px ${oldColors.gray[700]};
+    box-shadow: ${cVar('effectDividersBottom')};
     gap: ${sizes(8)};
-    grid-template: 1fr / auto 1fr 250px;
+    grid-template: 1fr / auto 1fr 193.333px;
```
drillprop: Why `193.333`? Where does it come from?

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -236,12 +256,17 @@ export const ChannelView: React.FC = () => {
     <ViewWrapper>
       <ChannelCover assetUrl={coverPhotoUrl} />
       <LimitedWidthContainer>
+        {smMatch ? (
+          <CollectorsBoxContainer>
+            {collectors.length > 0 && <CollectorsBox collectors={collectors} />}
```
drillprop, line 252: I still don't like the fact we have this empty div that just taking the space 😅 
image
I mean, I don't have a better solution for this, so if you don't come up with an idea of how to remove this guy, just leave it as it is.

drillprop (changes requested):

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
eldiegod: I don't have an issue with it it just keeps the space for the collector box, a solution for it would be more complicated than just not doing nothing

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
eldiegod: 250px was the size of the sort select + label but they removed the label so 193.333 is the remaining size the select should have without it

src/components/Avatar/Avatar.styles.ts
toiletgranny: Talked to @Lenorette, waiting for her response.

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
drillprop: This looks odd, and I still don't see the point of setting this to exactly 193.333 px.  
Why can't we just set this to a rounded value like 200px?

eldiegod:
Had a chat with Adam from design and decided to keep the Avatar variant as it is for now.

Lenorette:
Hey @eldiegod I've talked to Adam and I've changed avatar size on 768px breakpoint to 88px. 
https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=510%3A24776

I did the same changes also on member profile, so it's the same everywhere 
https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=923%3A136157

Let me know if its ok now 🙌

eldiegod:
> Let me know if its ok now 🙌

@Lenorette awesome, yes its great, thanks a lot for the quickness ⚡️

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
eldiegod: @drillprop So, made it 160px as per design, as it no longer has a value label

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -213,6 +217,22 @@ export const ChannelView: React.FC = () => {
...
+    },
+    { assetUrl: 'https://thispersondoesnotexist.com/image', tooltipText: 'William', nftsAmount: 6 },
+    {
+      assetUrl: 'https://eu-central-1.linodeobjects.com/atlas-assets/channel-posters/2.jpg',
+      tooltipText: 'Someone',
+      nftsAmount: 1,
+    },
+    { assetUrl: 'https://thispersondoesnotexist.com/image', tooltipText: 'Someone else', nftsAmount: 7 },
+  ]
+
```
drillprop, line 226: Let's maybe set this as an empty array for now and add the `todo` comment here? 
I'm a little worried that we deploy this by accident.

drillprop (approved):

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
kdembler: Certainly! @eldiegod for now please let's hide this section and also the NFTs tab 🙏

eldiegod: its done @kdembler @drillprop

PR #1964, opened 2022-01-20
Adjust step colors

### LINKED ISSUE #1677: Apply color adjustments to the "Step" component
### Description

Based on information gathered from @KubaMikolajczyk, as part of the NFT MVP scope of work.

### Figma link

https://www.figma.com/file/uY98FOmQPgozpNOtAiklBg/Issuing-NFT?node-id=796%3A48187

### How urgent this is?

This will block me soon

drillprop:
Fix #1677 
Summary of changes: 
- removed `active` and `completed` props, replaced them with `stepType` which could be either `current`, `future` or `completed`. 
- adjusted colors, some storybook refactor

Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4678%3A144966
Changes are very subtle, and it could be hard to notice what is exactly updated. So, just for your convenience, I took these screenshots:
Before: 
image
After: 
image

~**One thing I need to mention:**
I can't change the background color of the step, because we have the same background in StepperModal. Colors `Core / Neutral / 700 / Default` and `Background / Default / Strong / Default` are the same colors(#272D33). If I change this, those steps will be invisible. I'll ask Kuba what should we do with this(just FYI).~

### DIFF
--- a/src/components/Step/Step.tsx
+++ b/src/components/Step/Step.tsx
@@ -11,6 +11,7 @@ import {
   StepStatus,
   StepTitle,
   StepType,
+  StepVariant,
   StepWrapper,
 } from './Step.styles'
 
@@ -19,8 +20,8 @@ import { IconButton } from '../_buttons/IconButton'
 
 export type StepProps = {
   title: string
-  variant?: 'file' | 'default'
-  stepType?: StepType
+  variant?: StepVariant
+  type?: StepType
   isLoading?: boolean
   disabled?: boolean
   number?: number
@@ -29,7 +30,7 @@ export type StepProps = {
 }
 
 export const Step = forwardRef<HTMLDivElement, StepProps>(
-  ({ variant = 'default', isLoading, disabled, title, number, onDelete, className, stepType = 'current' }, ref) => {
+  ({ type = 'default', isLoading, disabled, title, number, onDelete, className, variant = 'current' }, ref) => {
     const [circularProgress, setCircularProgress] = useState(0)
 
     useEffect(() => {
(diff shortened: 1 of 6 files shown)


drillprop:
> I can't change the background color of the step, because we have the same background in StepperModal. Colors `Core / Neutral / 700 / Default` and `Background / Default / Strong / Default` are the same colors(#272D33). If I change this, those steps will be invisible. I'll ask Kuba what should we do with this(just FYI).

Fixed, we decided to use `colorBackgroundStrongAlpha` after all

src/components/Step/Step.tsx
```diff
@@ -3,23 +3,33 @@ import React, { forwardRef, useEffect, useState } from 'react'
...
+  StepWrapper,
+} from './Step.styles'
 
+import { Text } from '../Text'
 import { IconButton } from '../_buttons/IconButton'
 
 export type StepProps = {
   title: string
   variant?: 'file' | 'default'
-  completed?: boolean
```
eldiegod, line 13: I'm not sure we should combine `active` and `completed` props into `stepType`. I think they mean pretty different things, and even tho it might not be the case right now; it could very well be the case in the future

eldiegod (changes requested):
Visually looks good, tho had a comment about the prop change

src/components/Step/Step.tsx
drillprop: hmm, I don't agree, for now, the design specifies that there will be only three possibilities and I think we should follow the design as much as possible.
image
If we actually need this in the future we can always add these props. We don't need to make this more complicated. 

I did a small change in naming, though. `type` is now either `file` or `default` and `variant` is now either `current`, `future`, or `completed`

eldiegod: I think the naming is better for sure; whether we should follow designs as close as possible, I agree when it comes to styles, but logic wise, it is up to us to make sure it is a sound API; as we said, in this case, it doesn't matter too much but just something to keep in mind 👌

eldiegod (approved):

Issue #1863, opened 2022-01-05
Context menu component should use ListItem component as its items


drillprop:
Currently, we have some custom styled component inside Context Menu: 
![image](https://user-images.githubusercontent.com/51168865/148199666-becdb63b-9dbb-41d2-8a9b-251571967321.png)

We should make use of the recently added ListItem instead.

toiletgranny:
Please note that there's also a change in the background color that needs to happen. Designs for this component, along with a changelog outlining what exactly has changed (November 23, 2021) [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3158%3A103636).

PR #1979, opened 2022-01-24
use ListItem in ContextMenu

lcoenengithub:
**What is this?**

This update the MenuContext component to use the new ListItem instead of a custom styling. This also uses the new color scheme. 

**Issue** 

https://github.com/Joystream/atlas/issues/1863

**Illustration**

contextmenu

### DIFF
--- a/src/components/_overlays/ContextMenu/ContextMenu.styles.ts
+++ b/src/components/_overlays/ContextMenu/ContextMenu.styles.ts
@@ -1,25 +1,13 @@
 import styled from '@emotion/styled'
 
 import { Text } from '@/components/Text'
-import { cVar, oldColors, sizes, transitions } from '@/styles'
+import { cVar, sizes, transitions } from '@/styles'
 
 export const StyledContainer = styled.div`
-  background-color: ${oldColors.gray[800]};
+  background-color: ${cVar('colorBackgroundStrong')};
   width: 200px;
-  color: ${oldColors.white};
+  color: ${cVar('colorText')};
   word-break: break-all;
-`
-
-export const StyledMenuItem = styled.div`
-  display: flex;
-  align-items: center;
-  padding: ${sizes(4)};
-  transition: background-color 200ms ${transitions.easing};
-
-  &:hover {
-    cursor: pointer;
-    background-color: ${oldColors.gray[700]};
-  }
 `
 
 export const StyledText = styled(Text)`
--- a/src/components/_overlays/ContextMenu/ContextMenu.styles.ts
+++ b/src/components/_overlays/ContextMenu/ContextMenu.styles.ts
@@ -1,18 +1,9 @@
 import styled from '@emotion/styled'
 
-import { Text } from '@/components/Text'
-import { cVar, sizes, transitions } from '@/styles'
+import { cVar } from '@/styles'
 
 export const StyledContainer = styled.div`
   background-color: ${cVar('colorBackgroundStrong')};
   width: 200px;
-  color: ${cVar('colorText')};
   word-break: break-all;
-`
-
-export const StyledText = styled(Text)`
-  font: ${cVar('typographyDesktopH200')};
-  letter-spacing: ${cVar('typographyDesktopH200LetterSpacing')};
-  text-transform: ${cVar('typographyDesktopH200TextTransform')};
-  margin-left: ${sizes(3)};
 `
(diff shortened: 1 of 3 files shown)


src/components/_overlays/ContextMenu/ContextMenu.styles.ts
```diff
@@ -1,27 +1,15 @@
 import styled from '@emotion/styled'
 
 import { Text } from '@/components/Text'
-import { cVar, oldColors, sizes, transitions } from '@/styles'
+import { cVar, sizes, transitions } from '@/styles'
```
drillprop:
```suggestion
import { cVar, sizes } from '@/styles'
```

src/components/_overlays/ContextMenu/ContextMenu.styles.ts
```diff
@@ -1,27 +1,15 @@
...
-  padding: ${sizes(4)};
-  transition: background-color 200ms ${transitions.easing};
-
-  &:hover {
-    cursor: pointer;
-    background-color: ${oldColors.gray[700]};
-  }
-`
-
 export const StyledText = styled(Text)`
```
drillprop: You should also remove this component.

src/components/_overlays/ContextMenu/ContextMenu.tsx
```diff
@@ -12,12 +14,7 @@ export type MenuItemProps = {
 }
 
 export const ContextMenuItem: React.FC<MenuItemProps> = React.memo(({ icon, onClick, title }) => {
```
kdembler: I don't think we need this component anymore, seems like we could use `ListItem` directly in `ContextMenu` component

src/components/_overlays/ContextMenu/ContextMenu.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { ReactNode, useRef } from 'react'
 
-import { StyledContainer, StyledMenuItem, StyledText } from './ContextMenu.styles'
+import { ListItem } from '@/components/ListItem'
+
+import { StyledContainer, StyledText } from './ContextMenu.styles'
```
drillprop:
```suggestion
import { StyledContainer } from './ContextMenu.styles'
```

src/components/_overlays/ContextMenu/ContextMenu.styles.ts
kdembler: This doesn't seem to be used anymore

src/components/_overlays/ContextMenu/ContextMenu.styles.ts
```diff
@@ -1,27 +1,15 @@
...
 import { Text } from '@/components/Text'
-import { cVar, oldColors, sizes, transitions } from '@/styles'
+import { cVar, sizes, transitions } from '@/styles'
 
 export const StyledContainer = styled.div`
-  background-color: ${oldColors.gray[800]};
+  background-color: ${cVar('colorBackgroundStrong')};
   width: 200px;
-  color: ${oldColors.white};
+  color: ${cVar('colorText')};
```
kdembler: I don't think we need this style here, the text color will come from the list item component

src/components/_overlays/ContextMenu/ContextMenu.tsx
drillprop: We don't need ContextMenuItem anymore. Let's remove this and just use `ListItem` inside `ContextMenu`.

drillprop (changes requested):
Needs some cleaning. Besides that, looks good

kdembler (changes requested):
The menu items also doesn't have a pointer cursor. We should probably add this style to the `ListItem` component

kdembler:
Lol me and @drillprop made very similar comments, I didn't see that before posting my review 😅

src/components/_overlays/ContextMenu/ContextMenu.tsx
lcoenengithub: Good call!

kdembler:
@lcoenengithub I think you missed my comment:

> The menu items also doesn't have a pointer cursor. We should probably add this style to the ListItem component

drillprop (approved):
LGTM 👍🏻

kdembler (approved):
Congrats on the first merge @lcoenengithub 🎉

Issue #1710, opened 2021-12-03
Update Side navigation component to reflect updated Sign up flow


toiletgranny:
### Description

Based on information gathered from @Lenorette 

### How urgent this is?

It's blocking me right now

kdembler:
What changes are we expecting here?

toiletgranny:
@kdembler We'll just be updating the "Joystream Studio" button at the bottom, as we need to add some logic in case the user is not signed in, etc.

toiletgranny:
@kdembler I decided to do a little bit more than updating the buttons at the bottom of the Side navigation component, assuming that there might not be a better opportunity to tackle this component in the near future. 

![image](https://user-images.githubusercontent.com/35307309/148533139-5fa9f105-759d-4314-a83d-9adeb382a3bb.png)

Designs and documentation for this component can be found here:
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4322%3A173750

Also, please make sure to tackle https://github.com/Joystream/atlas/issues/1880 in the first place.

toiletgranny:
@kdembler There aren't a lot of changes (mostly: applying our new color design tokens, spacing & margins refinements, etc.) but in case you guys need to stick to only NFT and the updated Sign up flow-related changes, here's the recap:

1. When you're not "signed in", you should now see the blue "Sign in" button, instead of the current "Joystream studio". The button should do the exact same thing as its recently implemented counterpart in the Top navigation component.
![image](https://user-images.githubusercontent.com/35307309/148534705-5ca4273e-2fb3-454e-9654-3fa4fde77a7d.png)

2. When you're "signed in", the currently "Joystream studio" button should be replaced with "Go to Studio" button. Again, this button should  do the exact same thing as its recently implemented counterpart in the Top navigation component.
![image](https://user-images.githubusercontent.com/35307309/148534676-f2fdd651-05d0-4eed-b541-e83e10074918.png)

3. In the Studio, you should now see two buttons stacked on top of each other: 
![image](https://user-images.githubusercontent.com/35307309/148534639-4943d0e2-1a9f-4267-926b-0a9317e23186.png)

kdembler:
Let's do the full thing and also include changes from #1880 here

PR #1994, opened 2022-01-25
Update side navigation component to reflect

lcoenengithub:
**What is this?**

This update sign up flow to include direct link to signing in in the main view.

**Issue**

https://github.com/Joystream/atlas/issues/1710

**Illustration**

image
image
image


drillprop (changes requested):
I think we should close the side nav when clicking `Upload Video` in the studio.
image

Besides that, looks good.

Issue #1803, opened 2021-12-16
Layout of Members Page


dmtrjsg:
**AC:**

- [ ] Page layout together with all tabs
- [ ] About Tab

**Designs:**

Page [Link to Figma](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=919%3A130033)

About Tab [Link to Figma](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=890%3A130332)

Issue #2015, opened 2022-01-27
Fix topbar dropdown animation


drillprop:
happens on `signin` branch

To reproduce: 
Just refresh the site a few times.

https://user-images.githubusercontent.com/51168865/151388252-d9f1b68a-2ace-4027-b0db-0d71cf0cc3ef.mov

https://user-images.githubusercontent.com/51168865/151388412-e17c4a02-337c-48ed-8c9d-2132fec3e542.mov

PR #2020, opened 2022-01-27
Layout of Members Page

eldiegod:
Fixes:  #1803, #2015


packages/atlas/src/views/viewer/MemberView/MemberAbout.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <GridItem colSpan={{ base: 12, sm: 3 }} colStart={{ sm: -4 }}>
+        <DetailsText variant="h500">Details</DetailsText>
+
+        <Details>
+          <Text variant="t100" secondary>
+            Join date
+          </Text>
+          <Text variant="t300">
+            6 Jan 2019
+            {/* {member?.createdAt ? formatDate(new Date(member.createdAt)) : ''} */}
```
kdembler: This should use real data

packages/atlas/src/views/viewer/MemberView/MemberAbout.tsx
```diff
@@ -0,0 +1,84 @@
...
+          </Text>
+        </Details>
+
+        <Details>
+          <Text variant="t100" secondary>
+            Num. of channels
+          </Text>
+          <Text variant="t300">
+            {typeof member?.channels.length === 'number' ? formatNumberShort(member?.channels.length) : ''}
```
kdembler: Why do we need `typeof` here? I think we could just do `member?.channels.length ? ... : '0'`

packages/atlas/src/views/viewer/MemberView/MemberView.styles.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  ${media.xs} {
+    padding-top: ${sizes(8)};
+  }
+
+  ${media.sm} {
+    align-items: center;
+    border-bottom: solid 1px ${oldColors.gray[700]};
+    gap: ${sizes(8)};
+    grid-template: 1fr / 1fr 193.333px;
```
kdembler: Where is `193.333px` value coming from?

packages/atlas/src/views/viewer/MemberView/MemberView.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { ViewWrapper } from '@/components/ViewWrapper'
+import { Select } from '@/components/_inputs/Select'
+import { SORT_OPTIONS } from '@/config/sorting'
+import { useUser } from '@/providers/user'
+import { SentryLogger } from '@/utils/logs'
+
+import { MemberAbout } from './MemberAbout'
+import { SortContainer, StyledMembershipInfo, StyledTabs, TabsContainer } from './MemberView.styles'
+
+const TABS = ['NFTs', 'Activity', 'About'] as const
```
kdembler: Let's hide `NFTs` and `Activity` tabs for now so we can release member profile before those are ready

packages/atlas/src/views/viewer/MemberView/MemberView.tsx
```diff
@@ -0,0 +1,112 @@
...
+  if (error) {
+    return <ViewErrorFallback />
+  }
+  return (
+    <ViewWrapper>
+      <LimitedWidthContainer>
+        <StyledMembershipInfo
+          avatarUrl={member?.avatarUri ?? undefined}
+          handle={member?.handle}
+          // address={member?.controllerAccount}
```
kdembler: This shouldn't be commented?

kdembler (changes requested):
Other than the comments:
1. Let's merge this into `signin`
2. Let's make sure this works well with the edit membership view, as discussed in RocketChat

packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
```diff
@@ -94,7 +99,13 @@ export const MembershipInfo: React.FC<MembershipInfoProps> = ({
             (loading ? (
               <SkeletonLoader width={smMatch ? 148 : '100%'} height={48} />
             ) : (
-              <Button icon={<SvgActionEdit />} size="large" variant="secondary" fullWidth={!smMatch}>
+              <Button
+                onClick={() => navigate(absoluteRoutes.viewer.editMembership())}
```
kdembler: Why use `navigate` instead of regular `a` link?

packages/atlas/src/config/routes.ts
```diff
@@ -30,7 +30,8 @@ export const relativeRoutes = {
     channel: (id = ':id') => `channel/${id}`,
     channels: () => 'channels',
     video: (id = ':id') => `video/${id}`,
-    editMembership: () => 'membership/edit',
+    editMembership: () => 'member/edit',
```
kdembler: What if someone names their membership `edit`? :P

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -70,6 +73,9 @@ export const EditMembershipView: React.FC = () => {
       },
     })
     refetchActiveMembership()
+    if (success) {
+      navigate(absoluteRoutes.viewer.member(activeMembership?.handle))
+    }
```
kdembler: We don't await `refetchActiveMembership` so `activeMembership?.handle` will be the old handle if we edit it, right?

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -100,7 +106,10 @@ export const EditMembershipView: React.FC = () => {
           secondaryButton={{
             visible: true,
             text: 'Cancel',
-            onClick: resetForm,
+            onClick: () => {
+              resetForm()
+              navigate(absoluteRoutes.viewer.member(activeMembership?.handle))
```
kdembler: I'd suggest to go back here instead of the main member page, `react-router` should give us some method to do that. Also is there any value in resetting the form given that we leave it?

packages/atlas/src/views/viewer/MemberView/MemberAbout.tsx
```diff
@@ -0,0 +1,80 @@
...
+            Join date
+          </Text>
+          <Text variant="t300">{member?.createdAt ? formatDate(new Date(member.createdAt)) : ''}</Text>
+        </Details>
+
+        <Details>
+          <Text variant="t100" secondary>
+            Num. of channels
+          </Text>
+          <Text variant="t300">{member?.channels.length ? formatNumberShort(member?.channels.length) : ''}</Text>
```
kdembler: Let's use '0' as fallback. As it is currently for 0 channels we will display nothing, right?

packages/atlas/src/views/viewer/MemberView/MemberView.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  useEffect(() => {
+    if (currentTabName) {
+      setCurrentTab(currentTabName)
+    }
+  }, [currentTabName])
+
+  if (error) {
+    return <ViewErrorFallback />
+  }
```
kdembler, line 86: I think we don't handle the member not found in any way?

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
eldiegod: go back where exactly?

packages/atlas/src/config/routes.ts
eldiegod: maybe `my-membership/edit` ? don't have too many great ideas on this

packages/atlas/src/views/viewer/MemberView/MemberView.tsx
eldiegod: Added the usual we do for that

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
kdembler: To where you came from. To get to this page you will click the `Edit profile` button in the membership header right? When you did that, you was on some membership view, so that's where I'm suggesting we go back. We can just also go to main member view to simplify, in that case let's use regular navigation instead of `navigate`

kdembler: Ah I see you already did that, let's leave it as it is then

packages/atlas/src/config/routes.ts
kdembler: I think this should be `member/:handle/edit` actually, but we will need to adjust the logic a bit so that if I try to go to this route for someone else's profile, I get redirected

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
eldiegod: yeah I think thats best

packages/atlas/src/config/routes.ts
eldiegod: yeah, the issue for me is that that handle in URL seems redundant because we already know the member that's trying to edit it, and as you mention it creates the issue of having to address edge cases of people trying to edit other members profiles

we would also benefit from a global edit your membership link that would work for everyone

kdembler (approved):

PR #2024, opened 2022-01-28
update meta server for giza

### LINKED ISSUE #2023: Update social previews server to Giza

kdembler:
Resolves #2023

Summary of changes:
- Updated meta server to use Giza asset resolution
- Added graphql-codegen for type-safe queries
- Reworked tags generation
  - Refactored in a way so that it's easy to get just a tags list, this may be useful in regular Atlas
  - Added some additional meta tags
  - Added schema.org tags as that's what Discord uses (at least partially)
  - Tried to make embedded playback work on FB/Discord but failed
- Added Github action to build and publish docker image


packages/atlas-meta-server/src/api/index.ts
```diff
@@ -0,0 +1,23 @@
+import { GraphQLClient } from 'graphql-request'
+
+import { getSdk } from './__generated__/sdk'
+
+const { GRAPHQL_URL } = process.env
```
rafalpawlow: I cannot comment `docker-compose` here but I think we should add `GRAPHQL_URL` also to some env file for development purposes.

rafalpawlow (changes requested):

rafalpawlow (approved):

Issue #510, opened 2021-04-26
Allow dynamic page title


kdembler:
Dynamic page title would be a nice addition to Atlas - the browser tab name could change depending on the content displayed. (video, channel, "add new video", etc)

MIPB:
Can I try to solve this issue?

kdembler:
Sure thing, feel free to give it a try. I'm not exactly sure how this would work though - if we want the title to be truly dynamic, then we would need to use the data used on a given page. For example, how do you set it to the title of the video? It's doable, but may be a bit tricky. I suggest we discuss the solution here first to make sure it fits nicely and start implementation then

brabbuss:
@MIPB @kdembler I should've probably piped in here before working on this issue. I had begun adding dynamic titles and interpolating in relevant information as well as copy for while the video is loading. 

I was unsure of procedure, but I can hold off from coding any further just let me know! Anything I should know going forward? Still new to open source projects.

ex: 
![Screen Shot 2021-04-26 at 1 09 59 PM](https://user-images.githubusercontent.com/66697338/116137593-fae8e300-a690-11eb-8489-15bd0ab99f49.png)
![Screen Shot 2021-04-26 at 1 09 47 PM](https://user-images.githubusercontent.com/66697338/116137600-fde3d380-a690-11eb-8fce-e94d28934991.png)

MIPB:
I see @brabbuss has already started the improvement so I'll let him work on it. Looks good 👍

kdembler:
Oh wow, got a little crowded in here! @brabbuss if you're already working on it, please continue, but going forward it's a good idea to mention when you start working on something so we don't run into situation where two community members work on the same thing 😅 

Since you already have something working, I guess at this point, just open a draft PR once you feel good about it (doesn't have to be yet implemented for all the views) so we can discuss there. Thanks! 👍

brabbuss:
@kdembler @MIPB thanks! And sorry for the lack of communication there, I'm very new to the open source workflow.

@kdembler Is there a specific timeline where you'd prefer this issue be resolved by, or some general cadence to resolving issues? And will do on the draft PR - thanks all 😅

kdembler:
No worries, we managed to resolve this on time! And welcome to the open source community :)

Regarding the timeline, there isn't any really, that's just something that came to my mind today, it's not part of any bigger release. So as long as it doesn't drag for over a month, it's fine with me 😄

brabbuss:
[...]

PR #2050, opened 2022-02-01
add dynamic page titles and rich metadata

kdembler:
Summary of changes:
- Add dynamic page title on each view, resolves https://github.com/Joystream/atlas/issues/510
- Add rich meta tags on channel and video views
- Use [Media Session API](https://developer.mozilla.org/en-US/docs/Web/API/Media_Session_API) on video view to provide playback metadata to the browser:<br>Screenshot 2022-02-01 at 14 37 48


packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
```diff
@@ -11,6 +12,8 @@ import { VideoWorkspaceForm } from './VideoWorkspaceForm'
 import { VideoWorkspaceTabsBar } from './VideoWorkspaceTabsBar'
 
 export const VideoWorkspace: React.FC = React.memo(() => {
+  const headTags = useHeadTags('Edit video')
```
rafalpawlow: Shouldn't we distinguish between adding end editing?

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -53,6 +54,8 @@ export const EditMembershipView: React.FC = () => {
     }
   }, [activeMembership, activeMembershipLoading, resetForm])
 
+  const headTags = useHeadTags('Edit membership')
```
rafalpawlow, line 57: Same question about the distinction between editing and adding

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -108,6 +117,27 @@ export const VideoView: React.FC = () => {
...
+      return
+    }
+
+    const artwork: MediaImage[] = thumbnailUrl ? [{ src: thumbnailUrl, type: 'image/webp', sizes: '640x360' }] : []
+
+    navigator.mediaSession.metadata = new MediaMetadata({
+      title: video.title || '',
+      artist: video.channel.title || '',
+      album: '',
+      artwork: artwork,
```
rafalpawlow, line 133:
```suggestion
      artwork,
```

rafalpawlow (changes requested):

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
kdembler: Here you can only edit

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
kdembler: Wanted to keep it simple, but updated now, also for channel edit

rafalpawlow (approved):

PR #2051, opened 2022-02-01
Update TopbarStudio and other minor tweaks

### LINKED ISSUE #2038: Use new member/channel indicator in studio topbar
We currently have:
Screenshot 2022-01-31 at 16 11 17
 
We should have:
Screenshot 2022-01-31 at 16 12 29

### LINKED ISSUE #2066: Duplicated loader
**Describe the bug**

When you try to log in, 2 loaders appear

**To Reproduce**

1. go to studio
2. logout (remove localStorage entry)
3. refresh page
4. sign in

**Expected behavior**

**Screenshots**
Zrzut ekranu 2022-02-3 o 17 05 45 (2)

- OS: macOS Monterey 12.1
- Browser Chrome 97.0.4692.99

drillprop:
fix #2038 and #2066
Designs: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3173%3A135756

### DIFF
--- a/packages/atlas/src/components/Avatar/Avatar.styles.ts
+++ b/packages/atlas/src/components/Avatar/Avatar.styles.ts
@@ -102,7 +102,18 @@ const getAvatarSizeCss = ({ size }: ContainerProps): SerializedStyles => {
   }
 }
 
-const getBorderStyles = ({ isLoading, withoutOutline }: ContainerProps) => {
+export const sharedAvatarHoverStyles = css`
+  ::after {
+    box-shadow: inset 0 0 0 2px ${cVar('colorCoreNeutral200')};
+  }
+`
+export const sharedAvataActiveStyles = css`
+  ::after {
+    box-shadow: inset 0 0 0 2px ${cVar('colorCoreBlue500')};
+  }
+`
+
+const getBorderStyles = ({ isLoading, withoutOutline }: Omit<ContainerProps, 'size'>) => {
   if (withoutOutline) {
     return
   }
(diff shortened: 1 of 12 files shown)


eldiegod:
top bar looks good but not sure whats in scope here

drillprop:
> top bar looks good but not sure whats in scope here

hmm, I thought that it was self-explanatory 😄 . Look at the issue.  
image

Anyway, the scope of this task was just to update the topbar in the studio. 
I didn't add the button with notifications though, because we currently don't have this feature available.

eldiegod (changes requested):
Alright, looks and works well overall, just one thing, I don't think it was reflected in the designs but it triggers me a bit that when there are 2 Avatars and I want to click to open the dropdown they hover individually. 

I think these should hover at the same time and behave like a unit in there because they essentially are, wdyt?

drillprop:
> Alright, looks and works well overall, just one thing, I don't think it was reflected in the designs but it triggers me a bit that when there are 2 Avatars and I want to click to open the dropdown they hover individually.
> 
> I think these should hover at the same time and behave like a unit in there because they essentially are, wdyt?

Or maybe when you click channel avatar we should navigate to user channel? I will ask designers about this because I'm not sure what we should do in this case. The design doesn't explain that at all.
Anyway, good point!

drillprop:
@eldiegod I've talked to Adam, and he agrees with you. I'll implement what you proposed. Updated design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=4735%3A139148

drillprop:
Updated

eldiegod (approved):

PR #2059, opened 2022-02-01
show avatar hover styles only when it's clickable

kdembler:
Like in the title, a lot of avatars currently are misleading because they have hover/focus styles but don't allow any interaction

### DIFF
--- a/packages/atlas/src/components/Avatar/Avatar.styles.ts
+++ b/packages/atlas/src/components/Avatar/Avatar.styles.ts
@@ -12,6 +12,7 @@ export type AvatarSize = 'preview' | 'cover' | 'default' | 'fill' | 'bid' | 'sma
 type ContainerProps = {
   size: AvatarSize
   isLoading?: boolean
+  isClickable: boolean
   withoutOutline?: boolean
 }
 
@@ -107,38 +108,37 @@ export const sharedAvatarHoverStyles = css`
     box-shadow: inset 0 0 0 2px ${cVar('colorCoreNeutral200')};
   }
 `
-export const sharedAvataActiveStyles = css`
+export const sharedAvatarActiveStyles = css`
   ::after {
     box-shadow: inset 0 0 0 2px ${cVar('colorCoreBlue500')};
   }
 `
 
-const getBorderStyles = ({ isLoading, withoutOutline }: Omit<ContainerProps, 'size'>) => {
-  if (withoutOutline) {
+const getBorderStyles = ({ isLoading, isClickable, withoutOutline }: Omit<ContainerProps, 'size'>) => {
+  if (withoutOutline || isLoading) {
     return
   }
-  if (!isLoading) {
-    return css`
-      ::after {
-        ${square('100%')};
 
-        content: '';
-        display: block;
-        border-radius: 50%;
-        z-index: ${zIndex.overlay};
-        pointer-events: none;
-        box-shadow: inset 0 0 0 1px ${cVar('colorBorderMutedAlpha')};
-      }
+  return css`
+    ::after {
+      ${square('100%')};
 
-      :hover {
-        ${sharedAvatarHoverStyles};
-      }
+      content: '';
+      display: block;
+      border-radius: 50%;
+      z-index: ${zIndex.overlay};
+      pointer-events: none;
+      box-shadow: inset 0 0 0 1px ${cVar('colorBorderMutedAlpha')};
+    }
 
-      :active {
-        ${sharedAvataActiveStyles}
-      }
-    `
-  }
+    :hover {
+      ${isClickable && sharedAvatarHoverStyles};
+    }
+
+    :active {
+      ${isClickable && sharedAvatarActiveStyles};
+    }
+  `
 }
 
 export const EditButton = styled.button<EditButtonProps>`
--- a/packages/atlas/src/components/Avatar/Avatar.styles.ts
+++ b/packages/atlas/src/components/Avatar/Avatar.styles.ts
@@ -14,6 +14,8 @@ type ContainerProps = {
   isLoading?: boolean
   isClickable: boolean
   withoutOutline?: boolean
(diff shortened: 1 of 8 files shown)


eldiegod (dismissed):
Looks good but build is failing, also remember to update the story because there is no way rn to check the hover styles on there

packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.stories.tsx
```diff
@@ -62,7 +62,7 @@ const RegularTemplate: Story<ImageCropModalProps> = () => {
 
   return (
     <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'start', gap: '24px' }}>
-      <Avatar assetUrl={avatarImageUrl} editable onEditClick={() => avatarDialogRef.current?.open()} size="cover" />
+      <Avatar assetUrl={avatarImageUrl} editable onClick={() => avatarDialogRef.current?.open()} size="cover" />
```
rafalpawlow, line 65:
```suggestion
      <Avatar assetUrl={avatarImageUrl} editable onClick={avatarDialogRef.current?.open} size="cover" />
```

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -415,7 +415,7 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
                 assetUrl={avatarAsset?.url}
                 hasAvatarUploadFailed={hasAvatarUploadFailed}
                 size="fill"
-                onEditClick={() => avatarDialogRef.current?.open()}
+                onClick={() => avatarDialogRef.current?.open()}
```
rafalpawlow, line 422:
```suggestion
                onClick={avatarDialogRef.current?.open}
```

packages/atlas/src/components/Avatar/Avatar.tsx
```diff
@@ -38,26 +37,32 @@ export const Avatar: React.FC<AvatarProps> = ({
...
+    <Container
+      onClick={onClick}
+      size={size}
+      className={className}
+      isLoading={loading}
+      withoutOutline={withoutOutline}
+      isClickable={!!onClick}
+    >
       {isEditable && (
         <EditButton size={size} onClick={handleEditClick} type="button">
```
rafalpawlow: I'm wondering if we could simplify this. Do we really need 2 onClick handlers since we get rid of `onEditClick`?

rafalpawlow (changes requested):

packages/atlas/src/components/Avatar/Avatar.tsx
kdembler: I would like to keep `onClick` on button since it may be more natural click target than `div` for screen readers etc

rafalpawlow: Maybe container should be a button then if Avatar is clickable?

packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.stories.tsx
kdembler: This would produce TS errors since `onClick` will pass `MouseEvent` as an argument and that's not what `open` expects

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
kdembler: Same as above

packages/atlas/src/components/Avatar/Avatar.tsx
kdembler: Yeah that would make more sense, let me refactor that

packages/atlas/src/components/Avatar/Avatar.tsx
```diff
@@ -38,27 +37,30 @@ export const Avatar: React.FC<AvatarProps> = ({
...
-      onEditClick(event)
-    }
-  }
   const isEditable = !loading && editable && size !== 'default' && size !== 'bid'
 
   return (
-    <Container onClick={onClick} size={size} className={className} isLoading={loading} withoutOutline={withoutOutline}>
+    <Container
+      as={onClick ? 'button' : 'div'}
+      // @ts-ignore 'type' prop will only be applied if the rendered tag is 'button' so it is valid here
```
rafalpawlow: I'm thinking, maybe we could just override `type` in `ContainerProps` instead of ignoring?

packages/atlas/src/components/Avatar/Avatar.styles.ts
```diff
@@ -180,11 +179,15 @@ export const Container = styled('div', { shouldForwardProp: isPropValid })<Conta
...
+  padding: 0;
+  border: 0;
+  background: none;
+
   display: flex;
   justify-content: center;
   align-items: center;
   position: relative;
-  cursor: ${({ onClick }) => (onClick ? 'pointer' : 'unset')};
+  cursor: ${({ isClickable }) => (isClickable ? 'pointer' : 'inherit')};
```
rafalpawlow, line 192: WDYT?
```suggestion
  cursor: ${({ isClickable }) => (isClickable ? 'pointer' : 'default')};
```

kdembler: Would this help with some issue? If we use `default`, it may override inherited styles. For example, if you nest avatar inside a link, with `default` it wouldn't inherit the pointer

packages/atlas/src/components/Avatar/Avatar.tsx
kdembler: Done

rafalpawlow (approved):

Issue #1804, opened 2021-12-16
Activity Of a member


dmtrjsg:
**AC:**
- [x] [Activity Panel](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=919%3A133678)
- [x] [Hover on panels ](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=1001%3A37467)in activity
- [x] [Sorting on page](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=919%3A133678). Sort by options: - Newest (defualt) - Oldest
- [x] [Learn more ](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=919%3A133679)opens in new tab more detailed information of member profile in Pionnier, as it happen now in Atlas when you click on member name.

PR #2065, opened 2022-02-02
Member Activity tab

eldiegod:
fixes: #1804


packages/atlas/src/components/Text/Text.styles.ts
```diff
@@ -6,11 +6,23 @@ import { cVar } from '@/styles'
...
   color?: string
   isSecondary?: boolean
+  clampAfterLine?: number
 }
 
-const baseStyles = ({ color, isSecondary }: TextProps) => css`
+const clampStyles = ({ clampAfterLine }: TextProps) => css`
+  /* stylelint-disable */
+  display: -webkit-box;
+  /* stylelint-enable */
```
drillprop:
```suggestion
  /* stylelint-disable-next-line value-no-vendor-prefix */
  display: -webkit-box;
```
should work

packages/atlas/src/views/viewer/MemberView/MemberActivity.styles.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  ${GridRowWrapper}:last-of-type > & {
+    box-shadow: none;
+
+    ${media.sm} {
+      box-shadow: ${cVar('effectDividersBottom')};
+    }
+  }
+
+  ${GridRowWrapper}
```
drillprop: What is doing this GridRowWrapper here?

packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -0,0 +1,203 @@
...
+  title: string
+  description: string
+  thumnailUri: string
+}
+const ActivityItem: React.FC<ActivityItemProps> = ({ date, type, title, description, thumnailUri }) => {
+  const [thumbnailLoaded, setThumbnailLoaded] = useState(false)
+  const smMatch = useMediaMatch('sm')
+  const lgMatch = useMediaMatch('lg')
+
+  React.useEffect(() => {
```
drillprop: Let's import `useEffect` for consistency.

packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -0,0 +1,203 @@
...
+    ;(async () => {
```
drillprop: Let's declare the function and call it inside the body of `useEffect`

packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -0,0 +1,203 @@
...
+      const res = await imageUrlValidation(thumnailUri)
+      setThumbnailLoaded(res)
+    })()
+  }, [thumnailUri])
+
+  const getTitleTextVariant = () => {
+    if (smMatch) return 'h300'
+    else if (lgMatch) return 'h400'
+    else return 'h200'
```
drillprop: I'm not a big fan of not using brackets in long if statements. Could we add them here?

packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -0,0 +1,203 @@
...
+      </TitleAndDescriptionContainer>
+      {isLoading ? (
+        <PillSkeletonLoader />
+      ) : (
+        <PillAndDateContainer>
+          <div>
+            <Pill label={type} size="small" />
+          </div>
+          <DateText variant="t100" secondary>
+            {formatDateTime(new Date(date))}
```
drillprop: image
`formatDateTime` doesn't work well. Look at the day

packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -0,0 +1,203 @@
...
+      type: 'Sale',
+      title: 'closer',
+      description: 'Bedeho sold NFT for ツ 98,6K to Lenorette ',
+      thumnailUri: 'https://atlas-dev.joystream.app/distributor-1/api/v1/assets/21',
+    },
+  ]
+
+  return (
+    <section>
+      {activity.length === 0 ? (
```
drillprop, line 61: If the activity will be undefined we will see error. Let's maybe add `.?`

drillprop (changes requested):
Looks stunning! Found some minor issues. 
Also let's use `nft` as a base branch here. Not sure what to do with dummy data here, but let's keep them for now. I guess we can clean them later.

packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -0,0 +1,203 @@
...
+}
+
+type ActivityItemProps = {
+  date: string
+  type: string
+  title: string
+  description: string
+  thumnailUri: string
+}
+const ActivityItem: React.FC<ActivityItemProps> = ({ date, type, title, description, thumnailUri }) => {
```
drillprop: One last thing. Maybe we could add this particular component to the storybook? It will be easier to test it out.

packages/atlas/src/views/viewer/MemberView/MemberActivity.styles.ts
eldiegod: it was supposed to be part of the css selector below but somehow it ended up lines above, odd

drillprop:
Two things: 
1. Sorry, I missed that before, but I'm getting: 
image
2. I'd rather move `ActivityItem` to another file(you don't need to create a new directory). Maybe it's just me, but standalone `ActivityItem.stories.tsx` looks a little odd to me without designed file `ActivityItem.tsx`.  It gave me an expression that we placed some random file in this directory by mistake.

drillprop (changes requested):
Two things: 
1. Sorry, I missed that before, but I'm getting: 
image
2. I'd rather move `ActivityItem` to another file(you don't need to create a new directory). Maybe it's just me, but standalone `ActivityItem.stories.tsx` looks a little odd to me without designed file `ActivityItem.tsx`.  It gave me an impression that we placed some random file in this directory by mistake.

packages/atlas/src/views/viewer/MemberView/ActivityItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+  DateText,
+  DescriptionSkeletonLoader,
+  PillAndDateContainer,
+  PillSkeletonLoader,
+  Thumbnail,
+  ThumbnailSkeletonLoader,
+  Title,
+  TitleAndDescriptionContainer,
+  TitleSkeletonLoader,
+} from './MemberActivity.styles'
```
drillprop: Could you also move those to `ActivityItem.styles.ts`?

drillprop (changes requested):

drillprop (approved):

packages/atlas/src/views/viewer/MemberView/ActivityItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const isLoading = !date || !type || !title || !thumbnailLoaded
+  return (
+    <ActivityItemContainer loading={isLoading}>
+      {isLoading ? <ThumbnailSkeletonLoader /> : <Thumbnail src={thumnailUri} />}
+      <TitleAndDescriptionContainer>
+        {isLoading ? (
+          <TitleSkeletonLoader />
+        ) : (
+          <div>
```
kdembler: What do we need this `div` for?

packages/atlas/src/views/viewer/MemberView/ActivityItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <Text variant={lgMatch ? 't300' : 't200'} secondary>
+            {description}
+          </Text>
+        )}
+      </TitleAndDescriptionContainer>
+      {isLoading ? (
+        <PillSkeletonLoader />
+      ) : (
+        <PillAndDateContainer>
+          <div>
```
kdembler: Same as above, why do we need a `div` wrapper?

packages/atlas/src/views/viewer/MemberView/ActivityItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+  PillSkeletonLoader,
+  Thumbnail,
+  ThumbnailSkeletonLoader,
+  Title,
+  TitleAndDescriptionContainer,
+  TitleSkeletonLoader,
+} from './ActivityItem.styles'
+
+export type ActivityItemProps = {
+  date: string
```
kdembler: I think this should be `Date` right? Our cache automatically wraps all ISO dates in `Date` type

kdembler (commented):
It seems some of the styles for the Activity item list component on the highest breakpoint have not been applied:
- padding is not changed to 24px
- same for margins between sections of an activity item
- some typography differences

Also, the margins between the image and the title work a bit strange - we use 16px gap from grid but then additionally apply 8px margin on the image to stretch the spacing. Let's make sure all the spacing is defined in a single place, either grid gaps or margins so you don't have to do mental calculations to see what the spacing is

packages/atlas/src/views/viewer/MemberView/ActivityItem.tsx
eldiegod: Was part of a previous layout iteration but not needed anymore

eldiegod:
@kdembler I can't find the text styling different from design you mentioned, I think all else should be there

kdembler (approved):

Issue #1961, opened 2022-01-20
Video NFT: Component - Recommendations


dmtrjsg:
![Screenshot 2022-01-20 at 11.50.08.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/3e6ca001-5f90-47c3-af29-48b8a6f9762a)

[Link to Figma >](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1210%3A77287)

PR #2072, opened 2022-02-06
update video page

### LINKED ISSUE #1960: Video Page: Layout changes
**AC:**

- [ ] Update the description/license/title section
- [ ] Include CTAs at the bottom at the bottom of the page
- [ ] Placeholders for recommendations (#1961)

[Link to Figma >](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1288%3A146214)

rafalpawlow:
closes #1960 

**Acceptance criteria**
- [x] Update the description/license/title section
- [x] Include CTAs at the bottom at the bottom of the page
- [x] Placeholders for recommendations (#1961)

**Design**

https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=0%3A1


packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -166,61 +179,96 @@ export const VideoView: React.FC = () => {
...
+                  )}
+                </LicenseContainer>
+                <CategoryWrapper>
+                  <DescriptionTitle variant="h100">Category</DescriptionTitle>
+                  <Category>
+                    {category?.icon}
+                    <Text variant={mdMatch ? 't300' : 't200'} secondary>
+                      {category?.name}
+                    </Text>
+                  </Category>
```
drillprop: This definitely should be a link that leads to `/category/:id`

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -166,61 +179,96 @@ export const VideoView: React.FC = () => {
...
+                  <SkeletonLoader height={24} width={200} />
+                )}
+              </Meta>
+            </TitleContainer>
+            <ChannelContainer>
+              <ChannelLink followButton id={channelId} textVariant="h300" />
+            </ChannelContainer>
+            <DetailsWrapper>
+              <DescriptionContainer>
+                <DescriptionTitle variant="h100">Description</DescriptionTitle>
```
drillprop: I see the small problem here. When there is no description, you see just a title. It looks really odd on smaller screens: 
image

Could you ask the designer what we should do here? 
Maybe we shouldn't show `Description` when there is none?

packages/atlas/src/components/CategoryLink/CategoryLink.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  background-color: ${({ color }) => `rgba(${hexToRgb(color)}, 0.2) `};
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin-right: ${({ withHandle }) => (withHandle ? sizes(3) : 0)};
+
+  path {
+    fill: ${({ color }) => color};
+  }
+`
```
drillprop, line 37: We have the same element in VideoCategoryCard. Can we just export this and use it here? 
image

packages/atlas/src/views/viewer/VideoView/MoreVideos.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <ChannelLink id={channelId} customTitle={`More from ${channelName}`} textVariant="h400" />
+        ) : (
+          <CategoryLink name={categoryName} id={categoryId} textVariant="h400" />
+        )}
+        <Button to={linkUrl} variant="secondary" icon={<SvgActionChevronR />} iconOnly />
+      </MoreFrom>
+      <LayoutGrid>{gridContent}</LayoutGrid>
+      <SeeMoreButton to={linkUrl} variant="tertiary" icon={<SvgActionChevronR />} iconPlacement="right">
+        See the {type === 'channel' ? 'channel' : 'category page'}
+      </SeeMoreButton>
```
drillprop, line 70: Size is different on smaller screens mobile. You should use `large` button here: 
image

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
rafalpawlow: ~~@Lenorette~~ @toiletgranny We need your advice here

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -61,25 +76,19 @@ export const VideoView: React.FC = () => {
   }, [video, thumbnailUrl])
   const headTags = useHeadTags(video?.title, videoMetaTags)
 
-  const [startTimestamp, setStartTimestamp] = useState<number>()
+  const { startTimestamp, setStartTimestamp } = useVideoStartTimestamp(timestampFromQuery, video?.duration)
```
kdembler: I think we should be fine with including the hook that gets value of `timestampFromQuery` inside this hook as well?

packages/atlas/src/views/viewer/VideoView/MoreVideos.tsx
```diff
@@ -0,0 +1,84 @@
...
+    </MoreVideosContainer>
+  )
+}
+
+type MoreVideosVideoTileProps = {
+  id?: string
+}
+
+const MoreVideosVideoTile: React.FC<MoreVideosVideoTileProps> = ({ id }) => {
```
kdembler: I see this is why the video tile is missing the play button. Do we really need a separate component here, wouldn't regular viewer video tile work?

kdembler (changes requested):
Looks really sexy! Some points:
1. I think we should be little more conservatives on link areas. For example, this full area is a link, I think we should mostly avoid clickable whitespace. The same occurs for license and category links.<br>Screenshot 2022-02-09 at 11 14 06
2. The video tile in More videos section seems to be missing the play icon on hover?<br>Screenshot 2022-02-09 at 11 15 20
3. I noticed data for "More" sections takes quite a while to fetch. I think we need to start paying more attention to amounts of data we fetch. You use a full videos connection but I think we could use `GetBasicVideos` instead (we may need to add some fields there possibly).
4. There are issues when the description section is loading. License header have a skeleton loader, other headers don't. Also the description skeleton loaders overflow its container<br>Screenshot 2022-02-09 at 11 21 36
 5. There is some layout shift when more videos from channel are loading. As far as I can tell, it occurs for a split second when the channel info finishes loading. Managed to capture it on recording:

https://user-images.githubusercontent.com/12646744/153180087-a4254cec-4a47-4cd2-9eac-cd2e585f412e.mov

toiletgranny:
Because @Lenorette is currently off, I'll be picking up reviewing the implementation of her designs from now on.

So great to see the new designs for this page coming to life! Overall, as always, astonishing implementation quality from my perspective, but as there were quite a lot of elements to be implemented, I did find some minor issues here and there.

1. The size of the channel avatar:
   - As is: 32px,
   - To be: 40px,
2. Distance between the channel avatar and the channel name & the number of followers:
   - As is: 12px,
   - To be: 16px,
3. Distance between the "License" and "Category" columns:
   - As is: there's none,
   - To be: 24px,
4. Distance between the license name and the description:
   - As is: there's none,
   - To be: 8px,
5. License name letter case:
   - As is: all caps,
   - To be: sentence case,
6. License name link:
   - As is: it links to the current URL,
   - To be: we should not have any links over there?,
7. Horizontal Video tile component's context menu:
   - As is: there's no context menu button,
   - To be: there should be a context menu button, just like on the vertical version. This might be a component-level issue that I missed then reviewing the implementation a while ago. @drillprop can you confirm or deny?
8. Color of the category icon in the "More from {{category name}}" section:
   - As is: it's super bright,
   - To be: in a color of a category, with no extra brightening, 
9. Missing icon & color in the "Discover videos" CTA card at the bottom of the page:
   - As is: there's a placeholder icon in white color,
   - To be: based on[ Figma designs](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A44367) (please note, we missed this in Figma as well, I fixed it just now).
10. No video description:
    - As is: there's just a "Description" label with a blank space underneath [(example)](https://atlas-j440ni033-joystream.vercel.app/video/10226),
    - To be: the "Description" label should not be displayed if there is no description provided with the video. (please note, we missed this in Figma as well),
11. "Show more" button on SM and lower breakpoints:
    - As is: implementation-wise it's all good 👌 
    - To be: I think the variant of the button used in designs was a bit excessive, seeing it live, so I'd recommend using a white text version of this component. Designs are already updated, so you can [check it out over here](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1211%3A87591).

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
toiletgranny: Exactly what I recommended in my comment below, thank you!

rafalpawlow:
@toiletgranny I fixed most of your comments. However I have 2 questions:
> Distance between the license name and the description:
>
I'm not sure exactly which distance you're talking about
Zrzut ekranu 2022-02-9 o 18 10 24

> License name letter case:
>
I'm really not sure about this. Is this correct? Btw, upper case comes from query node, this is not intentional styled.
Zrzut ekranu 2022-02-9 o 18 23 43

toiletgranny:
@rafalpawlow
1. Right, this one can be tricky to find if you're not testing on the Main Testnet. [Here's an example video](https://atlas-o27u5xzeo-joystream.vercel.app/video/9621), make sure to switch to the Main Testnet. Not only there's no distance between the license name ("Custom") and its description ("Free to consume and share..."), but they are also presented in one line.
2. Oh, it comes from the query node! Sorry, I thought we're capitalizing this text on the front-end. If that's the case, let's not apply any `text-transform` here, the "Cc0" is a good example why. Excuse my ignorance, but is adjusting license names on the query node something within our reach? So that we have "Custom" instead of "CUSTOM", without manipulating that with CSS?

drillprop:
> 7. Horizontal Video tile component's context menu:
>    
>    * As is: there's no context menu button,
>    * To be: there should be a context menu button, just like on the vertical version. This might be a component-level issue that I missed then reviewing the implementation a while ago. @drillprop can you confirm or deny?

Just answering - it wasn't. We didn't miss that. Rafał just used tile without context menu items. Anyway, he already fixed that. 😉

kdembler:
@toiletgranny @rafalpawlow I'm a bit lost what we're discussing here when it comes to the licenses text case. The strings of licenses are actually defined in Atlas, query node only sends us the ID so we're free to manipulate this. However, I'm not sure what you are saying should be different, I think we only have `CUSTOM` capitalized, all the other values are abbreviations. So should we just update that one value to `Custom` and we're good? (@rafalpawlow this is defined in `src/data/knownLicenses.json`

packages/atlas/src/views/viewer/VideoView/MoreVideos.tsx
```diff
@@ -0,0 +1,64 @@
...
+  categoryId?: string
+  categoryName?: string | null
+  type: 'channel' | 'category'
+}
+
+const NUMBER_OF_VIDEOS = 6
+
+export const MoreVideos: React.FC<MoreVideosProps> = ({ channelId, channelName, categoryId, categoryName, type }) => {
+  const where = type === 'channel' ? { channel: { id_eq: channelId } } : { category: { id_eq: categoryId } }
+  const { videos = [], loading } = useBasicVideos({ where, limit: NUMBER_OF_VIDEOS })
```
kdembler: I think we may want to include sorting here (or in the query directly) to make sure we get the latest videos

packages/atlas/src/api/queries/videos.graphql
```diff
@@ -91,8 +91,8 @@ query GetVideos($offset: Int, $limit: Int, $where: VideoWhereInput, $orderBy: Vi
   }
 }
 
-query GetBasicVideos($where: VideoWhereInput) {
-  videos(where: $where) {
+query GetBasicVideos($limit: Int, $where: VideoWhereInput) {
+  videos(limit: $limit, where: $where) {
     ...BasicVideoFields
```
kdembler, line 114: This currently doesn't provide all the fields required by the `VideoTile`. You can see that this has the effect that once this query finishes, separate `GetVideo` is fired for each video to get missing details (check out network tab). We will need to add some missing fields here to make sure subsequent requests are not required

packages/atlas/src/views/viewer/VideoView/MoreVideos.tsx
```diff
@@ -17,17 +17,25 @@ type MoreVideosProps = {
   channelName?: string | null
   categoryId?: string
   categoryName?: string | null
+  videoId?: string
   type: 'channel' | 'category'
 }
 
-const NUMBER_OF_VIDEOS = 6
+const NUMBER_OF_VIDEOS = 7
```
kdembler: I suggest we keep it as 6 and use for both placeholder and slicing (you currently have 6 explicit there) and just add one for fetched items with maybe small comment to describe why we get additional one

toiletgranny (approved):
> So should we just update that one value to `Custom` and we're good?

@kdembler If all the other license types are abbreviations, then yes, let's do exactly that. Sorry for the confusion!

> Just answering - it wasn't. We didn't miss that. Rafał just used tile without context menu items.

@drillprop Ah, ok! Glad to know it was a simple thing. On another note, if using a video tile component without a context menu is a simple thing to do, perhaps we should make it... more difficult? You know, just to make sure we don't run into similar problems in the future? 

Other than this, @rafalpawlow, I reviewed your changes and it's all good from my perspective!❤️

toiletgranny:
Ok, hold on! I just stumbled upon something. It seems like designs don't actually provide any guidelines on how videos in the "More from {{category name}}" section should be sorted, and judging by the implementation I'm assuming currently they're sorted based on their upload date, right? 

I guess that's fine _for the time being_, but I'm wondering if sorting by popularity would be... even finer? If feasible? @kdembler @rafalpawlow 

Again, _for the time being_, because ideally, way in the future, we have a super sophisticated sorting based on a combination of similarity and rating and the user's taste and so on... 🤤

rafalpawlow:
@toiletgranny actually we have top 10 week and top 10 month. They base on number of views in given period. But thb, `More from` doesn't sound to me like `Most popular from` 😛. I don't think there are any obstacles to implementing this, right @kdembler ?

kdembler:
This would be tricky for channel videos, we don't keep track of most popular videos from a given channel at the moment, let's not overcomplicate this for the time being 🙏

kdembler (approved):

PR #2076, opened 2022-02-07
add channel's nft tab

### LINKED ISSUE #1851: NFT Tab
**Context:** Channel NFTs View

**✅  AC**
- [ ] Add new Tab "NFTs" to the channel tabs. Shows all Videos minted as NFTs by given channel
- [ ] Sorting same as for NFTs owned by Member (Default: Newest first; alt: Oldest first by date minted)
- [ ] Filtering
- [ ] Empty State = Same as Empty search results but diff copy ->  show same image as for zero search, but change copy to "This channel does not have any NFTs issued yet."
Screenshot 2022-01-04 at 15 14 34

**Designs**

[Link to NFT Tab Figma](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=510%3A31946) 

**⛔  OOS**

- [ ] Sorting by price
Screenshot 2022-01-04 at 16 11 39

rafalpawlow:
closes #1851

https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=510%3A31946


eldiegod (changes requested):
I'm not sure if this task was meant to have dummy data but probably it should for now at least. 

Are these NFTs items going to be paginated, infinite scrolled, not clear in the design.

If you switch tabs with filters open the filter bar stays open.

rafalpawlow:
@eldiegod this is not specified in task but I think both things you mention will be covered when API for nft will be ready

eldiegod (changes requested):
@rafalpawlow clicking on filters button now quickly opens and closes the filters bar, happened to me in my PR #2085 and I solved it but can't recall how exactly.

Mobile view also has some grid issues.

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -287,38 +306,54 @@ export const ChannelView: React.FC = () => {
...
-                value={!isSearching ? sortVideosBy : 0}
-                placeholder={isSearching ? 'Best match' : undefined}
-                items={!isSearching ? SORT_OPTIONS : []}
-                onChange={!isSearching ? handleSorting : undefined}
-              />
-            </SortContainer>
-          )}
-        </TabsContainer>
+            {currentTab === 'Videos' ||
+              (currentTab === 'NFTs' && (
```
eldiegod: This condition with parenthesis on `currentTab === 'NFTs'` makes it so video tab has no sorting or search, removing parenthesis should fix it

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -216,15 +228,22 @@ export const ChannelView: React.FC = () => {
           />
         </PaginationContainer>
       </>
+    ) : currentTab === 'NFTs' ? (
+      <VideoSection className={transitions.names.slide}>
+        {isSearching && <EmptyFallback title={`No videos matching "${searchQuery}" query found`} variant="small" />}
```
eldiegod: Copy here is talking about videos which is wrong

eldiegod (changes requested):
Few details left

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
rafalpawlow: This come from the design https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=581%3A43083

What copy should we have here?

eldiegod: I guess "no NFTs"..., @toiletgranny ?

rafalpawlow: I've added your suggestion, sounds reasonable. But let's wait for @toiletgranny

toiletgranny: Hey guys, sorry for my late response, I've been off last week. You're correct — the copy in designs was wrong. I've changed it to be `No NFTs matching “Border Collie” query found`.

drillprop:
Summary of changes
- created ChannelViewTabs which will keeps all the tab components in a separate file
- created ChannelNfts tab, and populated it with real data
- added GetNfts query and refactored GetNft query
- did some general refactoring and cleaning in ChannelView(search, pagination)

Out of scope: 
- searching on the NFTs tab
- filtering? I mean, I added category filters, but not for auctions. Query node has some issues and it's really hard to do it without some crazy workarounds, so I decided to omit this. Read more: https://github.com/Joystream/joystream/issues/3342, https://github.com/Joystream/hydra/issues/476

Also FYI - we won't be using infinite scrolling here, just regular pagination  - consulted this with the designer.

packages/atlas/src/api/client/cache.ts
```diff
@@ -124,6 +125,15 @@ const queryCacheFields: CachePolicyFields<keyof Query> = {
...
   },
+  ownedNftByUniqueInput: (existing, { toReference, args }) => {
+    return (
+      existing ||
+      toReference({
+        __typename: 'OwnedNft',
+        id: args?.where.id,
+      })
+    )
+  },
```
kdembler, line 136: Why do we need that?

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -1,10 +1,82 @@
...
+  const idIn = videosConnectionData?.videosConnection.edges.map((edge) => edge.node.id) || []
+
+  const { data: nftsData, loading: nftsLoading } = useGetNftsQuery({
+    variables: {
+      where: {
+        id_in: idIn,
+      },
+    },
+    skip: !idIn.length || videosConnectionLoading,
+  })
```
kdembler: This is temporary, yes? I'd assume once we can filter NFTs by owner, we will get rid of the videos query? If that's the case, could you add `TODO` comment?

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -1,10 +1,82 @@
...
+    loading: videosConnectionLoading,
+    ...rest
+  } = useGetVideosConnectionQuery({
+    variables: {
+      orderBy: opts?.sortNftsBy,
+      where: {
+        channel: { id_eq: channelId },
+        nft: {
+          metadata_contains: '',
+        },
```
kdembler: This is meant to say "NFT exists"?

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -28,35 +25,39 @@ import { transitions } from '@/styles'
...
   TitleContainer,
   TitleSection,
   TitleSkeletonLoader,
-  VideoSection,
 } from './ChannelView.styles'
+import { ChannelAbout, ChannelNfts, ChannelVideos } from './ChannelViewTabs'
+
+export const TABS = ['Videos', 'NFTs', 'Information'] as const
+export const INITIAL_FIRST = 50
+export const INITIAL_TILES_PER_ROW = 2
```
kdembler: Why do we use 2 as default now?

packages/atlas/src/views/viewer/ChannelView/ChannelViewTabs/ChannelNfts.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const placeholderItems = Array.from(
+    { length: loading ? tilesPerPage - (paginatedNfts ? paginatedNfts.length : 0) : 0 },
+    () => ({
+      id: undefined,
+    })
+  )
+
+  if (error) {
+    return <ViewErrorFallback />
+  }
```
kdembler: Maybe better to have this next to the normal `return`?

packages/atlas/src/views/viewer/ChannelView/ChannelViewTabs/ChannelVideos.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { VideoFieldsFragment, VideoOrderByInput } from '@/api/queries'
+import { EmptyFallback } from '@/components/EmptyFallback'
+import { Grid } from '@/components/Grid'
+import { Pagination } from '@/components/Pagination'
+import { ViewErrorFallback } from '@/components/ViewErrorFallback'
+import { VideoTileViewer } from '@/components/_video/VideoTileViewer'
+import { transitions } from '@/styles'
+import { SentryLogger } from '@/utils/logs'
+
+import { INITIAL_FIRST } from '..'
```
kdembler: This creates circular dependencies I think?

kdembler (commented):
Minor stuff, looks good

packages/atlas/src/api/client/cache.ts
drillprop: To reduce the number of requests to Orion. Without this chunk of code, I was fetching all the NFTs using `useChannelNfts`, and then fetching them again for every tile using `useNft`(which is called inside NftTileViewer).

packages/atlas/src/api/hooks/nfts.ts
drillprop: yes, temporary - I'll add a comment

packages/atlas/src/api/hooks/nfts.ts
drillprop: yeah, tested it out and it seems to be working. I'll add here a small comment as well.

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
drillprop: oops, this is probably some oversight.

packages/atlas/src/api/client/cache.ts
kdembler: Okay, makes sense!

kdembler (approved):
LGTM! Let's get approval from @eldiegod as well since review was requested

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
```diff
@@ -128,40 +121,59 @@ export const StyledButton = styled(Button)`
   width: 100%;
 `
 
-export const PaginationContainer = styled.div`
-  padding-top: ${sizes(6)};
+export const TabsWrapper = styled.div<{ isFiltersOpen: boolean }>`
+  z-index: 99;
```
eldiegod:
```suggestion
  z-index: ${zIndex.transactionBar};
```

Use this so it doesn't mess with the member dropdown

https://github.com/Joystream/atlas/pull/2200/files

eldiegod (approved):
Looks good to me! just one last detail

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
kdembler: @drillprop this need to be applied

Issue #1752, opened 2021-12-06
View all NFTs owned as member


dmtrjsg:
AC:
- [ ] [NFTs Owned Tab](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=919%3A130033)
- [ ] [Sorting](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=919%3A130033)
- [ ] Filtering

PR #2085, opened 2022-02-08
Member nft tab

eldiegod:
fixes: #1752


packages/atlas/src/views/viewer/MemberView/MemberView.styles.ts
```diff
@@ -31,23 +31,38 @@ export const StyledTabs = styled(Tabs)`
 
 export const TabsContainer = styled.div`
   display: grid;
-  margin-bottom: ${sizes(8)};
+  padding-top: ${sizes(8)};
   gap: ${sizes(2)};
-  grid-template: 'tabs tabs tabs' 1fr 'search search search' auto 'sort sort sort' auto / 1fr 1fr;
+  grid-template: 'tabs tabs tabs' 1fr 'search search search' auto 'sort sort sort' auto / 1fr 1fr 'filter filter filter' auto / 1fr 1fr;
```
drillprop: Something is very wrong on smaller screens: 
image

Second thing - can we just do this without declaring grid-areas? `  grid-template: 1fr auto auto / 1fr 1fr;`? should work the same in this case, right?

packages/atlas/src/views/viewer/MemberView/MemberView.styles.ts
```diff
@@ -2,7 +2,7 @@ import styled from '@emotion/styled'
 
 import { MembershipInfo } from '@/components/MembershipInfo'
 import { Tabs } from '@/components/Tabs'
-import { media, oldColors, sizes } from '@/styles'
+import { cVar, media, oldColors, sizes, transitions } from '@/styles'
```
drillprop: let's get rid of `oldColors` in this file.

drillprop (changes requested):

packages/atlas/src/views/viewer/MemberView/MemberView.styles.ts
eldiegod: Grid areas is just easier due the mobile grid layout difference where tabs take 2 columns, I think they are fine it was just not properly indented/wrong

packages/atlas/src/views/viewer/MemberView/MemberView.styles.ts
```diff
@@ -13,41 +13,46 @@ export const NotFoundMemberContainer = styled.div`
...
+    'tabs tabs' auto
+    'sort filter' auto / 1fr auto;
+  background-color: #000;
 
   ${media.sm} {
     align-items: center;
-    border-bottom: solid 1px ${oldColors.gray[700]};
-    gap: ${sizes(8)};
-    grid-template: 1fr / 1fr 160px;
+    box-shadow: ${cVar('effectDividersBottom')};
```
drillprop, line 36: Just one thing left.
image

drillprop (changes requested):

drillprop (approved):

drillprop (approved):

PR #2097, opened 2022-02-10
Introduce logic for estimating block time

### LINKED ISSUE #1867: Introduce logic for estimating block time
Prepare logic for estimating hour:minute:seconds time from a number of blocks.

drillprop:
Fix #1867


packages/atlas/src/hooks/useBlockTimeEstimation.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { useJoystream } from '@/providers/joystream'
+
+const ESTIMATED_BLOCK_TIME = 6000
+
+export const useBlockTimeEstimation = () => {
+  const { joystream, proxyCallback } = useJoystream()
+
+  const [currentBlock, setCurrentBlock] = useState(0)
+  const [timeofTheLastBlock, setTimeOfTheLastBlock] = useState(0)
```
kdembler: Small typo here:
```suggestion
  const [timeOfTheLastblack, setTimeOfTheLastBlock] = useState(0)
```

packages/atlas/src/hooks/useBlockTimeEstimation.ts
```diff
@@ -0,0 +1,58 @@
...
+          setTimeOfTheLastBlock(Date.now())
+        })
+      )
+    }
+    init()
+
+    return unsubscribe
+  }, [joystream, proxyCallback])
+
+  const convertBlockToDate = (block: number) => {
```
kdembler: I'm wondering, could we make those functions constant? Currently those functions will get updated on every block. If we keep the block number and date inside some ref instead of state, we could make both those convert functions static, so they wouldn't trigger re-renders

packages/atlas/src/hooks/useBlockTimeEstimation.ts
drillprop: You also did a typo 😅  `timeOfTheLastblack`

packages/atlas/src/hooks/useBlockTimeEstimation.ts
drillprop: hmm, I moved basically everything to refs. Let me know what you think.

kdembler: I would move function to regular `useCallback`, other than that, looks good

kdembler (approved):

Issue #1391, opened 2021-09-22
Don't overly truncate channel names


kdembler:
As (correctly IMO) reported by a community member, we overly truncate channel titles when those are being displayed in a various contexts. Example of this:

![](https://cdn.discordapp.com/attachments/859752198166282280/888054931850932264/atlasbug2.jpg)

![](https://cdn.discordapp.com/attachments/859752198166282280/888055114177327154/atlasbug3.jpg)

I think the design should provide examples of longer channel titles for both of those cases and then we should improve the implementation.

toiletgranny:
@kdembler I went through the app and, using Chrome inspect tools, "stress-tested" all places in which a channel name is being displayed. Keeping in mind that there's a global characters limit for a channel name (40 chars), my recommendations are as follows:

### Channel page / Membership profile
1. AS IS: The channel name itself has the following CSS properties:  `white-space: nowrap`, `text-overflow: ellipsis`, `overflow: hidden`, `max-width: 600px`.
2. TO BE: Channel name should be allowed to wrap if too lengthy, no max-width is necessary. I updated Figma designs for both the [Channel](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=510%3A31266) and the [Membership profile](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=919%3A130033) pages so that when inspecting **Figma layer constraints**, you can see the recommendation is to let the channel title be displayed with no constraints whatsoever.

![CleanShot 2022-02-03 at 11 10 03](https://user-images.githubusercontent.com/35307309/152322400-e75e4c57-4e02-49c4-85c3-479241b8c6fa.png)

### Browse channels page
1. AS IS: Under the "Discover new channels" section, channel names themselves have the following CSS properties: `white-space: nowrap`, `text-overflow: ellipsis`, `overflow: hidden`, `max-width: 200px`.
2. TO BE: Channel names should be allowed to wrap if too lengthy, no max-width is necessary. Please find and inspect **Figma layer constraints** in [historic designs for this page](https://www.figma.com/file/Vk2Z4QOiVa5bB6q3cBIG5J/Joystream-Atlas?node-id=8408%3A3917) to see that this has been the assumption from the beginning. 
![CleanShot 2022-02-03 at 11 14 12](https://user-images.githubusercontent.com/35307309/152323097-95ae4a83-872c-4a81-a371-ebf90acf8270.png)

---

[...]

PR #2098, opened 2022-02-10
don't truncate channels name

lcoenengithub:
**What is this?**

This removes `no-wrap` constraints around channel title per #1391 

**Illustration**

image

### DIFF
--- a/packages/atlas/src/components/_channel/ChannelWithVideos/ChannelWithVideos.styles.ts
+++ b/packages/atlas/src/components/_channel/ChannelWithVideos/ChannelWithVideos.styles.ts
@@ -27,9 +27,6 @@ export const InfoWrapper = styled.div`
 
 export const ChannelTitle = styled(Text)`
   max-width: 200px;
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
 `
 
 export const ChannelFollows = styled(Text)`
--- a/packages/atlas/src/components/_channel/ChannelWithVideos/ChannelWithVideos.styles.ts
+++ b/packages/atlas/src/components/_channel/ChannelWithVideos/ChannelWithVideos.styles.ts
@@ -25,10 +25,6 @@ export const InfoWrapper = styled.div`
   align-items: start;
 `
 
-export const ChannelTitle = styled(Text)`
-  max-width: 200px;
-`
-
 export const ChannelFollows = styled(Text)`
   margin-top: ${sizes(1)};
   max-width: 200px;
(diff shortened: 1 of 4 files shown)


rafalpawlow:
@lcoenengithub base branch is wrong, should be `dev`

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
```diff
@@ -58,9 +58,6 @@ export const TitleContainer = styled.div`
 
 export const Title = styled(Text)`
   margin-bottom: 0;
-  white-space: nowrap;
-  text-overflow: ellipsis;
-  overflow: hidden;
   max-width: 600px;
```
drillprop: The designer said that there should be no max-width here, so you should remove this line. You'll need to update also the container grid to make this work if I'm not mistaken.

packages/atlas/src/components/_channel/ChannelWithVideos/ChannelWithVideos.styles.ts
```diff
@@ -27,9 +27,6 @@ export const InfoWrapper = styled.div`
 
 export const ChannelTitle = styled(Text)`
   max-width: 200px;
```
drillprop: same as above

drillprop (changes requested):

packages/atlas/src/views/viewer/ChannelView/ChannelView.styles.ts
lcoenengithub: Thanks! Fixed

kdembler (changes requested):
Channel name on channel page doesn't wrap and overflows the layout:
Screenshot 2022-02-21 at 09 11 56

lcoenengithub:
@kdembler I modified the grid layout to make it wrap again. Thanks for your feedback!

image

kdembler (approved):

drillprop (changes requested):
...

Edit: made a mistake - ignore this comment for now

packages/atlas/src/components/_channel/ChannelWithVideos/ChannelWithVideos.styles.ts
```diff
@@ -6,6 +6,10 @@ import { Text } from '@/components/Text'
 import { Button } from '@/components/_buttons/Button'
 import { sizes } from '@/styles'
 
+export const Title = styled(Text)`
+  margin-right: ${sizes(6)};
+`
```
drillprop: Last thing. Why do we need this margin? I think we could safely remove this.

drillprop (changes requested):

drillprop (approved):

PR #2109, opened 2022-02-11
refactor VideoWorkspace and VideoForm

kdembler:
**Summary of changes:**

1. Cleaned up `VideoWorkspace` of leftovers from tabs
2. Renamed `VideoWorkspaceForm` to `VideoForm` so we can add `NFTForm` later
3. Cleaned up form state - now all state is kept inside the `VideoForm` component
4. All transaction handling and action bar moved to main `VideoWorkspace`
5. Added `VideoWorkspaceFormStatus` so that form can communicate its status back to `VideoWorkspace`
6. Moved bigger pieces of logic to separate hooks to make the files more readable (900 lines is a bit much)

**Left to fix:**

- [ ] If you open `/video-workspace` url directly, it breaks


packages/atlas/src/components/_inputs/MultiFileSelect/MultiFileSelect.tsx
```diff
@@ -21,15 +22,15 @@ type InputFile = {
   title?: string
 }
 
-export type VideoInputMetadata = {
+export type MediaInputMetadata = {
```
kdembler, line 25: Changed this type since it was conflicting with the `VideoInputMetadata` from `joystream-lib`. I think going forward we need to use `media` to distinct when we are talking about video as entity vs it's media file

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
```diff
@@ -1,46 +1,36 @@
...
-    assetsCache &&
-    (assetsCache.thumbnail.cropContentId || assetsCache.video.contentId)
-  )
-
-  // transaction management
-  const [thumbnailHashPromise, setThumbnailHashPromise] = useState<Promise<string> | null>(null)
-  const [videoHashPromise, setVideoHashPromise] = useState<Promise<string> | null>(null)
-  const [dialogState, setDialogState] = useState<VideoWorkspaceState>('unset')
-
-  useEffect(() => {
```
rafalpawlow, line 38: This was the fix for opening form directly from `/video-workspace`. It doesn't open now because on initial render `isWorkspaceOpen === true` and `CSSTransition` doesn't see change.

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -0,0 +1,489 @@
...
+}))
+
+type VideoFormProps = {
+  onSubmit: (data: VideoFormData) => void
+  setFormStatus: Dispatch<SetStateAction<VideoWorkspaceFormStatus | null>>
+}
+
+export const VideoForm: React.FC<VideoFormProps> = React.memo(({ onSubmit, setFormStatus }) => {
+  const [moreSettingsVisible, setMoreSettingsVisible] = useState(false)
+  const [cachedEditedVideoId, setCachedEditedVideoId] = useState<string>('')
```
drillprop:
```suggestion
  const [cachedEditedVideoId, setCachedEditedVideoId] = useState('')
```

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
```diff
@@ -99,22 +72,94 @@ export const VideoWorkspace: React.FC = React.memo(() => {
...
+        icon: <SvgControlsCancel width={16} height={16} />,
+      }}
+      draftBadge={{
+        visible: !isEdit,
+        text: mdMatch ? 'Drafts are saved automatically' : 'Saving drafts',
+        tooltip: {
+          text: 'Drafts system can only store video metadata. Selected files (video, thumbnail) will not be saved as part of the draft.',
+        },
+      }}
+      fullWidth={true}
```
drillprop: This `fullWidth` doesn't do anything. This can be safely removed here and in ActionBarProps

drillprop (commented):
found only two minor things, works pretty well! I'm really glad that you reduced lines of code per file, this was really hard to manage!

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
kdembler: I think it's something else, when you open `/video-workspace`, the route changes to `/videos`, that wouldn't be caused by lack of this effect I think

kdembler: Okay that route change happens on prod as well :P

kdembler: All right, fixed by adding `appear` prop on `CSSTransition`

drillprop (approved):

rafalpawlow (approved):

PR #2115, opened 2022-02-14
remove duplicate imports from same module

kdembler:


.eslintrc.js
```diff
@@ -22,27 +22,18 @@ module.exports = {
     },
   },
   rules: {
-    // taken care of by typescript
-    'react/prop-types': 'off',
+    // force using Logger object
+    'no-console': ['warn'],
+    'no-duplicate-imports': ['warn'],
```
kdembler, line 27: Only this rule was added, other than just grouped them together

rafalpawlow (approved):

PR #2135, opened 2022-02-16
trim titles&descriptions on submit

kdembler:
I've noticed a couple videos published with a trailing space in the title, I suspect adding an empty space at the start/end will never be user's intention so let's save them the trouble

### DIFF
--- a/packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
+++ b/packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
@@ -230,8 +230,8 @@ export const CreateEditChannelView: React.FC<CreateEditChannelViewProps> = ({ ne
     setIsWorkspaceOpen(false)
 
     const metadata: ChannelInputMetadata = {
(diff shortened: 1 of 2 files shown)


rafalpawlow:
I'm thinking maybe we should trim all additional whitespaces by adding something like `.replace(/\s+/g,' ')` to avoid any glitches?

drillprop:
I think Rafał made a good point here. For example, In our video form you actually can set a title like this: 
image
But on the other hand, you won't see these whitespaces on the video tile, because HTML doesn't render them. 
image
So maybe it's a good idea to get rid of them when the user is submitting. I'm also curious how youtube handle such cases

drillprop:
> I'm also curious how youtube handle such cases

Okay, Adam checked this for me. It seems that they are only using `trim()` i.e they are removing white spaces from the start and from the end. You can actually create a movie with a title:
```
long                  title
```
 but on youtube, it will be displayed as `long title`

kdembler:
I don't know about removing inner spaces, I would just trim it for now to stay on the safe side and we can revisit in the future

rafalpawlow (approved):

PR #2148, opened 2022-02-17
1595: squished video tiles

lcoenengithub:
**What is this?**

This is restyling numbered tiles inside the "popular" views. Responsivity of this screen have been restyled as well.

**Illustration**

image

**Figma**

https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A127191
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5178%3A259806

### DIFF
--- a/packages/atlas/src/components/Carousel/Carousel.styles.ts
+++ b/packages/atlas/src/components/Carousel/Carousel.styles.ts
@@ -43,9 +43,6 @@ export const Arrow = styled(IconButton)`
 `
 
 export const GliderContainer = styled.div`
-  padding-left: ${sizes(2)};
-  padding-top: ${sizes(2)};
-
   /* hides scrollbar on firefox */
   scrollbar-width: none;
 `
@@ -62,7 +59,6 @@ export const Track = styled.div`
 
 export const Dots = styled.div`
   padding: ${sizes(3.5)} 0;
-  margin-top: ${sizes(12)};
   display: none;
 
   ${media.md} {
--- a/packages/atlas/src/components/Carousel/Carousel.styles.ts
+++ b/packages/atlas/src/components/Carousel/Carousel.styles.ts
@@ -45,6 +45,7 @@ export const Arrow = styled(IconButton)`
 export const GliderContainer = styled.div`
   /* hides scrollbar on firefox */
   scrollbar-width: none;
+  overflow: visible;
 `
 
 export const Track = styled.div`
--- a/packages/atlas/src/components/RankingNumberTile/RankingNumber.stories.tsx
+++ b/packages/atlas/src/components/RankingNumberTile/RankingNumber.stories.tsx
@@ -1,6 +1,12 @@
+import { createApolloClient } from '@/api'
+import { ApolloProvider } from '@apollo/client'
 import styled from '@emotion/styled'
 import { Meta, Story } from '@storybook/react'
 import React from 'react'
+import { Router } from 'react-router'
+import { BrowserRouter } from 'react-router-dom'
+import { VideoTile } from '../_video/VideoTile'
+import { VideoTileViewer } from '../_video/VideoTileViewer'
 
 import { RankingNumberTile } from './RankingNumberTile'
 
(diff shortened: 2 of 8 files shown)


toiletgranny:
Hey @lcoenengithub, I can't seem to be able to find the [Number tile component](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5164%3A153589) in Storybook in order to review the implementation over there. If it's not there, can I ask you to add it and send me the link?

lcoenengithub:
> Hey @lcoenengithub, I can't seem to be able to find the [Number tile component](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5164%3A153589) in Storybook in order to review the implementation over there. If it's not there, can I ask you to add it and send me the link?

Hey @toiletgranny ! Thanks, I added it under `RankingNumberTile`, link: https://atlas-storybook-rar90yqu3-joystream.vercel.app/?path=/story/other-rankingnumbertile--ranking-number

kdembler:
@lcoenengithub I think we would really benefit for slightly bigger story, including real video tiles, and more items that can be scrolled inside a carousel

lcoenengithub:
> @lcoenengithub I think we would really benefit for slightly bigger story, including real video tiles, and more items that can be scrolled inside a carousel

I added a story under `VideoGallery` :)

toiletgranny:
@lcoenengithub thanks! Will review this later today!

toiletgranny (changes requested):
## Number tile
1. The width of the number wrapper is off. As is: `~80px`, to be: `64px` (up until the `SM` breakpoint the numbers are: `56px` and `40px`).
2. Border color of the number is incorrect. As is: `--color-core-neutral-500`, to be: `--color-border`.
3. Incorrect `letter-spacing` set on the number. As is: `letter-spacing: -0.17em;`, to be: none. For some reason, this also affects the vertical position of the number. Removing the letter-spacing makes the vertical position match the designs in 100%, so this is yet another reason to remove this property.
4. Regarding the `doubleDigits` property, the `line-height` is currently set to `0`. To be: `85%`. 
5. Regarding the `doubleDigits` property, numbers should be centered vertically relative to the parent container's height. Currently, they don't seem to be, e.g., when I type the number `233` numbers are going downwards.
6. The number wrapper should have the `overflow` property set to `hidden` so that numbers don't _bleed out_ to the right side of the wrapper. 
CleanShot 2022-02-28 at 16 51 45@2x


## Carousels (VideoGallery)
1. The number of video tiles matches the design guidelines, so it's all good here! 
2. The structure of the `RankingNumberTile` component seems to be rather complicated and difficult to analyze. At certain breakpoints, numbered tiles start to overlap, e.g., when the viewport width is `768px`. 
CleanShot 2022-02-28 at 16 34 08
3. This is perhaps partially related to the above: The width of the video tile components nested inside numbered tiles doesn't match designs. E.g., on the `425px` breakpoint, the width of the video tile component is `240px`, [whereas it should be](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5178%3A262594) `285px`. 
4. The scrollable contents of carousels are clipped at the edges, not allowing the numbered tile components to reach the edge of the viewport. In the designs, the contents of the carousels span beautifully from one edge of the viewport to another. Can we have that in the implementation since we're revisiting this component? Ideally, the inner padding in the carousel matches the globally applied one, which is `16px` up until the MD breakpoint, and then `32px` going forward. 
5. There appears to be subtle top & bottom padding applied on the carousel's content. Why is that? Designs do not have any additional vertical padding.
7. Carousels with channel card components inside them have incorrect heights above the `SM` viewport. There's plenty of empty space below.
8. (Optional) Carousels with channel card components inside don't seem to follow the responsiveness rules defined in Figma. Some of the breakpoint changes seem to be kicking in too soon and are not synchronized. Can we investigate that?

toiletgranny:
@lcoenengithub I just realized that implementing this component could be a tricky process in being able to spot all the design nuances that are going on "under the hood" just by looking at the raw Figma designs. Since this is the first time we're working together, I'd love to jump on a quick Tandem call with you and walk you through the things I listed above and perhaps show you a few cool Figma tricks you can use when inspecting designs next time. Let me know if you're interested! :)

toiletgranny:
Thanks for the updates! I can see some of the things I listed above being addressed (thank you!), but I'm afraid this might have caused some other parts of the implementation to break. So below, I listed everything I found.

## Number tile

1. On the smaller breakpoint variant, I've noticed two issues:
CleanShot 2022-03-03 at 15 59 36
   - The number seems to be centered horizontally, whereas it should always be left-aligned (always meaning no matter which variant or breakpoint it is),
   - The Video tile component seems to be overlapping the number wrapper.
2. The number wrapper is clearly changing its width depending on other external factors (?), whereas its width should always be constant (a reminder: `40px` and then `64px` from the `SM` breakpoints going forward).
https://user-images.githubusercontent.com/35307309/156591706-701db0f8-e9b5-4218-8bac-9093ab7f248e.mp4
3. There are still some issues with the `doubleDigits` property being set to `true` both on the smaller and the larger breakpoint. The digits are overly spaced apart, and they're not centered vertically as they're supposed to be. Please see the below table for easy comparison: 
   | Breakpoint | Implementation (as is) | Design (to be) |
   |---|---|---|
   | SM | image | image |
   | XXS | image | image |
4. I've just noticed that because of how `-webkit-text-stroke` works, the outline around the digits is added both inside and outside letters, making them look thinner than they should. The way I see it, there are two ways out of here. The first solution would be to use an [8-axis text-shadow](https://stackoverflow.com/a/47511171) instead of `-webkit-text-stroke`, another one would be to simply accommodate for the reduced thickness by increasing the `font-weight` property from `600` to `700`. Let me know if that would work!

## Carousels (VideoGallery)
I wasn't able to review the updated implementation of this, because the scrolling inside of the carousel stopped working. Now, the whole page is scrolled horizontally, which — of course — wasn't our goal. Would you let me know once this is addressed so that I can take a look? :pray:
https://user-images.githubusercontent.com/35307309/156597154-5ef6a12d-2fba-4523-950c-079250b5825b.mp4

kdembler:
Closing in favor of #2219, will implement changes there

PR #2149, opened 2022-02-17
get rid of aspect-ratio

rafalpawlow:


packages/atlas/src/views/viewer/VideoView/VideoView.styles.ts
```diff
@@ -62,6 +62,8 @@ const getPlayerContainerCinematicStyles = ({ cinematicView }: CinematicView) =>
         }
       `
     : css`
+        padding-top: 56.3%;
```
kdembler: For consistency with other usages:

```suggestion
        padding-top: 56.25%;
```

kdembler (approved):

Issue #1822, opened 2021-12-20
Create NFT widget component


Lenorette:
Link to final design in Figma
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1242%3A109695

kdembler:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Atlas-600edbb8799cc700103bc2d7/issues/Joystream/atlas/1822?planning-poker) with ZenHub @drillprop @eldiegod @lcoenengithub @rafalpawlow

PR #2162, opened 2022-02-18
NFT Widget for video page

eldiegod:
Fix: #1822


packages/atlas/src/components/_nft/NftWidget/NftWidget.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+      control: { type: 'select', options: ['small', 'medium'] },
+    },
+    status: {
+      control: { type: 'select', options: ['iddle', 'buy-now', 'auction'] },
+    },
+    nftState: { table: { disable: true } },
+  },
+  args: {
+    ownerHandle: 'ye 🖤',
+    ownerAvatarUri: 'https://54.172.37.177.nip.io/distributor-1/api/v1/assets/3',
```
kdembler: Maybe we use some other URL? This one may become dead once we switch to newer playground

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+  canWithdrawBid?: boolean
+  auctionPlannedEndDate?: Date
+}
+
+export type NftWidgetProps = {
+  ownerHandle?: string
+  ownerAvatarUri?: string
+  isOwner: boolean
+  nftState:
+    | { status: 'iddle'; lastPrice?: number; lastTransactionDate?: Date }
```
kdembler:
```suggestion
    | { status: 'idle'; lastPrice?: number; lastTransactionDate?: Date }
```

This will need update in other places as well

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+    | { status: 'buy-now'; buyNowPrice: number }
+    | Auction
+}
+
+// TODO: Update Joy icon with the right variant once it is exported correctly
+// TODO: remove dummy dollar values
```
kdembler: The logic for converting has been added already, do you think we could include this? https://github.com/Joystream/atlas/pull/2170

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+          label="Buy now"
+          content={
+            <>
+              <JoyTokenIcon />
+              <Text variant={contentTextVariant}>{formatNumberShort(buyNowPrice)}</Text>
+            </>
+          }
+          secondaryText="$9,629.25"
+        />
+      ) : null
```
kdembler: Could we extract this out of the component?

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+  nftState:
+    | { status: 'iddle'; lastPrice?: number; lastTransactionDate?: Date }
+    | { status: 'buy-now'; buyNowPrice: number }
+    | Auction
+}
+
+// TODO: Update Joy icon with the right variant once it is exported correctly
+// TODO: remove dummy dollar values
+export const NftWidget: React.FC<NftWidgetProps> = ({ ownerHandle, isOwner, nftState, ownerAvatarUri }) => {
+  const [containerRef, { width = 281 }] = useMeasure()
```
kdembler: Why do we default to 281?

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+  const size: Size = width > 280 ? 'medium' : 'small'
```
kdembler: Could we place this in a named constant? (can be in this file)

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+  const timeInSeconds = Math.max(0, Math.round((time.getTime() - new Date().getTime()) / 1000))
+  const lessThanAMinuteLeft: boolean = timeInSeconds < 60
+
+  useEffect(() => {
+    const interval = setInterval(forceUpdate, 1000)
+    return () => {
+      clearInterval(interval)
+    }
+  }, [forceUpdate])
```
kdembler: Maybe we just keep something like `currentTime` state and update it in this interval? Then we don't need the `rerender` logic. WDYT?

kdembler (changes requested):

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
```diff
@@ -0,0 +1,136 @@
...
+  padding: ${sizes(6)};
+
+  &[data-size='small'] {
+    grid-template-columns: minmax(0, 1fr);
+    padding: ${sizes(4)};
+    gap: ${sizes(4)};
+  }
+`
+
+export const NFTOwnerContainer = styled.div<SizeProps>`
```
drillprop: This part has a slighlty different height in the design: 
image

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+                {!nftState.auctionPlannedEndDate && !!nftState.topBid && (
+                  <GridItem colSpan={buttonColumnSpan}>
+                    <Button fullWidth size={buttonSize}>
+                      Review and accept bid
+                    </Button>
+                  </GridItem>
+                )}
+              </>
+            ) : nftState.buyNowPrice ? (
+              <>
```
drillprop: We don't need this fragment here

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+                        <Button fullWidth size={buttonSize} variant="destructive-secondary">
+                          Withdraw a bid
+                        </Button>
+                      </GridItem>
+                    )}
+                  </ButtonGrid>
+                </GridItem>
+              </>
+            ) : (
+              <>
```
drillprop: and here

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,320 @@
...
+      <SecondaryText as="p" variant="t100" secondary data-size={size}>
+        {secondaryText}
+      </SecondaryText>
+    </InfoItemContainer>
+  )
+}
+
+const NFTTimerItem: React.FC<{ size: Size; time: Date }> = ({ size, time }) => {
+  const [, rerender] = React.useState({})
+  const forceUpdate = React.useCallback(() => rerender({}), [])
```
drillprop: this looks very hacky to me. Why we can just set a time left every second inside useEffect?

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
eldiegod: This is the smallest value that will trigger the medium variant, basically making the medium variant the default.

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
eldiegod: I liked it a bit more coz I didn't have to care about handling the first value of the useState, made it into a useState now.

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
eldiegod: It's because I used our button text variant instead of a non-navigable Text component there for the owners name, any objection @toiletgranny?

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
eldiegod: I don't think there is anything too unusual going on here, I liked it a bit more coz I didn't have to care about handling the first value of the useState; made it into a useState now.

toiletgranny:
Hey @eldiegod, 2 quick questions before I begin reviewing the implementation: 
1. Where can I preview the NFT widget in Atlas, on the video page? Is it possible? 
2. On Storybook, I'm not able to expand the "History" section. Is it because it's not implemented yet?

eldiegod:
> 2\. m not able to expand the "History" section. Is it because it's not implemented y

Hey @toiletgranny, only in storybook so far, and yes that will be implemented in a different pull request

toiletgranny (changes requested):
## Adjustments
1. **As is/To be:** The way an NFT owner is rendered in the header part of the NFT widget does not match designs in terms of color, text style, and the underline visible upon hovering. I think I understand your reasoning behind changing this, but can I ask to stick to the designs here this time around?
2. **As is/To be:** There's an incorrect joy token icon displayed next to the NFT price. We've recently added missing variants of joy token icons to our Figma API file, so now it is possible to use [this icon](https://www.figma.com/file/1uxugLps6oooI38PGmdrn9/Figma-API--%3E-Storybook?node-id=637%3A380).
**To be:**
4. **As is/To be:** Wherever an NFT is mentioned in the copy, it should always be spelled out with capital letters, e.g. `Start sale of this nft` -> `Start sale of this NFT`, or `This nft is owned by` -> `This NFT is owned by`.
7. **As is:** There's a blank piece of space left underneath the timer.
**To be:** There should be no unused, blank space underneath the timer or any other element in the NFT widget.
CleanShot 2022-03-01 at 10 15 23@2x
8. **As is:** Colons in the timer are of the same color as the digits.
**To be:** Colons should be of the color `--color-text`, whereas digits should be `--color-text-strong`.
9. **As is:** The padding applied on the "History" section is currently set to `28px 24px`.
**To be:** The padding should be `24px` on all sides.
10. **As is:** The downwards chevron in the "History" section is implemented as a raw icon.
**To be:** This should be an icon button.

## Other
1. I wasn't able to see what the NFT widget looks like if no bid is placed yet, both from the viewer's as well as the owner's perspective. Can I ask for guidance on how should I change the properties in Storybook in order to get to this? These are the variants I'm looking for: 
image
2. As an NFT owner, I can't seem to trigger the "Remove from sale" button to appear in the NFT widget. Again, can I ask for guidance on how should I change the properties in Storybook in order to get to this?
3. I know that the background color of the NFT widget in designs was set to `--color-core-neutral-900`, but it was a mistake that we overlooked. Can I ask for this to be changed to `--color-background-muted`, so that we always use our semantic color design tokens whenever possible? I will update the designs shortly to reflect this.

packages/atlas/src/components/_nft/NftWidget/NftInfoItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [timeInSeconds, setTimeInSeconds] = useState<number>(getTimeInSeconds(time))
+
+  const lessThanAMinuteLeft: boolean = timeInSeconds < 60
+
+  useEffect(() => {
+    const interval = setInterval(() => setTimeInSeconds(getTimeInSeconds(time)), 1000)
+    return () => {
+      clearInterval(interval)
+    }
+  }, [time])
```
kdembler: We just merged https://github.com/Joystream/atlas/pull/2204, if you rebase you should reuse that logic

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
```diff
@@ -0,0 +1,85 @@
...
+    gap: ${sizes(2)};
+  }
+
+  &[data-two-columns='true'] {
+    grid-template-columns: 1fr 1fr;
+  }
+`
+
+export const JoyTokenIcon = styled(SvgActionJoyToken)`
+  /* /TODO: unnable to style svg like design */
```
kdembler: New icons have been added so we should be able to use final ones

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -0,0 +1,261 @@
...
+    | { status: 'idle'; lastPrice?: number; lastTransactionDate?: Date }
+    | { status: 'buy-now'; buyNowPrice: number }
+    | Auction
+}
+
+const SMALL_VARIANT_MAXIMUM_SIZE = 280
+
+// TODO: Update Joy icon with the right variant once it is exported correctly
+export const NftWidget: React.FC<NftWidgetProps> = ({ ownerHandle, isOwner, nftState, ownerAvatarUri }) => {
+  const [containerRef, { width = 281 }] = useMeasure()
```
kdembler: Could we use `SMALL_VARIANT_MAXIMUM_SIZE + 1` instead?

packages/atlas/src/hooks/useDeepMemo.ts
```diff
@@ -0,0 +1,21 @@
+import { equal as isEqual } from '@wry/equality'
```
kdembler: I think `lodash` exports `isEqual` as well, can we use that, since it's already our direct dependency?

kdembler (commented):
Couple last comments from my side, other than that looks great! Still need visual updates for Adam's points

eldiegod:
@toiletgranny 

> As is: The downwards chevron in the "History" section is implemented as a raw icon.
To be: This should be an icon button.

Noted I will implement the History related changes in the upcoming widget history PR

> As an NFT owner, I can't seem to trigger the "Remove from sale" button to appear in the NFT widget. Again, can I ask for guidance on how should I change the properties in Storybook in order to get to this?

For this you will have to do the next:
- set it to auction
- set topBid to `0`
- type `null` on auctionPlannedEndDate and click outside so it applies
- click on set Boolean for isOwner and set it to true

> As is/To be: There's an incorrect joy token icon displayed next to the NFT price. We've recently added missing variants of joy token icons to our Figma API file, so now it is possible to use [this icon](https://www.figma.com/file/1uxugLps6oooI38PGmdrn9/Figma-API--%3E-Storybook?node-id=637%3A380).

There seems to be something wrong with the new exported Icons, could you check on your side @kdembler? specifically size and the gradients.

>I wasn't able to see what the NFT widget looks like if no bid is placed yet, both from the viewer's as well as the owner's perspective. Can I ask for guidance on how should I change the properties in Storybook in order to get to this? These are the variants I'm looking for:

After some discussion, I removed the implementation of those states because they were impossible to reach since a startingPrice for auctions is mandatory thus always showing `Minimum bid` if there are no bids.


I think that should be it for clarifications and the rest should be implemented

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
eldiegod: Mentioned in a comment below, I tried them but they were not working size-wise and gradient wise

kdembler:
I think @toiletgranny makes a good point here regarding cancelling the auction.

These are the scenarios when the owner should be able to cancel the auction:
1. It's an open auction, you can cancel those at any time
2. It's an english auction and either:
a) no bids have been made yet
b) the auction has expired with no bids made

I think we don't handle 1. fully, it lets you cancel only if there are no bids, there is not constraint like this for open auction.

And it seems we're not handling 2. at all. 2.b) will be included in later designs (see https://github.com/Joystream/atlas/issues/2195), but 2.a) should be possible right away

eldiegod:
> designs (see #2195), but 2.a) should be possible right away

just implemented the states except for 2.b for now!

kdembler (approved):

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
toiletgranny: @eldiegod they should be fixed by now!

toiletgranny:
@kdembler @eldiegod Regarding missing auction canceling scenarios, I'll be addressing all of that (and more) in designs as part of https://github.com/Joystream/atlas/issues/2195 but also https://github.com/Joystream/atlas/issues/2182, so I agree it probably doesn't make sense to focus on this too much right now, in this PR.

@eldiegod regarding the implementation, now it's almost flawless, thank you very much! The "almost" comes from an incorrect size of the JOY token icon used in the "Medium" size of the NFT widget. Currently, it's `16px`, should be `24px`. Again, please note that this size should be applied only to the JOY token icons used in the "Medium" size of the NFT widget. 

Once the above is addressed, it's good to go from my perspective.

eldiegod:
> @kdembler @eldiegod Regarding missing auction canceling scenarios, I'll be addressing all of that (and more) in designs as part of #2195 but also #2182, so I agree it probably doesn't make sense to focus on this too much right now, in this PR.
> 
> @eldiegod regarding the implementation, now it's almost flawless, thank you very much! The "almost" comes from an incorrect size of the JOY token icon used in the "Medium" size of the NFT widget. Currently, it's `16px`, should be `24px`. Again, please note that this size should be applied only to the JOY token icons used in the "Medium" size of the NFT widget.
> 
> Once the above is addressed, it's good to go from my perspective.

good catch; expanded history PR and widget implemented in video page soon to come

PR #2170, opened 2022-02-21
feat: add tjoy price logic

lcoenengithub:
**What is this?**

This adds logic to grasp tJoy price in `useJoystream`.


packages/atlas/src/providers/joystream/provider.tsx
```diff
@@ -35,6 +39,14 @@ export const JoystreamProvider: React.FC = ({ children }) => {
 
   const proxyCallback = useCallback(<T extends object>(callback: T) => proxy(callback), [])
 
+  useEffect(() => {
```
kdembler, line 42: Your `useEffect` doesn't provide a dependency array, so it will run every time the provider renders

packages/atlas/src/providers/joystream/provider.tsx
```diff
@@ -35,6 +39,14 @@ export const JoystreamProvider: React.FC = ({ children }) => {
...
+    const getPrice = async () => {
+      const data = await fetch(JOYSTREAM_STATUS_URL)
```
kdembler: This needs error handling

packages/atlas/src/providers/joystream/provider.tsx
```diff
@@ -87,7 +99,7 @@ export const JoystreamProvider: React.FC = ({ children }) => {
   }, [activeAccountId, accounts, initialized])
 
   return (
-    <JoystreamContext.Provider value={{ joystream: initialized ? joystream.current : undefined, proxyCallback }}>
+    <JoystreamContext.Provider value={{ joystream: initialized ? joystream.current : undefined, proxyCallback, price }}>
```
kdembler, line 106: This exposes the price but I would expect some hook to convert between tJOY's and formatted dollar string. Please take a look at what Bartosz did here: https://github.com/Joystream/atlas/pull/2097 So some hook for conversion would be useful.

kdembler (changes requested):
Some points apart from the comment:
1. The target for this PR needs to be `nft` branch
2. It would be useful to add some playground to test this out, you can also refer to mentioned Bartosz's PR to see how to add a playground

packages/atlas/src/providers/joystream/provider.tsx
lcoenengithub: Thanks! That's fixed

packages/atlas/src/providers/joystream/provider.tsx
lcoenengithub: I added `useTokenPrice` below

lcoenengithub:
@kdembler I added a playground page (`playground/TJoyPrice`) and a hook for conversion! Thanks :)

kdembler (changes requested):
Your hook only returns a fixed string with the price. This isn't really useful though, we don't display the price itself anywhere in the UI. What we do display, for example, is a price of an NFT, shown both in tJOY and in dollars. The goal of this task is to create logic that will be able to **convert** between those values. If you need a concrete example, take a look at the figma link I posted in the issue. So I would expect some function that accepts tJOY amount and returns value in dollars, possibly formatted. Again, please take a look at what was done in #2097, I expect this to end up pretty similar.

lcoenengithub:
> Your hook only returns a fixed string with the price. This isn't really useful though, we don't display the price itself anywhere in the UI. What we do display, for example, is a price of an NFT, shown both in tJOY and in dollars. The goal of this task is to create logic that will be able to **convert** between those values. If you need a concrete example, take a look at the figma link I posted in the issue. So I would expect some function that accepts tJOY amount and returns value in dollars, possibly formatted. Again, please take a look at what was done in #2097, I expect this to end up pretty similar.

Hi @kdembler, thanks, I couldn't locate any code around currency conversion in #2097, but I added a converter: 
image
It's using `convertToUSD` and `convertToTJoy` from `useTokenPrice`

kdembler (approved):

PR #2179, opened 2022-02-23
add MemberComboBox component

### LINKED ISSUE #1921: Issue NFT with Whitelisting of selected members
**Context**

NFT can be offered to a controlled / selected group of Joystream members. For this, Publisher needs to look up the members by member ID and add to the list on Step 2 and review on Step 3.

**AC:**
- [ ] On Step 2, the list is editable, on Step 3 the list is not editable
- [ ] Margins are slightly different on both pages @KubaMikolajczyk is that intentional ❓ 

**Designs**

Link to user stories in Figma:
 [Add user to whitelist](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1000%3A29308)
[Remove user from whitelist](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1003%3A32597)

drillprop:
Fix #1921
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=969%3A17302
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=919%3A21501
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=921%3A17135

Summary of changes: 
- created MemberBadge component
- created TextFieldWithDropdown component
- created WhiteListTextField component(which should handle all the logic behind whitelisting members)
- added playground and stories

Since we don't have the form yet, I just created all the components. We cannot add this to issue NFT flow yet, this is out of the scope of this PR.


packages/atlas/src/components/_inputs/TextFieldWithDropdown/TextFieldWithDropdown.tsx
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+  const getTextFieldBottomEdgePosition = () => {
+    if (!textFieldRef.current || !textFieldWithDropdownWrapperRef.current) {
+      return
+    }
+    const { y: wrapperY } = textFieldWithDropdownWrapperRef.current.getBoundingClientRect()
+    const { y: inputY, height: inputHeight } = textFieldRef.current.getBoundingClientRect()
+    return inputY - wrapperY + inputHeight
+  }
```
eldiegod: Why is this calculation necessary?

packages/atlas/src/components/_inputs/WhiteListTextField/WhiteListTextField.tsx
```diff
@@ -0,0 +1,120 @@
...
+        const selectedMembersLookup = selectedMembers ? createLookup(selectedMembers) : {}
+        const filteredMembers = memberships
+          .map(({ handle, id, metadata: { avatar } }) => ({
+            handle,
+            avatarUri: avatar?.__typename === 'AvatarUri' ? avatar.avatarUri : undefined,
+            id,
+          }))
+          .filter((member) => {
+            return !selectedMembersLookup[member.id]
+          })
```
eldiegod: Since you are filtering in the debounced fn when a user is selected and type the same name again quickly it appears again for a brief second to be selected. Maybe we should filter outside of it

packages/atlas/src/components/_inputs/WhiteListTextField/WhiteListTextField.tsx
```diff
@@ -0,0 +1,120 @@
...
+        setMembers(filteredMembers)
+      } catch (error) {
+        SentryLogger.error('Failed to fetch memberships', 'WhiteListTextField', error)
+        setIsError(true)
+      }
+    }, 500)
```
eldiegod: can we make this 250ms? 500ms feels like waiting for an eternity to me 👨‍💻

eldiegod (changes requested):
Looks very nice, added some tiny comments 👌

packages/atlas/src/components/_inputs/TextFieldWithDropdown/TextFieldWithDropdown.tsx
drillprop: This will position the dropdown list below the input box. This is needed because if we add helper text to the text field, the list will be moved down below this helper text.
image
I couldn't find a cleaner solution for this. I did a small refactoring and made sure that this will be not called when the helper text is not set. I provided a comment as well.

packages/atlas/src/components/_inputs/WhiteListTextField/WhiteListTextField.tsx
drillprop: very good point 👍🏻 , fixed

eldiegod (approved):

packages/atlas/src/components/_inputs/TextFieldWithDropdown/TextFieldWithDropdown.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { ListWrapper, TextFieldWithDropdownWrapper } from './TextFieldWithDropdown.styles'
+
+import { TextField, TextFieldProps } from '../TextField'
+
+export type TextFieldWithDropdownProps<ItemType = unknown> = {
+  items?: (ListItemProps & ItemType)[]
+  loading?: boolean
+  onSelectedItemChange?: (item?: ListItemProps & ItemType) => void
+  onInputValueChange?: ((item?: string) => void) | ((item?: string) => Promise<void>)
```
kdembler: I think this would be simpler:
```suggestion
  onInputValueChange?: (item?: string) => void | Promise<void>
```

packages/atlas/src/components/_inputs/TextFieldWithDropdown/TextFieldWithDropdown.tsx
```diff
@@ -0,0 +1,106 @@
...
+  resetOnSelect?: boolean
+  notFoundNode?: ListItemProps | null
+} & Omit<TextFieldProps, 'charactersCount'>
+
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
+export const TextFieldWithDropdown = <ItemType extends unknown>(props: TextFieldWithDropdownProps<ItemType>) => {
```
kdembler: I think you use `extends unknown` here for the same reason we do in `Select.tsx`? If so, could you add the same comment here so it's clear to someone who reads it?

kdembler: Maybe we could use a simpler name for this component? `Combobox` seems to be actually quite popular, what do you think about this name?

packages/atlas/src/components/_inputs/WhiteListTextField/WhiteListTextField.tsx
```diff
@@ -0,0 +1,118 @@
...
+          handle,
+          avatarUri: avatar?.__typename === 'AvatarUri' ? avatar.avatarUri : undefined,
+          id,
+        }))
+        setMembers(members)
+      } catch (error) {
+        SentryLogger.error('Failed to fetch memberships', 'WhiteListTextField', error)
+        setIsError(true)
+      }
+    }, 250)
```
kdembler: I think we could increase this slightly. I opened devtools in the playground and started typing quite normally and during that a bunch of requests were fired. Maybe we could try 500?

packages/atlas/src/components/_inputs/WhiteListTextField/WhiteListTextField.tsx
```diff
@@ -0,0 +1,118 @@
...
+          data: { memberships },
+        } = await client.query<GetMembershipsQuery, GetMembershipsQueryVariables>({
+          query: GetMembershipsDocument,
+          variables: { where: { handle_startsWith: val } },
+        })
+        setIsLoading(false)
+
+        const members = memberships.map(({ handle, id, metadata: { avatar } }) => ({
+          handle,
+          avatarUri: avatar?.__typename === 'AvatarUri' ? avatar.avatarUri : undefined,
```
kdembler: This will not work if the avatar is uploaded through the storage system right? I think we should handle that case as well to be future-proof

kdembler (commented):
Very minor comments as well, works like a charm overall

packages/atlas/src/components/_inputs/TextFieldWithDropdown/TextFieldWithDropdown.tsx
drillprop: Agree, updated

packages/atlas/src/components/_inputs/WhiteListTextField/WhiteListTextField.tsx
```diff
@@ -28,6 +34,16 @@ export const WhiteListTextField: React.FC<WhiteListTextFieldProps> = ({ selected
...
   const client = useApolloClient()
   const [isError, setIsError] = useState(false)
+  const resolveAsset = useRawAssetResolver()
+
+  const getAvatarUri = (avatar?: BasicMembershipFieldsFragment['metadata']['avatar']) => {
+    if (avatar?.__typename === 'AvatarUri') {
+      return avatar.avatarUri
+    }
+    if (avatar?.__typename === 'AvatarObject' && avatar.avatarObject?.id) {
+      return resolveAsset(avatar.avatarObject?.id)?.url
```
kdembler: If you take a look at the raw asset resolver, you will notice that it will return an asset URL only if the asset has only been resolved, it doesn't schedule resolution for unknown assets. It'd be best to just pass either full member or full avatar object and use `useMemberAvatar`

drillprop: Should work now

kdembler (approved):

PR #2183, opened 2022-02-23
feat: add snackbar on URL copy

lcoenengithub:
**What is this?**

This adds a snackbar message at video URL copy

**Illustration**
Peek 2022-02-23 16-52 - video


packages/atlas/src/components/_video/VideoTileViewer/VideoTileViewer.tsx
```diff
@@ -8,6 +8,7 @@ import { useVideoTileSharedLogic } from '@/hooks/useVideoTileSharedLogic'
 import { copyToClipboard } from '@/utils/browser'
 import { formatDurationShort } from '@/utils/time'
 
+import { useSnackbar } from '../../../providers/snackbars/snackbar'
```
drillprop: you should use: 
```
import { useSnackbar } from '@/providers/snackbars'
```

drillprop (changes requested):
Couple things:
- we should use `dev` as a base branch.
- snackbar should have an icon: https://www.figma.com/file/LF7LOMhOA46bRcYHWlw1Bq/My-videos-page?node-id=792%3A31187 
- snackbar should be displayed also when you copy the URL in VideoTilePublisher. You missed this.
- I would add a snack bar also in MembershipInfo. We're copying the address there.  Maybe we could create a hook(`useCopyToClipboard`) that will be handling copying to the clipboard and displaying the snack bar. @kdembler wdyt?

drillprop:
Also, I have one more request. Maybe it's just oversight, but please add the link to the issue and to the design(if it's available) in your PRs.

kdembler:
> I would add a snack bar also in MembershipInfo. We're copying the address there. Maybe we could create a hook(useCopyToClipboard) that will be handling copying to the clipboard and displaying the snack bar. @kdembler wdyt?

Very good point, agreed

lcoenengithub:
@drillprop @kdembler thanks! I applied your points.

packages/atlas/src/providers/clipboard/hooks.ts
```diff
@@ -0,0 +1,12 @@
+import { useSnackbar } from '@/providers/snackbars'
+import { copyToClipboard } from '@/utils/browser'
+
+export const useClipboard = () => {
+  const { displaySnackbar } = useSnackbar()
+  return {
+    copyToClipboard: (text: string) => {
+      copyToClipboard(text)
+      displaySnackbar({ title: 'Video URL have been copied', iconType: 'info' })
+    },
```
drillprop: Firstly, I think, this should be "Video URL copied to clipboard". 

Secondly, when you click the copy button on the member page, you're copying the account address, not the video URL.
image
You need to add one more parameter to `copyToClipboard` function to display the correct information. Something like: 
```tsx
copyToClipboard: (text: string, snackbarText: string) => {
      copyToClipboard(text)
      displaySnackbar({ title: snackbarText, iconType: 'info' })
    },
```

drillprop (changes requested):
Works well 👍🏻 , just one thing left

packages/atlas/src/providers/clipboard/hooks.ts
lcoenengithub: Thanks @drillprops, I fixed it (I added some args with default values).

packages/atlas/src/providers/clipboard/hooks.ts
```diff
@@ -0,0 +1,12 @@
+import { SnackbarIconType, useSnackbar } from '@/providers/snackbars'
+import { copyToClipboard } from '@/utils/browser'
+
+export const useClipboard = () => {
+  const { displaySnackbar } = useSnackbar()
+  return {
+    copyToClipboard: (text: string, title = 'Video URL copied to clipboard', iconType: SnackbarIconType = 'info') => {
```
drillprop: hmm, I don't think we need to add `iconType` as a parameter - this should be always an informational snack bar.  I don't see the reason why we would change that. 

I also don't agree that we should set the default title.

packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
```diff
@@ -46,6 +46,7 @@ export const MembershipInfo: React.FC<MembershipInfoProps> = ({
   editable,
   className,
 }) => {
+  const { copyToClipboard } = useClipboard()
```
drillprop, line 49: we need to pass suitable snackbarText on line 57 i.e: `Account address copied to clipboard`

drillprop (changes requested):

drillprop:
@lcoenengithub I'm confused,  you requested my review, but I don't see any new commits or comments.  That was a misclick or did you forget to push your changes?

lcoenengithub:
> @lcoenengithub I'm confused, you requested my review, but I don't see any new commits or comments. That was a misclick or did you forget to push your changes?

Hey! Sorry, I forgot to push indeed :)

packages/atlas/src/components/MembershipInfo/MembershipInfo.tsx
```diff
@@ -46,14 +46,15 @@ export const MembershipInfo: React.FC<MembershipInfoProps> = ({
...
   const [copyButtonClicked, setCopyButtonClicked] = useState(false)
   const smMatch = useMediaMatch('sm')
 
   const handleCopyAddress = () => {
     if (!address) {
       return
     }
-    copyToClipboard(address)
+    copyToClipboard(address, 'Video URL copied to clipboard')
```
drillprop:
```suggestion
    copyToClipboard(address, 'Account address copied to clipboard')
```

drillprop (changes requested):

packages/atlas/src/providers/clipboard/hooks.ts
```diff
@@ -0,0 +1,12 @@
+import { useSnackbar } from '@/providers/snackbars'
+import { copyToClipboard } from '@/utils/browser'
+
+export const useClipboard = () => {
```
kdembler: This isn't a provider though, right? We keep only context providers in `src/providers`, I think this hook makes more sense in `src/hooks`

lcoenengithub: Good point!

lcoenengithub: Good point, I moved it there!

drillprop (approved):

packages/atlas/src/hooks/clipboard/hooks.ts
```diff
@@ -0,0 +1,12 @@
+import { useSnackbar } from '@/providers/snackbars'
```
kdembler: It's in `hooks/` now but it doesn't match how we name files in that directory, please take a look at other hooks, we should name appropriately and remove unnecessary directory

kdembler (approved):

PR #2189, opened 2022-02-25
Viewer NFT tile

### LINKED ISSUE #2175: Use QN data for NFT Tile component
Currently our NFT Tile component only accepts data via custom props. We should create some additional wrapper component that will accept just a video ID and it will get the data from the cache and pass all necessary data to NFT Tile. We can call it `ViewerNftTile` or something else. This should mirror the behaviour of `ViewerVideoTile`

drillprop:
fix #2175


packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { useAsset } from '@/providers/assets'
+
+import { NftTile, NftTileProps } from '../NftTile'
+
+type NftTileViewerProps = {
+  nftId?: string
+}
+
+export const NftTileViewer: React.FC<NftTileViewerProps> = ({ nftId }) => {
+  const { nft, loading } = useNft(nftId || '')
```
kdembler, line 17: We should probably add `skip` if `nftId` is not set

kdembler (commented):
Haven't found anything else other than this small comment, good job!

packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
drillprop: We already have this inside the `useNft` hook

kdembler: I wrote it, should have known lol

kdembler (approved):
Approve from me, let's also get one from Rafał since you requested a review

rafalpawlow (changes requested):
We should update storybook, it doesn't work properly now

packages/atlas/src/components/NftTile/NftTile.stories.tsx
```diff
@@ -20,16 +19,14 @@ export default {
...
-    auction: { control: { type: 'select', options: ['none', 'minBid', 'topBid', 'waiting'] }, defaultValue: 'none' },
-    role: { control: { type: 'radio', options: ['owner', 'viewer'] } },
-    timeLeft: { type: 'number' },
   },
   args: {
-    buyNow: false,
     role: 'owner',
+    title: 'Did An Alternate Reality Game Gone Wrong Predict QAnon?',
     auction: 'none',
     bid: 1234,
```
rafalpawlow:
```suggestion

```

rafalpawlow (changes requested):

packages/atlas/src/components/NftTile/NftTile.tsx
```diff
@@ -12,85 +12,65 @@ import { Container } from './NftTile.styles'
...
   duration?: number | null
   views?: number | null
-  bid: number
-  minBid?: number
-  topBid?: number
-  timeLeft?: number
+  buyNowPrice?: number | null
+  minBid?: number | null
+  topBid?: number | null
+  timeleft?: number
```
rafalpawlow: Why did you change it from camel case?

rafalpawlow (changes requested):
Auction counter, when you type `timeleft` doesn't count down anymore, is it intentional?

packages/atlas/src/components/NftTile/NftTile.tsx
drillprop: oh - I don't know I just thought that `timeleft` is an existing word 😆 . I'll fix this in a sec. 

Regarding counting - yes this was intentional, I move counting logic to NftTileViewer, because the `auction` prop relies on time left. You could see this here https://github.com/Joystream/atlas/pull/2189/files#diff-88aede35f9de0cbd66bbf341f5187f3c6c21bcea7abd030ab61299132e809cf7R100

rafalpawlow (approved):

Issue #2167, opened 2022-02-20
Update the video player's initial idle screen


toiletgranny:
Based on the comments from Robert and @kdembler, the goal here is to adjust the video player's initial idle screen so that the dark overly is not overly obstructing the video's thumbnail from view (eg. when embedded), at the same time making sure that the play button remains visible in most scenarios. 

![Untitled2](https://user-images.githubusercontent.com/35307309/154834528-a6ba1ec6-8e98-40f3-8ec0-d727cc1f51d4.jpg)

toiletgranny:
@kdembler The updated player component is ready to be reviewed. I added an "Idle" state to [the "Video player" component](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A76840) in Figma, and together with it — [a dedicated "Play button" component](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=216%3A76115) with subtle changes to its background and icon size upon hovering and pressing.

A simple prototype for this [can be found here](https://www.figma.com/proto/Pf31tuYpozYmpq163U2ho8/Web-Components?page-id=110%3A18273&node-id=5943%3A183057&viewport=418%2C48%2C0.37&scaling=min-zoom&starting-point-node-id=5943%3A183057&show-proto-sidebar=1), and the flow description [is here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5943%3A183059).

After quick explorations, I ended up using a semitransparent button, as it's turned out it's a good balance of visibility and not being too intrusive when placed against a busy, "designed" video thumbnail.

![image](https://user-images.githubusercontent.com/35307309/154986623-24ea5630-2a3c-4494-a329-b2b7928a84dc.png)

kdembler:
Fantastic, thanks Adam!

PR #2190, opened 2022-02-25
update big play button overlay

kdembler:
Resolves https://github.com/Joystream/atlas/issues/2167


drillprop (changes requested):
Generally, looks good to me, but what about the compact view? 
In the design, there is no big play button in the idle state
image

kdembler:
I don't want to increase the scope of what initially was a hotfix for the onboarding experience. I create a separate issue for this: https://github.com/Joystream/atlas/issues/2193

drillprop (approved):

Issue #2159, opened 2022-02-18
`Play in Joystream` in VideoTile context menu has no effect


kdembler:
Like in the title, this option in publisher video tile doesn't do anything
<img width="301" alt="Screenshot 2022-02-18 at 12 19 57" src="https://user-images.githubusercontent.com/12646744/154673205-563cc2e0-873b-4cba-90f1-1323e4a4a1cf.png">

PR #2197, opened 2022-02-28
fix: open in joystream bug

lcoenengithub:
**What is this?**

This fixes a bug when clicking "Open in joystream" in the publisher view

**Illustration**

Peek 2022-02-28 13-20

**Issue:**
#2159


packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
```diff
@@ -130,6 +131,10 @@ export const VideoTilePublisher: React.FC<VideoTilePublisherProps> = React.memo(
         },
       ]
 
+      const onOpenInTabClick = () => {
+        videoHref && navigate(videoHref || '')
```
kdembler: Let's open this in a new tab. You can use `openInNewTab` from `@/utils/browser`

kdembler (changes requested):

kdembler (approved):

PR #2201, opened 2022-02-28
Fix issues with upload status in VideoTilePublisher

### LINKED ISSUE #2108: Upload status on video tile incorrect
**Steps to reproduce:**
1. Upload new video
2. Wait for all assets to upload
3. In the same session, modify the thumbnail of this new video
4. When the updated thumbnail is uploading, the video tile has wrong status, showing both success and uploading at the same time

**Expected behavior:**
VideoTile goes back to regular uploading state when updated image is uploading

**Video:**

https://user-images.githubusercontent.com/12646744/153619728-6261c91b-2cb2-4b6d-8e7e-3db39bded978.mov

drillprop:
Fix #2108

### DIFF
--- a/packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
+++ b/packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
@@ -71,7 +71,7 @@ export const VideoTilePublisher: React.FC<VideoTilePublisherProps> = React.memo(
     const isUnlisted = video?.isPublic === false
 
     const getSlots = useCallback<() => undefined | SlotsObject>(() => {
-      if ((isUploading && !hasAssetUploadFailed) || loading) {
+      if (!hasAssetUploadFailed || loading) {
         return
       }
       const slots: SlotsObject = {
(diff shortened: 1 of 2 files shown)


kdembler (commented):
Seems to fix the mentioned issue and also additional problems, so great job. I noticed one additional problem though, not sure if it was there before or introduced by your changes. Here are the steps to reproduce:
1. I published a new video
2. I refresh the app before any of the assets finished uploading
3. After I refresh I got the `Upload failed` tile and 2 files were reported missing in uploads (all good)
4. I re-uploaded the thumbnail
5. The tile was in a normal state even though it should still be in `Upload failed` since video file was still missing

Could you also fix this case?

drillprop:
> noticed one additional problem though, not sure if it was there before or introduced by your changes

Hmm, good catch.  I'm pretty sure that I introduced this here and I think I know, why this is happening. I'll try to find a solution.

drillprop:
> Seems to fix the mentioned issue and also additional problems, so great job. I noticed one additional problem though, not sure if it was there before or introduced by your changes. Here are the steps to reproduce:
> 
> 1. I published a new video
> 2. I refresh the app before any of the assets finished uploading
> 3. After I refresh I got the `Upload failed` tile and 2 files were reported missing in uploads (all good)
> 4. I re-uploaded the thumbnail
> 5. The tile was in a normal state even though it should still be in `Upload failed` since video file was still missing
> 
> Could you also fix this case?

This should be fixed. 

Unfortunately, there is still one problem left. And the same happens on `dev`. If you reupload a broken thumbnail, the new thumbnail isn't set and you see infinite loading on the VideoTile:

image

The problem of course disappears when you refresh the page. 
I did some investigation, and found that `thumbnailPhotoUrl` is `undefined` after reuploading, but didn't figure out why.

Don't want to increase the scope of this PR much more. I'll create an issue with the steps on how to reproduce the bug. @kdembler wdyt?

drillprop:
Should work now. I created a follow-up here https://github.com/Joystream/atlas/issues/2208.

kdembler (approved):

Issue #2118, opened 2022-02-15
Use single ResizeObserver library


kdembler:
We use both `react-use-measure` and `use-resize-observer`, let's get rid of one

PR #2203, opened 2022-03-01
chore: get rid of useMeasure

lcoenengithub:
**What is this?**

This uses `useResizeObserver` instead of `useMeasure` to lean dependencies list.

**Issue**

#2118

### DIFF
--- a/packages/atlas/src/components/_inputs/TextField/TextField.tsx
+++ b/packages/atlas/src/components/_inputs/TextField/TextField.tsx
@@ -43,11 +43,6 @@ const TextFieldComponent: React.ForwardRefRenderFunction<HTMLInputElement, TextF
   },
   ref
 ) => {
-  /*
-  const [nodeLeftRef, nodeLeftBounds] = useMeasure()
-  const [nodeRightRef, nodeRightBounds] = useMeasure()
-*/
-
   const { ref: nodeLeftRef, width: nodeLeftBoundsWidth = 1 } = useResizeObserver()
   const { ref: nodeRightRef, width: nodeRightBoundsWidth = 1 } = useResizeObserver()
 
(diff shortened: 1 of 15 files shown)


kdembler:
Changing this to a draft - unnecessary comments and build is failing

packages/atlas/src/components/_inputs/TextField/TextField.tsx
```diff
@@ -54,8 +55,8 @@ const TextFieldComponent: React.ForwardRefRenderFunction<HTMLInputElement, TextF
           </NodeContainer>
         )}
         <TextInput
-          leftNodeWidth={nodeLeftBounds.width}
-          rightNodeWidth={nodeRightBounds.width}
+          leftNodeWidth={nodeLeftBoundsWidth}
+          rightNodeWidth={nodeRightBoundsWidth}
```
kdembler, line 59: This does not work properly, see Storybook `?path=/story/inputs-textfield--with-icons`

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
```diff
@@ -117,10 +117,10 @@ const VideoWorkspaceActionBar: React.FC<VideoWorkspaceActionBarProps> = ({
   onResize,
 }) => {
   const mdMatch = useMediaMatch('md')
-  const [actionBarRef, actionBarBounds] = useMeasure()
+  const { ref: actionBarRef, height: actionBarBoundsHeight = 1 } = useResizeObserver()
```
kdembler: This gets measured as 48px. Seems this library by default measure only the content-box, so padding doesn't get included in the measurement. Check https://github.com/ZeeCoder/use-resize-observer#options This will need to be taken into account in other places as well

kdembler (changes requested):

drillprop (approved):
Didn't find any issue, looks good

kdembler (approved):

PR #2204, opened 2022-03-01
add useMsTimestamp hook, clean up time estimation

kdembler:
Summary of changes:
- added `useMsTimestamp` hook that sends time updates to components
- updated block time estimation to always use ms timestamps
- cleaned up block time estimation logic slightly
- moved block subscription to joystream context provider so we always use a single one
- changed `NftTile` to accept timestamp instead of `timeLeft`

### DIFF
--- a/packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
+++ b/packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
@@ -5,6 +5,7 @@ import { useNft } from '@/api/hooks'
 import { AllNftFieldsFragment } from '@/api/queries'
 import { absoluteRoutes } from '@/config/routes'
 import { useBlockTimeEstimation } from '@/hooks/useBlockTimeEstimation'
+import { useMsTimestamp } from '@/hooks/useMsTimestamp'
 import { useAsset } from '@/providers/assets'
 import { useJoystream } from '@/providers/joystream'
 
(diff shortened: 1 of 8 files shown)


packages/atlas/src/hooks/useMsTimestamp.ts
```diff
@@ -0,0 +1,15 @@
...
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setTimestamp(Date.now())
+    }, 1000)
+
+    return () => clearInterval(interval)
+  }, [])
+
+  return timestamp
+}
```
drillprop, line 15: Maybe we could name this differently, just `useTimestamp` and pass to it a parameter which could be `ms`, `second`, maybe even `minutes`? 
Something like 
```tsx
 const msTimestamp = useTimestamp('ms')
 const secTimestamp = useTimestamp('second')
//...
```

packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
```diff
@@ -82,31 +64,29 @@ export const NftTileViewer: React.FC<NftTileViewerProps> = ({ nftId }) => {
...
-          ...nftCommponProps,
+          ...nftCommonProps,
           auction: 'none',
           buyNowPrice: nft.transactionalStatus.price,
         }
       case 'TransactionalStatusAuction': {
-        const isEnded =
-          (nft.transactionalStatus.auction?.endedAtBlock || timeLeft <= 0) &&
-          nft.transactionalStatus.auction?.auctionType.__typename === 'AuctionTypeEnglish'
+        const isEnded = auctionPlannedEndBlock && getCurrentBlock() >= auctionPlannedEndBlock
```
drillprop, line 73: When an auction expires I still see the `Less than the minute` slot.  I'm pretty sure that it worked before.

drillprop (changes requested):

packages/atlas/src/hooks/useMsTimestamp.ts
kdembler: Do we have a good use case for using different unit? I think it may be easier to maintain if we just pick one (`ms`) and stick with it

drillprop: ok, fair point 👍🏻

drillprop (approved):

Issue #1942, opened 2022-01-17
History expanded in NFT widget


Lenorette:
###Figma
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1264%3A100578

PR #2206, opened 2022-03-01
NFT Widget history

eldiegod:
fixes: #1942


eldiegod:
@toiletgranny you can take a look at the expanded history on this story book

toiletgranny (changes requested):
Looks stunning! I appreciate the transition animation you added, it's a nice detail and I hope it won't affect the performance once the NFT widget is placed on an actual video page with all the video tiles underneath it.

Here are some minor things I found:
1. Size of the JOY token icon is incorrect, currently it's `24px`, it should be `16px`,
2. Can we have the avatar component being inactive (no hover, pressed states) in history row components?
3. It seems like the scrolling container is set up incorrectly compared to designs. The key thing that's missing is the gradient fading effect on the top & bottom of the list container.
image

eldiegod:
Hey @toiletgranny, thanks for the feedback.

> Can we have the avatar component being inactive (no hover, pressed states) in history row components?

I assumed clicking on the Avatar or the member handle would be a way to go to this member page, was I wrong on that?

> It seems like the scrolling container is set up incorrectly compared to designs. The key thing that's missing is the gradient fading effect on the top & bottom of the list container.

I tried implementing the transparency but I had some issues with it and the animation not working too well together, so I wanted to know if it would be a big deal not to have it, wdyt?

kdembler:
@eldiegod Like mentioned during our sync yesterday, let's improve the layout and include the history list inside the element's node.

toiletgranny:
> I assumed clicking on the Avatar or the member handle would be a way to go to this member page, was I wrong on that?

Oh right, of course. Yes, your assumption is correct. Let's skip it.

> I tried implementing the transparency but I had some issues with it and the animation not working too well together, so I wanted to know if it would be a big deal not to have it, wdyt?

While this doesn't break the experience, it surely breaks my heart. 😅  I guess it's not such a big deal.

toiletgranny (approved):

eldiegod:
> While this doesn't break the experience, it surely breaks my heart. 😅 I guess it's not such a big deal.

Couldn't allow that, we had to remove the anim temporarily so I implemented the transparency

eldiegod:
@kdembler done!

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
```diff
@@ -0,0 +1,178 @@
...
+        <CopyContainer>
+          <Text variant={size === 'medium' ? 'h300' : 'h200'} secondary>
+            {copy}{' '}
+            <OwnerHandle to={absoluteRoutes.viewer.member(memberHandle)} variant="secondary" textOnly>
+              <Text variant={size === 'medium' ? 'h300' : 'h200'}>{memberHandle}</Text>
+            </OwnerHandle>
+          </Text>
+        </CopyContainer>
+        <Text variant="t100" secondary>
+          {format(date, "d MMM yyyy 'at' HH:mm")}
```
rafalpawlow: Date formatting is a little bit different in figma
Zrzut ekranu 2022-03-9 o 20 05 46

packages/atlas/src/components/_nft/NftWidget/NftHistory.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+  transform: rotate(0);
+  transform-origin: center;
+  transition: ${cVar('animationTransitionFast')};
+
+  &[data-open='true'] {
+    transform: rotate(180deg);
+  }
+`
+
+export const FadingBlock = styled.div<{ width: number } & { 'data-bottom'?: boolean } & SizeProps>`
```
rafalpawlow: Why not this way? Also maybe we could move declare it above? Did you intentionally not use `OpenProps`?
```suggestion
export const FadingBlock = styled.div<{ width: number; 'data-bottom'?: boolean } & SizeProps>`
```

packages/atlas/src/components/_nft/NftWidget/NftHistory.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+  height: ${sizes(6)};
+  background: linear-gradient(0deg, rgb(11 12 15 / 0) 0%, #0b0c0f 100%);
```
rafalpawlow:
```suggestion
  background: linear-gradient(0deg, rgb(11 12 15 / 0) 0%, ${cVar('colorCoreNeutral900')}100%);
```

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
```diff
@@ -0,0 +1,178 @@
...
+  JoyPlusIcon,
+  NftHistoryHeader,
+  StyledChevronButton,
+  TextContainer,
+  ValueContainer,
+} from './NftHistory.styles'
+import { Size } from './NftWidget'
+import { OwnerHandle } from './NftWidget.styles'
+
+export const NftHistory: React.FC<{ size: Size; width: number }> = ({ size, width }) => {
```
rafalpawlow: Let's declare types above

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
```diff
@@ -0,0 +1,178 @@
...
+  ]
+
+  return (
+    <>
+      <NftHistoryHeader data-open={isOpen} data-size={size} onClick={toggleIsOpen}>
+        <Text variant={size === 'small' ? 'h300' : 'h400'}>History</Text>
+        <StyledChevronButton data-open={isOpen} iconOnly variant="tertiary" icon={<SvgActionChevronB />} />
+      </NftHistoryHeader>
+      {isOpen && (
+        <div style={{ position: 'relative' }}>
```
rafalpawlow: We shouldn't use inline styles

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
```diff
@@ -0,0 +1,178 @@
...
+          <FadingBlock data-size={size} width={width} />
+          <HistoryPanel width={width} data-size={size} data-open={isOpen}>
+            {/* TODO: remove dummy data */}
+            {dummyData.map((props: HistoryItemProps, index) => (
```
rafalpawlow: I'm pretty sure there is no need to add this type, it's taken directly from `dummyData`. Correct me if I'm wrong
```suggestion
            {dummyData.map((props, index) => (
```

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
```diff
@@ -0,0 +1,178 @@
...
+              <HistoryItem key={index} {...props} />
+            ))}
+          </HistoryPanel>
+          <FadingBlock data-size={size} width={width} data-bottom />
```
rafalpawlow, line 121: Do we need `data-bottom` if it doesn't provide any value?

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
```diff
@@ -0,0 +1,178 @@
...
+type HistoryItemProps = {
+  size: Size
+  memberAvatarUri: string
+  date: Date
+  joyAmount?: number
+  dollarValue?: string
+  copy: string
+  memberHandle: string
+}
+export const HistoryItem: React.FC<HistoryItemProps> = ({
```
rafalpawlow, line 137: I think we could move this component to separate file, this one is big enough. WDYT?

rafalpawlow (changes requested):

toiletgranny:
> > While this doesn't break the experience, it surely breaks my heart. 😅 I guess it's not such a big deal.
> 
> Couldn't allow that, we had to remove the anim temporarily so I implemented the transparency

Just saw it. I have no words. It's beautiful. I got nothing more to add. Oh wait, I do. Thank you.

giphy

packages/atlas/src/components/_nft/NftWidget/NftHistory.styles.ts
eldiegod: yes, open props was not needed in this component

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
eldiegod: yes for these styles
image

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
eldiegod: I think is fine like this, once the dummy data is removed the file is going to be fairly small anyways

rafalpawlow (approved):

Issue #1927, opened 2022-01-14
NFT preview section of Issue NFT flow


KubaMikolajczyk:
**Context**

NFT preview (re-using NFT video tile) in the issue NFT flow

**AC**
- [ ] NFT preview depends on the populated form fields in the issue flow
- [ ] NFT preview interaction is inherited from NFT video tile
- [ ] Scope of this issue covers all steps and hover on owner group described in [this flow ](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1006%3A40082) Pls view design comments from Kuba

**Design**

Link to NFT[ Preview interactions](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1006%3A40082)

PR #2214, opened 2022-03-03
Nft Preview

eldiegod:
fixes: #1927


drillprop (changes requested):
Just a quick functionality/visual feedback - I'll look at the code later. 
I've noticed that the NFT Tile doesn't react to changes in the form.

https://user-images.githubusercontent.com/51168865/156548028-3249c2eb-fd97-431e-8951-204e68b0d92c.mov

Also, we're missing the `NFT Preview` text below the tile(this is probably my fault,  I forgot about this, when I worked on layout, sorry 😅 )
 
image

eldiegod:
> I've noticed that the NFT Tile doesn't react to changes in the form.

no worries it should be there now 👍

packages/atlas/src/components/NftTile/NftTile.tsx
```diff
@@ -1,23 +1,23 @@
...
   assetUrl?: string | null
-  name: string
+  name?: string
   onClick?: () => void
   loading?: boolean
 }
 
 export type NftTileProps = {
-  auction?: 'none' | 'minBid' | 'topBid' | 'waiting'
+  nftState?: 'idle' | 'on-sale' | 'auction'
```
drillprop: maybe we could rename this - `nftStatus` or `nftAuctionStatus` or `nftSaleStatus`? Wdyt?

packages/atlas/src/components/NftTile/NftTileDetails.tsx
```diff
@@ -35,22 +36,24 @@ export type NftTileDetailsProps = {
...
 type TileSize = 'small' | 'medium'
 
-type DetailsContent = {
+type DetailsContentProps = {
   caption: string
   icon: React.ReactNode
   content: string | number
   secondary?: boolean
-  tileSize?: TileSize
+  tileSize: TileSize | undefined
```
drillprop:
```suggestion
  tileSize?: TileSize
```

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.styles.ts
```diff
@@ -21,16 +21,17 @@ export const NftWorkspaceFormWrapper = styled.div`
 export const NftPreview = styled.div`
   background: black;
   display: flex;
+  gap: ${sizes(4)};
```
drillprop: We shouldn't use `gap` when the container display is set to flex.

drillprop (changes requested):
Looks good. 
I have one comment on this. When you're in the first step you're seeing NFT tile which looks like this: 
image
Maybe we should set `nftState` to `idle` as default or maybe even select the first option from the form? Maybe you could ask designers - it looks like in the designs first option should be checked, but Im not sure about this.

packages/atlas/src/components/NftTile/NftTile.tsx
eldiegod: I called it nftState on the NFT widget as well but we would def benefit from having this type defined somewhere to unify NFT logic a bit as everyone seems to be doing it differently.

packages/atlas/src/components/NftTile/NftTileDetails.tsx
eldiegod: no, I want this to be passed always but I also allow it to be undefined

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.styles.ts
eldiegod: right

packages/atlas/src/components/NftTile/NftTile.tsx
eldiegod: nftAuctionStatus or nftSaleStatus I think wouldn't cut it because those are subtypes within the type

eldiegod:
Removed `gap` and selected auction as default option.

Also added a margin prop to Text component that will come very useful

packages/atlas/src/components/NftTile/NftTileDetails.tsx
```diff
@@ -213,11 +214,14 @@ export const NftTileDetails: React.FC<NftTileDetailsProps> = ({
       onMouseEnter={toggleContentHover}
       onMouseLeave={toggleContentHover}
       tileSize={tileSize}
+      shouldHover={(contentHovered || hovered) && interactable}
```
drillprop, line 210: Getting warning in the console:
image
I think it comes from here.

packages/atlas/src/styles/sizes.ts
```diff
@@ -4,9 +4,11 @@ import { toPx } from '@/utils/styles'
 
 const base = 4
 
+export function sizes(n: number): string
 export function sizes<B extends boolean>(n: number, raw?: B): B extends false ? string : number
+
 export function sizes(n: number, raw?: boolean) {
-  return raw ? base * n : `${base * n}px`
+  return raw === true ? base * n : (`${base * n}px` as string)
```
drillprop: Hmm, why do we need this change? I think it should work as before, right? We don't need to add `as string` and check if raw is true.

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
```diff
@@ -146,6 +144,34 @@ export const NftForm: React.FC<NftFormProps> = ({ setFormStatus, onSubmit, setIs
...
+
+  const nftTileProps: NftTileProps = {
+    nftStatus: getNftStatus(),
+    thumbnail: { thumbnailUrl: thumbnailPhotoUrl },
+    title: video?.title,
+    owner:
+      getNftStatus() === 'auction'
+        ? { assetUrl: null, name: 'Owner: Auction winner' }
+        : { assetUrl: memberAvatarUri, name: activeMembership?.handle },
+    creator: { assetUrl: memberAvatarUri, name: activeMembership?.handle },
```
drillprop: The creator field should point to the channel, not to the member.

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
```diff
@@ -146,6 +144,34 @@ export const NftForm: React.FC<NftFormProps> = ({ setFormStatus, onSubmit, setIs
...
+  }
+
+  const nftTileProps: NftTileProps = {
+    nftStatus: getNftStatus(),
+    thumbnail: { thumbnailUrl: thumbnailPhotoUrl },
+    title: video?.title,
+    owner:
+      getNftStatus() === 'auction'
+        ? { assetUrl: null, name: 'Owner: Auction winner' }
+        : { assetUrl: memberAvatarUri, name: activeMembership?.handle },
```
drillprop: Why the owner is set when the status isn't an `auction`?

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
```diff
@@ -64,22 +70,14 @@ export const NftForm: React.FC<NftFormProps> = ({ setFormStatus, onSubmit, setIs
...
-    {
-      variant: 'future',
-      title: 'Accept listing terms',
-    },
-  ]
+
+  const { video, loading: loadingVideo } = useVideo(videoId, { fetchPolicy: 'cache-only' })
+  const { url: thumbnailPhotoUrl } = useAsset(video?.thumbnailPhoto)
+  const memberAvatarUri =
+    activeMembership?.metadata.avatar?.__typename === 'AvatarUri' ? activeMembership?.metadata.avatar.avatarUri : null
```
drillprop: you could use `useMemberAvatar` here to properly get the `avatarUri`, if I'm not mistaken

packages/atlas/src/components/NftTile/NftTileDetails.tsx
```diff
@@ -35,22 +36,24 @@ export type NftTileDetailsProps = {
...
   title?: string | null
   hovered?: boolean
+  interactable?: boolean
+  videoHref?: string
 }
 
 type TileSize = 'small' | 'medium'
 
-type DetailsContent = {
+type DetailsContentProps = {
```
drillprop: This is just a preference, but could you move this closer to the DetailContent component?

drillprop (changes requested):

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
eldiegod: I saw it somewhere in the design, @kdembler can you clarify this?

eldiegod: I saw it on the designs but it's not very clear, @kdembler thoughts?

https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1006%3A40082

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
eldiegod: I thought the owner was just gonna be the member, for now, maybe I missed something. @kdembler can u clarify?

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
eldiegod: nicee

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
kdembler: Creator and member are always separate fields.

Creator is always a channel that published the video and issued an NFT for it. You can get this via `nft.video.channel`, there is a query node issue that will add something like `nft.creator: Channel` for easier access.

Owner is almost always a member that is currently in possession of the NFT. That is `nft.ownerMember` field. There is one special case here when the NFT has been issued but haven't been sold/transferred yet. The question (mostly open at this point) is who do we show as an owner in this case. Atm, query node will return `null` for `ownerMember` field, but there is an issue open for this to be populated with the member that controls the channel, so there will always be both creator and owner.

So to sum up, creator is a channel, owner is a member. I hope this clears it out, let me know if something is still unclear

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.tsx
kdembler: Hmm, I think maybe we should override the design here and always show the current owner here

packages/atlas/src/components/NftTile/NftTileDetails.tsx
```diff
@@ -85,20 +82,25 @@ export const NftTileDetails: React.FC<NftTileDetailsProps> = ({
     },
   })
 
+  const handleCopyVideoURLClick = useCallback(() => {
```
kdembler, line 86: This should use the new hook we introduced recently, `const { copyToClipboard } = useClipboard()`

packages/atlas/src/components/Text/Text.styles.ts
```diff
@@ -3,127 +3,146 @@ import styled from '@emotion/styled'
...
+    }
+  | string
+
+const marginStyles = ({ margin }: TextBaseProps) =>
+  typeof margin !== 'string' && !!margin
+    ? css`
+        margin: ${margin.top} ${margin.right} ${margin.bottom} ${margin.left};
+      `
+    : css`
+        margin: ${margin ?? 0};
```
kdembler: I think this is conflicting with similar functionality Rafał is introducing in his PR: https://github.com/Joystream/atlas/pull/2218/files#diff-f4908e2ae287f2ff9c86b053a80805d5fc9ca31f1ff659566496391e1020cb0fR30

@eldiegod @rafalpawlow Please sync up on that so we use the same approach in both PRs

eldiegod:
All feedback should be in now!

kdembler (approved):
LGTM, let's get approval from @drillprop as well

drillprop (approved):
LGTM as well!

PR #2218, opened 2022-03-04
place bid view

### LINKED ISSUE #1859: NFT Purchase Layout with States and Validation
**Context**

Place the Bid View
Screenshot 2022-01-04 at 17 31 08

**AC**
Different states of the screen
- [ ] English Auction no bids placed ⚠️  _Transaction fee_ instead of gas fee ⚠️ 
- [ ] Only the number of blocks left
- [ ] Open Auction [(no timer)](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=965%3A93201) 
- [ ] Insufficient [Funds  ](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=1022%3A102687)
- [ ] Amount entered is less than [minimum bid](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=1022%3A104907)
- [ ] Amount entered is [more than Buy now price ](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=1024%3A112415)
- [ ] Responsive [link to Figma page](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=993%3A96351)

**OOC:**
- [ ] Displaying an hour:minute timer
- [ ] English Auction Timer view [with < 1 minute to go](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=965%3A90622)

**Designs**

[Link to Figma >](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=964%3A134035)

rafalpawlow:
Closes #1859 

Place bid is open by default right now for testing purposes. I'll change it before merge. https://github.com/Joystream/atlas/pull/2218/files#diff-c87c442e1af7c143a8096b3e3ea3d753e79ff0ca8630b32a6fa8c31261888c5cR14

**Design**
https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=964%3A134035


packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
```diff
@@ -0,0 +1,332 @@
...
+                </PaymentSplitWrapper>
+                <Divider />
+                <Text variant="h400" spacing={{ bottom: 4 }}>
+                  Price breakdown
+                </Text>
+                <Row>
+                  <Text variant="t100" secondary>
+                    Your balance
+                  </Text>
+                  {accountBalance ? (
```
drillprop: You didn't handle the scenario when you have insufficient balance: https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=985%3A96792

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
```diff
@@ -0,0 +1,332 @@
...
+const MINIMUM_BID = 301
+const TIME_LEFT = 80
+const BID = {
+  bidder: {
+    name: 'Mike Shipa',
+  },
+  amount: 341,
+}
+
+export const PlaceBid = () => {
```
drillprop:
```suggestion
export const PlaceBid: React.FC = () => {
```
I mean, I don't agree with this rule to adding React.FC everywhere, I think we shouldn't use React.FC at all, but Klaudiusz will definitely point this out 😄

drillprop: Also, I think it should be named differently. maybe we could name this AuctionView like in in figma?
image

packages/atlas/src/hooks/useCountdown.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (seconds) {
+      interval = setInterval(() => {
+        setTimeLeft((prevState) => {
+          if (prevState && prevState > 0) {
+            return prevState - 1
+          }
+          return 0
+        })
+      }, 1000)
+    }
```
drillprop:
```suggestion
  useEffect(() => {
    if (!seconds) {
      return
    }
    const interval = setInterval(() => {
      setTimeLeft((prevState) => {
        if (prevState && prevState > 0) {
          return prevState - 1
        }
        return 0
      })
    }, 1000)
```

Also, do we actually need first useEffect?

packages/atlas/src/providers/placeBid/provider.tsx
```diff
@@ -0,0 +1,35 @@
...
+type ContextValue = {
+  isPlaceBidOpen: boolean
+  setIsPlaceBidOpen: React.Dispatch<React.SetStateAction<boolean>>
+}
+
+export const PlaceBidContext = React.createContext<ContextValue | undefined>(undefined)
+PlaceBidContext.displayName = 'PlaceBidContext'
+
+export const PlaceBidProvider: React.FC = ({ children }) => {
+  // TODO: default open for preview purposes
```
drillprop: Could you move it to the Playground then?

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.styles.ts
```diff
@@ -0,0 +1,236 @@
...
+  ${media.md} {
+    max-width: 550px;
+    padding-bottom: ${sizes(18)};
+  }
+  ${media.lg} {
+    max-width: 560px;
+  }
+`
+
+export const StyledJoyTokenIcon = styled(JoyTokenIcon)`
```
drillprop: I think we have a separate component with this filter already applied If I'm not mistaken.

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
```diff
@@ -0,0 +1,332 @@
...
+    init()
+  }, [activeMembership, joystream])
+
+  const onCloseClick = () => setIsPlaceBidOpen(false)
+
+  const bidError = !!(placedBid && placedBid < MINIMUM_BID)
+  const timeLeftUnderMinute = timeLeft && timeLeft < 60
+
+  return (
+    <>
```
drillprop: We don't need `Fragment` here

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
```diff
@@ -0,0 +1,332 @@
...
+      <CSSTransition
+        in={isPlaceBidOpen}
+        appear
+        mountOnEnter
+        unmountOnExit
+        timeout={{ enter: 0, exit: parseInt(cVar('animationTimingSlow', true)) }}
+        classNames="place-bid"
+      >
+        <Container role="dialog">
+          <div style={{ position: 'fixed', top: '10px', left: '10px' }}>
```
drillprop: We should create styled component instead of doing `style={{ ... }}`

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.styles.ts
```diff
@@ -0,0 +1,236 @@
...
+  box-shadow: ${cVar('effectDividersTop')};
+  margin-top: ${sizes(6)};
+  margin-bottom: ${sizes(4)};
+`
+
+export const StyledActionBar = styled(ActionBar)`
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  z-index: 999;
```
drillprop: let's use `zIndex` from `sizes.ts`

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.styles.ts
```diff
@@ -0,0 +1,236 @@
...
+export const BidderName = styled(Text)`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`
+
+export const ActiveBidWrapper = styled.div`
+  display: flex;
+
+  > div {
```
drillprop: Instead pointing to a `div` I think we should create a styled component here

drillprop (changes requested):

drillprop (changes requested):
Also, Im getting: 
image

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
rafalpawlow: Good point

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
rafalpawlow: This is only for switching between types, it needs to be removed before merge.

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
kdembler: I would say `AuctionView` is a bit ambiguous, `PlaceBid` is more explicit. Or maybe `NftPurchaseView`? As explicit as we can get it ;)

kdembler: > I think we shouldn't use React.FC at all

@drillprop Why?

drillprop: There are several reasons for that, for example, React.FC always implies `children`. I think we had a short discussion about it before. This article sums it up nicely.
https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/

packages/atlas/src/hooks/useCountdown.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react'
+
+import Timeout = NodeJS.Timeout
+
+export const useCountdown = (seconds?: number) => {
```
kdembler: Where is this used?

packages/atlas/src/hooks/useCountdown.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react'
+
+import Timeout = NodeJS.Timeout
```
kdembler: This doesn't seem needed

packages/atlas/src/views/viewer/PlaceBid/PlaceBid.tsx
kdembler: Ah yes, remember that. If we rewrite the codebase at some point, let's revisit, but for now let's continue with what we have :D

packages/atlas/src/hooks/useCountdown.ts
rafalpawlow: https://github.com/Joystream/atlas/pull/2218/files#diff-452a81aec0cd71561d35a7d0c694335b700ad5cf020154dacde2f45fca339e40R79

We talked some time ago about creating 1 hook for counting down

kdembler: This has been already done here: #2204. This hook will provide you current time and will update it every second. So using this and target date, you can just find time left in the component

rafalpawlow: Sorry, haven't noticed

kdembler:
I think current naming is a bit too ambiguous (`AuctionView`/`useAuction`). Like mentioned in my previous comment (https://github.com/Joystream/atlas/pull/2218#r819627420), I think we should be more specific to not run into any conflicts later on. I suggested `NftPurchaseView`, could we use that? And `useNftPurchase` accordingly?

packages/atlas/src/providers/auction/provider.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [isAuctionOpen, setAuctionOpen] = useState(false)
+  const { incrementOverlaysOpenCount, decrementOverlaysOpenCount } = useOverlayManager()
+
+  useEffect(() => {
+    if (isAuctionOpen) {
+      incrementOverlaysOpenCount()
+    } else {
+      decrementOverlaysOpenCount()
+    }
+  }, [decrementOverlaysOpenCount, incrementOverlaysOpenCount, isAuctionOpen])
```
kdembler: This will also run the first time it's rendered, reducing overlays even though the provider hasn't opened any yet. I think you will need some `cached` state to be able to compare

packages/atlas/src/views/viewer/AuctionView/AuctionView.styles.ts
```diff
@@ -0,0 +1,230 @@
...
+    flex: 50%;
+    padding: 0;
+  }
+`
+
+export const PlaceBidWrapper = styled.div`
+  padding-bottom: ${sizes(6)};
+
+  ${media.sm} {
+    height: calc(100% - 80px);
```
kdembler: What is this `80px`?

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
```diff
@@ -0,0 +1,339 @@
...
+
+  useEffect(() => {
+    const init = async () => {
+      if (activeMembership) {
+        const balance = await joystream?.getAccountBalance(activeMembership?.controllerAccount)
+        setAccountBalance(balance)
+      }
+    }
+    init()
+  }, [activeMembership, joystream])
```
kdembler: I think it may be safer to use a subscription here? Currently we will fetch this only on first render, but the lifetime of this component may be long since we render it at all times, just not show it

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                    <div>
+                      <Text variant="t100" secondary>
+                        Buy now: {FIXED_PRICE} tJOY
+                      </Text>
+                    </div>
+                  </MinimumBidWrapper>
+                  <TextField
+                    placeholder={`Min. ${MINIMUM_BID} tJOY`}
+                    nodeStart={<JoyTokenIcon variant="silver" size={24} />}
+                    nodeEnd={!!placedBid && <Pill variant="overlay" label={`$${convertToUSD(placedBid)}`} />}
```
kdembler: We need to round this number, the current precision is way too precise, I think 2 numbers for fractional part is max. We should also handle really low values do we don't display `0.00` if the value is `0.004`. @TCzechowski any ideas here?

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
```diff
@@ -0,0 +1,339 @@
...
+        const balance = await joystream?.getAccountBalance(activeMembership?.controllerAccount)
+        setAccountBalance(balance)
+      }
+    }
+    init()
+  }, [activeMembership, joystream])
+
+  const onCloseClick = () => setAuctionOpen(false)
+
+  const bidError = !!(placedBid && placedBid < MINIMUM_BID)
```
kdembler: This validation doesn't work really well - it will start validating as soon as the user starts typing. If I want to input 100, I shouldn't get an error for typing `1` along the way. We probably should use `react-hook-form` here that will take care of all the input validations

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                      variant="h200"
+                      spacing={{ left: 4, right: 2 }}
+                      color={timeLeftUnderMinute ? cVar('colorTextError', true) : undefined}
+                    >
+                      {timeLeft && (!timeLeftUnderMinute ? formatDurationShort(timeLeft, true) : 'Under 1 min')}
+                    </Timer>
+                  </EndingTime>
+                  <FlexWrapper>
+                    <Text variant="t100" secondary spacing={{ left: 2, right: 1 }}>
+                      1508 blocks
```
kdembler: This needs to be updated as well

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
```diff
@@ -0,0 +1,339 @@
...
+  const [type, setType] = useState<'auction' | 'buy_now'>('auction')
+  const { isAuctionOpen, setAuctionOpen } = useAuction()
+  const mdMatch = useMediaMatch('md')
+  const { convertToUSD } = useTokenPrice()
+  const { joystream } = useJoystream()
+  const { activeMembership } = useUser()
+  const [placedBid, setPlacedBid] = useState<number | undefined>()
+  const [accountBalance, setAccountBalance] = useState<number | undefined>()
+  const timestamp = useMsTimestamp()
+  const timeLeft = Math.trunc((END_TIME - timestamp) / 1000)
```
kdembler: Let's include unit in the variable name so it's clear what this stores

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
```diff
@@ -0,0 +1,339 @@
...
+    const init = async () => {
+      if (activeMembership) {
+        const balance = await joystream?.getAccountBalance(activeMembership?.controllerAccount)
+        setAccountBalance(balance)
+      }
+    }
+    init()
+  }, [activeMembership, joystream])
+
+  const onCloseClick = () => setAuctionOpen(false)
```
kdembler: Should be `handleCloseClick`

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                      bidError
+                        ? 'Your bid must be higher than minimum bid'
+                        : insufficientFoundsError
+                        ? 'Insufficient funds.'
+                        : undefined
+                    }
+                  />
+                  {!!placedBid && placedBid > FIXED_PRICE && (
+                    <Text variant="t100" spacing={{ top: 2 }}>
+                      Max bid cannot be more than buy now price. Bidding for amount higher than Buy now will
```
kdembler: This doesn't fully follow the design, please check this user story: https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-Purchase-view?node-id=985%3A96806

packages/atlas/src/components/Text/Text.styles.ts
```diff
@@ -1,12 +1,20 @@
...
+  right?: number
+  top?: number
+  bottom?: number
+}
 
 type TextProps = {
   color?: string
   isSecondary?: boolean
   clampAfterLine?: number
+  spacingConfig?: Spacing
```
kdembler: What's the reasoning behind this? I'm not sure how I feel about this change, this way we move styling into the components markup, instead of style files like we do currently

packages/atlas/src/views/viewer/NftPurchaseView/NftPurchaseView.styles.ts
```diff
@@ -0,0 +1,230 @@
...
+    right: 0;
+    width: 1px;
+    height: 32px;
+    background-color: ${cVar('colorBorderMutedAlpha')};
+  }
+`
+
+export const Timer = styled(Text)`
+  padding-right: ${sizes(4)};
+`
```
kdembler: This is not how this is handled in the design

kdembler (changes requested):
One additional comment: Mobile views are close to unusable because the NFT card takes up most of the space on the screen. I'm pretty sure the card is supposed to be shown at the bottom of the page, not visible at all times, you can confirm with @TCzechowski

packages/atlas/src/components/Text/Text.styles.ts
rafalpawlow: We have a lot of `Text` components which have own styled component only because of margin. I thought it would save some work in the future.

kdembler: Okay, let's see how it goes

packages/atlas/src/views/viewer/AuctionView/AuctionView.tsx
TCzechowski: We've discussed that internally within the design team, and we agree to display only 2 numbers for fractional part, but if it's less than 0.01, can we show as many numbers as there are zeroes? What I mean by that is for example, if the converted amount is $0.0003623482.... then we display $0.0004 (rounded)?

packages/atlas/src/views/viewer/NftPurchaseView/NftPurchaseView.styles.ts
```diff
@@ -0,0 +1,239 @@
+import { css } from '@emotion/react'
+import styled from '@emotion/styled'
+
+import { ActionBar } from '@/components/ActionBar'
+import { Avatar } from '@/components/Avatar'
+import { Text } from '@/components/Text'
+import { JoyTokenIcon } from '@/components/_icons/JoyTokenIcon'
+import { cVar, media, sizes, zIndex } from '@/styles'
+
+const ACTION_BAR_HEIGHT = '80px'
```
kdembler: This is measured in all the other places in the app, best to do the same here IMO

packages/atlas/src/views/viewer/NftPurchaseView/NftPurchaseView.tsx
```diff
@@ -0,0 +1,377 @@
...
+                      <Text variant="h300" secondary>
+                        Ending in:
+                      </Text>
+                      <Timer
+                        variant="h200"
+                        spacing={{ left: 4, right: 2 }}
+                        color={timeLeftUnderMinute ? cVar('colorTextError', true) : undefined}
+                      >
+                        {timeLeftSeconds &&
+                          (!timeLeftUnderMinute ? formatDurationShort(timeLeftSeconds, true) : 'Under 1 min')}
```
kdembler: This will display 0 for a second when the auction is ending

packages/atlas/src/views/viewer/NftPurchaseView/NftPurchaseView.tsx
```diff
@@ -0,0 +1,377 @@
...
+                      </Timer>
+                    </EndingTime>
+                    <FlexWrapper>
+                      <Text variant="t100" secondary spacing={{ left: 2, right: 1 }}>
+                        {convertMsTimestampToBlock(END_TIME)} blocks
```
kdembler: This is supposed to show how many blocks are left till the auction end. Currently, you show ending block estimate based on the end time. This value should be dynamically updated as time passes

packages/atlas/src/views/viewer/NftPurchaseView/NftPurchaseView.tsx
```diff
@@ -0,0 +1,377 @@
...
+                  <SkeletonLoader width={82} height={16} />
+                )}
+              </Row>
+              <Row>
+                <Text variant="t100" secondary>
+                  {!placedBid && type === 'auction' ? 'You need to fill out the amount first' : 'Your bid'}
+                </Text>
+                {placedBid && (
+                  <Text variant="t100" secondary>
+                    {type === 'auction' ? placedBid : FIXED_PRICE}
```
kdembler: This is missing the `tJOY` unit

packages/atlas/src/views/viewer/NftPurchaseView/NftPurchaseView.tsx
```diff
@@ -0,0 +1,377 @@
...
+                    <Text variant="t100" secondary>
+                      {TRANSACTION_FEE} tJOY
+                    </Text>
+                  </Row>
+                  <Row>
+                    <Text variant="h500" secondary>
+                      You will pay
+                    </Text>
+                    <Text variant="h500">
+                      {type === 'buy_now' ? FIXED_PRICE : (Number(placedBid) || 0) + TRANSACTION_FEE} tJOY
```
kdembler: Why do you include fee in calculation only for auctions?

packages/atlas/src/views/viewer/NftPurchaseView/NftPurchaseView.tsx
```diff
@@ -0,0 +1,377 @@
...
+                      required: true,
+                      validate: {
+                        bidTooLow: (value) =>
+                          Number(value) >= MINIMUM_BID ? true : 'Your bid must be higher than minimum bid',
+                        bidTooHigh: (value) =>
+                          accountBalance
+                            ? Number(value) + TRANSACTION_FEE > accountBalance
+                              ? 'Insufficient funds.'
+                              : true
+                            : true,
```
kdembler, line 242: Not sure if this is included in the design but we also should handle this case:
1. The auction type is English
2. The bid is lower than the current top one

In open auction we don't care since users can post bids lower than the current winning one, but in English auction you always have to outbid the top bid

rafalpawlow: After short a discussion on rocket chat, we decided that when integrating we will set current bid as minimum bid.

kdembler (approved):

drillprop (approved):

PR #2219, opened 2022-03-04
Improve numbered carousels

### LINKED ISSUE #1595: Squished video tiles in numbered carousels
I think we could potentially improve the presentation of content in numbered carousels. I think the number of views broken into 2 lines doesn't look great. Also in general, seems like not enough horizontal space is given to those tiles, maybe we should rethink the sizing/positioning of numbers?
Screenshot 2021-11-16 at 13 58 28

kdembler:
Resolves #1595

### DIFF
--- a/packages/atlas/src/components/Carousel/Carousel.styles.ts
+++ b/packages/atlas/src/components/Carousel/Carousel.styles.ts
@@ -61,7 +61,7 @@ export const Track = styled.div`
 `
 
 export const Dots = styled.div`
-  margin-top: 48px;
+  margin-top: ${sizes(12)};
   display: none;
 
   ${media.md} {
--- a/packages/atlas/src/components/Carousel/Carousel.styles.ts
+++ b/packages/atlas/src/components/Carousel/Carousel.styles.ts
@@ -43,9 +43,6 @@ export const Arrow = styled(IconButton)`
 `
 
 export const GliderContainer = styled.div`
-  padding-left: ${sizes(2)};
-  padding-top: ${sizes(2)};
-
   /* hides scrollbar on firefox */
   scrollbar-width: none;
 `
@@ -62,7 +59,6 @@ export const Track = styled.div`
 
 export const Dots = styled.div`
   padding: ${sizes(3.5)} 0;
-  margin-top: ${sizes(12)};
   display: none;
 
   ${media.md} {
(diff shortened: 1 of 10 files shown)


toiletgranny:
_A fly does not sit._ I have nothing more to add, it's just perfect. 

There was this one thing I asked for back in #2148, which was clipping carousels on the edges and if we can do something about it:
> The scrollable contents of carousels are clipped at the edges, not allowing the numbered tile components to reach the edge of the viewport. In the designs, the contents of the carousels span beautifully from one edge of the viewport to another. Can we have that in the implementation since we're revisiting this component? Ideally, the inner padding in the carousel matches the globally applied one, which is 16px up until the MD breakpoint, and then 32px going forward.

Do you reckon it's something we could tackle while we are here?

kdembler:
The issue you mentioned has actually been with us since the very beginning of Atlas I think, solving this may be outside of the scope of this specific problem

toiletgranny (approved):
>The issue you mentioned has actually been with us since the very beginning of Atlas I think, solving this may be outside of the scope of this specific problem

Ah, of course, I understand. In my head, as I was experimenting with this myself, the solution seemed quite straightforward ([see the recording](https://cln.sh/HKSQUU), and the code snippet below), but if there are other implications we need to be careful about, then sure — let's tackle this some other time.

```
margin: 0 calc(var(--size-global-horizontal-padding) * -1);
padding: 0 var(--size-global-horizontal-padding);
```

kdembler:
I'm not able to replicate what you're showing in the video. Applying the CSS you provided I see that on some breakpoints some of the tiles get cut. This likely interfere with the size calculations the carousel is doing to know how many items to fit and how wide they should be.

packages/atlas/src/components/Carousel/Carousel.styles.ts
```diff
@@ -61,8 +61,7 @@ export const Track = styled.div`
 `
 
 export const Dots = styled.div`
-  padding: ${sizes(3.5)} 0;
-  margin-top: ${sizes(12)};
+  margin-top: 48px;
```
rafalpawlow: Didn't we agree for margins and padding we use `sizes`?

rafalpawlow (changes requested):

packages/atlas/src/components/Carousel/Carousel.styles.ts
kdembler: You're correct, fixed now

rafalpawlow (approved):

PR #2223, opened 2022-03-07
NFT issuance follow up fixes

### LINKED ISSUE #2212: NFT issuance follow up fixes
Some things left to update in the NFT form after initial implementation:

- [ ] Lack of margin between header and info icon on summary screen:
Screenshot 2022-03-02 at 16 22 27

- [ ] No formatting for price, check Figma
Screenshot 2022-03-02 at 16 22 23

- [ ] No `blocks` suffix for number of blocks for auction duration
Screenshot 2022-03-02 at 16 22 50

- [ ] Shifts in action bar when moving between video and NFT forms (button values change at different times I think, also the action bar occasionally disappears):

https://user-images.githubusercontent.com/12646744/156392286-9af56c80-1cfa-43e2-af60-fc057e039650.mov

- [ ] Auction duration selects shifting layout

https://user-images.githubusercontent.com/12646744/156391498-3cbb50d8-3a6d-47a0-b204-4716a43721db.mov

- [ ] Video `Visibility` setting is not preserved when moving between video and NFT form

https://user-images.githubusercontent.com/12646744/156393127-c4eb82fc-9cda-4806-a86c-752375f7b1bf.mov

rafalpawlow:
Closes #2212 

- [x] Lack of margin between header and info icon on summary screen
- [x] No formatting for price, check Figma
- [x] No blocks suffix for number of blocks for auction duration
- [x] Shifts in action bar when moving between video and NFT forms (button values change at different times I think, also the action bar occasionally disappears):
- [x] Auction duration selects shifting layout
- [x] Video Visibility setting is not preserved when moving between video and NFT form

### DIFF
--- a/packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.hooks.ts
+++ b/packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.hooks.ts
@@ -6,7 +6,7 @@ import { Listing } from '@/views/studio/VideoWorkspace/NftForm/types'
 export const useNftForm = () => {
   const [activeInputs, setActiveInputs] = useState<string[]>([])
   const [termsAccepted, setTermsAccepted] = useState(false)
-  const [listingType, setListingType] = useState<Listing>(undefined)
+  const [listingType, setListingType] = useState<Listing>('Auction')
   const [currentStep, setCurrentStep] = useState(0)
 
   const getTotalDaysAndHoursText = (start: Date, end: Date) => {
--- a/packages/atlas/src/views/studio/VideoWorkspace/VideoForm/utils.ts
+++ b/packages/atlas/src/views/studio/VideoWorkspace/VideoForm/utils.ts
@@ -1,7 +1,11 @@
 import { VideoFormData, VideoWorkspaceVideoFormFields } from '@/providers/videoWorkspace'
 
+type Public = {
+  isPublic?: boolean
+}
+
 export const convertVideoFormDataToFormFields = (
-  videoFormData: VideoFormData
+  videoFormData: Omit<VideoFormData, 'isPublic'> & Public
 ): Partial<VideoWorkspaceVideoFormFields> => {
   const {
     metadata: {
@@ -57,7 +61,7 @@ export const convertVideoFormDataToFormFields = (
     ...(category ? { category: category?.toString() } : {}),
     ...(hasMarketing ? { hasMarketing } : {}),
     ...(language ? { language } : {}),
-    ...(isPublic ? { isPublic } : {}),
+    ...(isPublic !== undefined ? { isPublic } : {}),
     ...(isExplicit ? { isExplicit } : {}),
     ...(license?.attribution ? { licenseAttribution: license?.attribution } : {}),
     ...(license?.code ? { licenseCode: license?.code } : {}),
--- a/packages/atlas/src/views/studio/VideoWorkspace/VideoForm/utils.ts
+++ b/packages/atlas/src/views/studio/VideoWorkspace/VideoForm/utils.ts
@@ -1,11 +1,7 @@
 import { VideoFormData, VideoWorkspaceVideoFormFields } from '@/providers/videoWorkspace'
 
-type Public = {
-  isPublic?: boolean
-}
-
 export const convertVideoFormDataToFormFields = (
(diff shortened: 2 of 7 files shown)


kdembler:
1. I edit a video that has "Visibility" set to public
2. I don't make any changes, only toggle "Issue as NFT"
3. I click Next
4. I click Back
5. Now no option for "Visibility" is selected

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.hooks.ts
```diff
@@ -6,7 +6,7 @@ import { Listing } from '@/views/studio/VideoWorkspace/NftForm/types'
 export const useNftForm = () => {
   const [activeInputs, setActiveInputs] = useState<string[]>([])
   const [termsAccepted, setTermsAccepted] = useState(false)
-  const [listingType, setListingType] = useState<Listing>(undefined)
+  const [listingType, setListingType] = useState<Listing>('Auction')
```
kdembler, line 9: This determines which listing type will be selected by default, right? Maybe would make sense to use "Not for sale" here to reduce friction if the user wants to just issue an NFT? It may be weird though since it's the last option. @KubaMikolajczyk any opinions?

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/utils.ts
```diff
@@ -1,7 +1,11 @@
 import { VideoFormData, VideoWorkspaceVideoFormFields } from '@/providers/videoWorkspace'
 
+type Public = {
+  isPublic?: boolean
+}
+
 export const convertVideoFormDataToFormFields = (
-  videoFormData: VideoFormData
+  videoFormData: Omit<VideoFormData, 'isPublic'> & Public
```
kdembler: What exactly is happening here?

1. `isPublic` property doesn't exists on `VideoFormData` so no point in excluding it
2. You add `Public` to the type but it's never used - when `videoFormData` is destructured, `isPublic` from `metadata` is used

Seems this change isn't really needed

rafalpawlow: Nevermind, you are right, I don't what happend with my IDE yesterday, I had some errors related to `videoFormData` types.

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.hooks.ts
kdembler: Actually now that we're moving issuance to a separate flow, maybe we don't need "Not for sale" option at all? Since now that flow will be about putting on sale, it assumes the NFT is not on sale atm

rafalpawlow: So, should we remove "Not for sale" and stay with "Auction" as default selected option?

toiletgranny: @kdembler As a general rule of thumb, the fewer clicks and decisions in the process, the better. And so if you need a decision for now, I'd say it makes 100% sense to have the "Not for sale" option selected by default, as you suggested.

But then I think you're correct that we're most likely going to drop the "Not for sale" option with the upcoming design changes. And so if we're left with "Auction" and "Fixed price" as the only options to choose from, then my recommendation would be not to have anything selected by default.

toiletgranny: cc @KubaMikolajczyk

kdembler: Okay, to keep things simple let's leave "Auction" selected by default for now and don't worry too much about it @rafalpawlow, we can revisit in a future PR

kdembler (approved):

packages/atlas/src/views/studio/VideoWorkspace/NftForm/NftForm.hooks.ts
KubaMikolajczyk: I don't know if we should change this to not for sale being a default - Im 100% for leaving Auction as a default mode and revisiting it in future when we get more data about frequency of user choice here

Issue #2216, opened 2022-03-03
Data fetching for NFT widget


eldiegod:
Fetch real data for NFT widget and place it on the video view

Out of scope:
- Expanded history data fetching

https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1242%3A109695

PR #2246, opened 2022-03-09
Nft widget data fetching

eldiegod:
fixes: #2216

### DIFF
--- a/packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
+++ b/packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
@@ -1,13 +1,17 @@
 import React from 'react'
 import useResizeObserver from 'use-resize-observer'
 
+import { useNft } from '@/api/hooks'
 import { GridItem } from '@/components/LayoutGrid'
 import { Text } from '@/components/Text'
 import { Button } from '@/components/_buttons/Button'
 import { JoyTokenIcon } from '@/components/_icons/JoyTokenIcon'
 import { absoluteRoutes } from '@/config/routes'
+import { useBlockTimeEstimation } from '@/hooks/useBlockTimeEstimation'
 import { useDeepMemo } from '@/hooks/useDeepMemo'
-import { useTokenPrice } from '@/providers/joystream'
+import { useMemberAvatar } from '@/providers/assets'
+import { useJoystream, useTokenPrice } from '@/providers/joystream'
+import { useUser } from '@/providers/user'
 import { formatNumberShort } from '@/utils/number'
 import { formatDateTime } from '@/utils/time'
 
@@ -33,14 +37,15 @@ export type Auction = {
   topBid?: number
   isCompleted?: boolean
   canWithdrawBid?: boolean
+  needsSettling: boolean
   auctionPlannedEndDate?: Date
 }
 
 export type NftWidgetProps = {
   ownerHandle?: string
-  ownerAvatarUri?: string
-  isOwner: boolean
-  nftStatus:
+  ownerAvatarUri?: string | null
+  isOwner?: boolean
+  nftStatus?:
     | { status: 'idle'; lastPrice?: number; lastTransactionDate?: Date }
     | { status: 'buy-now'; buyNowPrice: number }
     | Auction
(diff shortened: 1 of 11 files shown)


packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -269,3 +276,73 @@ export const NftWidget: React.FC<NftWidgetProps> = ({ ownerHandle, isOwner, nftS
     </Container>
   )
 }
+
+type HookReturn = NftWidgetProps & { shouldRenderWidget: boolean }
+export const useNftWidget = (videoId?: string): HookReturn => {
```
drillprop:
```suggestion
type UseNftWidgetReturn = NftWidgetProps & { shouldRenderWidget: boolean }
export const useNftWidget = (videoId?: string): UseNftWidgetReturn => {
```
wdyt? I think it's better, this is analogous to how we name props - we don't name our props types just `Props` or `ComponentProps`, we naming them as `<ComponentNameHere>Props`

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -269,3 +276,73 @@ export const NftWidget: React.FC<NftWidgetProps> = ({ ownerHandle, isOwner, nftS
...
+        ownerAvatarUri,
+        isOwner,
+        nftState: {
+          status: 'idle',
+        },
+        shouldRenderWidget: true,
+      }
+  }
+
+  return { shouldRenderWidget: false }
```
drillprop: Can we just return `null` instead of adding this `shouldRenderWidget` property?

packages/atlas/src/providers/joystream/useJoystream.ts
```diff
@@ -25,18 +25,18 @@ export const useTokenPrice = () => {
         currency: 'USD',
       })
 
-      const price = getTokenPrice()
+      const price = tokenPrice
```
drillprop: We don't need to assign this to another variable, we can just write `formatter.format(tokens * tokenPrice)`

packages/atlas/src/providers/joystream/useJoystream.ts
```diff
@@ -25,18 +25,18 @@ export const useTokenPrice = () => {
...
       return formatter.format(tokens * price)
     },
-    [getTokenPrice]
+    [tokenPrice]
   )
   const convertToTJoy = useCallback(
     (dollars: number) => {
-      const price = getTokenPrice()
+      const price = tokenPrice
```
drillprop: same as above

drillprop (commented):
I just noticed one thing more. I think we overlooked this in your previous PR
image
Buy now shouldn't be visible if the value is set to 0.
You can test this on video with id `6`

eldiegod:
> I just noticed one thing more. I think we overlooked this in your previous PR image Buy now shouldn't be visible if the value is set to 0. You can test this on video with id `6`

Added a patch for this but it's actually an orion bug where it returns buyNowPrice as a string where the graphql types says it should be type number

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -71,7 +71,7 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
     const buttonSize = size === 'small' ? 'medium' : 'large'
     const buttonColumnSpan = size === 'small' ? 1 : 2
     const BuyNow = ({ buyNowPrice }: { buyNowPrice?: number }) =>
-      buyNowPrice ? (
+      buyNowPrice && ~~buyNowPrice ? ( // TODO: remove `&& ~~buyNowPrice` when the payload type of buyNowPrice is fixed and it doesn't return strings anymore
```
drillprop: That's a nice hack, but can we just do a `Number(buyNowPrice)`? It wasn't clear at all to me - I had to google what actually `~~` does 😅 . I don't think it's common javascript knowledge.

drillprop (changes requested):
Just one thing. Besides that, looks good to me. Needs rebase as well.

drillprop (approved):
LGTM, but you need to rebase again, unfortunately.

packages/atlas/src/providers/joystream/provider.tsx
```diff
@@ -147,8 +147,10 @@ const useJoystreamUtilFns = (joystream: Remote<JoystreamLib> | undefined, proxyC
   const getCurrentBlockMsTimestamp = useCallback(() => currentBlockMsTimestampRef.current, [])
 
   return {
+    tokenPrice: tokenPriceRef.current,
```
kdembler: There's a slight problem with this approach: The context will not be updated when this value changes, i.e. changes won't trigger a rerender across users of this value. I think the approach with getter functions is safer since you are always guaranteed to get latest value when you trigger the function

kdembler: Ah but there's problem with my approach as well, the price can be not fetched yet when the component renders so it will get the old value. But current code has the same problem, what can we do to solve that? I guess we could keep the price inside a state instead of a ref?

kdembler: On some refreshes, I will get constant `$0` in the widget

kdembler (approved):

PR #2265, opened 2022-03-10
add BottomDrawer component

kdembler:
This adds a reusable `BottomDrawer` component that can be used for any kind of action overlay that slides from the bottom. This will include:
- Video workspace
- NFT workspace
- NFT purchase view
- Settling an auction
- Anything similar

The code is mostly copied from the video workspace, with some adjustments. The component includes support for:
- Entering/exiting the screen with animation and background overlay
- Managing scroll lock
- Drawer header
- Optional action bar
- Scrollable content inside the overlay, with space reserved for action bar (no need to measure manually)
- Updated tab title when the overlay is opened

You can test it out with the `BottomDrawer` story. I haven't used it in the video workspace yet, since it needs to be slightly rebuilt first, will do that later

### DIFF
--- a/packages/atlas/src/components/_overlays/BottomDrawer/BottomDrawer.styles.ts
+++ b/packages/atlas/src/components/_overlays/BottomDrawer/BottomDrawer.styles.ts
@@ -1,0 +1,79 @@
+import styled from '@emotion/styled'
+
+import { ActionBar } from '@/components/ActionBar'
+import { cVar, zIndex } from '@/styles'
+
+export const DrawerOverlay = styled.div`
+  position: fixed;
+  z-index: ${zIndex.videoWorkspaceOverlay};
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background-color: ${cVar('colorBackgroundOverlay')};
+  transition: opacity ${cVar('animationTransitionSlow')};
+
+  &.bottom-drawer-overlay {
+    &-enter-active,
+    &-exit {
+      opacity: 1;
+      pointer-events: initial;
+    }
+
+    &-enter,
+    &-exit-active {
+      opacity: 0;
+      pointer-events: none;
+    }
+  }
+`
+
+export const Container = styled.div`
+  transform: translateY(100%);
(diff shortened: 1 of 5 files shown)


packages/atlas/src/components/_overlays/BottomDrawer/BottomDrawer.styles.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+`
+
+export const Container = styled.div`
+  transform: translateY(100%);
+  opacity: 0;
+  position: fixed;
+  z-index: ${zIndex.videoWorkspaceOverlay};
+  top: var(--size-topbar-height);
```
rafalpawlow: this value cannot be fixed, e.g. in nft purchase drawer takes up the whole screen

packages/atlas/src/components/_overlays/BottomDrawer/BottomDrawer.styles.ts
```diff
@@ -0,0 +1,79 @@
...
+    &-exit-active {
+      transform: translateY(100%);
+      opacity: 0;
+    }
+  }
+`
+
+type ScrollContainerProps = {
+  actionBarHeight?: number
+  isEdit?: boolean
```
rafalpawlow: I think it's unnecessary, we don't use it anywhere

rafalpawlow (changes requested):

packages/atlas/src/components/_overlays/BottomDrawer/BottomDrawer.styles.ts
kdembler: 👍 I missed that. Added a `coverTopbar` prop that if true will make the drawer fill the full screen height

packages/atlas/src/components/_overlays/BottomDrawer/BottomDrawer.styles.ts
kdembler: Removed

rafalpawlow (approved):

PR #2271, opened 2022-03-10
NFT form date validation

### LINKED ISSUE #1919: Handle date-related validation in NFT issuing
**AC:**

- [ ] Handle validation of dates in past/future
- [ ] Handle summary screen when the start date is in the past - https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1251%3A93666
- [ ] Write unit tests for DatePicker component to test validation

drillprop:
Fix #1919 
Designs: https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1080%3A87052

### DIFF
--- a/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
+++ b/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
@@ -8,14 +8,15 @@ import 'react-datepicker/dist/react-datepicker.css'
 import { SvgControlsCalendar } from '@/components/_icons'
 import { Popover, PopoverImperativeHandle } from '@/components/_overlays/Popover'
 import { cVar } from '@/styles'
+import { EndDate, StartDate } from '@/views/studio/VideoWorkspace/NftForm/types'
 
 import { Select, SelectProps } from '../Select'
 
 export type AuctionDatePickerProps = {
   minDate?: Date | null
-  value: Date | string | null
-  onChange: (value: Date | string | null) => void
-} & Omit<SelectProps<Date | string | null>, 'onChange'>
+  value: StartDate | EndDate
+  onChange: (value: StartDate | EndDate) => void
+} & Omit<SelectProps<StartDate | EndDate>, 'onChange'>
 
 const PICK_DATE = 'pick-date'
 
@@ -25,7 +26,6 @@ export const AuctionDatePicker: React.FC<AuctionDatePickerProps> = ({
   onChange,
   label,
   minDate = new Date(),
-  size = 'small',
   ...rest
 }) => {
   const selectRef = useRef(null)
@@ -64,7 +64,7 @@ export const AuctionDatePicker: React.FC<AuctionDatePickerProps> = ({
   return (
     <Container>
       <Select
-        size={size}
+        size="small"
         label={label}
         labelTextProps={{ variant: 'h100', color: cVar('colorTextMuted'), secondary: true }}
         iconLeft={isPickDate ? <SvgControlsCalendar /> : undefined}
@@ -74,7 +74,18 @@ export const AuctionDatePicker: React.FC<AuctionDatePickerProps> = ({
         ref={selectRef}
         {...rest}
       />
-      <Popover offset={[0, 8]} ref={popOverRef} triggerMode="manual" triggerTarget={selectRef.current} trigger={null}>
+      <Popover
+        offset={[0, 8]}
+        ref={popOverRef}
+        triggerMode="manual"
+        triggerTarget={selectRef.current}
+        trigger={null}
+        onHide={() => {
+          if (value === 'pick-date') {
+            onChange(null)
+          }
+        }}
+      >
         <DatePicker
           open
           inline
--- a/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
+++ b/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
@@ -67,7 +67,7 @@ export const AuctionDatePicker: React.FC<AuctionDatePickerProps> = ({
           ...items,
           {
             value,
-            name: format(new Date(pickedValue?.date || 0), 'd MMM yyyy, HH:mm'),
+            name: format(new Date(pickedValue.date), 'd MMM yyyy, HH:mm'),
             hideInMenu: true,
             pickDateItem,
           },
--- a/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
+++ b/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
@@ -119,7 +119,7 @@ export const AuctionDatePicker: React.FC<AuctionDatePickerProps> = ({
         size="small"
         label={label}
         labelTextProps={{ variant: 'h100', color: cVar('colorTextMuted'), secondary: true }}
-        iconLeft={isPickDate ? <SvgControlsCalendar /> : undefined}
+        iconLeft={isPickDate && pickedValue.type !== 'default' ? <SvgControlsCalendar /> : undefined}
         value={pickedValue || { type: 'default' }}
         items={mappedItems}
         onChange={handleSelect}
--- a/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
+++ b/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
@@ -22,7 +22,7 @@ export type AuctionDatePickerValue =
     }
   | null
 
-type SelectValue = Date | 'pick-date' | 'default' | number | null | undefined
+export type SelectValue = Date | 'pick-date' | 'default' | number | null | undefined
 
 export type AuctionDatePickerProps = {
   minDate?: Date | null
--- a/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
+++ b/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import { addMonths, format } from 'date-fns'
-import { isValid } from 'date-fns/esm'
+import { isEqual } from 'lodash-es'
 import React, { useMemo, useRef, useState } from 'react'
 import DatePicker from 'react-datepicker'
 import 'react-datepicker/dist/react-datepicker.css'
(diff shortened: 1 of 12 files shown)


rafalpawlow (changes requested):
If the difference is less than an hour, no summary is shown
Zrzut ekranu 2022-03-11 o 17 38 23

Also when you pick expiration date earlier than starting date, there is no modal with warning

packages/atlas/src/views/studio/VideoWorkspace/NftForm/AcceptTerms/AcceptTerms.tsx
```diff
@@ -38,17 +37,14 @@ export const AcceptTerms: React.FC<AcceptTermsProps> = ({
...
-  const { getTotalDaysAndHoursText } = useNftForm()
-  const { convertDurationToBlocks } = useBlockTimeEstimation()
+  const { getNumberOfBlocksAndDaysLeft } = useNftForm()
 
-  const startDate = formData.startDate as Date
-  const endDate = formData.endDate as Date
+  const { startDate, endDate } = formData
 
-  const validDate = isValid(formData.startDate) && isValid(formData.endDate)
+  const isStartDateValid = startDate instanceof Date && isDateValid(startDate)
```
rafalpawlow: Why do we need `instanceof` here? Doesn't `isValid` do the job comprehensively? Isn't this something that isValid should take care of?

drillprop: If a user chooses `1 day/3 days/5 days` as the expiration date in the select dropdown, the `endDate` will be set to a `number`.  If you pass primitive number to `isValid` it will return true - I want to avoid that. 

Anyway, I'll use another utility function from `date-fns` - `isDate` which should take care of date validation here and in the other functions.

drillprop: ok, changed my mind. I'll use just `instanceof Date` everywhere. If I start using `isDate` I'll have to add `as Date` to every valid date, which is annoying.

rafalpawlow (approved):

kdembler:
We synced up with @drillprop and we will make some changes here to make those dates easier to consume by the extrinsic

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/AcceptTerms/AcceptTerms.tsx
```diff
@@ -31,26 +30,22 @@ type AcceptTermsProps = {
   toggleTermsAccept: () => void
 }
 
-const DATE_FORMAT = 'dd MMM yyyy, HH:mm OOOO'
+const DATE_FORMAT = 'dd MMM yyyy, HH:mm'
```
kdembler: There is utility `formatDateTime` in `@/utils/time`, can we use that?

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
```diff
@@ -25,6 +29,8 @@ import {
 import { NftFormData, NftFormStatus } from './NftForm.types'
 import { SetUp } from './SetUp'
 
+const DATE_FORMAT = 'dd MMM yyyy, HH:mm'
```
kdembler: Same here

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.hooks.ts
```diff
@@ -9,21 +14,56 @@ export const useNftForm = () => {
...
   const getTotalDaysAndHoursText = (start: Date, end: Date) => {
     const { days, hours } = intervalToDuration({
       start,
       end,
     })
-    const parsedDays = days ? `${days} ${days > 1 ? 'Days' : 'Day'}` : ''
-    const parsedHours = hours ? `${hours} ${hours > 1 ? 'Hours' : 'Hour'}` : ''
+
+    const parsedDays = days ? pluralizeNoun(days, 'Day') : ''
+    const parsedHours = hours !== undefined ? (hours >= 1 ? pluralizeNoun(hours, 'Hour') : 'Less than an hour') : ''
```
kdembler: This is a bit error prone, for example it can produce "Total: 8 Days Less than an hour" which isn't very correct. I think we don't need to do this manually, we can use for example `formatDistance` from `date-fns`

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.hooks.ts
```diff
@@ -9,21 +14,56 @@ export const useNftForm = () => {
...
   const previousStep = useCallback(() => setCurrentStep((step) => step - 1), [])
 
+  const getNumberOfBlocksAndDaysLeft = (startDate: AuctionDatePickerValue, endDate: AuctionDatePickerValue) => {
+    const isStartDateAndEndDateValid = startDate?.type === 'date' && endDate?.type === 'date'
+    const now = new Date(Date.now())
+
+    if (isStartDateAndEndDateValid) {
+      return {
+        blocks: convertDurationToBlocks(differenceInMilliseconds(endDate.date, startDate.date)),
+        daysAndHoursText: getTotalDaysAndHoursText(startDate.date, endDate.date),
```
kdembler: As a follow up to my point above, I think it's better for the UI to handle text to display, the hook should just return raw data

packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
```diff
@@ -1,93 +1,135 @@
...
-              setStartDate(date)
-              onChange?.(date)
-            }
-          }}
+          timeFormat="HH:mm"
+          onChange={handlePickDate}
           openToDate={minDate ?? undefined}
           minDate={minDate} // TODO: value to be discussed
-          maxDate={addMonths(new Date(), 5)} // TODO: value to be discussed
+          maxDate={maxDate || addMonths(new Date(), 5)} // TODO: value to be discussed
```
kdembler: I think both of those have to be defined by the consumer, not the component itself. Let's remove the `addMonths` and `TODO` comments and discuss that in `SetUp` usage

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
```diff
@@ -154,66 +163,66 @@ export const SetUp: React.FC<SetUpProps> = ({
...
+                <Controller
+                  name="startDate"
+                  control={control}
+                  render={({ field: { onChange, value }, fieldState: { error } }) => (
+                    <AuctionDatePicker
+                      size="regular"
+                      label="Starting date"
+                      error={!!error}
+                      helperText={error?.message}
+                      minDate={new Date()}
```
kdembler, line 188: Maybe we add `useMsTimestamp` and supply `minDate={new Date(msTimestamp)}` to make sure it updates?

kdembler: Although need to make sure it doesn't break anything

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
```diff
@@ -154,66 +163,66 @@ export const SetUp: React.FC<SetUpProps> = ({
...
+                      maxDate={(endDate?.type === 'date' && endDate.date) || undefined}
```
kdembler: There is a runtime parameter that dictates how far in the future the auction can start. You can access this value by `api.query.content.auctionStartsAtMaxDelta()` from the Polkadot API. This returns a number of blocks. We will need to take this into account here and also in validation. I'm aware this wasn't mentioned in the initial scope, so possibly let's do it as a followup to not drag this PR for too long

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
```diff
@@ -154,66 +163,66 @@ export const SetUp: React.FC<SetUpProps> = ({
...
+                <Controller
+                  name="endDate"
+                  control={control}
+                  render={({ field: { onChange, value }, fieldState: { error } }) => (
+                    <AuctionDatePicker
+                      size="regular"
+                      label="expiration date"
+                      error={!!error}
+                      helperText={error?.message}
+                      minDate={(startDate?.type === 'date' && startDate.date) || new Date()}
```
kdembler, line 211: Likewise to the point above, there is also a parameter that dictates for how long the auction can run, `api.query.content.maxAuctionDuration()`, we can also handle this in a followup

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
```diff
@@ -154,66 +163,66 @@ export const SetUp: React.FC<SetUpProps> = ({
...
                 <Text variant="t200-strong" color={cVar('colorTextMuted', true)}>
                   Total:
                 </Text>
                 &nbsp;
-                <Text variant="t200-strong">{daysAndHours}</Text>
+                <Text variant="t200-strong">{numberOfBlocksAndDaysLeft.daysAndHoursText}</Text>
                 &nbsp;
                 <Text variant="t200-strong" secondary>
-                  / {numberOfBlocks?.toLocaleString('no', { maximumFractionDigits: 1 })} Blocks
+                  / {numberOfBlocksAndDaysLeft.blocks?.toLocaleString('no', { maximumFractionDigits: 1 })} Blocks
```
kdembler: Let's update the implementation of `formatNumber` in `@/utils/number` and use it here

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
drillprop: opened https://github.com/Joystream/atlas/issues/2303

packages/atlas/src/utils/number.ts
```diff
@@ -1,5 +1,5 @@
-export const formatNumber = (num: number): string => {
-  return num.toLocaleString('en-US').split(',').join(' ')
+export const formatNumber = (num: number, opts?: Intl.NumberFormatOptions): string => {
+  return num.toLocaleString('no', opts).split(',').join(' ')
```
kdembler: The `.split.join` shouldn't be necessary anymore, that's handled by the format

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
```diff
@@ -79,22 +77,37 @@ export const NftForm: React.FC<NftFormProps> = ({ setFormStatus, onSubmit, video
...
+      )
+      .refine(
+        (val) => {
+          const endDate = getValues('endDate')
+          if (val?.type === 'date' && endDate?.type === 'date') {
+            if (endDate.date > val.date) {
+              return true
+            } else {
+              return false
+            }
```
kdembler: This can be just `return endDate.date > val.date`

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
```diff
@@ -79,22 +77,37 @@ export const NftForm: React.FC<NftFormProps> = ({ setFormStatus, onSubmit, video
...
       z.object({
         type: z.literal('duration'),
-        durationDays: z.number().nullable(),
+        durationDays: z.number(),
       }),
     ])
     .nullable()
     .optional()
 
   const schema = z.object({
```
kdembler: Could we maybe move this to a function that creates this schema given necessary values from the form? It's crowding the component a bit

kdembler (approved):

Issue #2188, opened 2022-02-25
Handle starting NFT auction as non-video-owner


kdembler:
We will need to be able to trigger the `NftForm` from the video page to allow the NFT owner to start an auction when they aren't the owner of the video itself.

kdembler:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Atlas-600edbb8799cc700103bc2d7/issues/Joystream/atlas/2188?planning-poker) with ZenHub @drillprop @eldiegod @lcoenengithub @rafalpawlow

PR #2281, opened 2022-03-11
create NftSaleBottomDrawer, clean up VideoWorkspace

kdembler:
This PR splits starting NFT sale from the video workspace, allowing it to be triggered in viewer context.

**Scope of changes:**

- Created `NftSaleBottomDrawer` independent of video edit
- Added NFT royalties field to `VideoWorkspace`
- Used `BottomDrawer` in `VideoWorkspace`, cleaned it up

Resolves: #2188


packages/atlas/src/providers/nftActions/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+    [setCurrentAction, setCurrentNftId]
+  )
+
+  const openNftSettlement = useCallback(
+    (nftId: string) => {
+      setCurrentNftId(nftId)
+      setCurrentAction('settle')
+    },
+    [setCurrentAction, setCurrentNftId]
+  )
```
drillprop, line 35: We're creating three functions that basically do the same. Is there any reason why we do that? Can we make just one and pass, `settle, purchase` or `putOnSale` as a second argument?

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
```diff
@@ -73,183 +49,58 @@ export const VideoWorkspace: React.FC = React.memo(() => {
...
-            canSubmit={!!formStatus?.isDisabled}
-            canReset={formStatus?.isDirty || isIssuedAsNft || false}
-            onPrimaryButtonClick={formStatus?.triggerFormSubmit}
-            onSecondaryButtonClick={onSecondaryButtonClick}
-            onResize={setActionBarHeight}
-          />
-        </Container>
-      </CSSTransition>
-    </>
+    <BottomDrawer
```
drillprop, line 94: With the new bottom drawer, we will experience a layout shift when the scrollbar shows up. I think we should make the scrollbar always visible as we did previously. 

https://user-images.githubusercontent.com/51168865/157914817-c21cb1c0-3a5e-4aa0-8aa3-cd4bd7222cb5.mov

drillprop (changes requested):
Looks good, found two minor things

packages/atlas/src/providers/nftActions/hooks.ts
kdembler: Don't know, just thought that maybe in the future we may want to pass separate params to different types of actions. I guess it doesn't hurt us to have them separate

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.tsx
kdembler: Good catch! I've added a `fixedScrollbar` prop to `BottomDrawer` that will set `overflow-y: scroll`. Did that because we shouldn't have the scrollbar there always since some of the drawer contents will not need scrolling

packages/atlas/src/providers/nftActions/hooks.ts
drillprop: ok, that sounds fair.

rafalpawlow (changes requested):
Something is wrong with primary button, when you edit video and toggle `Mint an NFT`, primary button has wrong text.

https://user-images.githubusercontent.com/13706246/158164123-23ff4cc4-ecf8-483a-a71d-9dd91ed5e80c.mov

kdembler:
What text would you expect here?

rafalpawlow:
When you are uploading new video there is `Publish & mint`

kdembler:
Yes, changed the condition slightly so it's always `Publish & mint` if you toggle the NFT switch

drillprop (approved):

rafalpawlow (approved):

PR #2291, opened 2022-03-14
handle remove from sale

### LINKED ISSUE #1754: Cancel NFT sale from NFT tile
**AC**

- [ ] [Remove from sale](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=972%3A224906)

rafalpawlow:
Closes #1754 

**Design**
https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?node-id=972%3A218668


packages/atlas/src/components/NftTile/NftTileDetails.tsx
```diff
@@ -63,11 +68,18 @@ export const NftTileDetails: React.FC<NftTileDetailsProps> = ({
...
   const { copyToClipboard } = useClipboard()
   const [contentHovered, setContentHovered] = useState(false)
   const toggleContentHover = () => setContentHovered((prevState) => !prevState)
   const [tileSize, setTileSize] = useState<TileSize>()
+  const { canCancelSale, canPutOnSale, canBuyNow, canMakeBid, videoId, isBuyNow } = useNftState(nftId)
+  const { refetch } = useNft(videoId || '')
+  const [removeFromSaleDialogOpen, setRemoveFromSaleDialogOpen] = useState(false)
+  const { activeMemberId } = useAuthorizedUser()
+  const { joystream, proxyCallback } = useJoystream()
+  const handleTransaction = useTransaction()
```
kdembler: I think those shouldn't be in this component. So far both `NftTile` and `NftTileDetails` have been dumb components, only working based on props received. This makes them smart and quite a lot. We probably should pass all the necessary data via props and include all the logic/hooks in `NftTileViewer` instead

packages/atlas/src/hooks/useNftState.ts
```diff
@@ -0,0 +1,65 @@
+import { useMemo } from 'react'
+
+import { useNft } from '@/api/hooks'
+import { useUser } from '@/providers/user'
+
+export const useNftState = (id?: string) => {
```
kdembler: I think this could accept the nft data directly, the user of this hook should be responsible for fetching it already

packages/atlas/src/hooks/useNftState.ts
```diff
@@ -0,0 +1,65 @@
...
+  const { nft, loading } = useNft(id || '')
+  const { memberships } = useUser()
+
+  const nftFetched = nft && !loading
+  const isOwner = !memberships?.find((membership) => membership.handle == nft?.ownerMember?.handle)
+
+  const isBuyNow = useMemo(() => {
```
kdembler: What's the point of all the `useMemo` here? We don't do any complicated operations and we have stable references anyway since all of those are booleans

kdembler (changes requested):

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
 } from '@/api/queries'
 import { createLookup } from '@/utils/data'
 
-export const useNft = (id: string) => {
+export type NftStatus =
+  | {
+      status: 'auction'
+      startingPrice: number
+      buyNowPrice?: number
+      topBid?: number
```
kdembler, line 20: Probably needs to include the member as well

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
 } from '@/api/queries'
 import { createLookup } from '@/utils/data'
 
-export const useNft = (id: string) => {
+export type NftStatus =
+  | {
+      status: 'auction'
```
drillprop, line 16: Can we also add here type of auction - i.e. `open-auction` and `english-auction`?

packages/atlas/src/components/NftTile/NftTile.tsx
```diff
@@ -17,7 +17,7 @@ export type Member = {
 }
 
 export type NftTileProps = {
-  nftStatus?: 'idle' | 'on-sale' | 'auction'
+  status?: 'idle' | 'on-sale' | 'auction'
```
drillprop: I'm aware that it's not your change, but I'm confused here. What means `on-sale`? Is it `buy-now`? If yes, can we make this consistent with changes in `useNft`?

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
...
+          ...commonProperties,
+          status: 'idle',
+        }
+    }
+  }
 
   return {
-    nft: data?.ownedNftByUniqueInput,
+    nft,
+    nftStatus: getNftProperies(),
```
drillprop: typo:
```suggestion
    nftStatus: getNftProperties(),
```

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
...
   const { data, ...rest } = useGetNftQuery({ variables: { id }, skip: !id })
+  const nft = data?.ownedNftByUniqueInput
+
+  const commonProperties = {
+    title: nft?.video?.title,
+    duration: nft?.video?.duration,
+    views: nft?.video?.views,
+  }
+
+  const getNftProperies = (): NftStatus => {
```
drillprop: typo:
```suggestion
  const getNftProperties = (): NftStatus => {
```

drillprop (changes requested):

packages/atlas/src/components/NftTile/NftTile.tsx
eldiegod: yes `on-sale` means that you can buy it now

drillprop: hmm, so I think `buy-now` will be a much better name.

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
...
+      duration?: number | null
+      views?: number
+      auctionPlannedEndBlock?: number
+      needsSettling?: boolean
+    }
+  | {
+      status: 'idle'
+      lastPrice?: number
+      lastTransactionDate?: Date
+      auctionPlannedEndBlock?: number
```
eldiegod: Why do we need auctionPlannedEndBlock on a non auction state?

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
...
+    }
+  | {
+      status: 'buy-now'
+      buyNowPrice: number
+      auctionPlannedEndBlock?: number
```
eldiegod: same as above

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -46,8 +47,21 @@ export type NftWidgetProps = {
   ownerAvatarUri?: string | null
   isOwner?: boolean
   nftStatus?:
-    | { status: 'idle'; lastPrice?: number; lastTransactionDate?: Date }
-    | { status: 'buy-now'; buyNowPrice: number }
+    | {
+        status: 'idle'
+        lastPrice?: number
+        lastTransactionDate?: Date
+        needsSettling?: boolean
```
eldiegod: If these last 3 props are common to all nft statuses we can just move them up to the general NftWidgetProps

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
...
 import { createLookup } from '@/utils/data'
 
-export const useNft = (id: string) => {
+export type NftStatus =
+  | {
+      status: 'auction'
+      startingPrice: number
+      buyNowPrice?: number
+      topBid?: number
+      isCompleted?: boolean
```
kdembler: This will be set only for finished auction which should never make their way into active transactional status, I think we can remove this field

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
...
+      title?: string | null
+      duration?: number | null
+      views?: number
```
kdembler, line 24: Those aren't included in other variants of this type

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +10,73 @@ import {
...
+      auctionPlannedEndBlock?: number
+      needsSettling?: boolean
```
kdembler: Same here, not used

packages/atlas/src/components/NftTile/NftTile.tsx
eldiegod: so I don't agree because the state in which the nft is is called being on sale the buy now is the action you can take, is like if you call the auction `bid-now`, but granted that we have `TransactionalStatusBuyNow` then i guess it would make sense to call it the same as that 🤷‍♂️

packages/atlas/src/api/hooks/nfts.ts
kdembler: Planned end doesn't exist for `idle` state

packages/atlas/src/api/hooks/nfts.ts
kdembler: Planned end doesn't exist for `buy-now` state

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -274,18 +288,17 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
 
 type UseNftWidgetReturn = NftWidgetProps | null
 export const useNftWidget = (videoId?: string): UseNftWidgetReturn => {
```
kdembler, line 284: Seems like almost everything that is in this hook should now just rely on `useNftState`?

eldiegod (changes requested):
There is a couple more states from the nfts that we are going to need, I suggest taking a look at my PR for the updated types
https://github.com/Joystream/atlas/pull/2304/files#diff-b713a61de81d4417e5d506e571b7f2a11d0f02ce71f84a797faf398b22e9bf52

packages/atlas/src/components/NftTile/NftTile.tsx
drillprop: I'm not sure if I understand. So, you're saying that `on-sale` means either `TransactionalStatusBuyNow ` or `TransactionalStatusAuction` **but with buy now price set**, right? 
If yes, this is very confusing. Maybe we could differentiate this even more - status could be `idle` or `only-buy-now` or `auction-buy-now` or `auction`?

eldiegod:
no, I mean that the states that the Nft can be in are `idle` you can't do
much, `on-sale` you can buy now for a fixed price; and `auction` you can
bid and also buy now depending on params.

What I mean is that the wording `buy-now` is not a state of the nft itself
but an action taken by the user just like bidding is and `auction` would be
the state, does that make sense?

packages/atlas/src/components/NftTile/NftTile.tsx
kdembler: I think we should reflect transactional status. So `idle`, `auction`, `buy-now`, I think it's quite clear. Potentially we could split auction into separate auction types if we wanted but up to discussion

kdembler:
Diego, `buy-now` is a state though, as defined by runtime and query node. I agree it may not be the most fortunate name, but I think it may be the best to follow that

eldiegod:
> s defined by runtime and query node. I agree it may not be the most fortunate name, but I think it may be the best to follow that

Yeah, that's what I meant 2 comments back, it's called like that in runtime and query node but it's confusing as well as you can also `buy-now` in auction. I don't mind just making it clear

packages/atlas/src/components/NftTile/NftTile.tsx
eldiegod: I wouldn't split auction but add a type inside of 'open' | 'english', did this in my PR with the new changes

kdembler: That works as well

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +11,78 @@ import {
...
+  | {
+      status: 'idle'
+      lastPrice?: number
+      lastTransactionDate?: Date
+      views?: number
+    }
+  | {
+      status: 'buy-now'
+      buyNowPrice: number
+      views?: number
```
kdembler: `duration` and `title` are missing from those variants but are returned in the data. If they are needed, let's include them

packages/atlas/src/components/NftTile/NftTile.tsx
```diff
@@ -30,13 +30,17 @@ export type NftTileProps = {
...
   topBid?: number | null
   timeLeftMs?: number
-  role: 'owner' | 'viewer'
   fullWidth?: boolean
   interactable?: boolean
+  canPutOnSale?: boolean
+  canCancelSale?: boolean
+  canBuyNow?: boolean
+  canMakeBid?: boolean
+  handleRemoveFromSale?: () => void
```
kdembler: We use `on*` for props (we should at least 😅)

packages/atlas/src/components/NftTile/NftTileDetails.tsx
```diff
@@ -36,14 +36,19 @@ export type NftTileDetailsProps = {
...
-  nftStatus?: 'idle' | 'on-sale' | 'auction'
+  nftStatus?: 'idle' | 'buy-now' | 'auction'
   buyNowPrice?: number | null
   startingPrice?: number | null
   topBid?: number | null
   title?: string | null
   hovered?: boolean
   interactable?: boolean
   videoHref?: string
+  handleRemoveFromSale?: () => void
```
kdembler: Same here, `onRemoveFromSale`

packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
```diff
@@ -16,20 +18,41 @@ type NftTileViewerProps = {
...
+          cancelNftSale(false)
+          closeModal()
+        },
+      },
+      secondaryButton: {
+        variant: 'secondary',
+        text: 'Cancel',
+        onClick: () => closeModal(),
+      },
+    })
```
kdembler: Would make most sense for this modal to live in some kind of hook so that it can be reused for widget/tile/etc.

packages/atlas/src/hooks/useNftState.ts
```diff
@@ -0,0 +1,92 @@
...
+    (isBuyNow ||
+      (nft.transactionalStatus.__typename === 'TransactionalStatusAuction' &&
+        !!nft.transactionalStatus.auction?.buyNowPrice))
+
+  const canMakeBid = nft && !isOwner && nft.transactionalStatus.__typename === 'TransactionalStatusAuction'
+
+  const canCancelSale =
+    nft &&
+    isOwner &&
+    ['TransactionalStatusAuction', 'TransactionalStatusBuyNow'].includes(nft.transactionalStatus.__typename)
```
kdembler: English auction can be cancelled only if no bids have been made yet

packages/atlas/src/hooks/useNftState.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const userBid =
+    nft &&
+    nft.transactionalStatus.__typename === 'TransactionalStatusAuction' &&
+    nft.transactionalStatus?.auction?.bids.find((bid) => !bid.isCanceled && bid.bidder.id === activeMembership?.id)
+
+  const isOwner = nft?.ownerMember?.id === activeMembership?.id
+
+  const isBuyNow = nft && nft?.transactionalStatus?.__typename === 'TransactionalStatusBuyNow'
+  const isAuction = nft && nft.transactionalStatus.__typename === 'TransactionalStatusAuction'
```
kdembler: You define those handy vars but later in the code keep checking the full `nft?.transactionalStatus?.__typename`, let's make use of them

packages/atlas/src/hooks/useNftTransactions.ts
```diff
@@ -0,0 +1,32 @@
...
+import { useTransaction } from '@/providers/transactionManager'
+import { useAuthorizedUser } from '@/providers/user'
+
+export const useNftTransactions = (videoId?: string) => {
+  const { refetch } = useNft(videoId || '')
+  const { activeMemberId } = useAuthorizedUser()
+  const { joystream, proxyCallback } = useJoystream()
+  const handleTransaction = useTransaction()
+
+  const cancelNftSale = (isBuyNow?: boolean, cb?: () => Promise<unknown>) => {
```
kdembler: Could this go into the `useNftActions` hook as well?

kdembler: Should use `useCallback`

kdembler (changes requested):
Shaping up, really like it

packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
kdembler: It can be a part of `cancelNftSale`

packages/atlas/src/components/NftTile/NftTile.tsx
rafalpawlow: Sorry, it's not the first time. I was in the project before where we had rule for `handle*` 😅

kdembler: No worries. `on*` for props, `handle*` for handlers

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +11,82 @@ import {
...
+    }
+  | {
+      title?: string | null
+      duration?: number | null
+      status: 'buy-now'
+      buyNowPrice: number
+      views?: number
+    }
+export type Nft = AllNftFieldsFragment | null
```
kdembler: Do we need this type? I'm afraid it may be a bit ambigous what it represents

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -7,11 +11,82 @@ import {
...
+      case 'TransactionalStatusAuction': {
+        return {
+          ...commonProperties,
+          status: 'auction',
+          type:
+            nft.transactionalStatus.auction?.auctionType.__typename === 'AuctionTypeOpen'
+              ? 'open-auction'
+              : 'english-auction',
+          startingPrice: Number(nft.transactionalStatus.auction?.startingPrice) ?? 0,
+          buyNowPrice: Number(nft.transactionalStatus.auction?.buyNowPrice) ?? undefined,
```
kdembler: If `buyNowPrice` is missing, this will produce `NaN`. You could fix it by changing to `||` instead of `??`

packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
```diff
@@ -16,20 +17,27 @@ type NftTileViewerProps = {
...
+  const thumbnail = useAsset(nft?.video?.thumbnailPhoto)
+  const creatorAvatar = useAsset(nft?.video?.channel.avatarPhoto)
+  const { canPutOnSale, canMakeBid, canCancelSale, canBuyNow } = useNftState(nft)
+  const { cancelNftSale } = useNftTransactions()
+
+  const onRemoveFromSale = () => {
+    if (!nftId || !nft?.video.id) {
+      return
+    }
+    cancelNftSale(nftId, nft.video.id, false)
```
kdembler: nft ID and video ID are the same, you need only single one

packages/atlas/src/components/NftTileViewer/NftTileViewer.tsx
```diff
@@ -16,20 +17,27 @@ type NftTileViewerProps = {
...
+  const { nftStatus, nft, loading } = useNft(nftId || '')
   const navigate = useNavigate()
-  const thumbnail = useAsset(nft?.video.thumbnailPhoto)
-  const creatorAvatar = useAsset(nft?.video.channel.avatarPhoto)
+  const thumbnail = useAsset(nft?.video?.thumbnailPhoto)
+  const creatorAvatar = useAsset(nft?.video?.channel.avatarPhoto)
+  const { canPutOnSale, canMakeBid, canCancelSale, canBuyNow } = useNftState(nft)
+  const { cancelNftSale } = useNftTransactions()
+
+  const onRemoveFromSale = () => {
```
kdembler: `handleRemoveFromSale` :D

packages/atlas/src/hooks/useNftState.ts
```diff
@@ -0,0 +1,68 @@
...
+  const isAuction = !!auction
+
+  const canBuyNow = nft && !isOwner && (isBuyNow || !!auction?.buyNowPrice)
+
+  const canMakeBid = nft && !isOwner && isAuction
+
+  const canCancelSale =
+    auction?.auctionType.__typename === 'AuctionTypeEnglish'
+      ? !auction?.bids.length
+      : auction?.auctionType.__typename === 'AuctionTypeOpen'
```
kdembler: This should also include buy now status

packages/atlas/src/hooks/useNftTransactions.ts
```diff
@@ -0,0 +1,70 @@
...
+        handleTransaction({
+          txFactory: async (updateStatus) =>
+            (await joystream.extrinsics).cancelNftSale(
+              videoId,
+              activeMemberId,
+              isBuyNow || false,
+              proxyCallback(updateStatus)
+            ),
+          onTxSync: async (_) => _refetchData(id),
+          onTxFinalize: cb ? () => cb() : undefined,
```
kdembler: This isn't needed

packages/atlas/src/hooks/useNftState.ts
kdembler: Only owner can cancel sale

packages/atlas/src/hooks/useNftTransactions.ts
```diff
@@ -0,0 +1,70 @@
...
+export const useNftTransactions = () => {
+  const { activeMemberId } = useAuthorizedUser()
+  const { joystream, proxyCallback } = useJoystream()
+  const handleTransaction = useTransaction()
+  const [openModal, closeModal] = useConfirmationModal()
+  const client = useApolloClient()
+
+  const _refetchData = useCallback(
+    (id: string) => {
+      client.query<GetNftQuery, GetNftQueryVariables>({
```
kdembler, line 19: Let's add `return` here. If you don't return the promise, then it cannot be awaited and `useTransaction` won't be able to catch error if it happens

kdembler (approved):
Leaving an approve here so my review doesn't block merging. However, let's get approval from @drillprop and @eldiegod before merging

drillprop (approved):
Looks good to me! Great work

eldiegod (approved):
Looks good, maybe we can revisit the part about having 3 hooks for nfts later

PR #2292, opened 2022-03-14
Fix ending overlay bug and update styled

### LINKED ISSUE #2213: Bug in summary screen for embedded player
Screenshot 2022-03-02 at 17 33 32

rafalpawlow:
Fixes #2213 

**Design**
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2090%3A119949

### DIFF
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.styles.ts
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.styles.ts
@@ -5,6 +5,7 @@ import { Text } from '@/components/Text'
 import { Button } from '@/components/_buttons/Button'
 import { IconButton } from '@/components/_buttons/IconButton'
 import { ChannelLink } from '@/components/_channel/ChannelLink'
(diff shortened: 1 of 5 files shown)


packages/atlas/src/components/_video/VideoPlayer/VideoOverlays/EndingOverlay.styles.ts
```diff
@@ -107,27 +108,29 @@ export const StyledChannelLink = styled(ChannelLink)<StyledChannelLinkProps>`
     margin-top: ${({ noNextVideo }) => (noNextVideo ? sizes(2) : sizes(4))};
   }
 
+  ${TitleWrapper} {
```
kdembler: We shouldn't access nested element on `ChannelLink` like this, we don't want this view to break if something changes inside the component. If this style is really necessary, let's apply it in `ChannelLink` based on some prop

rafalpawlow:
@kdembler Fixed, by the way I added story for ChannelLink

kdembler (approved):

Issue #2195, opened 2022-02-28
Expired English auction state for NFT components


kdembler:
### Description

While developing some of the NFT components we've noticed that they're missing states for an English auction that has expired. What I mean by that is:
- the auction has finished - it's past its end block and new bids are not accepted
- at the same time, the auction hasn't been settled yet.

The following components are affected:
1. `NftTile` doesn't seem to have any variant for that case
2. `NftWidget` (displayed on Video page) has a state for when you (as a buyer) have won and the auction is awaiting settlement. However, it's not specified what should be presented in this case to:
a) NFT owner
b) random user not participating in the auction

One additional case to handle is when the auction has expired but there were no bids made to it. In that case, the auction needs to be cancelled by the owner.

### How urgent this is?

It's blocking me right now

toiletgranny:
This has been addressed in https://github.com/Joystream/atlas/issues/2182, see examples below.

![image](https://user-images.githubusercontent.com/35307309/157063574-3073b09c-f849-4ced-861b-dc95c8c5722f.png)
![image](https://user-images.githubusercontent.com/35307309/157064210-9f2e5517-04f5-4ac0-8e95-2ca90d19e616.png)
![image](https://user-images.githubusercontent.com/35307309/157063620-416bb6e4-804e-4025-aa37-36ae55e78490.png)

Issue #2196, opened 2022-02-28
NFT Widget doesn't mention current auction winner


kdembler:
### Description

The NFT Widget component shows the NFT auction status, including the current top bid. However, it doesn't show the current auction winner, i.e. who has made the current top bid. If I'm participating in the auction it would be useful to see at a glance whether I'm currently the winner, without analyzing the NFT's history.

### Figma link

https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1242%3A109726

### How urgent this is?

This will block me soon

toiletgranny:
This has been addressed in https://github.com/Joystream/atlas/issues/2182, see the preview below.

![image](https://user-images.githubusercontent.com/35307309/157063142-bb8eb7ff-3367-47b7-91e3-a5d6ed6e1c4b.png)

Issue #2180, opened 2022-02-23
UI for when an auction will start in the future - NFT Widget


eldiegod:
### Description

Fairly self-explanatory we need a UI state for the NFT widget to show that an auction exists but has not yet started.

### Figma link

https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1242%3A109695

### How urgent this is?

This is blocking me now

kdembler:
I think the same goes for NftTile?

toiletgranny:
This has been addressed in https://github.com/Joystream/atlas/issues/2182, see examples below. Please note, that this also requires changes to the NFT tile component.

![image](https://user-images.githubusercontent.com/35307309/157064535-5cbb9fb1-4ab5-4544-af2e-a5e901a396fa.png)
![image](https://user-images.githubusercontent.com/35307309/157064643-a3b83123-beec-4f65-8816-97d5d97e1541.png)

PR #2304, opened 2022-03-16
Widget design changes

eldiegod:
Fixes: #2195  #2180  #2196

Missing small text below some buttons

### DIFF
--- a/packages/atlas/src/api/hooks/nfts.ts
+++ b/packages/atlas/src/api/hooks/nfts.ts
@@ -39,7 +39,7 @@ export type NftStatus = (
 ) &
   CommonNftProperties
 
-export type UseNftData = Omit<QueryResult, 'data'> & { nft?: AllNftFieldsFragment | null; nftStatus: NftStatus }
+export type UseNftData = Omit<QueryResult, 'data'> & { nft?: AllNftFieldsFragment | null; nftStatus?: NftStatus }
 
 export const useNft = (id: string): UseNftData => {
   const { data, ...rest } = useGetNftQuery({ variables: { id }, skip: !id })
@@ -51,7 +51,7 @@ export const useNft = (id: string): UseNftData => {
     views: nft?.video?.views,
   }
 
-  const getNftProperties = (): NftStatus => {
+  const getNftProperties = (): NftStatus | undefined => {
     switch (nft?.transactionalStatus.__typename) {
       case 'TransactionalStatusAuction': {
         return {
@@ -71,11 +71,13 @@ export const useNft = (id: string): UseNftData => {
           status: 'buy-now',
           buyNowPrice: Number(nft.transactionalStatus.price),
         }
-      default:
+      case 'TransactionalStatusIdle':
         return {
           ...commonProperties,
           status: 'idle',
         }
+      default:
+        return undefined
     }
   }
 
--- a/packages/atlas/src/api/hooks/nfts.ts
+++ b/packages/atlas/src/api/hooks/nfts.ts
@@ -26,7 +26,6 @@ export type NftStatus = (
       topBid: number | undefined
       topBidder: BasicMembershipFieldsFragment | undefined
       auctionPlannedEndBlock?: number
-      startsAtDate: Date | undefined
     }
   | {
       status: 'idle'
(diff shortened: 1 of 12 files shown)


packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -277,48 +433,86 @@ export const useNftWidget = (videoId?: string): UseNftWidgetReturn => {
...
   const { url: ownerAvatarUri } = useMemberAvatar(owner)
+  const { url: topBidderAvatarUri } = useMemberAvatar(
+    nft?.transactionalStatus.__typename === 'TransactionalStatusAuction'
+      ? nft?.transactionalStatus?.auction?.lastBid?.bidder
+      : undefined
+  )
 
   const isOwner = owner?.id === activeMembership?.id
 
+  //TODO:
```
drillprop: what means `TODO` here?

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -277,48 +433,86 @@ export const useNftWidget = (videoId?: string): UseNftWidgetReturn => {
...
+      const auctionPlannedEndDate = nft.transactionalStatus.auction?.plannedEndAtBlock
+        ? new Date(convertBlockToMsTimestamp(nft.transactionalStatus.auction?.plannedEndAtBlock))
+        : undefined
+      const englishTimerState: EnglishTimerState = isExpired
+        ? 'expired'
+        : isRunning
+        ? 'running'
+        : isUpcoming
+        ? 'upcoming'
+        : null
```
drillprop: Let's merge Rafal's PR with useNft hook first and clean this up.

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -147,39 +196,100 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
...
           <>
             {nftStatus.topBid ? (
               <NftInfoItem
                 size={size}
                 label="Top bid"
                 content={
                   <>
-                    <JoyTokenIcon size={size === 'small' ? 16 : 24} variant="silver" />
+                    <TopBidderContainer>
+                      <Avatar assetUrl={ownerAvatarUri} size="bid" />
```
drillprop, line 299: It's hard to spot, but there is a small outline that comes from avatar, 
image
We can easily fix this by changing TopBidderTokenContainer zIndex

drillprop (commented):
In general, looks good, but in my opinion, we should merge for Rafał's PR first, and then do some refactoring to `useNftWidget` hook(or even remove it).

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
eldiegod: ahh, this one was driving me crazy, thanks for the tip

eldiegod:
Couldn't add the logic for `bidFromPreviousAuction` as can't issue or bid on nfts

packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -11,35 +11,33 @@ import {
...
       startingPrice: number
-      buyNowPrice?: number
-      topBid?: number
-      topBidder?: BasicMembershipFieldsFragment
-      title?: string | null
-      duration?: number | null
-      views?: number
+      buyNowPrice: number | undefined
+      topBid: number | undefined
+      topBidder: BasicMembershipFieldsFragment | undefined
```
drillprop, line 27: I'm kinda don't understand this. Why did you get rid of question mark here and replaced it with `undefined`?  I think I  don't understand your intentions here - could you explain?

drillprop (changes requested):
Found two bugs, look at these videos:

https://user-images.githubusercontent.com/51168865/159445554-372618cb-c7fc-48b9-bbed-d2e5fbe4ef0a.mov


https://user-images.githubusercontent.com/51168865/159445583-8cb5fba7-7195-4aa7-81ba-4a80c5ab2f00.mov

You can test them out on videos with id: 20 and 21

packages/atlas/src/api/hooks/nfts.ts
eldiegod: The reasoning behind this is that I want these props to `always` be passed from the `useNftWidget` hook, they can be undefined that's fine, but since they are so many it's easy to forget to add a variable in there and the compiler wouldn't complain if we marked them as optional.

So in this case, if you forget to pass a variable the compiler will yell at you, so it's a way of preventing future bugs in this component where we don't pass props manually

drillprop: hmm okay, just tried to mess around with this hook, and now makes more sense. Thank you for the explanation!

drillprop (approved):

PR #2311, opened 2022-03-17
integrate NFT minting and sale extrinsics

### LINKED ISSUE #1922: Issue NFT transaction Handling
**Context**

This issue encapsulates the set of APi calls and responses handling across Atlas and other services to publish the nft. From user POV there is a requirement to sign transaction with the wallet to make it happen. In the very end of the flow the Issue NFT flow closes and user is shown a confirmation toaster with the new status of NFT. 

It is a single transaction and we can bundle all of the actions:

- Publish Video
- Issue as NFT
- Put to Auction

kdembler:
Resolves #1922

### DIFF
--- a/packages/atlas/src/joystream-lib/helpers.ts
+++ b/packages/atlas/src/joystream-lib/helpers.ts
@@ -5,7 +5,9 @@ import {
   AuctionType,
   EnglishAuctionDetails,
   InitTransactionalStatus,
+  NftIssuanceParameters,
   OpenAuctionDetails,
+  Royalty,
   StorageAssets,
 } from '@joystream/types/content'
 import { DataObjectCreationParameters } from '@joystream/types/storage'
@@ -242,7 +244,7 @@ export const createNftAuctionParams = (registry: Registry, inputMetadata: NftAuc
   throw new JoystreamLibError({ name: 'UnknownError', message: `Unknown auction type`, details: { inputMetadata } })
 }
 
-export const createNftIssuanceTransactionalStatus = (
+const createNftIssuanceTransactionalStatus = (
   registry: Registry,
   inputMetadata: NftIssuanceInputMetadata
 ): InitTransactionalStatus => {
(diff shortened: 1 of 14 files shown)


packages/atlas/src/joystream-lib/helpers.ts
```diff
@@ -257,3 +259,21 @@ export const createNftIssuanceTransactionalStatus = (
...
+): NftIssuanceParameters | null => {
+  if (!inputMetadata) {
+    return null
+  }
+
+  const initTransactionalStatus = createNftIssuanceTransactionalStatus(registry, inputMetadata)
+
+  return new NftIssuanceParameters(registry, {
+    nft_metadata: new Bytes(registry, '0x0'),
+    royalty: new Option(registry, Royalty, inputMetadata.royalty),
```
drillprop, line 275: Setting royalties still doesn't work, right? I lost of track what is working now and what's not. I mean, I tried to set royalties and failed with doing that.

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -181,35 +182,43 @@ export const VideoForm: React.FC<VideoFormProps> = React.memo(({ onSubmit, setFo
 
       const videoWidth = videoInputFile?.mediaPixelWidth
       const videoHeight = videoInputFile?.mediaPixelHeight
+      const assets: VideoFormAssets = {
```
drillprop, line 185: Okay, found a weird bug. I had a video that wasn't issued as NFT. I clicked `edit video` clicked `mint an NFT` and published this change. Everything seemed to be working as it should(no error) but when I checked the NFT in the query node I got `null`. 
But this is not the end. I tried again with the same video and next time I got an extrinsic error - `NFT for given video id already exists`. I assumed that the query node didn't get information somehow about this NFT. 
Could you reproduce this and see if it happens also to you? 
I guess this is another problem with the query node.

drillprop: btw, the problem happened on video with id 19, if you want to investigate this further.

drillprop (approved):
I have two questions, but I don't think they are fixable on our side. Besides, that works great, really love the experience.

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
kdembler: Yes, I forgot to mention, query node support for minting as part of edit is broken at the moment :( https://github.com/Joystream/joystream/issues/3424

PR #2317, opened 2022-03-17
clean up playground

kdembler:
Just a tiny housekeeping to remove some unused playgrounds so it's easier to use


eldiegod (approved):

drillprop:
I pushed another small change. This should lower the width of the NftTile in the NFT extrinsic playground. Let me know what you think

kdembler:
Fantastic idea @drillprop

PR #2322, opened 2022-03-21
Set correct max dates in NFT Issue form

### LINKED ISSUE #2303: Set correct maxDate in AuctionDatePicker
We need to set proper maxDate in step 2 of our NftIssueForm and also make sure that validation works properly.
To accomplish that we can use these two functions from Polkadot API
 - `api.query.content.auctionStartsAtMaxDelta()` -  indicates when in the future the auction can start(in blocks). 
- `api.query.content.maxAuctionDuration()` - self-explanatory, how long auction can run(also in blocks)

drillprop:
Fix #2303


drillprop:
Draft for now. It will need some refactoring after merging https://github.com/Joystream/atlas/pull/2326

kdembler (approved):

PR #2323, opened 2022-03-21
fix search input styling for iOS

### LINKED ISSUE #2093: Channel page search input - wrong styling on iOS
**Screenshots**
IMG_2572

- OS: iOS 15.2.1
- Browser: Safari 15.2

rafalpawlow:
Fixes #2093 

It doesn't have white background anymore (not sure if this is our fix or iOS), but it still has unnecessary border radius.
IMG-2617


drillprop (changes requested):
If we can't change the border radius, maybe we shouldn't use input with type `search` then? We're hiding the cancel button anyway, wdyt?

rafalpawlow:
As discussed on meeting, let's stay with search type. Mostly because of iOS behaviour.

drillprop (approved):

PR #2326, opened 2022-03-22
improve nft sale validation

kdembler:
This PR:
- Adds fetching of chain constants
- Validates the NFT minimum bid to be higher than chain variable
- Provides default values for "starting price" and "buy now" fields, depending on listing type
- Adds error messages if incorrect values (like negative number) are provided

### DIFF
--- a/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.utils.ts
+++ b/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.utils.ts
@@ -4,9 +4,9 @@ import { z } from 'zod'
 import { AuctionDatePickerValue } from '@/components/_inputs/AuctionDatePicker'
 import { pluralizeNoun } from '@/utils/misc'
 
-import { NftFormFields } from './NftForm.types'
+import { Listing, NftFormFields } from './NftForm.types'
 
-export const createValidationSchema = (data: NftFormFields) => {
+export const createValidationSchema = (data: NftFormFields, listingType: Listing, minStartingPrice: number) => {
   const auctionDateType = z
     .union([
       z.object({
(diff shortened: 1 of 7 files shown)


packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.utils.ts
```diff
@@ -21,6 +21,10 @@ export const createValidationSchema = (data: NftFormFields) => {
     .nullable()
     .optional()
 
+  const buyNowPrice = z
+    .number({ required_error: 'Buy now price must be provided', invalid_type_error: 'Buy now price must be a number' })
+    .min(1, 'Buy now price cannot be lower than 1')
```
drillprop, line 26: We still need to cover one case. Minimum bid shouldn't be higher than buy now price: 
image

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.utils.ts
```diff
@@ -63,8 +67,11 @@ export const createValidationSchema = (data: NftFormFields) => {
         { message: 'Expiration date cannot be earlier than starting date' }
       ),
     royalty: z.number().nullable().optional(),
-    startingPrice: z.number().nullable().optional(),
-    buyNowPrice: z.number().nullable().optional(),
+    startingPrice: z
+      .number({ invalid_type_error: 'Minimum bid must be a number' })
+      .min(minStartingPrice, `Minimum bid cannot be lower than ${minStartingPrice}`)
```
drillprop: Same as above I guess

drillprop (changes requested):

drillprop (approved):

PR #2332, opened 2022-03-22
Show sign in dialog if user takes NFT action

### LINKED ISSUE #2276: Display login dialog if non-logged-user takes NFT action
When the user takes some action that requires login, we should display a login modal, as seen [in Figma](https://www.figma.com/file/ymHYOFItrmFaLeBC5Z4tpQ/Sign-in?node-id=511%3A56699)

Component has been implemented in https://github.com/Joystream/atlas/pull/1775 along with a hook.

**AC:**

- [ ] Make sure the dialog is shown whenever NFT action is taken

drillprop:
fix #2276

To test it out: 
- clear local storage 
- go to the playground and try to open the NFT purchase drawer or NFT settlement drawer

### DIFF
--- a/packages/atlas/src/providers/nftActions/hooks.ts
+++ b/packages/atlas/src/providers/nftActions/hooks.ts
@@ -1,12 +1,19 @@
 import { useCallback, useContext } from 'react'
 
+import { useDisplaySignInDialog } from '@/hooks/useDisplaySignInDialog'
+
 import { NftActionsContext } from './provider'
+
+import { useUser } from '../user'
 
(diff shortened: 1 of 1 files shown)


rafalpawlow (changes requested):
I'm thinking, maybe instead of repeating this condition (DRY) we could create some function which will check this? Or some proxy function which will check whether user is logged in, and then fire action or open sign in dialog?

rafalpawlow (approved):

kdembler (approved):

PR #2342, opened 2022-03-23
add Olympia banner

kdembler:
This PR adds a banner informing users that Atlas will keep using Giza network for a couple of days and changes they are making will be lost until full migration. The banner is visible in Studio or when creating/editing a membership.

**It's hacky.** I don't want to do a proper layout with this banner on top so added a quick hack, once we migrate to Olympia, we can revert this commit

Screenshot 2022-03-23 at 18 52 49


rafalpawlow (changes requested):
it cuts navbar buttons on mobile

Zrzut ekranu 2022-03-23 o 19 04 36

rafalpawlow (changes requested):
No more mobile problem, it covers close button
Zrzut ekranu 2022-03-23 o 19 14 31

rafalpawlow (approved):

PR #2346, opened 2022-03-24
update context menu component

### LINKED ISSUE #2290: Update the Context menu component
**AC:**

- [ ] Add new border radius
- [ ] Add missing top/bottom padding

---

The goal of this quick task is to introduce a subtle border radius to the Context menu component, based on the recently added radii design tokens. Designs and docs [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3158%3A103636).

image

---

When revisiting this component's implementation, it would be great if we could add a missing top & bottom padding of `4px` in all variants.

image

rafalpawlow:
Closes #2290 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3158%3A103636

### DIFF
--- a/packages/atlas/src/components/_overlays/ContextMenu/ContextMenu.styles.ts
+++ b/packages/atlas/src/components/_overlays/ContextMenu/ContextMenu.styles.ts
@@ -38,5 +38,5 @@ export const StyledContainer = styled.div<StyledContainerProps>`
   word-break: break-all;
   border-radius: ${cVar('radiusSmall')};
   overflow: auto;
-  padding: ${({ scrollable }) => sizes(scrollable ? 2 : 1)} 0;
+  padding: ${sizes(1)} 0;
 `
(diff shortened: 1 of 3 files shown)


packages/atlas/src/components/_overlays/ContextMenu/ContextMenu.styles.ts
```diff
@@ -1,9 +1,42 @@
...
+  ${getHeightStyles};
 
-export const StyledContainer = styled.div`
   background-color: ${cVar('colorBackgroundStrong')};
-  width: 200px;
+  width: 192px;
   word-break: break-all;
+  border-radius: ${cVar('radiusSmall')};
+  overflow: auto;
+  padding: ${({ scrollable }) => sizes(scrollable ? 2 : 1)} 0;
```
drillprop: I just asked Adam about this, he made a mistake. We shouldn't change this padding based on if it's scrollable or not. This should be always 4px

drillprop (changes requested):

drillprop (approved):

PR #2348, opened 2022-03-24
update snackbar component

### LINKED ISSUE #2280: Update the Snackbar component
**AC:**

- [ ] Add rounded corners
- [ ] Make sure we use the proper transition, as per Figma

The goal of this quick task is to introduce a subtle border radius to the Snackbar component, based on the recently added radii design tokens. Designs and docs [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=4606%3A140876).

image

rafalpawlow:
Closes #2280 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=4606%3A140876

### DIFF
--- a/packages/atlas/src/providers/snackbars/snackbar.tsx
+++ b/packages/atlas/src/providers/snackbars/snackbar.tsx
@@ -10,7 +10,7 @@ import {
   SvgAlertsSuccess24,
   SvgAlertsWarning24,
 } from '@/components/_icons'
-import { cVar, sizes, zIndex } from '@/styles'
+import { cVar, media, sizes, zIndex } from '@/styles'
 
 import { SnackbarIconType, useSnackbarStore } from './store'
 
@@ -42,7 +42,7 @@ export const Snackbars: React.FC = () => {
     <SnackbarsContainer>
       <TransitionGroup>
         {snackbars.map(({ id, iconType, onActionClick, onExit, ...snackbarProps }) => (
-          <CSSTransition key={id} timeout={parseInt(cVar('animationTimingFast', true)) * 2} classNames="snackbar">
+          <CSSTransition key={id} timeout={parseInt(cVar('animationTimingMedium', true)) * 2} classNames="snackbar">
             <Snackbar
               {...snackbarProps}
               onActionClick={() => {
@@ -73,8 +73,12 @@ export const SnackbarsContainer = styled.div`
   left: var(--size-sidenav-width-collapsed);
   bottom: ${sizes(18)};
   margin-left: ${sizes(4)};
-  max-width: 360px;
-  width: 100%;
   display: grid;
   z-index: ${zIndex.nearVideoWorkspaceOverlay};
+  width: calc(100% - ${sizes(8)});
+
+  ${media.xs} {
+    width: 100%;
+    max-width: 360px;
+  }
 `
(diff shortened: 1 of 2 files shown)


packages/atlas/src/providers/snackbars/snackbar.tsx
```diff
@@ -42,7 +42,7 @@ export const Snackbars: React.FC = () => {
     <SnackbarsContainer>
       <TransitionGroup>
         {snackbars.map(({ id, iconType, onActionClick, onExit, ...snackbarProps }) => (
-          <CSSTransition key={id} timeout={parseInt(cVar('animationTimingFast', true)) * 2} classNames="snackbar">
+          <CSSTransition key={id} timeout={parseInt(cVar('animationTimingMedium', true)) * 2} classNames="snackbar">
```
drillprop, line 45: Shouldn't we set here `animationTransitionCallout`?

drillprop (changes requested):

packages/atlas/src/providers/snackbars/snackbar.tsx
rafalpawlow: But it's timing property, `animationTransitionCallout` uses `animationTimingMedium`

drillprop: oh, you're right!

drillprop (approved):

kdembler:
LGTM but let's get approval from @toiletgranny on this one

toiletgranny (approved):
**TL;DR**: The implementation of the above-listed changes is flawless, it's good to go, but we should address the lack of _dynamic_ (for the lack of a better word) Snackbar positioning in the near future, because it affects the animation.

---

While the transition animation seems to be correct on paper, it doesn't feel the same when you compare it to what led to us choosing the timing and easing values for this in Figma. You can experience this on yourself [in this prototype](https://www.figma.com/proto/Pf31tuYpozYmpq163U2ho8/Web-components?page-id=6544%3A129802&node-id=6544%3A129803&viewport=400%2C48%2C0.45&scaling=min-zoom&starting-point-node-id=6544%3A129803), if you compare the animation to the one in Storybook. 

There's a very specific reason for that and it's the exaggerated bottom margin in the Snackbar wrapper, compared to Figma designs, where we use `16px` of bottom and left margin. If I remember correctly, this increased margin had to be added to accommodate for displaying Snackbars inside video workspaces which come with the additional bottom footer that's taking some real estate. 

However, the unpleasant consequences of the above are that (a) this is how the Snackbar component is positioned everywhere now (it's not just the video workspace), and that (b) this increased bottom margin messes with the animation easing, specifically: the _overshoot_ that's at the end of the animation. The _overshoot_ in the below graph is what creates the _bouncing_ effect at the end of an animation. Now, how strong and pronounced the bounce effect is depends on (a) how far an element needs to travel and (b) at what time. In Figma, the bounce is soft and subtle because the Snackbar needs to stop only at 16px from the bottom of the screen. In Storybook, as the Snackbar needs to travel a bit further to the top, the bounce at the end is pushy, tactless, and distracting.

CleanShot 2022-03-25 at 11 18 14
 
And this is why I would love us to address the issue of this "dynamic" Snackbar positioning. Because of this, reviewing @rafalpawlow's implementation is a bit tricky today.

PR #2367, opened 2022-03-28
update nft widget timers

### LINKED ISSUE #2329: Update the NFT widget component to add loading state for timers and support more scenarios for displaying auction dates
We need to implement a loading state for the timers in the Nft Widget as the current block data is sometimes loaded after the nft data arrives.

https://user-images.githubusercontent.com/51168865/159445554-372618cb-c7fc-48b9-bbed-d2e5fbe4ef0a.mov

rafalpawlow:
Closes #2329

**Design**
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2077%3A112135

### DIFF
--- a/packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
+++ b/packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
@@ -1,3 +1,4 @@
+import { differenceInDays, differenceInSeconds } from 'date-fns'
 import React from 'react'
 import useResizeObserver from 'use-resize-observer'
 
@@ -17,7 +18,7 @@ import { NftSaleType } from '@/joystream-lib'
 import { useMemberAvatar } from '@/providers/assets'
 import { useTokenPrice } from '@/providers/joystream'
 import { formatNumberShort } from '@/utils/number'
-import { formatDateTime } from '@/utils/time'
+import { formatDateTime, formatDurationShort, formatTime } from '@/utils/time'
 
 import { NftHistory } from './NftHistory'
 import { NftInfoItem, NftTimerItem } from './NftInfoItem'
@@ -52,6 +53,7 @@ export type Auction = {
   englishTimerState: EnglishTimerState | undefined
   auctionPlannedEndDate: Date | undefined
   startsAtDate: Date | undefined
+  plannedEndAtBlock?: number | null
 }
 
 export type NftWidgetProps = {
@@ -83,7 +85,7 @@ export type NftWidgetProps = {
   onWithdrawBid?: () => void
 }
 
-const SMALL_VARIANT_MAXIMUM_SIZE = 280
+const SMALL_VARIANT_MAXIMUM_SIZE = 340
 
 export const NftWidget: React.FC<NftWidgetProps> = ({
   ownerHandle,
@@ -109,14 +111,22 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
   const { convertToUSD, isLoadingPrice } = useTokenPrice()
 
   const content = useDeepMemo(() => {
-    if (!nftStatus) return
+    if (!nftStatus) {
+      return
+    }
     const contentTextVariant = size === 'small' ? 'h400' : 'h600'
     const buttonSize = size === 'small' ? 'medium' : 'large'
     const buttonColumnSpan = size === 'small' ? 1 : 2
+    const timerColumnSpan = 2
+    const auctionBeginsDifference =
+      nftStatus?.status === 'auction' && nftStatus.startsAtDate
+        ? differenceInDays(nftStatus.startsAtDate, new Date())
+        : 0
     const BuyNow = ({ buyNowPrice }: { buyNowPrice?: number }) =>
       buyNowPrice ? (
         <NftInfoItem
           size={size}
+          loading={!buyNowPrice || convertToUSD(buyNowPrice) === '$0'}
           label="Buy now"
           content={
             <>
@@ -352,6 +362,7 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
             ) : (
               <NftInfoItem
                 size={size}
+                loading={!nftStatus.startingPrice || convertToUSD(nftStatus.startingPrice) === '$0'}
                 label="Starting Price"
                 content={
                   <>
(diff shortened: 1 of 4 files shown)


packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -78,7 +80,7 @@ export type NftWidgetProps = {
   onNftChangePrice?: () => void
 }
 
-const SMALL_VARIANT_MAXIMUM_SIZE = 280
+const SMALL_VARIANT_MAXIMUM_SIZE = 340
```
eldiegod, line 92: Why are we changing the sizing breakpoint?

packages/atlas/src/hooks/useBlockTimeEstimation.ts
```diff
@@ -9,9 +9,14 @@ export const useBlockTimeEstimation = () => {
 
   const convertBlockToMsTimestamp = useCallback(
     (targetBlock: number) => {
+      const currentBlockMsTimestamp = getCurrentBlockMsTimestamp()
```
eldiegod: we need to move away from `getCurrentBlockMsTimestamp()` which doesn;t updates because is a ref and instead use an `useState` just like we do for `currentBlock`

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -345,33 +354,44 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
...
-                }
-              />
+              <GridItem colSpan={timerColumnSpan}>
+                <NftInfoItem
+                  size={size}
+                  label="Auction ended on"
+                  loading={!nftStatus.auctionPlannedEndDate}
+                  content={
+                    nftStatus.auctionPlannedEndDate && (
+                      <Text variant={size === 'medium' ? 'h600' : 'h400'} secondary>
```
eldiegod: the variant you are looking for is this `contentTextVariant`

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -103,10 +105,17 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
...
+      return
+    }
     const contentTextVariant = size === 'small' ? 'h400' : 'h600'
     const buttonSize = size === 'small' ? 'medium' : 'large'
     const buttonColumnSpan = size === 'small' ? 1 : 2
+    const timerColumnSpan = 2
+    const auctionBeginsDifference =
+      nftStatus?.status === 'auction' && nftStatus.startsAtDate
+        ? differenceInDays(nftStatus.startsAtDate, new Date())
+        : 0
```
eldiegod: this state you can put directly under the `auction` switch case

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -345,33 +354,44 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
...
-                }
-              />
+              <GridItem colSpan={timerColumnSpan}>
+                <NftInfoItem
+                  size={size}
+                  label="Auction begins on"
+                  loading={!nftStatus.startsAtDate}
+                  content={
+                    nftStatus.startsAtDate && (
+                      <Text variant={size === 'medium' ? 'h600' : 'h400'} secondary>
```
eldiegod: same as above use the `contentTextVariant`

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
eldiegod: then you can also remove the check `nftStatus?.status === 'auction' &&`

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -103,10 +105,17 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
...
 
   const content = useDeepMemo(() => {
-    if (!nftStatus) return
+    if (!nftStatus) {
+      return
+    }
     const contentTextVariant = size === 'small' ? 'h400' : 'h600'
     const buttonSize = size === 'small' ? 'medium' : 'large'
     const buttonColumnSpan = size === 'small' ? 1 : 2
+    const timerColumnSpan = 2
```
eldiegod: this has to be `size === 'small' ? 1 : 2` just like the `buttonColumnSpan` otherwise the layout breaks on small variant

eldiegod (changes requested):
The screenshot below is video 46 and it doesn't show a timer even tho is a running English auction.

Screenshot 2022-03-29 at 21 58 30

As you can see in the video below the loading state for the timer is coming in way too late we should be showing a loading state for the timer as soon as we get the Nft data that tells us this is an English auction

https://www.loom.com/share/275127eeb4244bddb414c6a244ef5172

On the screenshot below, the timer, when there are less than 24 hours till the auction starts, is not updating once per second as we would like it to.

I was thinking if we should just move all the logic related to auction expired, running, upcoming.. to the timer component and handle it there.
As it will rerender itself every second like we need it to and will shave off some complexity from the main component.
Wdyt? 
I can explain in a call if needed too

Screenshot 2022-03-29 at 22 28 22

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
rafalpawlow: Yeah, I knew this will concern you. That's because we changed font size of timer and for example this: `5 Mar 2022 at 12:00` is too long for `281`

rafalpawlow:
> I was thinking if we should just move all the logic related to auction expired, running, upcoming.. to the timer component and handle it there.

@eldiegod I'm not sure if this is a good idea, we don't use `NftTimerItem` component for all time values, for example upcoming uses `NftInfoItem`. BTW, what

eldiegod (approved):
looks good now !

eldiegod:
Tho we will probably need to do something in another PR about the action buttons that now also depend on timers, tho we should probably be getting `englishTimerState` from the query node in the first place, wdyt @kdembler ?

kdembler:
You need to unpack this a bit @eldiegod:

> action buttons that now also depend on timers

what do you mean?

> tho we should probably be getting `englishTimerState` from the query node in the first place

same, what do you mean here? We load timers locally in the app so that state cannot be served by the query node

eldiegod:
@kdembler 
Check the action buttons on this vid, for example, it will make it clear

https://atlas-lf69rckhh-joystream.vercel.app/video/46

happens because we don't know the state of the English auction from the payload.

Maybe I meant runtime I get confused by both often, but we should be able to know in the payload if an English auction is currently running, expired, or upcoming before we load timers locally.

kdembler:
Ah I see what you mean now. Query node actually doesn't know when the auction has expired, same for runtime. If you try to place a bid after it has expired, a check will see that `current_block > expire_block` so the runtime will not allow it. But there isn't any state that says "okay this auction is expired now". QN could support that on its own, but it currently doesn't and I don't think it's realistic waiting for yet another feature so we will need to handle this locally. If there's a design issue, can you raise it with Adam @eldiegod?

I'm merging in the meantime since it's approved

PR #2368, opened 2022-03-28
style fixes: badges & search close icon

kdembler:


packages/atlas/src/components/_buttons/Button/Button.tsx
```diff
@@ -30,6 +30,7 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         textOnly={textOnly}
         variant={variant}
         iconOnly={iconOnly}
+        data-badge={badge}
```
eldiegod, line 33: we should add `data-badge` as an emotion prop in this component

eldiegod (approved):
Looks good added a suggestion

packages/atlas/src/components/_buttons/Button/Button.tsx
eldiegod: possibly even make it a string and use `formatNumberShort` so it doesn't break with big numbers

kdembler: Added to props, made it to always be string, let's leave it up to consumer to format this string if needed

eldiegod (approved):

eldiegod:
There is a type error on filters bar badge due to now being a string but besides LGTM

packages/atlas/src/components/_buttons/Button/Button.tsx
kdembler: Okay, changed that back to allow number as well, it's up to the consumer, doesn't make a difference for the component

PR #2373, opened 2022-03-29
create VideoDetails component, fix description line breaks

kdembler:
Even though the diff is a bit lengthy, the changes made are very small:
1. Extracted all the stuff about video details section (description + license + category) to `VideoDetails` component
2. Added `display: block` to each line of description text, so the line break is visible


drillprop (approved):
Looks good to me! 👍🏻

Issue #1740, opened 2021-12-06
Fetching NFT Notifications


dmtrjsg:
To fetch notification we will need to read QN queries that correspond to low-level events emitted by the blockchain. First, we will need to compile a list of kinds of notification we may want to display (initial list is [here](https://www.figma.com/file/Goequb9eZDL6WxK7T74oxS/Notifications?node-id=922%3A54970), but will likely need an update). Then, for each of those, we need to figure out how to get data for it. 

**Example:**

_Example here_

kdembler:
### Intro

I did a preliminary research into this and it is _tricky™_. First, here's the list of notifications the designs have envisioned for the NFT experience:
- Somebody bid on an NFT I'm selling
- Somebody purchased (BuyNow or buyNowPrice in an auction) an NFT I'm selling
- Somebody withdrew their bid from an open auction for an NFT I'm selling (I think we should skip this one, doesn't provide much value and creates unnecessary noise, I'm gonna ignore this in further analysis)
- Somebody outbid me in an english auction I'm participating in (I was the top bid, now somebody else is)
- A bid that I placed in an open auction became withdrawable (bid lock duration has passed)
- A bid that I placed in an open auction has been accepted by the seller (I'm the owner now)
- An english auction I participated in has expired (is past it's end block) and I won (I need to settle the auction to become the owner)

### Problems

- **Somebody bid on an NFT I'm selling**
- **Somebody purchased (BuyNow or buyNowPrice in an auction) an NFT I'm selling**
- **A bid that I placed in an open auction has been accepted by the seller (I'm the owner now)**
1. Let's use first notification as an example - "Somebody bid on an NFT I'm selling". The relevant event can be queried from the query node - `auctionBidMadeEvents` provides events for all the bids that were placed for an auction. However, we need those events only for NFTs the user was selling, not all the NFTs. This is how the filter could look like: `auctionBidMadeEvents(where: { video: { nft: { ownerMember: { id_eq: "user" } } } })`. This uses 3 levels of nested filters while the query node only supports filters that go a single level deep. It's not possible to filter those events in a desired way
[...]

PR #2374, opened 2022-03-29
add notifications fetching

### LINKED ISSUE #1739: Read/ Unread notification states
- [ ] clicking on notification from widget
- [ ] clicking on notification from notiff centre (page)
- [ ] displaying all notiffs as unread from new browser (when local state is empty)

kdembler:
Resolves #1739 and #1740.

Will create followups for those issues to tackle additional fetching later, once QN adds support


packages/atlas/src/providers/notifications/notifications.hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+        id: event.video.id,
+        title: event.video.title || '',
+      },
+    }
+  } else if (event.__typename === 'OpenAuctionBidAcceptedEvent') {
+    return {
+      id: event.id,
+      type: 'open-auction-ended',
+      date: event.createdAt,
+      // this will be accessible directly on event object in the future
```
drillprop: Maybe, let's add "TODO:" here if it changes in the future. It will be easier to spot/find this comment during the refactor.

packages/atlas/src/views/notifications/Notifications.tsx
```diff
@@ -74,12 +39,13 @@ export const Notifications = () => {
...
-          {NOTIFICATIONS.map((notification, idx) => (
+          {notifications.map((notification, idx) => (
             <StyledNotificationTile
               key={`notification-${notification.id}-${idx}`}
-              {...notification}
-              onSelect={() => toggleSelected(notification.id)}
+              notification={notification}
               selected={selectedNotifications.includes(notification.id)}
+              onSelect={() => toggleSelected(notification.id)}
+              onClick={() => markNotificationAsRead(notification.id)}
```
drillprop, line 48: I'm confused about this behavior and simultaneously couldn't find the right answer in the design. 
Currently, if I click NFT title, the tile notification is marked as read, but when I do this on the notification widget, nothing happens. 
Shouldn't we do the same on NFT widget?

packages/atlas/src/providers/notifications/notifications.store.ts
```diff
@@ -0,0 +1,41 @@
...
+        set((state) => {
+          state.readNotificationsIdsMap[id] = true
+        })
+      },
+    }),
+  },
+  {
+    persist: {
+      key: 'notifications',
+      whitelist: ['lastAllReadBlock', 'readNotificationsIdsMap'],
```
drillprop, line 32: let's maybe keep these values in const variable?
i.e: 

```
const LOCAL_STORAGE_KEY = 'notifications'
const WHITELIST: (keyof NotificationsStoreState)[] = ['lastAllReadBlock', 'readNotificationsIdsMap']
```

drillprop (changes requested):

packages/atlas/src/providers/notifications/notifications.store.ts
kdembler: Do we gain anything from that? The values are used in a single place only and they are described by the object key so not sure

packages/atlas/src/views/notifications/Notifications.tsx
kdembler: Yep, I missed that, fixed now

packages/atlas/src/providers/notifications/notifications.hooks.ts
kdembler: Done

rafalpawlow (changes requested):
Notification tile story doesn't work properly

packages/atlas/src/providers/notifications/notifications.store.ts
drillprop: ok, make sense

drillprop (approved):
LGTM!

rafalpawlow (changes requested):
I think this behaviour is wrong. When you select notification it changes it's state to read, but according to design it should keep unread state until you click `Mark as read` on floating panel.

https://user-images.githubusercontent.com/13706246/160889580-67b94ad7-4b18-4c92-9433-9312888e2f7e.mov

kdembler:
Yeah that was an issue with the component, fixed now

rafalpawlow (changes requested):
The problem still exists when you click tile outside of checkbox

https://user-images.githubusercontent.com/13706246/161089477-eae0230c-eced-4838-8695-4f869336f1b0.mov

kdembler:
That's desired, when you navigate to a notification, we want to mark it as read

rafalpawlow (approved):

PR #2377, opened 2022-03-30
Update purchase view

### LINKED ISSUE #2224: NFT purchase view changes
**AC:**

- [x] Fix issues 3,4,5 pointed out in Tomek's comment below
- [x] Update the section with top bid and your bid
- [x] Add change price flow
- [x] Make sure in open auction your new bid doesn't have to be higher than the top one

---

This issue gathers all my recommendations for things we need to address in [NFT purchase view designs](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-purchase-view?node-id=0%3A1) due to recent findings and changes to our understanding of NFTs in Joystream.

## Changing bid
With the recent design NFT changes and adjustments, it has become apparent that each member can have only _one_ bid placed under a particular NFT at a time.

For this reason, if I, as a user, have placed a bid in an open auction, and that bid got outbid later by someone else, to make a new offer I can either withdraw my last bid and place a new one, or change the one I placed previously. And the latter is a flow that we're currently missing in the designs for the NFT purchase view. 

In other words, designs for the NFT purchase view need to be updated so that the below user story is supported:
_**As a content viewer, I want to be able to change my bid in an open auction.**_

Please note that while changing my bid in an open auction I can change it to be either higher or lower, as long as it matches the starting price (min bid).

## Starting price & "Current bid"
1. In open auctions, unlike with english auctions, there's no concept of a "current bid". Any bid can win in an open auction, it doesn't need to be the most recent or the highest one. For this reason, perhaps we can rephrase this to say "Top bid"? I believe that this way we can have this working both for english and open auctions, without needing to introduce a logic that would change this label depending on the auction type. An additional bonus would be preserving consistency with the rest of the designs, as we're already using the term "Top bid" in a few places.
2. There's no concept of a minimum bid in open auctions in a way it exists in english auctions. The only requirement a bid needs to meet in an open auction is being equal or higher than the _starting price_, right? Perhaps we might want to consider changing the wording here for open auctions to say "starting price", rather than saying "minimum bid"? I don't think this is necessary, so I'll leave this one up to you to decide, as an author of the initial designs for this view. 

CleanShot 2022-03-08 at 12 12 46@2x

cc @kdembler @dmtrjsg

drillprop:
Fix #2224 

~Draft, changing bid flow not included yet~

E: FYI UI for changing bid flow is added, but some things could not work correctly. Currently, you need to outbid the top bid in the auction and this will change in the future. You can also bid on the same auction several times(and you'll see all your bids in the bids array)

Summary of changes 
- fixed purchase view timer:  
image
- fix scrolling issues, change size of the logo, updated pill variant in text field
- added change bid flow

### DIFF
--- a/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.styles.ts
+++ b/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.styles.ts
@@ -4,7 +4,7 @@ import styled from '@emotion/styled'
 import { Text } from '@/components/Text'
 import { JoyTokenIcon } from '@/components/_icons/JoyTokenIcon'
 import { SvgJoystreamLogoShort } from '@/components/_illustrations'
-import { cVar, media, sizes } from '@/styles'
+import { cVar, media, sizes, zIndex } from '@/styles'
 
 const flexStyles = css`
   display: flex;
@@ -15,7 +15,6 @@ export const FlexWrapper = styled.div`
   ${flexStyles};
 `
 export const Content = styled.div`
-  padding-top: ${sizes(8)};
   display: flex;
   flex-direction: column-reverse;
   min-height: 100%;
@@ -23,7 +22,7 @@ export const Content = styled.div`
   padding-bottom: 130px;
 
   ${media.md} {
-    padding-top: ${sizes(16)};
+    height: 100%;
     flex-direction: row;
     overflow-y: unset;
     padding-bottom: unset;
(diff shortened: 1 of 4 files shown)


packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.styles.ts
```diff
@@ -88,32 +92,36 @@ export const Header = styled.div`
...
-  ${flexStyles};
+export const CurrentBidWrapper = styled.div`
+  margin-bottom: ${sizes(6)};
+  background-color: ${cVar('colorBackgroundStrong')};
+`
 
+export const TokenWrapper = styled.div`
   position: relative;
+  left: -4px;
+  z-index: 10;
```
rafalpawlow: let's use `z-index` from our utils

rafalpawlow (changes requested):
Overall looks good, however I'm not sure about scrolling. Now when we have vertical scroll, it scrolls nft tile with form together, previously only form was scrollable. Did you confirm this behavior with the designers?

rafalpawlow (changes requested):
You are using wrong JoyTokenIcon variant inside input.

rafalpawlow (approved):

kdembler (commented):
1. There are some alignment issues with the payment split and the Joystream logo:
Screenshot 2022-04-06 at 14 14 42
2. The "Top bid" header uses H100 in the design, we have H300

kdembler (approved):

PR #2382, opened 2022-03-30
fix video tile skeleton

rafalpawlow:


packages/atlas/src/components/_video/VideoTileViewer/VideoTileViewer.tsx
```diff
@@ -47,7 +47,7 @@ export const VideoTileViewer: React.FC<VideoTileViewerProps> = ({ id, onClick, d
       videoHref={videoHref}
       channelHref={channelHref}
       onChannelAvatarClick={() => navigate(channelHref)}
-      loadingDetails={loading}
+      loadingDetails={loading || !video}
```
drillprop, line 46: Maybe we should use `internalIsLoadingState` that is returned from useVideoTileShared then? it seems, that we're doing the same thing twice: 
```tsx
  const internalIsLoadingState = loading || !video
```
You'll need to also update `VideoTilePublisher`

drillprop (changes requested):

packages/atlas/src/components/_video/VideoTileViewer/VideoTileViewer.tsx
rafalpawlow: I'm not sure if this is good idea, since `loading` and `video` are passed from `VideoTileViewer`
`VideoTilePublisher` already has this change.

packages/atlas/src/hooks/useVideoTileSharedLogic.ts
```diff
@@ -5,21 +5,16 @@ import { useAsset } from '@/providers/assets'
 type UseVideoSharedLogicOpts = {
   video?: BasicVideoFieldsFragment | VideoFieldsFragment | null
   loading?: boolean
```
drillprop: We need to also remove this prop and update VideoTileViewer and VideoTilePublisher

drillprop (changes requested):

drillprop (approved):

PR #2399, opened 2022-04-01
minimized transaction

### LINKED ISSUE #2225: New way of handling transaction with minimised snack bar
## Context

Reactions due to frequency require a more subtle way to show signing transaction toaster

## Scope

Update the way how toaster notification looks and works
- [x]  update card design
- [ ] only show at the point of triggering reaction, once signed we _don't show progress/success_ 
- [ ] Show toaster it again if transaction fails
- [ ] Write logic for handling the transaction in the background. Pay attention to the fact that the transaction handling cannot be longer done by a hook called by the view, because the view can be quit when the transaction is processing.

**Out of scope ⛔ :**
- [ ] In case of multiple transactions triggered, the toaster will show compound (aggregate counter of transactions pending signing) ⚠️ covered in comments designs which are WIP
- [ ] Skip Retry button in failed snackbar

rafalpawlow:
Closes #2225 

To test this, go to `/playground/minimized-transaction`

**Design**
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2282%3A163630


kdembler (commented):
Haven't looked at the code yet, just the experience but works pretty solid!
Some things I think could be improved:
1) If I cancel the transaction inside the extension, I get an "Unknown error", we already had this handled for regular transactions
2) The `Retry` button on the snackbar doesn't do anything, let's drop it (for now at least)

packages/atlas/src/providers/snackbars/snackbar.tsx
```diff
@@ -10,16 +10,23 @@ import {
   SvgAlertsSuccess24,
   SvgAlertsWarning24,
 } from '@/components/_icons'
+import { Spinner } from '@/components/_loaders/Spinner'
 import { cVar, media, sizes, zIndex } from '@/styles'
 
 import { SnackbarIconType, useSnackbarStore } from './store'
 
+export const StyledSpinner = styled(Spinner)`
```
kdembler: Does this needs to be exported?

kdembler (commented):
Left some feedback in Slack, we need to clean this up slightly

kdembler (approved):

PR #2412, opened 2022-04-04
Card interactions

### LINKED ISSUE #1829: Create NFT Card component interactions
Follow up to #1678 
As part of this task, let's add 3D interaction and the holographic effect to the component

drillprop:
Fix #1829


packages/atlas/src/components/_nft/NftCard/NftCard.tsx
```diff
@@ -25,22 +36,54 @@ export const NftCard: React.FC<NftCardProps> = ({
...
-            <>
-              <Members caption="Supporters" members={supporters} />
-              <Separator />
-            </>
+      {reflectionCells.map((_, idx) => {
+        const row = Math.floor(idx / 10)
+        const column = idx % 10
+
+        const lp = 50 + (column * 10 - 50) / 1
+        const tp = 50 + (row * 10 - 50) / 1
```
drillprop:
```suggestion
        const lp = column * 10
        const tp = row * 10
```
I just copied this chunk of code from somewhere and started to tweak and left this not refactored. Let's just simplify these calculations or even move to `ReflectionGridCell`

packages/atlas/src/components/_nft/NftCard/NftCard.tsx
```diff
@@ -25,22 +36,54 @@ export const NftCard: React.FC<NftCardProps> = ({
...
+        return (
+          <ReflectionGridCell
+            row={row}
+            column={column}
+            lp={lp}
+            tp={tp}
+            key={idx}
+            style={{
+              top: `calc(${row} * 10%)`,
+              left: `calc(${column} * 10%)`,
```
drillprop: This could be moved inside to `ReflectionGridCell`

drillprop (commented):
I know that I partly wrote this code but I allowed myself to add two comments 😅

packages/atlas/src/components/_nft/NftCard/NftCard.styles.ts
```diff
@@ -7,7 +7,137 @@ import { cVar, media, sizes } from '@/styles'
...
+  opacity: 0;
+`
+
+export const Pattern = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
+  background-image: url('https://i.imgur.com/AjBArhU.png');
```
rafalpawlow: I think we shouldn't take it from external server, it should live in our assets directory.

packages/atlas/src/components/_nft/NftCard/NftCard.styles.ts
```diff
@@ -7,7 +7,137 @@ import { cVar, media, sizes } from '@/styles'
...
+  background-position: 50% 50%;
+  background-size: 1000%;
+  z-index: 2;
```
rafalpawlow: Can we use `zIndex` from our utils?

packages/atlas/src/components/_nft/NftCard/NftCard.styles.ts
```diff
@@ -7,7 +7,137 @@ import { cVar, media, sizes } from '@/styles'
...
+}
+
+export const ReflectionGridCell = styled.div<{
+  row: number
+  column: number
+  lp: number
+  tp: number
+}>`
+  position: absolute;
+  z-index: 1;
```
rafalpawlow: Same question as above

rafalpawlow (changes requested):
Looks and works amazing. However I have a doubt though, it seems to me that there should be some suggestion that the card is interactive. Maybe it should animate itself slightly or be slightly tilted by default? @toiletgranny

toiletgranny:
@rafalpawlow I love this idea, brilliant! How about a subtle attention seeker [like this one](https://jsfiddle.net/toiletgranny/ym5Laujs/93/)? We could have it played only once, with... let's say a 2s delay, to make sure everything's properly loaded. The only question is, @eldiegod, are we going to be able to show all the lights, shines, and patterns in the animation like this one, even though the cursor might not be hovering over the NFT card?

eldiegod:
>@rafalpawlow I love this idea, brilliant! How about a subtle attention seeker [like this one](https://jsfiddle.net/toiletgranny/ym5Laujs/93/)? We could have it played only once, with... let's say a 2s delay, to make sure everything's properly loaded. The only question is, @eldiegod, are we going to be able to show all the lights, shines, and patterns in the animation like this one, even though the cursor might not be hovering over the NFT card?

@toiletgranny It's possible but it's not that easy because of what you mention, if we want this I think it's best to work on it in a separate ticket/PR

rafalpawlow (approved):

Issue #2393, opened 2022-03-31
Drawers should not cover the entire screen 


drillprop:
Currently, when we open the purchase drawer or the settlement drawer, we cover the entire screen. This is wrong.  

We basically need to do the same as we have in publisher drawers(VideoWorkspace or NftSaleBottomDrawer) - we should leave some space for the top bar(this means that the top bar should be visible).

PR #2413, opened 2022-04-04
Drawers should not cover the entire screen

eldiegod:
VideoWorkspace or NftSaleBottomDrawer should leave some space for the top bar

 fix: #2393


drillprop (approved):

rafalpawlow (changes requested):
Let's update styles and story also since this will be never used

kdembler:
Agreed, let's just remove the prop completely

rafalpawlow (approved):

PR #2417, opened 2022-04-05
Update royalties input in VideoForm

### LINKED ISSUE #2266: Divide minting and listing experiences in Video workspace
**AC:**

- [x] Update the visuals of the royalties input field
- [x] Use chain constants for royalties validation

---

# Overview
Due to new discoveries made during the implementation of NFT Issuance - New adjustments have to be done to the video workspace. 

# Tasks
- Re-design part of interface responsible for issuing NFT in video workspace
- Remove royalties screen from designs
- Remove royalties screen from user stories
- Correct all user stories accordingly to the new limitations
- Update video tile context menus with new actions
- Update all RWD
- Remove not for sale NFT listing
- Re-Design smooth experience for listing NFT on auction 
- Design onboarding(explanation) experience for a user minting NFT for the first time & showing how to list it after its done

drillprop:
Fix #2266 
The title of the issue could be misleading. The scope of this task is to just update NFT royalties input(basically everything what's listed in AC)

I also added `nftPlatformFeePercentage` to the `NftPurchaseBottomDrawer`

### DIFF
--- a/packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
+++ b/packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
@@ -106,6 +106,7 @@ export const VideoForm: React.FC<VideoFormProps> = React.memo(({ onSubmit, setFo
     getValues,
     setValue,
     watch,
+    trigger,
     reset,
     formState: { errors, dirtyFields, isDirty, isValid },
   } = useForm<VideoWorkspaceVideoFormFields>({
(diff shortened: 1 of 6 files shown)


packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -462,26 +467,36 @@ export const VideoForm: React.FC<VideoFormProps> = React.memo(({ onSubmit, setFo
...
+                      }
+                    },
                     required: {
                       value: watch('mintNft'),
                       message: 'Creator royalties must be set',
                     },
                     min: {
                       value: 0,
-                      message: 'Creator royalties cannot be lower than 0',
+                      message: 'Creator royalties cannot be lower than 0%',
```
eldiegod: As I understand we should use here `nftMinCreatorRoyaltyPercentage` both in the value and in the message

eldiegod (approved):
One small detail left, besides that all good!

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
kdembler: Yeah it's a good point, actually the minimum value for royalties is 1%. We allow 0 here, because we interpret this as "royalties disabled". But what if user provides `0.5`? Let me ping @drillprop on Slack, we can discuss

drillprop: Actually `nftMinCreatorRoyaltyPercentage` is `1`, and I handle "0.5 scenaro" on the line `475`. In these lines, I only make sure that no one set a negative value.  
We decided that we will wait for a response from Kuba because this is a kinda UX issue.

eldiegod (approved):

kdembler (approved):

PR #2420, opened 2022-04-05
Use new video counters from query node

### LINKED ISSUE #1944: Use new video counters from query node
```
query {
  videoCategories {
    id
    activeVideosCounter
  }
}
```
`activeVideosCounter` field is now available for categories and channels, we should use it instead of what we do current (Discover view with multiple queries / `useChannelVideosCount` hook)

drillprop:
Fix #1944

Just FYI, I didn't remove `useVideoCount` entirely because It's still needed in CategoryView to count filtered videos.


packages/atlas/src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -57,20 +55,10 @@ export const DiscoverView: React.FC = () => {
...
-    }
-  )
   const isMdBreakpoint = useMediaMatch('md')
 
   const headTags = useHeadTags('Discover')
 
-  if (error) {
-    return <ViewErrorFallback />
-  }
-
```
eldiegod, line 73: It wouldn't hurt to keep this error fallback but instead, check if an error comes from the `useCategory` hook

eldiegod (approved):
Looks solid!

PR #2421, opened 2022-04-05
update put on sale form

### LINKED ISSUE #2308: Explore Auction step page adjustments
## Context 

A recent discussion on rocketchat brought to our attention a new perspective on both English and open types of auctions.
After a brief Thursday meeting, we decided to prepare explorations of two solutions to the discovered problem.

## Scope

Details are in Loom walkthrough on all changes [Link > ](https://www.loom.com/share/a1af69f07cd24011a1903e3b93b12de8)

AC:
- [ ] Auction type selector updated to Open and Timed
- [ ] Timed auction has extra field for date range - Ending Date
- [ ] specific day/ shortcuts date pickers 
- [ ] Display starting date on auction confirmation page (last step before setting live)
- [ ] Update tJOY text fields (auction&buy now) to look the same as they do in purchase view (tJOY icon on the left, USD price in the pill on the right)

## Design

Changelog link: https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace

rafalpawlow:
Closes #2308 

**Scope**
- [x] Auction type selector updated to Open and Timed
- [x] Timed auction has extra field for date range - Ending Date
- [x] specific day/ shortcuts date pickers 
- [x] Display starting date on auction confirmation page (last step before setting live)
- [x] Update tJOY text fields (auction&buy now) to look the same as they do in purchase view (tJOY icon on the left, USD price in the pill on the right)

**Design**
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1%3A307

### DIFF
--- a/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
+++ b/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
@@ -3,15 +3,25 @@ import { Controller, useFormContext } from 'react-hook-form'
 
 import { Pill } from '@/components/Pill'
 import { Text } from '@/components/Text'
+import { JoyTokenIcon } from '@/components/_icons/JoyTokenIcon'
 import { AuctionDatePicker } from '@/components/_inputs/AuctionDatePicker'
 import { FormField } from '@/components/_inputs/FormField'
 import { MemberComboBox } from '@/components/_inputs/MemberComboBox'
+import { OptionCardRadio } from '@/components/_inputs/OptionCard'
 import { TextField } from '@/components/_inputs/TextField'
+import { useTokenPrice } from '@/providers/joystream'
 import { cVar } from '@/styles'
 import { pluralizeNoun } from '@/utils/misc'
 import { formatNumber } from '@/utils/number'
 
-import { AuctionDatePickerWrapper, DaysSummary, DaysSummaryInfo, Header, StyledFormField } from './SetUp.styles'
+import {
+  AuctionDatePickerWrapper,
+  DaysSummary,
+  DaysSummaryInfo,
+  Header,
+  OptionCardRadioWrapper,
+  StyledFormField,
+} from './SetUp.styles'
 
 import { useNftFormUtils } from '../NftForm.hooks'
 import { AuctionDurationTooltipFooter } from '../NftForm.styles'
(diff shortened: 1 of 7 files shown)


packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
```diff
@@ -223,15 +220,19 @@ export const SetUp: React.FC<SetUpProps> = ({
...
                       maxDate={maxEndDate}
-                      disabled={!activeInputs.includes('auctionDuration')}
                       onChange={onChange}
                       items={expirationDateItems}
-                      value={value}
+                      value={
+                        value || {
+                          type: 'duration',
+                          durationDays: 1,
+                        }
```
drillprop, line 226: I've noticed one issue with AuctionDatePickers. I'm not sure what's causes it. Maybe it's this useEffect on the line 70

1. Select a timed auction
2. pick a specific expiration date
3. click the next step
4. go back
5. You'll see that the expiration date is not set, but the number od blocks suggests that "1 day" is picked
image

drillprop (changes requested):
I pushed one small commit, which should fix the issue with not setting the min date on the auction date picker.

Generally looks and works good, besides this one bug.

drillprop (approved):
Everything works now, looks good to me! 👍🏻

kdembler (changes requested):
Works well but found some points worth addressing since we're making updates here already:

1. The buy now price USD pill is connected with the starting price input:

https://user-images.githubusercontent.com/12646744/161966975-b239235c-af91-4ae7-8b56-b1aece24c63f.mov

2. For the english auction, the summary screen has "Expiration date" and "Total auction duration" which are both expressed as a duration (1 day). Could you update this so that "Expiration date" is always a date (even if the user picked "1 day" as duration option) and the "Total auction duration" is always duration even if the user picked a specific date?
Screenshot 2022-04-06 at 13 48 19

3. Can we make this say Open auction / Timed auction?
Screenshot 2022-04-06 at 13 48 27

4. I think "Starting date" isn't very useful for Fixed price listing type, could we remove it from the summary?
Screenshot 2022-04-06 at 13 48 54

kdembler (commented):
Also just noticed that going through the sale flow with auction with default values (open auction) actually creates an english auction, can you check that?

kdembler (approved):

PR #2457, opened 2022-04-08
implement figma import for illustrations

rafalpawlow:
While doing another task, @toiletgranny came up with the idea that we could do an figma import for illustrations. So here it is.


kdembler (changes requested):
This imports a lot of illustrations that we won't be using in our codebase, I wouldn't want dead code to be laying around. I think we need to somehow restrict it to only illustration that Atlas is using

rafalpawlow:
@toiletgranny could you remove unsused illustrations from figma file?

toiletgranny:
@rafalpawlow done!

kdembler (changes requested):
@rafalpawlow Generated illustrations still include some unnecessary ones

kdembler (approved):

Issue #2312, opened 2022-03-17
Update the Dialog (Modal) component


toiletgranny:
**AC:**

- [ ] Make sure positioning, sizing, margins, and overflows are implemented correctly
- [ ] Update animation if needed
- [ ] Add corner radii to dialog

The goal of this task is to 
1. Update the Dialog component based on the designs, docs, and changelog which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6261%3A134667). Please note that this component utilizes a new set of design tokens for corner radii. A separate GitHub issue for that can be found here: https://github.com/Joystream/atlas/issues/2273. 
![image](https://user-images.githubusercontent.com/35307309/158852664-cece098e-dfd4-4a49-ad97-26aa72a41b4d.png)

2. Update the "Modal" component's implementation based on new documentation in Figma. A reminder that so far, there was no explicit design documentation for the behavior of _modal content_ in Atlas therefore I sense that support from @kdembler might be needed in translating the design docs into actionable tickets for engineers. The docs in Figma attempt to capture and document the current usage of modals in Atlas while taking the opportunity to propose new solutions and improvements. 
![image](https://user-images.githubusercontent.com/35307309/158853011-1bd3a724-1b13-4483-bc99-bd7d28e30ed8.png)
![image](https://user-images.githubusercontent.com/35307309/158854402-d0bb10af-0efd-4bdf-976c-48f48d7befcf.png)

kdembler:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Atlas-600edbb8799cc700103bc2d7/issues/Joystream/atlas/2312?planning-poker) with ZenHub @drillprop @eldiegod @rafalpawlow

PR #2462, opened 2022-04-11
Update the Dialog (Modal) component

eldiegod:
fixes: #2312

### DIFF
--- a/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
+++ b/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
@@ -103,7 +103,7 @@ export const footerDividersStyles = css`
 `
 
 type FooterProps = {
-  'data-hasadditionalactions'?: boolean
+  'data-has-additional-actions'?: boolean
   additionalActionsNodeMobilePosition?: 'bottom' | 'top'
 } & DividersProps
 
@@ -122,8 +122,8 @@ export const Footer = styled.div<FooterProps>`
     align-items: center;
   }
 
-  &[data-hasadditionalactions='false'],
-  &:not([data-hasadditionalactions]) {
+  &[data-has-additional-actions='false'],
+  &:not([data-has-additional-actions]) {
     justify-content: end;
   }
 `
--- a/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
+++ b/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
@@ -7,11 +7,11 @@ import { cVar, media, sizes } from '@/styles'
 export type DialogSize = 'default' | 'compact'
 
 type DividersProps = {
-  dividers: boolean
+  dividers?: boolean
 }
 
 type SizeProps = {
-  size: 'default' | 'compact'
+  size: DialogSize
 }
 
 type ContentProps = {
@@ -60,6 +60,16 @@ export const Header = styled.div<DividersProps>`
   align-items: flex-start;
   padding: var(--local-size-dialog-padding) var(--local-size-dialog-padding) 0;
 
+  button:last-of-type {
+    margin-left: ${sizes(6)};
+  }
+
+  ${media.sm} {
+    button:last-of-type {
+      margin-left: ${sizes(10)};
+    }
+  }
+
   ${({ dividers }) => dividers && headerDividersStyles};
 `
 
@@ -101,6 +111,7 @@ type FooterProps = {
 } & DividersProps
 
 export const Footer = styled.div<FooterProps>`
+  width: 100%;
   padding: 0 var(--local-size-dialog-padding) var(--local-size-dialog-padding);
   display: flex;
   justify-content: space-between;
--- a/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
+++ b/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
@@ -103,7 +103,7 @@ export const footerDividersStyles = css`
 `
 
 type FooterProps = {
-  hasAdditionalActions: boolean
+  hasAdditionalActions?: boolean
   additionalActionsNodeMobilePosition: 'bottom' | 'top'
 } & DividersProps
 
--- a/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
+++ b/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
@@ -103,7 +103,7 @@ export const footerDividersStyles = css`
 `
 
 type FooterProps = {
-  'data-hasAdditionalActions'?: boolean
+  'data-hasadditionalactions'?: boolean
   additionalActionsNodeMobilePosition?: 'bottom' | 'top'
 } & DividersProps
 
@@ -122,8 +122,8 @@ export const Footer = styled.div<FooterProps>`
     align-items: center;
   }
 
-  &[data-hasAdditionalActions='false'],
-  &:not([data-hasAdditionalActions]) {
+  &[data-hasadditionalactions='false'],
+  &:not([data-hasadditionalactions]) {
     justify-content: end;
   }
 `
--- a/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
+++ b/packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
@@ -103,8 +103,8 @@ export const footerDividersStyles = css`
 `
 
 type FooterProps = {
-  hasAdditionalActions?: boolean
-  additionalActionsNodeMobilePosition: 'bottom' | 'top'
+  'data-hasAdditionalActions'?: boolean
+  additionalActionsNodeMobilePosition?: 'bottom' | 'top'
 } & DividersProps
 
 export const Footer = styled.div<FooterProps>`
@@ -112,7 +112,7 @@ export const Footer = styled.div<FooterProps>`
   padding: 0 var(--local-size-dialog-padding) var(--local-size-dialog-padding);
   display: flex;
   justify-content: space-between;
-  flex-direction: ${({ additionalActionsNodeMobilePosition }) =>
+  flex-direction: ${({ additionalActionsNodeMobilePosition = 'top' }) =>
     additionalActionsNodeMobilePosition === 'bottom' ? 'column-reverse' : 'column'};
 
   ${({ dividers }) => dividers && footerDividersStyles};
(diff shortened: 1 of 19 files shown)


packages/atlas/src/components/_overlays/Modal/Modal.tsx
```diff
@@ -24,12 +26,12 @@ export const Modal: React.FC<ModalProps> = ({ children, show, className, noBoxSh
 
   return (
     <Portal containerRef={modalContainerRef}>
-      <CSSTransition in={show} timeout={200} classNames={transitions.names.fade} mountOnEnter unmountOnExit>
-        <ModalBackdrop />
+      <CSSTransition in={show} timeout={250} classNames={transitions.names.fade} mountOnEnter unmountOnExit>
```
drillprop: We should use `cVar` here: `timeout={parseInt(cVar('animationTimingMedium', true))}`

packages/atlas/src/components/_overlays/Modal/Modal.tsx
```diff
@@ -24,12 +26,12 @@ export const Modal: React.FC<ModalProps> = ({ children, show, className, noBoxSh
...
+        <ModalBackdrop onClick={onExitClick} />
       </CSSTransition>
       <CSSTransition
         in={show}
-        timeout={200}
+        timeout={250}
```
drillprop: Same as above

packages/atlas/src/providers/confirmationModal.tsx
```diff
@@ -57,15 +57,31 @@ export const useConfirmationModal = (modalProps?: DialogModalProps) => {
...
+          <DialogModal
+            {..._args}
+            onExitClick={
+              _args?.onExitClick
+                ? () => {
+                    _args?.onExitClick?.()
+                    _closeModal()
+                  }
+                : undefined
+            }
```
drillprop: Hmm, I don't like how it's written, can we create the handler above component? You moved `args || modalProps` to variable anyway. I would also write this little differently i.e: 
```tsx
        const handleClick = () => {
          if (!_args?.onExitClick) {
            return undefined
          }
          _args?.onExitClick?.()
          _closeModal()
        }
```
This is almost the same and in my opinion more readable

packages/atlas/src/styles/transitions.ts
```diff
@@ -18,7 +20,7 @@ export const transitions = {
     fadeAndSlide: 'fade-slide',
     slide: 'slide',
     slideDown: 'slide-down',
-    modal: 'modal',
+    modal: 'atlas-modal',
```
drillprop: Why do we need to change that name? Do we have another "modal" somewhere?

drillprop (changes requested):
These changes don't work well with the create member modal. This is how it looks on a 550px viewport

image

I'm also not sure if we should keep this particular dialog on the bottom of the screen. @toiletgranny wdyt?


PS: please add a link to the design in your PRs 🙏🏻

toiletgranny (changes requested):
I'm so happy to see this coming alive so quickly! ❤️ I did, however, find a few issues that must be addressed.

1. **Modal animations.** I wasn't able to see any desktop-sized dialog appearing on screen with the scale animation. Not in the Storybook, nor in the app. Are you sure the animation was implemented properly? As a reminder, new transition animations are described [in Figma, here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6341%3A146413) (the "Animation" section at the very bottom).
2. **Dialog header.** — Please note, that there should be a `24px` (or `40px` starting with the SM breakpoint) of distance between the close button and the dialog title in the Header section. Currently, there's no gap in between these elements.
3. **Sign in** — This dialog has been left unaffected by our changes, it seems. Would it be a lot of work to have this component inherit the recent changes we made to dialogs and overlays, such as new transitions, new responsive behavior, rounded corners, the styling of the footer, etc.? At the very least, can we make sure it doesn't stop shrinking horizontally when reaching `690px` of width? Of course, ideally, all future updates to the styling/behavior of our dialogs/modals in Atlas are also picked up in this dialog. 
CleanShot 2022-04-12 at 15 53 16
4. **Create membership** — Same comment as above.
CleanShot 2022-04-12 at 15 38 14
5. **Processing and errors** — In the below two examples, both dialogs are — obviously — too wide. As a reminder, from the docs in Figma: "_The width of the modal contents should also be freely customizable, with 480px being a default value and 768px and 1120px being preferable presets for extending._"
CleanShot 2022-04-12 at 15 43 08
CleanShot 2022-04-12 at 15 51 30

---

Now, @kdembler, I wasn't able to access Studio in a way that would allow me to inspect dialog components over there. No matter which Polkadot account or chain I tried, I would always get stuck at the "Oops! An error occurred." screen. For this reason, I don't know if there are other issues with dialogs that are a result of these changes. Is there anything else I can try to access Studio in this PR's... _deployment_ (if that's the word 😅 )?

eldiegod:
@toiletgranny did what I could on the existing dialogs I think most things should be fixed pls take a look again and lmk !

drillprop (changes requested):
Still doesn't feel right to me:
image
image

image

@toiletgranny I don't see the answer to my question. I'm not sure if this topic was raised.  Maybe you discussed this with Diego somewhere else, so I'll ask again:
Should we keep the member dialog(only this particular dialog) at the bottom of the page on smaller screens? 👇🏻 

image
 
Shouldn't we leave this centered as it was before?

toiletgranny:
Thanks, @eldiegod! I can confirm that the (1) Modal animations and (2) Dialog header issues I've listed in my previous comment are now fixed, thanks!

1. **Max-width** — As @drillprop pointed out, the issue of overly wide dialogs is still present, so for quick reference:
   - The max-width of both the [Transaction dialog](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=5207%3A204082) and [Alert dialog](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6278%3A138198) components, starting on the SM breakpoint, should be both `480px`. In Figma, you'll find out that the max-width of the Transaction dialog component is set to `440px`, but this is due to historic reasons and I'm 100% fine with it following a global rule. 
   - The max-width of both the Sign in and Create membership dialogs should be set to `768px`.
   - For the future, let's keep in mind that `480px` should be the _default_ max-width value for all Dialog components. 
2. **Sign in**
   - The footer in this dialog def looks more aligned with our recent changes now, and the only thing that's still missing is the top border. 
   - Also, notice how the entire window is scrollable when there's not enough vertical space to display it fully in the viewport. Ideally, we're following the same pattern as with all other scrollable Dialog components, which is that only the body part is scrollable.  You can see this very clearly on the screen recording below when I'm switching between the Sign in (incorrect scrolling behavior) dialog and the Create membership dialog (correct scrolling behavior). 

https://user-images.githubusercontent.com/35307309/165274933-d3e97ac4-df90-4a1e-8e53-0f1da4d0543b.mp4

3. **Create membership** — It seems like either not all changes to Dialog component footers have been picked up in the Create membership dialog, or that we might have missed that one change entirely. Namely, the size of Button components inside Dialog footers has changed from Large to Medium. In the Create membership dialog, however, the size of the button in the future is still Large.

---

To answer your question, @drillprop: [following the new designs](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6341%3A146413) ("Positioning, size, marings & overflow" section), all Dialog components (or, if you prefer, all components placed inside of the Modal component) should be placed at the bottom and be stretched to fill the entire width of the viewport. So **yes**, we should keep the member dialog (Sign in dialog) component at the bottom of the page on smaller screens, just like every other Dialog component up until the SM breakpoint.

eldiegod:
> To answer your question, @drillprop: [following the new designs](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6341%3A146413) ("Positioning, size, marings & overflow" section), all Dialog components (or, if you prefer, all components placed inside of the Modal component) should be placed at the bottom and be stretched to fill the entire width of the viewport. So **yes**, we should keep the member dialog (Sign in dialog) component at the bottom of the page on smaller screens, just like every other Dialog component up until the SM breakpoint.


@toiletgranny ok, I think I covered the bad widths and other stuff this time around, lmk if you find anything else!

drillprop (approved):
Looked at the code again and tested this visually and LGTM!  🎉

drillprop (changes requested):
Oh sorry, there is still a warning in the console. This is happening when I'm opening create member modal
image

toiletgranny (approved):
Now it's perfect! Good to go from my perspective!

eldiegod:
should be fixed

packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
```diff
@@ -81,25 +92,27 @@ const getDenseHeaderContentStyles = ({ denseHeader, noContentPadding }: ContentP
...
 export const footerDividersStyles = css`
-  box-shadow: inset 0 1px 0 0 ${oldColors.gray[600]};
+  box-shadow: ${cVar('effectDividersTop')};
   padding-top: var(--local-size-dialog-padding);
 `
 
 type FooterProps = {
-  hasAdditionalActions: boolean
-  additionalActionsNodeMobilePosition: 'bottom' | 'top'
+  'data-hasadditionalactions'?: boolean
```
drillprop:
```suggestion
  'data-has-additional-actions'?: boolean
```
This is not really readable

packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
```diff
@@ -108,12 +121,17 @@ export const Footer = styled.div<FooterProps>`
     flex-direction: row;
     align-items: center;
   }
+
+  &[data-hasadditionalactions='false'],
+  &:not([data-hasadditionalactions]) {
```
drillprop:
```suggestion
  &[data-has-additional-actions='false'],
  &:not([data-has-additional-actions]) {
```

packages/atlas/src/components/_overlays/Dialog/Dialog.tsx
```diff
@@ -104,7 +114,7 @@ export const Dialog: React.FC<DialogProps> = ({
       {hasFooter && (
         <Footer
           dividers={dividers || actionDivider}
-          hasAdditionalActions={!!additionalActionsNode}
+          data-hasadditionalactions={!!additionalActionsNode}
```
drillprop:
```suggestion
          data-has-additional-actions={!!additionalActionsNode}
```

drillprop (changes requested):

packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
eldiegod: yeah probably better

drillprop (approved):

Issue #2432, opened 2022-04-07
NFT widget, incorrect buttons when auction ended


drillprop:
if the auction has ended we shouldn't show `place a bid` and `buy now` buttons
![image](https://user-images.githubusercontent.com/51168865/162216532-83f3790f-50fe-4943-9736-cab3aeee822f.png)

User should see: 
![image](https://user-images.githubusercontent.com/51168865/162217277-7d56e098-5356-4b93-8df9-bbcd7e79a295.png)

Designs: https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2077%3A139886

As part of this task make sure that we see also set the correct buttons on the owner's side: 

![image](https://user-images.githubusercontent.com/51168865/162218099-d3ddb511-27e0-40d2-ab41-c9ba59312454.png)

PR #2466, opened 2022-04-11
Correct buttons on expired auction - Nft widget

eldiegod:
fixes: #2432

test at: `/video/10`

### DIFF
--- a/packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
+++ b/packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
@@ -52,6 +52,7 @@ export type Auction = {
   userBidUnlockDate: Date | undefined
   canWithdrawBid: boolean | undefined
   canChangeBid: boolean | undefined
+  hasTimersLoaded: boolean | undefined
   englishTimerState: EnglishTimerState | undefined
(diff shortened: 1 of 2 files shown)


rafalpawlow:
@eldiegod this PR needs rebase

eldiegod:
@rafalpawlow done!

rafalpawlow (changes requested):
Generally works well, however during loading `Place a bid` button is visible and it's actually clickable.

https://user-images.githubusercontent.com/13706246/163243415-4ef038d6-d1d2-4a4c-b1f2-c8843b43cfbe.mov

eldiegod:
@kdembler more problems with the timers loading, why don't we just wait till everything loads to show the widget? or alternatively have a `currentBlock` field in the payloads that we can compare with so we can at least know the initial state of the auction before timers kick in

kdembler:
> or alternatively have a currentBlock field in the payloads that we can compare with so we can at least know the initial state of the auction before timers kick in

I would love that, but it's not really feasible at this point, we gotta work with what we have :( Can we just not show any buttons until the timers are finished loading?

eldiegod:
Pushed the fix to not show buttons until timers load

kdembler (approved):

PR #2474, opened 2022-04-12
Comment component

### LINKED ISSUE #2452: Comment component
**AC:**

- [ ] Comment body component as seen in Figma: https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411%3A217091
  - [ ] Loading state
- [ ] Comment component combining ☝️ and Comment row (#2453)

**Out of scope:**

- [ ] Reactions
- [ ] Replies

drillprop:
Fix #2452 

Designs: 
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411%3A217092

The component is available in the storybook. Be aware that the reactions footer isn't in the scope of this PR.

### DIFF
--- a/packages/atlas/src/components/_comments/Comment/Comment.stories.tsx
+++ b/packages/atlas/src/components/_comments/Comment/Comment.stories.tsx
@@ -1,6 +1,6 @@
 import { Meta, Story } from '@storybook/react'
 import React from 'react'
-import { BrowserRouter } from 'react-router-dom'
+import { MemoryRouter } from 'react-router-dom'
 
 import { Comment, CommentProps } from './Comment'
 
@@ -18,9 +18,9 @@ export default {
   },
   decorators: [
     (Story) => (
-      <BrowserRouter>
+      <MemoryRouter>
         <Story />
-      </BrowserRouter>
+      </MemoryRouter>
     ),
   ],
   argTypes: {
--- a/packages/atlas/src/components/_comments/Comment/Comment.stories.tsx
+++ b/packages/atlas/src/components/_comments/Comment/Comment.stories.tsx
@@ -25,6 +25,9 @@ export default {
   ],
   argTypes: {
     createdAt: { control: 'date' },
+    onDeleteClick: { table: { disable: true } },
+    onEditClick: { table: { disable: true } },
+    onEditLabelClick: { table: { disable: true } },
   },
 } as Meta<CommentProps>
 
--- a/packages/atlas/src/components/_comments/Comment/Comment.stories.tsx
+++ b/packages/atlas/src/components/_comments/Comment/Comment.stories.tsx
@@ -1,0 +1,33 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+import { BrowserRouter } from 'react-router-dom'
+
+import { Comment, CommentProps } from './Comment'
+
+export default {
+  title: 'comments/Comment',
+  component: Comment,
+  args: {
+    memberAvatarUrl: 'https://placedog.net/100/100?random=2',
+    memberHandle: 'johndoe',
+    loading: false,
+    type: 'default',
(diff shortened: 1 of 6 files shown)


packages/atlas/src/components/_comments/Comment/Comment.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  margin-bottom: ${sizes(2)};
+`
+
+export const CommentHeaderDot = styled.div`
+  background-color: ${cVar('colorBackgroundAlpha')};
+  border-radius: 100%;
+  width: ${sizes(1)};
```
rafalpawlow: if I remember correctly we decided to not use `sizes` for width and height. Also, just a reminder that we have `square` helper 😛

packages/atlas/src/components/_comments/Comment/Comment.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+      'Lorem ipsum dolor sit amet consectetur adipisicing elit. Officia, facere molestias, iusto unde harum ducimus reprehenderit iste, magnam delectus aperiam impedit eligendi iure autem repellendus dolorum culpa sed cupiditate illum.',
+  },
+  decorators: [
+    (Story) => (
+      <BrowserRouter>
+        <Story />
+      </BrowserRouter>
+    ),
+  ],
+  argTypes: {
```
rafalpawlow, line 26: Let's remove unused args like: `onEditLabelClick` `onEditClick` `onDeleteClick`

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+import { CommentRow, CommentRowProps } from '../CommentRow'
+
+export type CommentProps = {
+  memberHandle?: string
+  createdAt?: Date
+  comment?: string
+  loading?: boolean
+  isEdited?: boolean
+  type: 'default' | 'deleted' | 'options'
```
rafalpawlow, line 33: I'm thinking, shouldn't we use some `isOwner` flag instead of `options` type? WDYT?

rafalpawlow (changes requested):

packages/atlas/src/components/_comments/Comment/Comment.tsx
drillprop: I'm not sure, I would rather stick to the types defined by designers in this case. Remember that also the creator of the video can delete a comment.

rafalpawlow (approved):

toiletgranny:
It was a pleasure reviewing this! @drillprop thank you for taking the time to put together such a fantastic Storybook page for this component, it made the process so much easier!

I have a few thoughts/questions for you to consider:
1. I might be missing something, but for now it seems like we have an unnecessary 8px of a bottom margin underneath the comment's content. `.css-19idom { margin-bottom: 8px; }` Or is it because of the reactions footer that's not there yet?
2. I realize this might not be relevant at the moment, but just in case: let's keep in mind that the timestamp should be a hyperlink to that comment.
3. Ideally, there's a 1-second delay for displaying a tooltip with the exact timestamp. Currently, the tooltip is displayed instantly. Is this something we can add?
4. I might have missed that in the documentation, so apologies for that in advance, but ideally, both the avatar and the member handle/name should be hyperlinks to that member profile page. Is this something we can add?

drillprop:
@toiletgranny 
1. Removed
2. For now it's impossible to implement because we only have this component available in the storybook. But I added a "to-do" comment in the code, to not forget about this. 
3. Added
4. Added(but it might be hard to test this in the storybook)

toiletgranny (approved):

PR #2475, opened 2022-04-12
update tooltip component

### LINKED ISSUE #2272: Update the Tooltip component
The goal of this task is to update the Tooltip component based on the designs, docs, and changelog which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6215%3A139642).

Please note that it's the first component to utilize a new set of design tokens for corner radii. A separate GitHub issue for that can be found here: https://github.com/Joystream/atlas/issues/2273.

image

rafalpawlow:
Closes #2272 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6215%3A139642

I talked to @toiletgranny about the tooltip "footer", we made some decisions but it's waiting for figma implementation.

### DIFF
--- a/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
@@ -17,6 +17,8 @@ export default {
     className: { table: { disable: true } },
     reference: { table: { disable: true } },
     hideOnClick: { table: { disable: true } },
+    delay: { table: { disable: true } },
+    showOnCreate: { table: { disable: true } },
     customContent: { type: 'boolean' },
   },
   args: {
@@ -26,6 +28,7 @@ export default {
     placement: 'bottom-start',
     text: 'Lorem ipsum dolor amet! Lorem ipsum dolor amet!',
     headerText: 'Lorem ipsum',
+    oneLine: false,
   },
 } as Meta<TooltipProps>
 
--- a/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
@@ -10,24 +10,40 @@ export default {
   title: 'other/Tooltip',
   component: Tooltip,
   argTypes: {
-    text: { defaultValue: 'Lorem ipsum dolor amet! Lorem ipsum dolor amet!' },
-    headerText: { defaultValue: 'Lorem ipsum' },
     placement: {
       control: { type: 'select', options: ['top-start', 'top-end', 'bottom-start', 'bottom-end'] },
-      defaultValue: 'bottom-start',
     },
     icon: { defaultValue: false },
-    offsetX: { defaultValue: 0 },
-    offsetY: { defaultValue: 0 },
     className: { table: { disable: true } },
     reference: { table: { disable: true } },
-    arrowDisabled: { table: { disable: true } },
-    footer: { control: { type: 'boolean' } },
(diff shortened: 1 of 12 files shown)


packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -10,29 +10,53 @@ export default {
...
       defaultValue: 'bottom-start',
     },
     icon: { defaultValue: false },
     offsetX: { defaultValue: 0 },
     offsetY: { defaultValue: 0 },
+    className: { table: { disable: true } },
+    reference: { table: { disable: true } },
+    arrowDisabled: { table: { disable: true } },
+    footer: { control: { type: 'boolean' } },
+  },
```
drillprop: could you clean the storybook here and use `args` instead of `argTypes` to set the default values and controls?

packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -10,29 +10,53 @@ export default {
...
+  args: {
+    footer: false,
   },
 } as Meta
```
drillprop:
```suggestion
} as Meta<TooltipProps>
```

packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -10,29 +10,53 @@ export default {
...
     placement: {
       control: { type: 'select', options: ['top-start', 'top-end', 'bottom-start', 'bottom-end'] },
       defaultValue: 'bottom-start',
     },
     icon: { defaultValue: false },
     offsetX: { defaultValue: 0 },
     offsetY: { defaultValue: 0 },
+    className: { table: { disable: true } },
+    reference: { table: { disable: true } },
+    arrowDisabled: { table: { disable: true } },
```
drillprop: we don't need `arrowDisabled` anymore, right?

packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
drillprop: Also, could you add a template where the tooltip is always visible? It's really hard to inspect them with dev tools and I think it could be useful for us and for designers.

drillprop (changes requested):

drillprop (changes requested):
I think we need to do more changes. 
In Figma, 
- there are multi-line and one-line variants with different paddings.
- there is also the so-called `custom type` tooltip, where you could add anything.

rafalpawlow:
@drillprop 
> there is also the so-called custom type tooltip, where you could add anything.

For this purpose there is `footer` prop where you can pass `ReactNode`

drillprop:
> @drillprop
> 
> > there is also the so-called custom type tooltip, where you could add anything.
> 
> For this purpose there is `footer` prop where you can pass `ReactNode`

if so, can we name this differently? This won't be always a footer, right?

drillprop (changes requested):
There is still one issue. If there is no text or header and you add you add custom content, you won't see the tooltip: 
image

I also found that something weird is happening with centering when you check these controls: 
image

image

I guess having a one-line tooltip with custom content will never happen, but maybe it's worth investigating and fixing(Im fine if you don't resolve this bug)

drillprop (approved):

PR #2494, opened 2022-04-13
bid lock duration when changing bid

### LINKED ISSUE #2429: Apply bid lock duration to changing bid
In an open auction, changing bid is subject to bid lock duration as well. However, it only applies if you want to change your bid to a lower one. Bids can be increased without any restrictions

rafalpawlow:
Closes #2429 any by mistake (I didn't see issue) #2498

**Design**
https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-purchase-view?node-id=1846%3A24907

### DIFF
--- a/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
+++ b/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
@@ -61,14 +61,14 @@ export const NftPurchaseBottomDrawer: React.FC = () => {
   const [showBuyNowInfo, setBuyNowInfo] = useState(false)
   const { currentAction, closeNftAction, currentNftId, isBuyNowClicked } = useNftActions()
   const { nft, nftStatus, loading, refetch } = useNft(currentNftId || '')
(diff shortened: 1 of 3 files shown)


packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
```diff
@@ -394,6 +400,21 @@ export const NftPurchaseBottomDrawer: React.FC = () => {
...
+                          value < Number(userBid?.amount) &&
+                          timeToUnlockSeconds &&
+                          timeToUnlockSeconds > 0
+                        ) {
+                          return `Your bid is locked from placing a bid lower than your previous one. You can bid for a higher amount or try again in ${formatDurationShort(
+                            Number(timeToUnlockSeconds.toFixed()),
+                            false,
+                            true
+                          )}`
+                        }
```
kdembler: This timer won't be updating unfortunately so I think maybe it's not worth having the duration here. Let's overwrite the design here and just use a timestamp. Something like `Your bid is currently locked and cannot be lowered. It will be unblocked at 14 Apr 2022 16:10.

kdembler (approved):

PR #2500, opened 2022-04-14
channel view NFT collectors

### LINKED ISSUE #2479: Reintroduce NFT collectors in channel view
We've hidden NFT collectors inside the channel view because the data for it wasn't available. This can be done now (see https://github.com/Joystream/joystream/pull/3563). Let's bring it back and integrate with real data

rafalpawlow:
Closes #2479 

**Design**
https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=510%3A33487

### DIFF
--- a/packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
+++ b/packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
@@ -1,5 +1,6 @@
 import { generateChannelMetaTags } from '@joystream/atlas-meta-server/src/tags'
 import React, { useEffect, useMemo, useRef, useState } from 'react'
+import { useNavigate } from 'react-router'
 import { useParams, useSearchParams } from 'react-router-dom'
 
 import { useChannel, useChannelNftCollectors } from '@/api/hooks'
@@ -55,6 +56,7 @@ export const ChannelView: React.FC = () => {
   const [tilesPerRow, setTilesPerRow] = useState(INITIAL_TILES_PER_ROW)
   const currentTabName = searchParams.get('tab') as typeof TABS[number] | null
   const videoRows = useVideoGridRows('main')
+  const navigate = useNavigate()
 
   const tilesPerPage = videoRows * tilesPerRow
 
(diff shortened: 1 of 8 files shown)


packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -180,9 +181,14 @@ export const ChannelView: React.FC = () => {
...
-  // TODO: replace with real NFT collector data
-  const collectors: Collector[] = []
+  const mappedChannelNftCollectors =
+    channelNftCollectors && channelNftCollectors.length
+      ? channelNftCollectors?.map(({ amount, member }) => ({
+          nftsAmount: amount,
+          url: member?.metadata.avatar?.__typename === 'AvatarUri' ? member?.metadata.avatar?.avatarUri : '',
+          tooltipText: member?.handle,
+        }))
+      : []
```
drillprop:
```suggestion
  const mappedChannelNftCollectors =
    channelNftCollectors?.map(({ amount, member }) => ({
      nftsAmount: amount,
      url: member?.metadata.avatar?.__typename === 'AvatarUri' ? member?.metadata.avatar?.avatarUri : '',
      tooltipText: member?.handle,
    })) || []
```
should do the job

packages/atlas/src/views/viewer/ChannelView/ChannelView.tsx
```diff
@@ -210,7 +216,9 @@ export const ChannelView: React.FC = () => {
       <LimitedWidthContainer>
         {smMatch ? (
           <CollectorsBoxContainer>
-            {collectors.length > 0 && <CollectorsBox collectors={collectors} />}
+            {mappedChannelNftCollectors && mappedChannelNftCollectors.length > 0 && (
```
drillprop:
```suggestion
            {mappedChannelNftCollectors.length > 0 && (
```
I think the result should be the same, especially if you apply my previous suggestion

drillprop (changes requested):
The code looks good but I've noticed that it's not useful at all. I think we should add an anchor tag to every member in the collector's box dialog and in the AvatarGroup as well. 

Although I'm not sure about the AvatarGroup. Could you ask a designer? Maybe this is described in the designs somewhere? If we decide that it shouldn't be interactable, we should make the AvatarGroup static(without this jumping transform effect, I'm almost sure that we have a prop for that)

drillprop (approved):

PR #2505, opened 2022-04-14
improve notifications fetching

### LINKED ISSUE #2424: Improve notifications fetching
Followup to #1740. We need to fetch notifications only for NFTs that the current user is owner of, not all of them

kdembler:
Fixes #2424


rafalpawlow (changes requested):
Looks good, but there is a problem with fetching new notifications when you are switching accounts.

1. switch to 2nd account
2. make some action on nft from 1st account
3. switch to 1st account
4. you need to refresh page to get new notifications

kdembler:
Changed the fetch policy so that we refetch when changing members. Also added polling every 10sec

rafalpawlow (approved):

Issue #2356, opened 2022-03-25
Withdraw a bid from a past auction


kdembler:
**AC**

- [ ] Handle withdrawing from a past auction
- [ ] Make sure NFT widget states are correct

Issue #2482, opened 2022-04-12
Fetch last sale price


kdembler:
QN added `lastSale` field on `OwnedNft` so we can see the last price for which it was sold. Let's use this data in the app

PR #2508, opened 2022-04-14
Withdraw a bid from a past auction

eldiegod:
fixes: #2356 and #2482


kdembler (changes requested):
Looks mostly good, but I think we should avoid the current behaviour of fetching bids for all the NFTs, this will result in a lot of requests (one additional request per NFT tile). Could we move the bids fetch to be made only for the NFT widget?

kdembler (approved):

Issue #2468, opened 2022-04-11
Testing ops for Atlas


dmtrjsg:
1. Create a staging env pointing to Rhodes network
2. Show block details and link to polkadot js in console after each transaction is signed. Similar to Pioneer implementation: https://github.com/Joystream/pioneer/issues/2289

kdembler:
1. Sounds pretty easy, np
2. Switching between networks is implemented already but not using `config.json` file since we had that feature long before those files were generated. We basically provide avaiable environments during build time. Whenever you have Atlas open, you can press Ctrl+Shift+D to open developer tools. With that, you can pick which network you want to connect to. If you need separate playground for testing, we can include it in our config and you will be able to use it. Does that work, or do you need to be able to move between different playground dynamically?
3. 👍

kdembler:
Additionally move `atlas-staging` Vercel deployment to Rhodes

PR #2509, opened 2022-04-14
show tx details in console

rafalpawlow:
Part of #2468


rafalpawlow:
@kdembler I'm not sure if the way I passed `endpoint` is correct. I feel that there must be an easiest way but didn't find any.

kdembler (approved):

PR #2521, opened 2022-04-19
add NFT history integration

### LINKED ISSUE #2217: NFT Widget history integration
What should (and can) we support as entries for the Nft widget history?

what I have thought of so far:

Current auction bids
State changes i.e -> nft put on sale/auction/idle
Previous owners
Minting

kdembler:
fixes #2217

### DIFF
--- a/packages/atlas/src/components/_nft/NftWidget/NftWidget.hooks.ts
+++ b/packages/atlas/src/components/_nft/NftWidget/NftWidget.hooks.ts
@@ -1,0 +1,240 @@
+import { differenceInCalendarDays, differenceInSeconds } from 'date-fns'
+import { useEffect } from 'react'
+
+import { useNft, useNftHistory } from '@/api/hooks'
+import { useBids } from '@/api/hooks/bids'
+import { NftWidgetProps } from '@/components/_nft/NftWidget/NftWidget'
+import { useNftState } from '@/hooks/useNftState'
+import { useMemberAvatar } from '@/providers/assets'
+import { useUser } from '@/providers/user'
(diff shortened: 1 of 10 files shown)


packages/atlas/src/components/_nft/NftWidget/NftWidget.hooks.ts
```diff
@@ -0,0 +1,209 @@
...
+  useEffect(() => {
+    const pollingInterval = 10000
+    startNftPolling(pollingInterval)
+    startHistoryPolling(pollingInterval)
+
+    return () => {
+      stopNftPolling()
+      stopHistoryPolling()
+    }
+  }, [startHistoryPolling, startNftPolling, stopHistoryPolling, stopNftPolling])
```
drillprop: I think the better option would be using `pollInterval`. https://www.apollographql.com/docs/react/data/queries/#polling 
We wouldn't need this effect here.

drillprop (changes requested):
Generally looks good. Found one thing to consider and I also noticed one small UI bug(?) - I mean, I'm not sure if it is supposed to look like that. 
This is how `history` looks like with one history entry: 
image

I know that's not your change, but could we also fix that? I think this should be relatively easy. You just need to find the HistoryPanel component and change `height` to `max-height` there. I did something like that in the browser dev tools.

packages/atlas/src/components/_nft/NftWidget/NftWidget.hooks.ts
kdembler: Much better this way, thanks!

drillprop (approved):

PR #2525, opened 2022-04-20
Activity member tab integration

### LINKED ISSUE #2472: Fetch data for Activity of Member tab on Member Profile Details
Pre-cursor: https://github.com/Joystream/atlas/issues/1804

drillprop:
Fix #2472 
Sorting and infinite scrolling aren't in the scope of this PR


eldiegod:
lets do a rebase on this one

packages/atlas/src/views/viewer/MemberView/MemberActivity.hooks.ts
```diff
@@ -0,0 +1,193 @@
...
+      from: BasicMembershipFieldsFragment | null
+    } & NftActivitiesRecord)
+  | ({
+      type: 'Price change'
+      from: BasicMembershipFieldsFragment | null
+      price: number
+    } & NftActivitiesRecord)
+
+const parseActivities = (
+  event: ReturnType<typeof useRawActivities>['activities'][0],
```
eldiegod: We don't need to grab the first item of the array we can just do: 
```suggestion
  event: ReturnType<typeof useRawActivities>['activities'][number],
```

eldiegod (approved):
looks pretty neat

kdembler (changes requested):
We're missing price here:
Screenshot 2022-04-21 at 20 11 40

Also, what do we do with sorting? The select is still there but has no effect. Making it work is probably super easy since we do sorting client side anyway

kdembler (approved):

PR #2533, opened 2022-04-21
update ListItem colors

### LINKED ISSUE #2470: Incorrect text & icon color in the ListItem component
**Describe the bug**

1. As is: The color of the label and icon in the ListItem component, in the default state, is set to `var(--color-core-neutral-300)`,
2. To be: The color of the label and icon in the ListItem component, in the default state, should be `var(--color-text)`.

rafalpawlow:
Closes #2470 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3152%3A103472

### DIFF
--- a/packages/atlas/src/components/ListItem/ListItem.styles.ts
+++ b/packages/atlas/src/components/ListItem/ListItem.styles.ts
@@ -73,7 +73,7 @@ const captionRightStyles = ({ captionPosition }: CaptionPosition) =>
     text-align: right;
   `
 export const Caption = styled(Text)<CaptionPosition>`
-  ${captionRightStyles}
+  ${captionRightStyles};
 `
 
 export const LabelContainer = styled.div`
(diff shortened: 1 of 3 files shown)


packages/atlas/src/components/ListItem/ListItem.styles.ts
```diff
@@ -93,7 +93,7 @@ const iconStyles = ({ destructive, isHovering, isSelected }: NodeContainerProps)
         ? cVar('colorTextError')
         : isHovering || isSelected
         ? cVar('colorCoreNeutral50')
-        : cVar('colorCoreNeutral300')};
+        : cVar('colorTextStrong')};
```
drillprop: This is the wrong color. We should use `colorText`

packages/atlas/src/components/ListItem/ListItem.tsx
```diff
@@ -73,7 +73,7 @@ export const ListItem = React.forwardRef<HTMLDivElement, ListItemProps>(
                   ? cVar('colorTextError')
                   : isHovering || selected
                   ? cVar('colorCoreNeutral50')
-                  : cVar('colorCoreNeutral300')
+                  : cVar('colorTextStrong')
```
drillprop: Same here

drillprop (changes requested):

drillprop (approved):

PR #2540, opened 2022-04-21
initial comments&reactions integration and playground

kdembler:
Adds initial integration of comments&reactions from QN. This only includes subset of extrinsic, no queries have been added yet. Seems the edit comment extrinsic isn't working properly atm, fix in QN is pending


packages/atlas/src/views/playground/Playgrounds/PlaygroundReactionsComments.tsx
```diff
@@ -0,0 +1,246 @@
...
+      {getTabContents()}
+    </div>
+  )
+}
+
+type CommonProps = {
+  videoId: string
+}
+
+const ReactToVideo: React.FC<CommonProps> = ({ videoId }) => {
```
drillprop, line 57: There is no option to remove the reaction. Is it possible? I thought when someone clicks a second time on the Like button, we should remove the reaction, right?

packages/atlas/src/views/playground/Playgrounds/PlaygroundReactionsComments.tsx
```diff
@@ -0,0 +1,246 @@
...
+        value={videoReaction}
+        onChange={(value) => value && setVideoReaction(value)}
+        items={reactionItems}
+        label="Reaction type"
+      />
+      <Button onClick={handleLike}>React</Button>
+    </div>
+  )
+}
+const ReactToComment: React.FC<CommonProps> = () => {
```
drillprop, line 95: A similar question to the above. Is it possible to cancel the reaction?

drillprop, line 95: A similar question to the above. Is it possible to cancel the reaction?

packages/atlas/src/views/playground/Playgrounds/PlaygroundReactionsComments.tsx
kdembler: Yes, and that's also how you do in the QN, you need to send your like again and it will cancel the original one

packages/atlas/src/views/playground/Playgrounds/PlaygroundReactionsComments.tsx
kdembler: Should be the same as above

drillprop (approved):

Issue #2450, opened 2022-04-08
Comment Input component


kdembler:
This should use Comment row inside (#2453)

https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411%3A217371

kdembler:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Atlas-600edbb8799cc700103bc2d7/issues/Joystream/atlas/2450?planning-poker) with ZenHub @drillprop @eldiegod @rafalpawlow

PR #2558, opened 2022-04-25
Comment input

eldiegod:
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411%3A217347

fixes: #2450


packages/atlas/src/components/_comments/CommentInput/CommentInput.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    ),
+  ],
+  argTypes: {},
+} as Meta<CommentInputProps>
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const Template: Story<CommentInputProps & any> = (args) => (
```
rafalpawlow: Why do we need `any` here?

eldiegod: because I'm using custom props in the story that are not part of `CommentInputProps`

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+    transition: all ${cVar('animationTransitionFast')};
+  }
+`
+export const Border = styled.div<{ 'data-show': boolean; 'data-processing': boolean }>`
+  overflow-x: hidden;
+  width: 100%;
+  height: 1px;
+  background-color: ${cVar('colorCoreNeutral800Lighten')};
+  transition: ${cVar('animationTransitionFast')};
+  transition-delay: cVar('animationTimingFast');
```
drillprop: You missed curly braces here, besides that, why do we need a delay here?

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  &[data-show='true'] {
```
drillprop: Hmm, I'm still unsure how I should treat this `data-` pattern. I'll just leave a comment here for now.

packages/atlas/src/components/_comments/CommentRow/CommentRow.stories.tsx
```diff
@@ -1,48 +0,0 @@
...
-      <Text variant="t200">
-        Lorem ipsum, dolor sit amet consectetur adipisicing elit. Aspernatur aut, ipsam sit impedit voluptatem ad libero
-        necessitatibus porro dignissimos voluptatum quis debitis tempora omnis, provident repellat iure accusantium,
-        blanditiis voluptate!{' '}
-      </Text>
-    </div>
-  </CommentRow>
-)
-
-export const Default = Template.bind({})
```
drillprop: Why did you remove these stories?

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+        data-show={show}
+        height={textAreaHeight}
+        onKeyDown={(e) => {
+          if ((e.nativeEvent.ctrlKey || e.nativeEvent.metaKey) && e.nativeEvent.code === 'Enter') {
+            onComment?.()
+          }
+        }}
+        onClick={() => {
+          setActive(true)
+          textAreaRef.current?.focus()
```
drillprop: Correct me if I'm wrong but I'm pretty sure that you don't need this. If Container had been a label, the text area would just focus automatically.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+} & CommentRowProps
+
+export const CommentInput: React.FC<CommentInputProps> = ({ processing, onCancel, onComment, ...rest }) => {
+  const smMatch = useMediaMatch('sm')
+  const containerRef = useRef<HTMLDivElement>(null)
+  const textAreaRef = useRef<HTMLTextAreaElement>(null)
+  const [active, setActive] = useState(false)
+  const [text, setText] = useState('')
+
+  const { ref: measureRef, height: textAreaHeight = 40 } = useResizeObserver({ box: 'border-box' })
```
drillprop, line 27: If I remember correctly how it works, you can pass ref to the `useResizesObserver` options. You don't need to create another `measureRef`.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+            onComment?.()
+          }
+        }}
+        onClick={() => {
+          setActive(true)
+          textAreaRef.current?.focus()
+        }}
+      >
+        <StyledTextArea
+          ref={mergeRefs([textAreaRef, measureRef])}
```
drillprop, line 70: This would be needed if my comment above is true.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const containerRef = useRef<HTMLDivElement>(null)
+  const textAreaRef = useRef<HTMLTextAreaElement>(null)
+  const [active, setActive] = useState(false)
+  const [text, setText] = useState('')
+
+  const { ref: measureRef, height: textAreaHeight = 40 } = useResizeObserver({ box: 'border-box' })
+
+  useEffect(() => {
+    if (active) textAreaRef.current?.focus()
+  }, [active])
```
drillprop: again - we could remove this if the container had been the label.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+        event.preventDefault()
+        event.stopPropagation()
+        setActive(false)
+      }
+    }
+    document.addEventListener('click', handleClickOutside, true)
+    return () => {
+      document.removeEventListener('click', handleClickOutside, true)
+    }
+  }, [active])
```
drillprop, line 46: onBlur on TextArea isn't enough?

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+    height: 2px;
+    background-color: ${cVar('colorBackgroundPrimary')} !important;
```
drillprop: I would avoid using `!important` as much as I can. I think we could avoid this by passing the prop to StyledCommentRow?

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  &[data-processing='true'] {
+    background-color: ${cVar('colorCoreNeutral700')} !important;
```
drillprop: Same as above

packages/atlas/src/components/_comments/CommentInput/CommentInput.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+    cancelButton: false,
+  },
+  decorators: [
+    (Story) => (
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    ),
+  ],
+  argTypes: {},
```
drillprop: Let's disable `onComment`, `onCancel` and `className` in controls

packages/atlas/src/components/_comments/CommentInput/CommentInput.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+import React from 'react'
+import { MemoryRouter } from 'react-router-dom'
+
+import { CommentInput, CommentInputProps } from './CommentInput'
+
+export default {
+  title: 'comments/CommentInput',
+  component: CommentInput,
+  args: {
+    cancelButton: false,
```
drillprop, line 17: I would add an image URL to `memberAvatarUrl` as default, just to make this story a little prettier and make sure that this prop works.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import { Border, ButtonsContainer, Container, Flex, StyledCommentRow, StyledTextArea } from './CommentInput.styles'
+
+import { CommentRowProps } from '../CommentRow'
+
+export type CommentInputProps = {
+  processing: boolean
+  onComment?: () => void
+  onCancel?: () => void
+} & CommentRowProps
```
drillprop, line 18: Do we need all the CommentRowProps? I think the highlight is not needed

drillprop (changes requested):
Looks really pretty! 
However, I found a couple of things that we could improve. 

In terms of functionality: 
 - You still can write inside the text area when it's processing. 
 - I don't think keyboard navigation is handled. https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2431%3A215801  We could omit the "confirm" scenario, we could do this during the integration PR, but I think "dismiss" scenario we should include in this PR.

drillprop:
Just last thing - I found that when I tap on the information button on my smartphone I don't see any tooltip.

packages/atlas/src/components/_comments/CommentInput/CommentInput.stories.tsx
drillprop: So let's just add `CommentInputProps & { cancelButton: boolean }` it would be just easier to spot which prop isn't part of the `CommentInputProps`

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
eldiegod: > I don't think keyboard navigation is handled. https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2431%3A215801 We could omit the "confirm" scenario, we could do this during the integration PR, but I think "dismiss" scenario we should include in this PR.

actually "confirm" is already there on the `onKeyDown` just need integrations.

For the dismiss I'm not too sure what should happen, I press scape and the whole comment I'm writing gets deleted? can you clarify @toiletgranny

packages/atlas/src/components/_comments/CommentInput/CommentInput.stories.tsx
drillprop: Maybe we could also add `description` to this particular control? I don't know how it works in a storybook - this is not super important,  just a suggestion.

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
drillprop: I think you replied to the wrong comment 😄   Just saying - I hope you won't overlook this 👇🏻 
> You missed curly braces here, besides that, why do we need a delay here?

eldiegod: oops, yeah meant to add a general comment; so I pretty much grabbed the anim from your like dislike component and the delay was there but in this case, don't think is needed

packages/atlas/src/components/_comments/CommentRow/CommentRow.stories.tsx
eldiegod: thought it became redundant since `Comment` and `CommentInput` showcased commentRow in a better scenario

drillprop: Hmm I see, I would opt for not removing it although. I think it's good to know that we have a separate wrapper that can be used in two different contexts.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
eldiegod: yeah, we don't need focus here but we do need to setActive sadly

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
eldiegod: I tried that at first but it gave be buggy results where it would stop measuring this way is how I got it to work

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
eldiegod: removed

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
eldiegod: don't think so, clicking on the container even if its a label will cause it to open and close quickly because it blurs and focuses again

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
eldiegod: not 100% sure but I don't think it will matter too much if not

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
kdembler: Let's accept it for now and we will discuss during next dev sync

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
toiletgranny: @eldiegod @drillprop Yeah, I can see how this part of the documentation was a bit unclear. And to answer your question, pressing escape should not erase your input from the "Comment input" component. It should feel as a regular _blur_ of an input, meaning the "Comment input" component changes its state back to Default.

toiletgranny:
@eldiegod Since Bartek and Rafał already requested some changes, can I ask you to let me know once these are addressed so that I can also jump in and do the design review of the implementation? It would be great if we could always request a review from the designer behind a particular feature! ❤️

packages/atlas/src/components/_comments/CommentRow/CommentRow.stories.tsx
```diff
@@ -2,14 +2,11 @@ import { Meta, Story } from '@storybook/react'
...
-import { cVar } from '@/styles'
-
-import { CommentRow, CommentRowProps } from './CommentRow'
+import { Comment, CommentProps } from '../Comment'
 
 export default {
-  title: 'comments/CommentRow',
-  component: CommentRow,
+  title: 'comments/Comment',
+  component: Comment,
```
drillprop: but this is a CommentRow, not a Comment. We have stories for comment already. This is a different thing.

packages/atlas/src/components/_comments/CommentRow/CommentRow.stories.tsx
```diff
@@ -25,24 +22,8 @@ export default {
...
-      <Text variant="h200">Please ignore styling of this blue box. It's for testing purposes only</Text>
-      <Text variant="t200">
-        Lorem ipsum, dolor sit amet consectetur adipisicing elit. Aspernatur aut, ipsam sit impedit voluptatem ad libero
-        necessitatibus porro dignissimos voluptatum quis debitis tempora omnis, provident repellat iure accusantium,
-        blanditiis voluptate!{' '}
-      </Text>
-    </div>
-  </CommentRow>
-)
+const Template: Story<CommentProps> = (args) => <Comment {...args} />
```
drillprop: same as above

drillprop (changes requested):

eldiegod:
> @eldiegod Since Bartek and Rafał already requested some changes, can I ask you to let me know once these are addressed so that I can also jump in and do the design review of the implementation? It would be great if we could always request a review from the designer behind a particular feature! ❤️

@toiletgranny yep you're right, I think it should be ready for u know once you are back 🤓

drillprop (approved):
LGTM 💪🏻  but let's wait for the final approval from Adam

drillprop:
Last thing:
I just noticed that you forgot to add the `index.ts` file in the `CommentInput` directory.

packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+`
+
+export const ButtonsContainer = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: end;
+  padding: ${sizes(4)} 0 0 0;
+
+  button:nth-last-child(2) {
```
drillprop: I'm getting an error from emotion: 
`The pseudo class ":nth-last-child" is potentially unsafe when doing server-side rendering. Try changing it to ":nth-last-of-type".`

drillprop (changes requested):
Found another thing 😅

drillprop (approved):

toiletgranny (changes requested):
This looks stunning! 🔥  A few minor issues I found:

1. The bottom border color should be set to `var(--color-border-muted-alpha)`; currently, it's var`(--color-core-neutral-800-lighten)`.
2. On hover, the bottom border color should change to `var(--color-border-alpha)`; currently, it's not happening.
3. Ideally, the bottom border color transition is animated using the same timing and easing as everything else in the Comment input component, which would be `var(--animation-transition-medium)`; currently, it's `var(--animation-transition-fast)`.
4. Ideally, the tooltip component is positioned _above_ the Information component, and aligned to the left. Currently, it's placed below the Information component, and aligned to the right side.
5. Not sure if this issue belongs here or if it's just a misuse of the Tooltip component, but it seems like we're missing the max-width in the Tooltip component (the one about comments being stored on the blockchain). Notice how the body text is wrapped in the designs and how it's not the case in the implementation. CC @rafalpawlow.

toiletgranny:
Much better! The bottom border is not quite there yet, however. Upon hovering over the Comment input, instead of changing the background-color property in the `.css-uja03m` class, you set a new background-color property on the `.css-6etn8w` class, which results in:
1. The lack of animation in the border color change,
2. Having a much brighter border color on hover, than anticipated (because the two transparent borders are now overlapping each other: the first one `var(--color-border-muted-alpha)` in the `.css-uja03m` class, and the second one `var(--color-border-alpha)`, in the `.css-6etn8w` class).

Once this minor issue is resolved, it's good to go from my perspective!

PR #2563, opened 2022-04-26
Comment reactions

### LINKED ISSUE #2520: Create comment reactions components
Implement:
- Reaction chip - https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2400%3A217561
- Reaction popover - https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2404%3A218315

drillprop:
Fix #2520 

If you want to test it out, check the ReactionChip, ReactionPopover, and Comment components in Storybook. 
Be aware that, I didn't use`Tweemojis` yet(I'm using default emojis), and I didn't add a reply button. Those will be implemented later.

### DIFF
--- a/packages/atlas/src/components/_comments/ReactionChip/ReactionChip.styles.ts
+++ b/packages/atlas/src/components/_comments/ReactionChip/ReactionChip.styles.ts
@@ -1,9 +1,57 @@
+import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 
-export const ReactionChipButton = styled.button`
+import { cVar, sizes } from '@/styles'
+
+export type ReactionChipState = 'default' | 'disabled' | 'processing'
+type ReactionChipButtonProps = {
+  state: ReactionChipState
+  active: boolean
+}
+
+const getStaticBackground = (props: ReactionChipButtonProps) => {
+  switch (props.state) {
+    case 'processing':
+      return props.active ? cVar('colorBackgroundStrongAlpha') : cVar('colorBackgroundAlpha')
+    default:
+    case 'default':
+    case 'disabled':
+      return props.active ? cVar('colorBackgroundStrongAlpha') : 'none'
+  }
+}
+
+const getHoverStyles = (props: ReactionChipButtonProps) => css`
+  :hover,
+  :focus {
+    background: ${cVar('colorBackgroundStrongAlpha')};
+    box-shadow: ${props.active ? `inset 0 0 0 1px ${cVar('colorBorderStrongAlpha')}` : 'none'};
+  }
+
+  :active {
+    background: ${cVar('colorBackgroundAlpha')};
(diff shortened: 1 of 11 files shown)


toiletgranny:
relieved-caminhos-do-cora%C3%A7%C3%A3o-feels-good-it-feels-nice-thats-the-spot-gif-16348311

I don't

toiletgranny (approved):
relieved-caminhos-do-cora%C3%A7%C3%A3o-feels-good-it-feels-nice-thats-the-spot-gif-16348311

I'm speechless. Not only there is not a single thing not matching the designs (and I checked on Chrome, Safari, Firefox, macOS and iOS!), it's actually better than this. The additional effort I can imagine you have gone through to change the cursor on disabled states, or to ensure a proper keyboard tabbing in the reaction popover is what truly motivates me to keep moving forward being on the same team. 

Maybe it's that I don't need much to have someone making my day, but oh boy — you def made my day. ❤️

toiletgranny:
There's a super tiny detail that on Firefox and Safari emojis are rendered slightly pushed off the center (only on non-retina displays, so with 72ppi), but I don't think it's something that should bother us now, that we don't have the final emoji set implemented yet. Could be a thing to keep an eye on for when we have Twemojis ready.

CleanShot 2022-04-27 at 14 49 15

packages/atlas/src/components/_comments/ReactionChip/ReactionChip.styles.ts
```diff
@@ -0,0 +1,68 @@
...
+  max-height: 16px;
+
+  /* TODO remove this media query once we implement tweemoji */
+
+  /* this is added because chrome has some issue with rendering emojis */
+
+  /* https://stackoverflow.com/questions/42016125/emoji-rendered-in-chrome-have-different-widths-than-in-other-browsers/44145771#44145771 */
+  /* stylelint-disable-next-line  media-feature-name-no-unknown */
+  @media not screen and (min-device-pixel-ratio: 2), not screen and (min-resolution: 192dpi) {
+    transform: translateX(-2px);
```
eldiegod, line 67: For the record, this is applying to other browsers (i.e. firefox) where the problem is not there and the shift is not needed.

then again it might be fixed by using tweemoji

eldiegod (approved):
Looks pretty good, I would suggest to throw in some `formatNumberShort` on the reaction chip for future proof

packages/atlas/src/components/_comments/ReactionChip/ReactionChip.styles.ts
drillprop: yeah, I will just ignore this for now. This would require doing some browser checking and I don't want to waste time on that. I think Klaudiusz already working on adding tweemoji anyway

PR #2576, opened 2022-04-28
fix sidenav scroll hiding

### LINKED ISSUE #2572: Rhodes - scrolling is blocked when you switch from studio to viewer
1. Open navigation, click go to the studio
2. Open navigation again, click go to joystream
3. You can’t scroll
This is happening only on `rhodes` branch.  It looks like `decrementOverlaysOpenCount` is not fired.

rafalpawlow:
Closes #2572

The problem was in `useOverlayManager`. When you change page, `overlayId` ref is changing and `prevSet.delete(overlayId)` is no longer working.

### DIFF
--- a/packages/atlas/src/providers/overlayManager.tsx
+++ b/packages/atlas/src/providers/overlayManager.tsx
@@ -78,7 +78,11 @@ export const useOverlayManager = () => {
 
   const decrementOverlaysOpenCount = useCallback(() => {
     setOverlaysSet((prevSet) => {
-      prevSet.delete(overlayId)
+      if (prevSet.size === 1) {
+        prevSet.clear()
+      } else {
+        prevSet.delete(overlayId)
+      }
       return new Set(prevSet)
     })
   }, [overlayId, setOverlaysSet])


packages/atlas/src/providers/overlayManager.tsx
```diff
@@ -78,7 +78,11 @@ export const useOverlayManager = () => {
 
   const decrementOverlaysOpenCount = useCallback(() => {
     setOverlaysSet((prevSet) => {
-      prevSet.delete(overlayId)
+      if (prevSet.size === 1) {
+        prevSet.clear()
+      } else {
+        prevSet.delete(overlayId)
+      }
```
kdembler, line 85: I'm afraid this may have some undesired consequences that we may not foresee right now, because it changes the mechanics of the manager. Take a look at this code I've added:
https://github.com/Joystream/atlas/blob/6355b6bb24e8b271c369a346ca8e31bc3c6bf62d/packages/atlas/src/components/_navigation/SidenavBase/SidenavBase.tsx#L51-L57

I think if we just added this to the `useEffect`, it would fix the bug?
```tsx
return decrementOverlaysOpenCount
```

rafalpawlow: @kdembler Yes it would do the job but I think my solution is more versatile, we won't have to worry about handling this case in the future.

kdembler (approved):

PR #2578, opened 2022-04-28
Fix circular dependency

eldiegod:


packages/atlas/src/components/_video/ReactionStepper/ReactionStepper.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 
-import { ReactionButton, ReactionSteppperState } from './ReactionButton'
+import { ReactionButton } from './ReactionButton'
+import { ReactionSteppperState } from './ReactionButton/ReactionButton.styles'
```
kdembler: Could this type be re-exported from `./ReactionButton/index.ts` so we don't need to import from `.styles`?

kdembler (approved):

PR #2582, opened 2022-04-28
fix dialog buttons

rafalpawlow:
Dialog buttons were on the left but should have been on the right, this PR fixes this.
@drillprop this is your code so please make sure I didn't destroy your work.


eldiegod:
I'll fix this in my pr, if you can do without better not to tweak it more. 

But also this fix won't work for the `reactionStepper`

PR #2586, opened 2022-04-29
video reactions integration

### LINKED ISSUE #2229: Integrate video reactions
### Context

Handling transaction for reactions. Depends on all components to be in place before doing it.

### Scope
- [ ] Handle transaction execution for leaving reaction on a video page

### Design
[Figma ](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2093%3A135230)

✋ 📖  link to be added perhaps with some help from @kdembler

rafalpawlow:
Closes #2229

### DIFF
--- a/packages/atlas/src/views/viewer/VideoView/VideoView.tsx
+++ b/packages/atlas/src/views/viewer/VideoView/VideoView.tsx
@@ -20,6 +20,7 @@ import { CTA_MAP } from '@/config/cta'
 import { absoluteRoutes } from '@/config/routes'
 import { useCategoryMatch } from '@/hooks/useCategoriesMatch'
 import { useClipboard } from '@/hooks/useClipboard'
+import { useDisplaySignInDialog } from '@/hooks/useDisplaySignInDialog'
 import { useHeadTags } from '@/hooks/useHeadTags'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
 import { useNftTransactions } from '@/hooks/useNftTransactions'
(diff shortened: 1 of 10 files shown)


packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -102,6 +110,46 @@ export const VideoView: React.FC = () => {
...
+
+    handleTransaction({
+      preProcess: () => setVideoReactionProcessing(true),
+      txFactory: async (updateStatus) =>
+        (await joystream.extrinsics).reactToVideo(activeMemberId, video.id, reaction, proxyCallback(updateStatus)),
+      minimized: {
+        signErrorMessage: 'Failed to react to video',
+      },
+      onTxFinalize: async () => {
+        await refetch()
```
drillprop, line 148: you should refetch only inside `onTxSync`, If you call refetch inside `onTxFinalize`, sometimes you don't get correct updated results.

drillprop (changes requested):
I see some issues
1. You cannot take back your reaction, you can only change your reaction from like do dislike or dislike to like. You should be able to remove your reaction by clicking a second time the reaction button.
2. If I cancel a transaction on Polkadot dialog, ReactionStepper is still in a processing state
3. Sometimes refetch doesn't update counters correctly
4. You didn't handle the scenario when a user isn't signed in. When the user is not logged in we should display our sign-in dialog(you should use `useDisplaySignInDialog`)

drillprop (changes requested):
Right now we're seeing the onboarding dialog first, and when we click `got it` we're seeing the sign-in dialog. This is not how it's supposed to work
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2093%3A135230

drillprop (changes requested):
We should show the tooltip when someone is hovering ReactionStepper when `processing` not always when the state is processing

drillprop (approved):
LGTM!

PR #2605, opened 2022-05-04
Comment reactions integration

### LINKED ISSUE #2448: Reaction to comments integration
CO/CV reacts to comments [link](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140555) link

- [ ] Existing Reaction
- [ ] New Reaction
- [ ] Undoing Reaction
- [ ] Multiple Reactions

drillprop:
Fix #2448

~Based on #2563.~ 
~I'll rebase once #2563 is merged~

Flows: 
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140555


packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -87,7 +87,8 @@ export const Comment: React.FC<CommentProps> = ({
   ]
 
   const reactionIsProcessing = reactions?.some(({ state }) => state === 'processing')
-  const allReactionsApplied = reactions && reactions?.length >= Object.values(REACTION_TYPE).length
+  const allReactionsApplied =
+    reactions && reactions.filter((r) => r.count)?.length >= Object.values(REACTION_TYPE).length
```
kdembler: I don't think that `?` is necessary, `.filter` will always return an array

packages/atlas/src/components/_comments/ReactionChip/ReactionChip.tsx
```diff
@@ -1,46 +1,94 @@
...
+import { ReactionsOnboardingPopover } from '@/components/_video/ReactionsOnboardingPopover'
+import { pluralizeNoun } from '@/utils/misc'
 import { formatNumberShort } from '@/utils/number'
 
 import { EmojiContainer, ReactionChipButton } from './ReactionChip.styles'
 
-export type ReactionType = 'amusment' | 'love' | 'laugh' | 'shock' | 'anger'
+export type ReactionId = 1 | 2 | 3 | 4 | 5
+
+export type ReactionType = 'like' | 'love' | 'laugh' | 'shock' | 'anger'
```
kdembler: 1. I see we have two types of identifiers here. Do we use both? If yes, what is each used for and do we need both?
2. I think we need to move this to a "more global" place like `@/config` or something. Best to bring `REACTION_TYPE` along

packages/atlas/src/components/_comments/ReactionChip/ReactionChip.tsx
```diff
@@ -1,46 +1,94 @@
...
+  reactionId: ReactionId
   state?: 'default' | 'disabled' | 'processing' | 'read-only'
-  onReactionClick?: (type: ReactionType) => void
+  onReactionClick?: (type: ReactionId) => void
+  onPopoverHide?: () => void
+  reactionPopoverDismissed?: boolean
+}
+
+export const REACTION_TYPE: Record<ReactionId, string> = {
+  1: '👍 ',
```
kdembler: Why do we have a space after the like?

packages/atlas/src/components/_comments/ReactionChip/ReactionChip.tsx
```diff
@@ -1,46 +1,94 @@
...
+    if (state === 'processing' && !reactionPopoverDismissed) {
+      popoverRef.current?.show()
+    }
+  }, [reactionPopoverDismissed, state])
+
+  const handleReact = (reactionPopoverDismissed: boolean) => {
+    if (reactionPopoverDismissed) {
+      onReactionClick?.(reactionId)
+    }
+  }
```
kdembler: If the value exists when we create the handler, is there a good reason to still pass it as an argument?

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -0,0 +1,79 @@
...
+  const { openSignInDialog } = useDisplaySignInDialog()
+  const [processingCommentReactionId, setProcessingCommentReactionId] = useState<string | null>(null)
+  const [videoReactionProcessing, setVideoReactionProcessing] = useState(false)
+
+  const handleReactToComment = (commentId: string, reactionId: ReactionId) => {
+    if (!joystream || !activeMemberId) {
+      ConsoleLogger.error('no joystream or active member')
+      openSignInDialog({ onConfirm: signIn })
+      return
+    }
```
kdembler: This should never happen if the app functions properly right? If yes, then let's skip opening the sign in dialog and just print a console error

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -0,0 +1,79 @@
...
+    setProcessingCommentReactionId(commentId + `-` + reactionId.toString())
+
+    handleTransaction({
+      preProcess: () => setProcessingCommentReactionId(commentId + `-` + reactionId.toString()),
```
kdembler: This line seems redundant? We do exactly same above

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -0,0 +1,79 @@
...
+    })
+  }
+
+  const handleLikeAndDislike = (videoId: string, reaction: VideoReaction) => {
+    if (!joystream || !activeMemberId) {
+      return
+    }
+
+    handleTransaction({
+      preProcess: () => setVideoReactionProcessing(true),
```
kdembler: Here probably makes sense to do this outside of the `handleTransaction` as well

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -0,0 +1,79 @@
...
+import { ReactionId } from '@/components/_comments/ReactionChip'
+import { VideoReaction } from '@/joystream-lib'
+import { useJoystream } from '@/providers/joystream'
+import { useTransaction } from '@/providers/transactionManager'
+import { useUser } from '@/providers/user'
+import { ConsoleLogger } from '@/utils/logs'
+
+import { useDisplaySignInDialog } from './useDisplaySignInDialog'
+
+export const useReactionTransactions = (refetch: QueryResult['refetch']) => {
```
kdembler: Is there a good reason to provide `refetch` via an argument versus doing it inside this hook, similar to what we do for nft transactions? One issue would be if you wanted to react to both video and comments from a single component, you would be able to only provide a single `refetch`

packages/atlas/src/components/_comments/ReactionChip/ReactionChip.tsx
drillprop: `ReactionId` is what we're getting from QN. This is not very useful, to be honest - I think the query node should return something like an enum, maybe? 
Anyway, I changed my approach a bit and I created an object which has ReactionId as a property and object as a value, which contains the name and the corresponding emoji. I moved this to @/config/reactions.ts
A reaction name is used to add a more accessible name to `ReactionChip` button - you can check the `title` prop. Let me know what you think

kdembler: We specifically didn't want to make it an enum so that apps are free to pick whatever reactions they want to use. Somebody may fork Atlas and use 🤮 as reaction instead and they should be free to do so :P It's our job as app developers to keep the list of reactions we want to use and assign some ids to them. Will take a look at the code again

packages/atlas/src/config/reactions.ts
```diff
@@ -0,0 +1,9 @@
+export type ReactionId = 1 | 2 | 3 | 4 | 5
```
kdembler: Maybe we extract it from the object below to avoid duplication? `type ReactionId = keyof REACTION_TYPE`?

packages/atlas/src/components/_comments/ReactionPopover/ReactionPopover.tsx
```diff
@@ -40,13 +40,14 @@ const getTranslateNumber = (idx: number) => {
 export const ReactionPopover: React.FC<ReactionPopoverProps> = ({ onReactionClick, disabled }) => {
   const [isOpen, setIsOpen] = useState(false)
   const smMatch = useMediaMatch('sm')
+  const reactionPopoverDismissed = usePersonalDataStore((state) => state.reactionPopoverDismissed)
   const reactions = Object.entries(REACTION_TYPE).map(([key, value]) => ({
-    name: key as ReactionType,
-    value,
+    reactionId: Number(key) as ReactionId,
```
kdembler, line 45: Why does `reactionId` need to be a number?

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -0,0 +1,157 @@
...
+    async (videoId: string, commentBody: string, parentCommentId?: string) => {
+      if (!joystream || !activeMemberId) {
+        ConsoleLogger.error('no joystream or active member')
+        return
+      }
+
+      return handleTransaction({
+        preProcess: () => {
+          setCommentInputProcessing(true)
+        },
```
kdembler: Let's do this outside of `handleTransaction`, `preProcess` should be used for async actions like computing file hash etc

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -0,0 +1,157 @@
...
+            commentBody,
+            parentCommentId || null,
+            proxyCallback(updateStatus)
+          ),
+        onTxSync: async ({ block }) => {
+          const refetchResult = await refetchComments()
+          const newCommentsQueryResult = refetchResult[0] as GetCommentsQueryHookResult
+          const newCommentId = newCommentsQueryResult.data?.comments.find(
+            (comment) => comment.commentcreatedeventcomment?.[0].inBlock === block
+          )?.id
```
kdembler: What if there's multiple comments added in the same block? I would suggest to skip this part in this PR as this will be very tricky to get right. Once we have https://github.com/Joystream/atlas/issues/2629, this should be much easier

kdembler: Or I guess we can leave it but add a TODO that this needs to be changed

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -0,0 +1,157 @@
...
+          setHighlightedComment(newCommentId || null)
```
kdembler: Should this really be handled by this hook that's concerned with logic of blockchain interactions? I think this fits more into view/component responsibilities, especially that we may want to highlight comments in other cases as well in the future

packages/atlas/src/components/_comments/ReactionPopover/ReactionPopover.tsx
drillprop: The QN also returns numbers in the `reactionId` field. `Object.entries` convert keys into a string, that's why I'm changing this to the number.  I want to make sure that type is the same as `ReactionId`.

kdembler: That's a bit weird since it accepts string as an input

kdembler: Nvm, it's a number as well

kdembler (approved):

PR #2609, opened 2022-05-05
Not found page follow up

eldiegod:


eldiegod:
@TCzechowski could you check the not found page and let me know if you spot any issues?

rafalpawlow (changes requested):
One more thing from my side. Let's use pattern from figma api

TCzechowski (approved):
The content is now centered both vertically and horizontally, and I didn't see any other issues previously, so I think we are good to go! :)

kdembler:
Agreed with @rafalpawlow here, let's use illustrations synced via Figma API

eldiegod:
> Agreed with @rafalpawlow here, let's use illustrations synced via Figma API

@rafalpawlow how do I do that again?

rafalpawlow:
@eldiegod just run `yarn svgs:illustrations`

rafalpawlow (approved):

PR #2632, opened 2022-05-06
Add comment integration

### LINKED ISSUE #2439: Add comment integration
Add comment to video [link](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140569)

drillprop:
Fix #2439

### DIFF
--- a/packages/atlas/src/components/_comments/Comment/Comment.styles.ts
+++ b/packages/atlas/src/components/_comments/Comment/Comment.styles.ts
@@ -68,6 +68,7 @@ export const StyledSvgActionTrash = styled(SvgActionTrash)`
 
 export const CommentBody = styled(Text)`
   display: flex;
+  white-space: pre-line;
   align-items: center;
 `
 
--- a/packages/atlas/src/api/queries/__generated__/comments.generated.tsx
+++ b/packages/atlas/src/api/queries/__generated__/comments.generated.tsx
@@ -54,6 +54,7 @@ export type GetCommentsQuery = {
       }
     }
     reactions: Array<{ __typename?: 'CommentReaction'; id: string; createdAt: Date; reactionId: number }>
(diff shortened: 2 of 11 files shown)


kdembler:
When not logged in :P

Screenshot 2022-05-09 at 17 16 05

drillprop:
> When not logged in :P
> 
> Screenshot 2022-05-09 at 17 16 05

fixed

packages/atlas/src/components/_comments/Comment/Comment.styles.ts
```diff
@@ -68,6 +68,7 @@ export const StyledSvgActionTrash = styled(SvgActionTrash)`
 
 export const CommentBody = styled(Text)`
   display: flex;
+  white-space: pre-line;
   align-items: center;
```
eldiegod, line 72: nice

eldiegod (changes requested):
Works as expected but there is an issue with focus/blur in CommentInput; I would guess it's either because of the change in the onFocus functions or by the custom placeholder, I remember I thought it was not worth it to have a custom placeholder and think it was because of this.

To reproduce the issue: click on the CommentInput -> don't type -> click multiple times in the container besides the buttons -> it will case the input to close and open flicker and blur/focus flciker

drillprop:
> I remember I thought it was not worth it to have a custom placeholder and think it was because of this.

Unfortunately, the design requires a custom placeholder:
image
It's impossible to change the styling of one word(in this case "bedeho") using only `::placeholder` selector. I had to add something custom. 

> To reproduce the issue: click on the CommentInput -> don't type -> click multiple times in the container besides the buttons -> it will case the input to close and open flicker and blur/focus flciker

I think I fixed this. Let me know what you think

eldiegod (approved):
yeah now it's good

PR #2633, opened 2022-05-06
update ComboBox

### LINKED ISSUE #2392: Create global combo box component
**AC:**
- Go through the implementation, make sure we match the designs

---

To support a few cases of a similar use a unified global combo box component needs to be created. This component will be used:
- In video workspace for adding subtitles to a video
- In NFT workspace for adding members to a whitelist
- In video workspace for adding video to a playlist

Todo:
- Create a flexible component to work in all 3 scenarios & in future ones
- Change the currently existing solution in final designs & user stories to this new component
- Document the whole component and its do's and dont's

rafalpawlow:
Closes #2392 

**Design**
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=2780%3A134836

### DIFF
--- a/packages/atlas/src/components/_inputs/ComboBox/ComboBox.styles.ts
+++ b/packages/atlas/src/components/_inputs/ComboBox/ComboBox.styles.ts
@@ -1,6 +1,8 @@
 import styled from '@emotion/styled'
 
-import { cVar, zIndex } from '@/styles'
+import { TextField } from '@/components/_inputs/TextField'
+import { Spinner } from '@/components/_loaders/Spinner'
+import { cVar, sizes, zIndex } from '@/styles'
 
 export const ComboBoxWrapper = styled.div`
   position: relative;
@@ -8,7 +10,7 @@ export const ComboBoxWrapper = styled.div`
 `
 
 export const ListWrapper = styled.ul<{ topPosition?: number }>`
-  max-height: 300px;
+  max-height: 188px;
   overflow-y: auto;
   background: ${cVar('colorBackgroundStrong')};
   padding: 0;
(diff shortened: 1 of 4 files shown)


packages/atlas/src/components/_inputs/ComboBox/ComboBox.tsx
```diff
@@ -84,11 +86,11 @@ export const ComboBox = <T extends unknown>(props: ComboBoxProps<T>) => {
...
     <ComboBoxWrapper ref={comboBoxWrapperRef}>
       <div {...getComboboxProps()}>
-        <TextField
+        <StyledTextField
           {...textFieldProps}
-          error={error}
+          error={error || !!noItemsFound}
           {...getInputProps({ ref: textFieldRef })}
-          nodeEnd={loading && <SvgActionLoader />}
+          nodeEnd={loading && <StyledSpinner size="small" />}
```
drillprop: We shouldn't use Spinner here - I think this is some legacy component, and we should actually remove this. We have a component called Loader and we should use variant `small` here.

packages/atlas/src/components/_inputs/ComboBox/ComboBox.styles.ts
```diff
@@ -19,3 +21,25 @@ export const ListWrapper = styled.ul<{ topPosition?: number }>`
   z-index: ${zIndex.globalOverlay};
   margin: 0;
 `
+
+export const StyledSpinner = styled(Spinner)`
```
drillprop: I think this won't be needed if we apply the above.

drillprop (changes requested):
Just found that the TextField has the wrong height. It has 50px instead of 48px. 
This 2 px difference comes from 1px border applied on the input. We should use box-shadow instead. Can we fix this in this PR?

drillprop (approved):

PR #2634, opened 2022-05-06
fix closing create member modal without accepting terms

### LINKED ISSUE #2608: `ControllerAccountRequired` error when taking any action after closing create member modal
**Describe the bug**

It's impossible to take any action in the app when a user closes the create member modal after the `connect account` step`(`?loginStep=2`). 

**To Reproduce**

1. Make sure that you're signed in
2. Click `Add new member` on `Member dropdown`
3. Select some account, click the `Connect account` button
4. Close Dialog on the next step
5. Try to take any action, for instance: try to upload video, bid on nft, mint nft
6. You'll notice that when you take any action you'll see that the Polkadot extension popup points to the wrong account(the account you chose during the `connect account` step)
7. You'll get an error `ControllerAccountRequired`

From the user's perspective, this could be very hard to fix. He would need to clear localstorage and sign in again to be able do any user actions. 

**Expected behavior**

All user actions should be possible without an error when the user closes the `create member` modal

- OS:  macOS Monterey
- Browser: Chrome

rafalpawlow:
Closes #2608

### DIFF
--- a/packages/atlas/src/components/_auth/SignInSteps/AccountStep.tsx
+++ b/packages/atlas/src/components/_auth/SignInSteps/AccountStep.tsx
@@ -1,4 +1,4 @@
-import React, { FormEvent, useState } from 'react'
+import React, { FormEvent } from 'react'
 import { useNavigate } from 'react-router'
 import { CSSTransition, SwitchTransition } from 'react-transition-group'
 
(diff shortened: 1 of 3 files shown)


kdembler:
Requesting review from @drillprop instead, as he's been touching that piece of code more recently and also he reported the bug

drillprop (changes requested):
This is not working correctly. If I don't have an account I can't create a membership because the create member dialog never shows up.

drillprop (approved):

kdembler (approved):

Issue #2463, opened 2022-04-11
Review and update Button component


rafalpawlow:
**Scope:**
- Include changes from #2123
- Make sure `textOnly` and `iconOnly` variants are needed, probably we can get rid of them and "compute" those values
- Rethink component props and API, see what variants make sense. Best to discuss with the dev team before doing the work
- Figure out what's the situation with `IconButton`, do we need it or use regular button with `iconOnly`
- Make sure all the variants look the same as in Figma
- Clean up storybook stories

---

### Description

We have mismatches between design and implementation, especially in the teritary variant. Please make sure storybook works well for all variants, especially for `textOnly` variant.

### Figma link

https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=216%3A6243

### How urgent this is?

Not a blocker, but still worth considering

eldiegod:
> Make sure textOnly and iconOnly variants are needed, probably we can get rid of them and "compute" those values

what does "compute" mean here?

for the record we should remove `IconButton` I marked it as deprecated as `IconButton` is a separate implementation from our buttons and makes it so we have to maintain 2 components everytime we add features to button

PR #2636, opened 2022-05-09
Show more/show less button on comment

### LINKED ISSUE #2564: Introduce character limits on comments
We need to update the designs so that:
1. There's a char limit for displaying comments and the "show more" button,
2. There's a chat limit for posting a new comment.

drillprop:
Fix #2564 
Designs: 
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=3677%3A265878

Be aware that the look of the show more button differs from the design. This won't be an issue once https://github.com/Joystream/atlas/issues/2463 is done

I didn't handle [exceeding characters limit story](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/?node-id=3677%3A254185). I'll create a separate issue just for that.

### DIFF
--- a/packages/atlas/src/components/_comments/CommentBody/CommentBody.tsx
+++ b/packages/atlas/src/components/_comments/CommentBody/CommentBody.tsx
@@ -1,0 +1,47 @@
+import React, { useEffect, useRef, useState } from 'react'
+
+import { SvgActionChevronB, SvgActionChevronT } from '@/components/_icons'
+
+import { ExpandButton, StyledCommentText } from './CommentBody.styles'
+
+export const CommentBody = ({ children }: { children?: string }) => {
+  const [commentExpanded, setCommentExpanded] = useState(false)
+  const [isTruncated, setIsTruncated] = useState(false)
+
+  const firstRender = useRef<boolean>(true)
+  const commentBodyRef = useRef<HTMLParagraphElement>(null)
+
+  useEffect(() => {
+    if (!firstRender.current || !commentBodyRef.current) {
+      return
+    }
+    setIsTruncated(commentBodyRef.current?.offsetHeight < commentBodyRef.current?.scrollHeight)
+    firstRender.current = false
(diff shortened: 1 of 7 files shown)


packages/atlas/src/components/_comments/CommentBody/CommentBody.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useEffect, useRef, useState } from 'react'
+
+import { SvgActionChevronB, SvgActionChevronT } from '@/components/_icons'
+
+import { ExpandButton, StyledCommentText } from './CommentBody.styles'
+
+export const CommentBody = ({ children }: { children?: string }) => {
```
rafalpawlow, line 7: Can't we just add proper type here?
```suggestion
export const CommentBody: React.FC = ({ children }) => {
```

rafalpawlow (changes requested):
1. there is wrong spacing between author and deleted comment. Can we fix this as part of this PR? We have `8px` but in figma it's `12px`
Zrzut ekranu 2022-05-10 o 13 37 37

2. I've noticed that Show more button has wrong color but I assume it will be fixed here https://github.com/Joystream/atlas/pull/2637 .

packages/atlas/src/components/_comments/CommentBody/CommentBody.tsx
drillprop: I want to be able to pass **only** string here, not React.ReactNode, that's why I wrote this.

drillprop:
> 1. there is wrong spacing between author and deleted comment. Can we fix this as part of this PR? We have `8px` but in figma it's `12px`

Fixed

> 2. I've noticed that Show more button has wrong color but I assume it will be fixed here https://github.com/Joystream/atlas/pull/2637 . 

correct

rafalpawlow (approved):

PR #2637, opened 2022-05-09
Button update

eldiegod:
Fixes: #2463

- IconButton component got removed; use the Button component with icon prop and no children.
- iconOnly prop got removed; for a while, this prop did nothing as we calculate this when an icon is passed, and no children are passed.
- I removed the ButtonBase component as it served no purpose.
- removed unused or duplicated styles.
- updated the text only variants

I spent hours trying to find a way around this just with types. Still, due to the text only variants being different from the default variants, in an attempt to properly handle this, I renamed the `textOnly` to `_textOnly` to be used internally only.
I also created the TextButton component, the same as our Button component but with correct typings for text variants.


packages/atlas/src/components/Banner/Banner.styles.ts
```diff
@@ -3,7 +3,7 @@ import styled from '@emotion/styled'
 import { Text } from '@/components/Text'
 import { cVar, sizes } from '@/styles'
 
-import { IconButton } from '../_buttons/IconButton'
+import { Button } from '../_buttons/Button'
```
kdembler: I think we should use absolute import path here

packages/atlas/src/components/Carousel/Carousel.styles.ts
```diff
@@ -1,15 +1,16 @@
 import styled from '@emotion/styled'
 
-import { IconButton } from '@/components/_buttons/IconButton'
 import { cVar, media, oldColors, sizes, transitions, zIndex } from '@/styles'
 
+import { Button } from '../_buttons/Button'
```
kdembler: I think we should use absolute import path here

packages/atlas/src/components/DrawerHeader/DrawerHeader.tsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react'
 
 import { Pill } from '@/components/Pill'
-import { IconButton } from '@/components/_buttons/IconButton'
 import { SvgActionClose } from '@/components/_icons'
 
 import { Tab, TabContainer, TabTitle, Tabbar } from './DrawerHeader.styles'
 
+import { Button } from '../_buttons/Button'
```
kdembler: I think we should use absolute import path here

packages/atlas/src/components/ProgressDrawer/ProgressDrawer.tsx
```diff
@@ -20,6 +19,8 @@ import {
   StyledSvgGlyphChevronDown,
 } from './ProgressDrawer.styles'
 
+import { Button } from '../_buttons/Button'
```
kdembler: I think we should use absolute import path here

packages/atlas/src/components/Searchbar/Searchbar.styles.ts
```diff
@@ -2,10 +2,11 @@ import isPropValid from '@emotion/is-prop-valid'
 import styled from '@emotion/styled'
 
 import { Text } from '@/components/Text'
-import { IconButton } from '@/components/_buttons/IconButton'
 import { SvgControlsSearchAlt } from '@/components/_icons'
 import { cVar, media, oldColors, sizes, square } from '@/styles'
 
+import { Button } from '../_buttons/Button'
```
kdembler: I think we should use absolute import path here

packages/atlas/src/components/Searchbar/Searchbar.tsx
```diff
@@ -23,6 +22,8 @@ import {
   StyledSvgOutlineSearch,
 } from './Searchbar.styles'
 
+import { Button } from '../_buttons/Button'
```
kdembler: I think we should use absolute import path here

packages/atlas/src/components/Step/Step.tsx
```diff
@@ -16,7 +16,7 @@ import {
 } from './Step.styles'
 
 import { Text } from '../Text'
-import { IconButton } from '../_buttons/IconButton'
+import { Button } from '../_buttons/Button'
```
kdembler: I think we should use absolute import path here

packages/atlas/src/components/_buttons/Button/Button.styles.ts
```diff
@@ -1,104 +1,421 @@
...
 import { SerializedStyles, css } from '@emotion/react'
 import styled from '@emotion/styled'
 
 import { smallBadgeStyles } from '@/components/Badge'
-import { Text } from '@/components/Text'
-import { sizes } from '@/styles'
+import { cVar, sizes } from '@/styles'
 
-import { ButtonBase, ButtonSize, ButtonVariant } from '../ButtonBase'
+const ButtonSizeObj = { large: 'large', medium: 'medium', small: 'small' }
```
kdembler: I don't think this object is needed at all?

packages/atlas/src/components/_buttons/Button/Button.styles.ts
```diff
@@ -1,104 +1,421 @@
...
+          opacity: ${iconOnly ? 0.25 : 0.5};
+        }
       `
-    case 'medium':
+    case 'tertiary':
       return css`
-        padding-left: ${sizes(4)};
-        padding-right: ${sizes(4)};
+        color: ${cVar('colorCoreNeutral50')};
+        ${iconOnly && `border-radius: 50%`};
```
kdembler: We agreed to not use this type of "style in a string" as it bypasses stylelint: https://github.com/Joystream/atlas/blob/master/docs/styleguide.md#conditional-styles

packages/atlas/src/components/_buttons/Button/Button.styles.ts
```diff
@@ -1,104 +1,421 @@
...
+  }
 `
 
-const getIconOnlyPadding = (size?: ButtonSize, iconOnly?: boolean) => {
-  if (!iconOnly) return null
-  switch (size) {
-    case 'large':
-      return css`
-        padding: ${sizes(1)};
+const ButtonSizeStyles = ({ size, iconOnly, textOnly }: ButtonBaseStyleProps) => {
```
kdembler: This shouldn't be Capitalized since it's not a component

packages/atlas/src/components/_buttons/Button/Button.styles.ts
```diff
@@ -1,104 +1,421 @@
...
-  size: ButtonSize
-}
+  margin-right: ${({ iconPlacement, iconOnly }) => (iconPlacement === 'left' && !iconOnly ? sizes(2) : 0)};
+  margin-left: ${({ iconPlacement, iconOnly }) => (iconPlacement === 'right' && !iconOnly ? sizes(2) : 0)};
+`
 
-type ButtonBaseProps = {
-  'data-badge'?: string | number | boolean
-} & ButtonSizeProps
+const ButtonVariantStyles = ({ variant, iconOnly }: ButtonBaseStyleProps): SerializedStyles => {
```
kdembler: Same here, shouldn't be capitalized

packages/atlas/src/components/_buttons/HamburgerButton/HamburgerButton.styles.ts
```diff
@@ -1,14 +1,15 @@
 // based on https://github.com/jonsuh/hamburgers licensed under MIT
 import styled from '@emotion/styled'
 
-import { IconButton } from '@/components/_buttons/IconButton'
 import { oldColors, square } from '@/styles'
 
+import { Button } from '../Button'
```
kdembler: Should use absolute path

packages/atlas/src/components/_inputs/FileSelect/FileSelect.styles.ts
```diff
@@ -108,7 +108,7 @@ export const Thumbnail = styled(animated.img, { shouldForwardProp: (prop) => pro
   transition: opacity 400ms ease-out;
 `
 
-export const DismissButton = styled(IconButton)`
+export const DismissButton = styled(Button)`
```
kdembler: This component seems to not be used anywhere, let's remove it

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
```diff
@@ -53,7 +53,7 @@ export const OwnerLabel = styled(Text)`
   grid-area: owner-label;
 `
 
-export const OwnerHandle = styled(Button)`
+export const OwnerHandle = styled(TextButton)`
```
kdembler: Should this be a button actually versus just a regular link?

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -355,12 +355,8 @@ export const NftWidget: React.FC<NftWidgetProps> = ({
...
                     <>
                       {convertToUSD(nftStatus.topBidAmount)} from{' '}
-                      <OwnerHandle
-                        to={absoluteRoutes.viewer.member(nftStatus.topBidderHandle)}
-                        variant="secondary"
-                        textOnly
-                      >
-                        <Text variant="t100">{nftStatus.isUserTopBidder ? 'you' : nftStatus.topBidderHandle}</Text>
+                      <OwnerHandle to={absoluteRoutes.viewer.member(nftStatus.topBidderHandle)} variant="secondary">
+                        <Text variant="t100">{nftStatus.topBidderHandle}</Text>
```
kdembler: You removed the `you` functionality here

packages/atlas/src/components/_overlays/Dialog/Dialog.tsx
```diff
@@ -92,12 +84,12 @@ export const Dialog: React.FC<DialogProps> = ({
...
           <FooterButtonsContainer additionalActionsNodeMobilePosition={additionalActionsNodeMobilePosition}>
             {secondaryButton && (
-              <Button variant={secondaryButton.variant || 'secondary'} {...buttonProps} {...secondaryButton}>
+              <Button variant={secondaryButton.variant || 'secondary'} size={buttonSize} {...secondaryButton}>
                 {secondaryButton.text}
               </Button>
             )}
             {primaryButton && (
-              <StyledPrimaryButton variant={primaryButton.variant || 'primary'} {...buttonProps} {...primaryButton}>
+              <StyledPrimaryButton variant={primaryButton.variant || 'primary'} {...primaryButton} size={buttonSize}>
```
kdembler: Let's destruct `primaryButton` at the end so that consumer is able to override if needed and so we're consistent with the secondary button

packages/atlas/src/utils/types.d.ts
```diff
@@ -0,0 +1,5 @@
+export {}
+
+declare global {
+  type Subset<K, T extends K> = T
+}
```
kdembler: This should go into `src/types`

kdembler (changes requested):
Good job in general, left some mostly cosmetic comments. From the visual standpoint, I've noticed that all the text buttons have the same height (24px) regardless of the size:
Screenshot 2022-05-10 at 14 50 09

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
eldiegod: ah fak

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
eldiegod: hmm i'm not sure rn but I think it was a button in design

eldiegod:
> ons have the same height (24px) regardless of the size:

@kdembler yeah, it looks like that but it's because of the flex parent component.

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
kdembler: I don't think it is, at least in some cases, just a text here:
Screenshot 2022-05-10 at 22 15 10

toiletgranny: I can confirm it is just a text in here. If it were a button, it would be marked as such in the layers panel.

toiletgranny (changes requested):
Overall, it's very nice! Here are the things I found:
1. When hovering over any button now, I can feel severe stuttering in the transition animation, no matter which CSS property is being animated. At the same time, in the "Actions" tab on Storybook, I can see quite a lot of onMouseMove events being recorded every time my cursor moves, even by 1px. Could it be connected?
2. The height of the last row of Text Only buttons (Tertiary) is off by 1px in all three sizes, eg. instead of `24px`, it's `23px`. I can see you apply a negative margin vertically on both sides (`margin-top: -0.5px; margin-bottom: -0.5px;`), and I assume it's to accommodate for the 1px bottom border in all other Text Only buttons, which brings me to my next point, which is that...
3. In my opinion, the current specification for CSS borders is just wrong. As far as I know, the `box-sizing: border-box` property would only do its job if you specifically set the element's width and height, which rarely ever happens in the context of UI components such as buttons. Therefore, in Figma, I used absolute positioning for rendering bottom borders. Perhaps you'll find this approach (or something else, eg. `:before`/ `:after` pseudo-elements) more suitable in this case? I'm leaving this as a loose suggestion.

drillprop:
> When hovering over any button now, I can feel severe stuttering in the transition animation, no matter which CSS property is being animated. At the same time, in the "Actions" tab on Storybook, I can see quite a lot of onMouseMove events being recorded every time my cursor moves, even by 1px. Could it be connected?

@eldiegod I used `onMouseMove` and `onMouseOut` in the ChannelCard(check the FollowButton), but I made a mistake with selecting the right handler. We should use `onMouseEnter` and `onMouseLeave` here instead, which is much more performant(it's triggered basically only once when you hover and unhover on the button). Could you replace it and remove `onMouseMove` and `onMouseOut` props?

eldiegod:
> 3\. context of UI components such as buttons. Therefore, in Figma, I used absolute positioning for rendering bottom borders. Perhaps you'll find this approach (or something else, eg. `:before`/ `:after` pseudo-elements) more suitable in this case? I'm leaving this as a loose suggestion.

@toiletgranny I think the issues should be solved now, point 2 was a really easy fix so I didn't wanna go about changing much of the border now but I agree with you nowadays I build this type of thing using :before/ :after

eldiegod:
> > When hovering over any button now, I can feel severe stuttering in the transition animation, no matter which CSS property is being animated. At the same time, in the "Actions" tab on Storybook, I can see quite a lot of onMouseMove events being recorded every time my cursor moves, even by 1px. Could it be connected?
> 
> @eldiegod I used `onMouseMove` and `onMouseOut` in the ChannelCard(check the FollowButton), but I made a mistake with selecting the right handler. We should use `onMouseEnter` and `onMouseLeave` here instead, which is much more performant(it's triggered basically only once when you hover and unhover on the button). Could you replace it and remove `onMouseMove` and `onMouseOut` props?

@drillprop yep fixed 👍

packages/atlas/src/components/_nft/NftWidget/NftWidget.styles.ts
eldiegod: replaced it with a normal Link + Text

kdembler (approved):
The hamburger button is no longer rounded:
Screenshot 2022-05-12 at 12 42 00

Haven't found any other issues though, LGTM once 👆 if fixed, let's get an approval from Adam as well

toiletgranny (approved):

PR #2638, opened 2022-05-09
auction datepicker - preselect closest time

### LINKED ISSUE #2590: Have the datetime picker preselecting the closest time option by default
- [ ] Set default time to now?

> Ideally, the datetime picker also selects the closest time option by default, based on the current time. Without it, I'm forced to select the time manually, and if I'm interested in picking an evening time, I need to scroll through a very long list of 30 minutes time intervals, which begin at 00:00.

CleanShot 2022-04-29 at 16 03 33

rafalpawlow:
Closes #2590

### DIFF
--- a/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
+++ b/packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled'
-import { format } from 'date-fns'
+import { format, roundToNearestMinutes } from 'date-fns'
 import React, { useEffect, useMemo, useRef, useState } from 'react'
 import DatePicker from 'react-datepicker'
 import 'react-datepicker/dist/react-datepicker.css'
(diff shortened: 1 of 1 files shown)


packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
```diff
@@ -135,25 +136,30 @@ export const AuctionDatePicker: React.FC<AuctionDatePickerProps> = ({
...
-          openToDate={minDate ?? undefined}
-          minDate={minDate || new Date(msTimestamp)}
-          maxDate={maxDate}
-          showDisabledMonthNavigation
-          showTimeSelect
-        />
+        {open && (
+          <DatePicker
+            inline
+            selected={startDate || roundToNearestMinutes(minDate || new Date(msTimestamp), { nearestTo: 30 })}
```
eldiegod: Maybe put this on a variable and lets add a comment to explain what is going on in here

eldiegod (approved):
Works good!

packages/atlas/src/components/_inputs/AuctionDatePicker/AuctionDatePicker.tsx
kdembler: Agreed on the comment above. Also, I'm not fully sure how the code works, if the current time is `16:10`, will it round it to `16:00`? That would be in the past

kdembler (approved):

Issue #2621, opened 2022-05-05
Auctions that start in the future accept bids


dmtrjsg:
Did not match expected result:
Even though the auction hasn't started yet, the "Place bid" button was displayed in the NFT widget, allowing me to open up the NFT purchase view. This was true in all of my accounts, even after refreshing the page.

PR #2645, opened 2022-05-09
Fix nft widget bid button showing in future auctions

eldiegod:
fixes: #2621

Testable video id: 9


drillprop (commented):
Looks good, but I think we should merge this into `dev`, though.

eldiegod:
> Looks good, but I think we should merge this into `dev`, though.

that's probably right

kdembler (approved):

Issue #2602, opened 2022-05-04
In channels sections, we should show only channels which have at least 4 videos.


drillprop:
This was originally raised by Bedeho. 

Right now channels section often takes too much space and most of the queried channels had just one video. This results in a bad user experience:
![image](https://user-images.githubusercontent.com/51168865/166666506-38c0ec70-b5b1-40da-ad72-0f0be42e356d.png)

 We should filter them out using the `   activeVideosCounter_gt` parameter in our queries and show only channels which have at least 4 videos. This applies to sections in `New & noteworthy`, `Homepage`, and `Channels`

Issue #1507, opened 2021-10-20
Poor new content visibility 


bedeho:
#### Objective

Introduce changes to New section of the App called "New and noteworthy"

- [x] Add new content section "Recently uploaded" to the top of "New" Top menu category
- [x] By Default show two rows of videos (up to 12) sorted desc by date added
- [x] Add Show more button loading two more rows of videos
> Upon clicking “Show more videos”, a new set of videos should be loaded, with the number matching the currently visible amount of videos. For example: if there are 8 videos displayed by default (2 rows, 4 videos each), clicking the “Show more videos” button should load another 2 rows of videos.

**Out of scope**
- [ ] Introduction of Playlists
- [ ] Adding tabs for different content types (Playlists, NFTs)

#### Designs

> Link to Figma 

#### Context
We are continuously seeing new content, lots of it every day, but its not really surfaced in any effective way on Atlas. The part of the app that comes closest is the `New` screen, but it basically has looked like this for the past 3 months

![Screenshot 2021-10-20 at 14 04 14](https://user-images.githubusercontent.com/437292/138089182-2c56a9ad-c923-45f1-8222-35f6f02d6b63.png)

Part of the problem here is certainly that `Videos worth watching` are not being updated by those responsible for featuring, but I think even if they really actively managed it, there probably should be much more prominence given to just exposing the last N uploads to the system. I think a super easy fix which could enhance what we have now would be to just add a section for the last N uploads, which could have a `Show more videos` type button if the user wants to expose more. The flow of new videos is so high that even if someone uploaded some really low quality stuff with bad metadat or previews, they would quickly disappear. Right now I have a really hard time understanding who is creating new content and channels, and this seems to solve that easier than any other intervention I can think of.

toiletgranny:
@bedeho, @dmtrjsg, @kdembler I added a "Recently uploaded" section featuring all videos uploaded to Atlas, sorted by their upload date, newest to oldest. A few things worth noting:
- My suggestion is to add this section at the top of the page, above the current "Videos worth watching" and "Promising new channels" sections. The reasoning behind this is: having all other Atlas pages being primarily focused on promoting high-quality, hand-curated content, this page, for a change, is focused on promoting the new content. By moving the "Recently uploaded" section to the top, we increase the chances of welcoming users with new & fresh content every time they land over there, thus creating an image of Atlas being a platform that's very much alive and which has a very active community (it's true!).
[...]

PR #2646, opened 2022-05-09
Show channels with >= 4 videos on channel sections

eldiegod:
also fixes hooks rule broken on `useChannelsListData`

fixes: #2602, #1507

### DIFF
--- a/packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
+++ b/packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
@@ -139,7 +139,7 @@ const useChannelsListData = (queryType: ChannelsQueryType, selectedLanguage: str
     {
       limit: 15,
       orderBy: ChannelOrderByInput.CreatedAtAsc,
-      where: { ...commonWhere.where, language: { iso_contains: selectedLanguage } },
+      where: { activeVideosCounter_gt: 1, language: { iso_contains: selectedLanguage } },
     },
     { ...commonOpts, skip: queryType !== 'regular' }
   )
--- a/packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
+++ b/packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
@@ -125,26 +125,32 @@ const useChannelsListData = (queryType: ChannelsQueryType, selectedLanguage: str
   const commonOpts: QueryHookOptions = {
     onError: (error) => SentryLogger.error('Failed to fetch channels', 'ExpandableChannelsList', error),
   }
+  const commonWhere = {
+    where: {
+      activeVideosCounter_gt: 4,
+    },
+  }
+
+  const discover = useDiscoverChannels(commonWhere, { ...commonOpts, skip: queryType !== 'discover' })
+  const popular = usePopularChannels(commonWhere, { ...commonOpts, skip: queryType !== 'popular' })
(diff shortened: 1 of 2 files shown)


packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
```diff
@@ -125,26 +125,32 @@ const useChannelsListData = (queryType: ChannelsQueryType, selectedLanguage: str
...
   }
+  const commonWhere = {
+    where: {
+      activeVideosCounter_gt: 4,
+    },
+  }
+
+  const discover = useDiscoverChannels(commonWhere, { ...commonOpts, skip: queryType !== 'discover' })
+  const popular = usePopularChannels(commonWhere, { ...commonOpts, skip: queryType !== 'popular' })
+  const promising = usePromisingChannels(commonWhere, { ...commonOpts, skip: queryType !== 'promising' })
```
drillprop, line 136: ❤️  nice one!

packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
```diff
@@ -125,26 +125,32 @@ const useChannelsListData = (queryType: ChannelsQueryType, selectedLanguage: str
...
+  // regular channels query needs explicit limit and sorting as it's not defined by Orion
+  const regular = useChannels(
+    {
+      limit: 15,
+      orderBy: ChannelOrderByInput.CreatedAtAsc,
+      where: { ...commonWhere.where, language: { iso_contains: selectedLanguage } },
+    },
+    { ...commonOpts, skip: queryType !== 'regular' }
+  )
```
drillprop, line 145: Maybe I wasn't completely clear on the issue, but I don't think we should apply `activeVideosCounter_gt` as default in regular channels query. We're using this query on `Channels in your language` and it could result in the `No channels found` screen on pretty much most of the languages. Maybe we could omit that, or just set this to just `1`?  @kdembler wdyt?

eldiegod: yeah I think since regular query is only used there is safe to make it a `1` we could have it as per case with a prop later if we need to

packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
kdembler: :+1:

packages/atlas/src/components/_channel/ExpandableChannelsList/ExpandableChannelsList.tsx
kdembler: Yeah I think `1` works as a default, we should just override to `4` for the discover channels section. Although we need to test it out on prod and see if 4 actually works well, we may need to adjust

drillprop: I checked these queries(promising/discover/popular) on the graphql playground, and we're getting a lot of results. 

There is another smaller issue, though - results from these queries(promising and discover) are pretty similar, but I don't know what we actually do about it. It's basically expected that the most followed channels will have the most views.

drillprop (approved):

kdembler:
Will wait with merging till the release (tomorrow) to be able to test on prod first

eldiegod:
Added the fix to https://github.com/Joystream/atlas/issues/1507 as well since it's related and a oneliner

PR #2654, opened 2022-05-11
Set correct avatar on comment input

drillprop:
This is just a minor fix. We were missing the member avatar on comment input.

image

### DIFF
--- a/packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
+++ b/packages/atlas/src/components/_comments/CommentInput/CommentInput.styles.ts
@@ -26,7 +26,7 @@ export const CustomPlaceholder = styled(Text)`
   opacity: 0;
   top: 0;
   color: ${cVar('colorTextMuted')};
-  user-select: none;
+  pointer-events: none;
   transition: all ${cVar('animationTransitionMedium')};
 `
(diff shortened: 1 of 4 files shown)


eldiegod (changes requested):
Looks like when the user is logged out the avatar is loading indefinitely.

Also for the selecting issue: it's much better, now we can select from the start of a sentence but we still can't start selecting from the middle of a sentence or double click on a word to select it.

 not part of the scope of this PR anyways but pointing it out

kdembler:
@drillprop Please make sure to create a followup issue if you decide to not include this part in this PR:
> Also for the selecting issue: it's much better, now we can select from the start of a sentence but we still can't start selecting from the middle of a sentence or double click on a word to select it.

drillprop:
> Looks like when the user is logged out the avatar is loading indefinitely.

fixed

> Also for the selecting issue: it's much better, now we can select from the start of a sentence but we still can't start selecting from the middle of a sentence or double click on a word to select it.

I can't reproduce this. Works for me on firefox and chrome. Can someone confirm this? @kdembler how does it work for you?

kdembler:
Yep, works for me as expected as well

eldiegod:
@drillprop can confirm it works perfect now

eldiegod (approved):

PR #2661, opened 2022-05-11
hide unlisted nfts

### LINKED ISSUE #2615: Hide unlisted NFTs from non-owners on Member Page and Channel Page
ER:

- [ ] Hide unlisted NFTs from member and channel views
- [ ] Only show Unlisted NFTs to owners of that membership/channel

AR:
Currently shows to everyone

rafalpawlow:
Closes #2615

### DIFF
--- a/packages/atlas/src/views/viewer/ChannelView/ChannelViewTabs/ChannelNfts.tsx
+++ b/packages/atlas/src/views/viewer/ChannelView/ChannelViewTabs/ChannelNfts.tsx
@@ -31,8 +31,10 @@ export const ChannelNfts: React.FC<ChannelNftsProps> = ({
   isFiltersApplied,
 }) => {
   const { currentPage, setCurrentPage } = usePagination(0)
-  const { activeChannelId } = useUser()
+  const { memberships } = useUser()
 
+  const userChannels = memberships?.map((membership) => membership.channels).flat()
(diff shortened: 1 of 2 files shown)


packages/atlas/src/views/viewer/ChannelView/ChannelViewTabs/ChannelNfts.tsx
```diff
@@ -38,6 +40,9 @@ export const ChannelNfts: React.FC<ChannelNftsProps> = ({
       creatorChannel: {
         id_eq: channelId,
       },
+      video: {
+        isPublic_eq: activeChannelId !== channelId ? true : undefined,
```
kdembler: Would it be a lot of hassle to actually check if the user is the owner of the channel instead of checking if the users is logged in as a channel? That would help in cases where I own multiple channels (but I can only be "logged in" as a single channel at a given time)

drillprop (changes requested):
Hmm, I can still see unlisted NFTs in the activity tab. Shouldn't we also hide them there? What about overview stats?  @kdembler? 
image

kdembler:
I think that may be going too far, I would ignore activities for now, I really doubt that somebody will even create an unlisted NFT in the foreseeable future

drillprop (approved):

kdembler (approved):

Issue #2440, opened 2022-04-08
Edit Comment component and integration


dmtrjsg:
Edit my comment + cancel action [link](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140596)
Cancel editing [link](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2302%3A299128)

Includes integration

Issue #2541, opened 2022-04-21
Disallow multiple transactions pending


toiletgranny:
I'm creating this task after my sync with @kdembler, so that we don't forget to address the issue of having multiple pending blockchain transactions in comments & reactions. Each proposed solution comes with different tradeoffs and time effort to implement, so that's why a collective decision on this is needed. 

Possible solutions to be discussed during a meeting:
1. Blocking all social interaction elements (reactions, comment inputs, video likes, etc.) on a page for when there's a pending blockchain transaction waiting for a signature.
   a. **Benefits**: the experience is not great, but at least it's solid and bulletproof = no sad surprises for the user,
   b. **Tradeoffs**: the inability of leaving multiple reactions under comment in one go,
   c. **Potential drawbacks:** what if the user navigates to another video page before signing a transaction for a like under the previous video?
   d. **Effort**: Design: Rather low? Design adjustments would take around 0.5 days.
2. Assigning each transaction consecutive unique ids/nonces, without waiting for the user to sign them.
   a. **Benefits**: allowing users to perform multiple "social interactions" on a video page in one go, one after another (eg. just like maniacally adding multiple emojis under a massage on Slack),
   b. **Tradeoffs**: transactions might fail if not signed in chronological order, which could happen quite often? This should be tested.
   c. **Potential drawbacks:** will that collide with NFT or video publishing-related transactions?
   d. **Effort**: Design: None.
3. (Transaction queue) Waiting for ~1/2/3 seconds (TBD) after clicking on a reaction (or like, or posting a comment), allowing the user to do this multiple times, and then triggering sort of a "transaction bundle".
   a. **Benefits, Tradeoffs, Potential drawbacks:** It's a mix of 1st and 2nd solution, depending on how exactly we're going to approach this.
   b. **Effort**: Design: Same as with 1st solution.

toiletgranny:
Draft note: 
1. Number two (except for the reactions chips)
2. Create a dialog asking the user to sign the outstanding polkadot transaction,

toiletgranny:
Alright, @dmtrjsg, @kdembler, I updated [designs, docs, and user stories for comments](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=1426%3A86220) to address the above.

Details can be found in the changelog in Figma, which I'm pasting below, for your convenience:

[...]

PR #2668, opened 2022-05-12
Edit comments integration

eldiegod:
fixes: #2440 #2720

### DIFF
--- a/packages/atlas/src/components/_comments/Comment/Comment.tsx
+++ b/packages/atlas/src/components/_comments/Comment/Comment.tsx
@@ -1,5 +1,5 @@
 import { format } from 'date-fns'
-import React, { useCallback, useRef, useState } from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 import { CSSTransition, SwitchTransition } from 'react-transition-group'
 
 import { Text } from '@/components/Text'
@@ -96,6 +96,17 @@ export const Comment: React.FC<CommentProps> = ({
     },
   ]
 
+  const domRef = useRef<HTMLDivElement>(null)
+  const [highlightedPreviously, setHighlightedPreviously] = useState<boolean | undefined>(false)
+
+  // scroll comment into view once the comment gets highlighted
+  useEffect(() => {
+    if (highlighted === true && !highlightedPreviously) {
+      domRef.current?.scrollIntoView({ behavior: 'smooth', block: 'center' })
+    }
+    setHighlightedPreviously(highlighted)
+  }, [highlightedPreviously, highlighted])
+
   const reactionIsProcessing = reactions?.some(({ state }) => state === 'processing')
   const allReactionsApplied =
     reactions && reactions.filter((r) => r.count).length >= Object.values(REACTION_TYPE).length
@@ -141,7 +152,7 @@ export const Comment: React.FC<CommentProps> = ({
       memberUrl={memberUrl}
       memberAvatarUrl={memberAvatarUrl}
     >
-      <CommentWrapper shouldShowKebabButton={shouldShowKebabButton}>
+      <CommentWrapper ref={domRef} shouldShowKebabButton={shouldShowKebabButton}>
         <SwitchTransition>
           <CSSTransition
             timeout={parseInt(cVar('animationTimingFast', true))}
--- a/packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
+++ b/packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
@@ -31,7 +31,7 @@ export type CommentInputProps = {
   hasInitialValueChanged: boolean
   onComment?: () => void
   onCancel?: () => void
-  onChange?: (value: string) => void
+  onChange?: (e: React.ChangeEvent<HTMLTextAreaElement>) => void
   onFocus?: () => void
 } & CommentRowProps
 
@@ -114,7 +114,7 @@ export const CommentInput: React.FC<CommentInputProps> = ({
             rows={1}
             value={value}
             placeholder={`Leave a public comment as ${memberHandle ? ` ${memberHandle}` : '...'}`}
-            onChange={(e) => !readOnly && onChange?.(e.target.value)}
+            onChange={(e) => !readOnly && onChange?.(e)}
             disabled={processing}
             data-processing={processing}
           />
--- a/packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
+++ b/packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
@@ -27,12 +27,12 @@ export type CommentInputProps = {
   processing: boolean
   readOnly?: boolean
   memberHandle?: string
+  value: string
+  hasInitialValueChanged: boolean
   onComment?: () => void
   onCancel?: () => void
   onChange?: (value: string) => void
   onFocus?: () => void
-  value?: string
-  initialValue?: string
 } & CommentRowProps
 
 const COMMENT_LIMIT = 50000
@@ -47,7 +47,7 @@ export const CommentInput: React.FC<CommentInputProps> = ({
   onChange,
   onFocus,
   value,
-  initialValue,
+  hasInitialValueChanged,
   ...rest
 }) => {
   const smMatch = useMediaMatch('sm')
@@ -57,14 +57,6 @@ export const CommentInput: React.FC<CommentInputProps> = ({
   const { displaySnackbar } = useSnackbar()
 
   const { ref: measureRef, height: textAreaHeight = 40 } = useResizeObserver({ box: 'border-box' })
-
-  useEffect(() => {
-    if (initialValue) {
-      onChange?.(initialValue)
-    }
-    // too difficult to make onChange stable but it kinda is already
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [initialValue])
 
   useEffect(() => {
     if (!active) {
@@ -94,8 +86,8 @@ export const CommentInput: React.FC<CommentInputProps> = ({
     onComment?.()
   }
 
-  const show = !!value || !!initialValue || active || processing
-  const canComment = !!value && initialValue !== value
+  const show = !!value || active || processing
+  const canComment = !!value && hasInitialValueChanged
   return (
     <StyledCommentRow {...rest} processing={processing} show={show} isMemberAvatarClickable={false}>
       <Container
--- a/packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
+++ b/packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
@@ -29,9 +29,10 @@ export type CommentInputProps = {
   memberHandle?: string
   onComment?: () => void
   onCancel?: () => void
-  onChange?: (e: React.ChangeEvent<HTMLTextAreaElement>) => void
+  onChange?: (value: string) => void
   onFocus?: () => void
   value?: string
+  initialValue?: string
 } & CommentRowProps
 
 const COMMENT_LIMIT = 50000
@@ -46,6 +47,7 @@ export const CommentInput: React.FC<CommentInputProps> = ({
   onChange,
   onFocus,
   value,
+  initialValue,
   ...rest
 }) => {
   const smMatch = useMediaMatch('sm')
@@ -57,15 +59,19 @@ export const CommentInput: React.FC<CommentInputProps> = ({
   const { ref: measureRef, height: textAreaHeight = 40 } = useResizeObserver({ box: 'border-box' })
 
   useEffect(() => {
+    if (initialValue) {
+      onChange?.(initialValue)
+    }
+    // too difficult to make onChange stable but it kinda is already
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [initialValue])
+
+  useEffect(() => {
     if (!active) {
       return
     }
     const handleClickOutside = (event: Event) => {
       if (containerRef.current && !containerRef.current.contains(event.target as Node)) {
-        // stop propagation so it doesn't get triggered again on button click
-        // prevent default so it doesn't trigger unwanted submits
-        event.preventDefault()
-        event.stopPropagation()
         setActive(false)
       }
     }
@@ -75,10 +81,8 @@ export const CommentInput: React.FC<CommentInputProps> = ({
     }
   }, [active])
 
-  const show = !!value || active || processing
-
   const validateLengthAndProcess = () => {
-    if (value && value.length > COMMENT_LIMIT) {
+    if (!!value && value.length > COMMENT_LIMIT) {
       displaySnackbar({
         title: 'Comment too long',
         description: `Your comment must be under 50 000 characters. Currently, it's ${formatNumber(value.length)}.`,
@@ -90,6 +94,8 @@ export const CommentInput: React.FC<CommentInputProps> = ({
     onComment?.()
   }
 
+  const show = !!value || !!initialValue || active || processing
+  const canComment = !!value && initialValue !== value
   return (
     <StyledCommentRow {...rest} processing={processing} show={show} isMemberAvatarClickable={false}>
       <Container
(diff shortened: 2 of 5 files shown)


drillprop (changes requested):
There were a lot of changes recently, so I'll review your PR after you resolve all the conflicts

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -12,7 +12,17 @@ import { ConsoleLogger } from '@/utils/logs'
 export const useReactionTransactions = () => {
   const { activeMemberId } = useUser()
   const { joystream, proxyCallback } = useJoystream()
-  const [commentInputProcessing, setCommentInputProcessing] = useState(false)
+  const [commentInputProcessing, _setCommentInputProcessing] = useState<Record<string, boolean>>({
+    'comment-box': false,
+  })
```
drillprop: What's going on here? Why `commentInputProcessing` is an object with this weirdly named property?

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -105,6 +125,34 @@ export const useReactionTransactions = () => {
...
+  const updateComment = useCallback(
+    async ({ commentId, videoId, commentBody }: { commentId: string; videoId: string; commentBody: string }) => {
+      if (!joystream || !activeMemberId || !videoId) {
+        ConsoleLogger.error('no joystream or active member')
+        return
+      }
+      await handleTransaction({
+        preProcess: () => {
+          setCommentInputProcessing({ commentId, value: true })
+        },
```
drillprop: Let's not put this in `preProcess`, let's move this before `handleTransaction` as we did in other methods.

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -12,7 +12,17 @@ import { ConsoleLogger } from '@/utils/logs'
...
+  const [commentInputProcessing, _setCommentInputProcessing] = useState<Record<string, boolean>>({
+    'comment-box': false,
+  })
+  const setCommentInputProcessing = ({ commentId, value }: { commentId: string; value: boolean }) => {
+    _setCommentInputProcessing((process) => {
+      const processing = { ...process }
+      processing[commentId] = value
+      return processing
+    })
+  }
```
drillprop, line 30: Could you explain why we need to set `commentId` in `commentInputProcessing `? This is because there could be multiple processing comments? This is impossible right now, we can't run multiple transactions at once. We don't block them yet, but I guess we will - https://github.com/Joystream/atlas/issues/2541

Also, I think, in that case, this should be named differently.

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -43,19 +43,21 @@ type GetCommentReactionsArgs = {
...
-  const [commentBody, setCommentBody] = useState('')
-  const { id } = useParams()
   const reactionPopoverDismissed = usePersonalDataStore((state) => state.reactionPopoverDismissed)
   const { activeMemberId, activeAccountId, signIn, activeMembership } = useUser()
   const { openSignInDialog } = useDisplaySignInDialog()
   const { isLoadingAsset: isMemberAvatarLoading, url: memberAvatarUrl } = useMemberAvatar(activeMembership)
   const [highlightedComment, setHighlightedComment] = useState<string | null>(null)
+  const { id: videoId } = useParams()
+  const [commentBody, _setCommentBody] = useState<Record<string, string>>({ 'comment-box': '' })
+  const [isEditingComment, _setIsEditingComment] = useState<Record<string, boolean>>({})
```
drillprop: Again, I'm a bit lost here.  
Why do we have `comment-box` property here and what does it mean(also why you didn't use `COMMENT_BOX_ID` above)?
Why `isEditingComment` is an object? This should be named differently. If I understand this correctly we could create an array or set of `editingCommentsIds` and If someone editing a comment we're adding commentId to this array when someone ends editing, we remove this commentId.  The name `isEditingComment` suggests that this is a primitive value, which could be only a boolean. The name of the variable should always inform you what's inside and what you should expect. 

I speak for myself but I highly doubt this will be understandable to someone who sees this code for the first time or revisits this after two weeks.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -27,9 +27,10 @@ export type CommentInputProps = {
   memberHandle?: string
   onComment?: () => void
   onCancel?: () => void
-  onChange?: (e: React.ChangeEvent<HTMLTextAreaElement>) => void
+  onChange?: (value: string) => void
   onFocus?: () => void
   value?: string
+  initialValue?: string
```
drillprop: Why do we need `initialValue`? I think `value` should be enough. Can't we just set `value` once the user triggers an edit comment?

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -50,16 +52,20 @@ export const CommentInput: React.FC<CommentInputProps> = ({
 
   const { ref: measureRef, height: textAreaHeight = 40 } = useResizeObserver({ box: 'border-box' })
 
+  useEffect(() => {
+    if (initialValue) {
+      onChange?.(initialValue)
+    }
+    // too difficult to make onChange stable but it kinda is already
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [initialValue])
```
drillprop: I think if my previous comment is true, we shouldn't need that effect

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -96,6 +96,16 @@ export const Comment: React.FC<CommentProps> = ({
     },
   ]
 
+  const domRef = useRef<HTMLDivElement>(null)
+  const [_highlighted, _setHighlighted] = useState<boolean | undefined>(false)
+
+  useEffect(() => {
+    if (highlighted === true && !_highlighted) {
```
drillprop: What means `highlighted` and not `_higlighted`?

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -168,9 +211,64 @@ export const CommentsSection: React.FC<CommentsSectionProps> = ({ disabled, vide
...
+          processing={commentInputProcessing[comment.id]}
+          readOnly={!activeMemberId}
+          memberHandle={activeMembership?.handle}
+          onFocus={() => !activeMemberId && openSignInDialog({ onConfirm: signIn })}
+          onComment={() => handleUpdateComment({ commentId: comment.id })}
+          value={commentBody[comment.id]}
+          initialValue={comment.text}
+          withoutOutlineBox
+          onChange={(value) => setCommentBody({ commentId: comment.id, comment: value })}
+          onCancel={() => {
```
drillprop: Let's create the handleCancelComment function.

packages/atlas/src/components/_comments/Comment/Comment.tsx
drillprop: I mean this works, but again I doubt that someone would know what's happening when he revisits this effect.

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -89,10 +91,29 @@ export const CommentsSection: React.FC<CommentsSectionProps> = ({ disabled, vide
     commentInputProcessing,
     deleteComment,
     moderateComment,
+    updateComment,
   } = useReactionTransactions()
 
   const mdMatch = useMediaMatch('md')
-  const placeholderItems = loading && !comments ? Array.from({ length: 4 }, () => ({ id: undefined })) : []
+
+  const setCommentBody = ({ commentId, comment }: { commentId: string; comment: string | undefined }) =>
```
drillprop: I'm not a big fan of creating variables like `_setSomething` and `setSomething`. Couldn't we rename this handler handleCommentBodyChange or smth?

drillprop (changes requested):
Works pretty good 💪🏻  but I have some concerns about naming. I got lost several times reading the code, I think we could improve this in a few places.

packages/atlas/src/hooks/useReactionTransactions.ts
drillprop: Oh, okay, nvm - I think I understand, why you need this.
Anyway, I sustain my opinion that this should be named differently though, and also I think we should keep an array or even [set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) of processing comments not an object.

packages/atlas/src/hooks/useReactionTransactions.ts
eldiegod: discussed below

packages/atlas/src/hooks/useReactionTransactions.ts
eldiegod: I disagree with the second part, we always use objects for indexing similar things, I don't think we have ever used a `set` probably because it doesn't play very well with `useState` and an array wouldn't allow for indexing by `CommentId` easily; anyways it could be done both ways

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
eldiegod: yes, it's exactly as you mention but instead of an array is a standard record (obj) for easier indexing.

I renamed the var for clarity

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
eldiegod: We need `initialValue` because we need to know if the comment being edited has changed from its `initialValue` before being able to publish it again

drillprop (changes requested):
The comments section is now broken, I don't see any comments. I'll review the code later.

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -136,41 +209,61 @@ export const CommentsSection: React.FC<CommentsSectionProps> = ({ disabled, vide
...
-          count: reactionsCount.find((r) => r.reactionId === reaction.reactionId)?.count || 0,
-          active: reactions.some((r) => r.reactionId === reaction.reactionId && r.member.id === activeMemberId),
-        }
-      })
-    },
-    [processingCommentReactionId, activeMemberId]
-  )
-  const memoizedComments = useMemo(() => {
     return comments?.map((comment, idx) => {
-      return (
```
drillprop, line 148: Ok found the issue - we don't see comments, because you remove `return` here

packages/atlas/src/hooks/useReactionTransactions.ts
drillprop: Actually, we're using the set in OverlayManager to get information on how many overlays are open. And it works pretty well. We're also using a set in the asset store, but this is a different case.

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
drillprop: I still sustain that you don't need that. You also don't need this awkward `useEffect` below.  
Hear me out:  
1. Remove initialValue and this useEffect
2. When a user clicks edit, we're calling: 
```
          onEditClick={() => setIsEditingComment({ commentId: comment.id, value: true })}
```
We could add here `setCommentInputTextRecord` and set the `comment.text`(basically what you do in this `useEffect`)
3. Instead of passing initialValue we could pass `isEditChanged` or something, which will be condition `commentInputTextRecord[commentId] === comment.text`
4. We're handling this `isEditChanged` inside the CommentInput and show a dialog if is false.

drillprop (changes requested):
1. The performance is really bad now. Writing in the comment is really laggy.
2. You reintroduce this issue:

https://user-images.githubusercontent.com/51168865/168980326-2af7cc93-c0c2-4b77-9866-dac4f8d6ef1e.mov

just FYI, this worked before.

drillprop:
>  The performance is really bad now. Writing in the comment is really laggy.

This could be a bigger problem and could be caused by something else also(not only your changes).  I've noticed that the same is happening on `reactions` on the video where we have a lot of comments(video/1). I'm looking at it right now.

eldiegod:
> > The performance is really bad now. Writing in the comment is really laggy.
> 
> This could be a bigger problem and could be caused by something else also(not only your changes). I've noticed that the same is happening on `reactions` on the video where we have a lot of comments(video/1). I'm looking at it right now.

@drillprop I'm looking for a solution for this but it's because every time an input triggers on change the commentInputTextRecord changes and it rerenders pretty much the whole comment section. 

But yeah the problem is still there regardless of my changes

eldiegod:
@drillprop I think it should all be good now, performance issues apart

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
```diff
@@ -27,11 +27,12 @@ export type CommentInputProps = {
   processing: boolean
   readOnly?: boolean
   memberHandle?: string
+  value: string
+  hasInitialValueChanged: boolean
   onComment?: () => void
   onCancel?: () => void
-  onChange?: (e: React.ChangeEvent<HTMLTextAreaElement>) => void
+  onChange?: (value: string) => void
```
drillprop: Now we don't need that change. I would bring back `onChange` which takes the event as a parameter.

packages/atlas/src/hooks/useTrackPropChange.ts
```diff
@@ -0,0 +1,19 @@
...
+      }
+      return ps
+    }, {})
+    if (Object.keys(changedProps).length > 0) {
+      // eslint-disable-next-line no-console
+      console.log('Changed props:', changedProps)
+    }
+    prev.current = props
+  })
+}
```
drillprop, line 19: I like that!

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -43,19 +43,23 @@ type GetCommentReactionsArgs = {
...
-  const [commentBody, setCommentBody] = useState('')
-  const { id } = useParams()
   const reactionPopoverDismissed = usePersonalDataStore((state) => state.reactionPopoverDismissed)
   const { activeMemberId, activeAccountId, signIn, activeMembership } = useUser()
   const { openSignInDialog } = useDisplaySignInDialog()
   const { isLoadingAsset: isMemberAvatarLoading, url: memberAvatarUrl } = useMemberAvatar(activeMembership)
   const [highlightedComment, setHighlightedComment] = useState<string | null>(null)
+  const { id: videoId } = useParams()
+  // indexed by commentId's
+  const [commentInputTextCollection, setCommentInputTextCollection] = useState(new Map<string, string>())
```
drillprop, line 59: You didn't have to change this to map 😅  I was fine with using an object in this case.

drillprop (changes requested):
Comment input is not cleared once you add a comment

image

packages/atlas/src/components/_comments/CommentInput/CommentInput.tsx
eldiegod: fixed

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -87,28 +109,54 @@ export const useReactionTransactions = () => {
...
+          setCommentInputIsProcessing({ commentInputId, value: false })
 
-          const newCommentsQueryResult = refetchResult[0] as GetCommentsConnectionQueryHookResult
+          const newCommentsQueryResult = refetchResult[0]
           // TODO - We probably shouldn't use inBlock here - it's possible that we could create multiple comments in one block
           // Update once https://github.com/Joystream/atlas/issues/2629 is done.
-          newCommentId = newCommentsQueryResult.data?.commentsConnection.edges.find(
-            (edge) => edge.node.commentcreatedeventcomment?.[0].inBlock === block
-          )?.node.id
+          newCommentId = newCommentsQueryResult.data?.userComments.find(
```
drillprop: Actually, we should use `videoCommentsConnection` here,  the userComments query ignores comments which are replies to another comment.

drillprop (changes requested):
Last thing

packages/atlas/src/hooks/useReactionTransactions.ts
eldiegod: fixed!

drillprop (approved):
Looks good to me, but I'd wait for final approval from Klaudiusz. 
I requested him to review it.

kdembler (approved):
Looks good to me as well! I found one small bug - if I start editing a comment, click the confirm button but cancel the transaction, the comments still gets highlighted. Let's fix it in some followup though. @eldiegod can you create an issue or a PR with a fix?

Issue #2446, opened 2022-04-08
Delete comment CV


dmtrjsg:
CV: Delete Comment I left in the past[ link](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140589)

PR #2674, opened 2022-05-13
Remove and moderate comment integration

### LINKED ISSUE #2447: Delete (moderate) comment CO
CO: Delete comment viewers left in the past [link](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140616)

drillprop:
Fix #2447 #2446 

delete comment: https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140589
moderate comment: https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2213%3A140616

Only deleted comments with replies should be "visible" in CommentSection. You'll find one in `video/10`

### DIFF
--- a/packages/atlas/src/api/queries/fragments.graphql
+++ b/packages/atlas/src/api/queries/fragments.graphql
@@ -296,4 +296,9 @@ fragment CommentFields on Comment {
   commentcreatedeventcomment {
     inBlock
   }
+  moderatedInEvent {
+    videoChannel {
+      title
+    }
+  }
 }
--- a/packages/atlas/src/api/queries/fragments.graphql
+++ b/packages/atlas/src/api/queries/fragments.graphql
@@ -296,9 +296,4 @@ fragment CommentFields on Comment {
   commentcreatedeventcomment {
     inBlock
   }
-  moderatedInEvent {
-    videoChannel {
-      title
-    }
-  }
 }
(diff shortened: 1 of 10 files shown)


packages/atlas/src/api/queries/fragments.graphql
```diff
@@ -296,4 +296,9 @@ fragment CommentFields on Comment {
   commentcreatedeventcomment {
     inBlock
   }
+  moderatedInEvent {
+    videoChannel {
+      title
+    }
+  }
```
kdembler: What do we need this event for?

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -527,7 +553,19 @@ export class JoystreamLibExtrinsics {
...
+    return this.sendMetaprotocolMemberExtrinsic(memberId, msg, cb)
+  }
+
+  async moderateComment(memberId: MemberId, channelId: MemberId, commentId: string, cb?: ExtrinsicStatusCallbackFn) {
+    await this.ensureApi()
+
+    const msg: IChannelOwnerRemarked = {
+      moderateComment: {
+        commentId,
+        rationale: 'I want to delete this comment',
```
kdembler: I suggest we leave this empty for now

packages/atlas/src/api/queries/fragments.graphql
drillprop: We need to know which channel moderated the comment. Check the `moderatedBy` prop in Comment.

kdembler: It will always be the channel that created the video, doesn't seem like we need to fetch that from the comment?

drillprop: Hmm, yeah, I guess we don't need this 👍🏻   Fixed.

kdembler (approved):

PR #2677, opened 2022-05-13
add NFT discovery section

kdembler:
Changes:
1. Added "New NFTs on sale" section on homepage
2. Added "Video NFTs" view with all the NFTs and filtering/pagination
3. Made NFT tile's full area clickable, not only the thumbnail
4. Changed NFT status filtering to use checkboxes again (you can select multiple statuses at once)


toiletgranny:
Perfect! As far as I can see — and I checked all the filters and counted the results — everything seems to be working fine!

The only thing that's bothering me is why can't we display the number of selected checkboxes in the blue badge in the "Status" filter group when the "Other filters" group manages this just fine? Not a blocker for me, though, in case that's a bigger issue.

toiletgranny (approved):

kdembler:
> The only thing that's bothering me is why can't we display the number of selected checkboxes in the blue badge in the "Status" filter group when the "Other filters" group manages this just fine? Not a blocker for me, though, in case that's a bigger issue.

I just missed that when changing from radio button. Fixed now

packages/atlas/src/views/viewer/NftsView/NftsView.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const smMatch = useMediaMatch('sm')
+  const mdMatch = useMediaMatch('md')
+
+  const filtersBarLogic = useFiltersBar()
+  const {
+    filters: { setIsFiltersOpen, isFiltersOpen },
+    canClearFilters: { canClearAllFilters },
+    ownedNftWhereInput,
+    videoWhereInput,
+  } = filtersBarLogic
```
drillprop, line 38: Hmmm, now we could apply the same filters on MemberView and ChannelView. Right now we can only filter by status there. 
Just saying, let's do this as a follow-up maybe?

drillprop (approved):
That's a big improvement, now NFTs are easier to find. That feels much better.

image

packages/atlas/src/views/viewer/NftsView/NftsView.tsx
kdembler: Yeah, let's do that later

PR #2681, opened 2022-05-13
fix clipped popovers

### LINKED ISSUE #2610: Fix clipped popovers in section headers
When opening the filters popover in the Channel page, the NFTs tab, the component is clipped if displayed above the trigger element.

https://user-images.githubusercontent.com/35307309/166924214-d2fa225e-3f0c-4f72-8991-7cea15ecb2fb.mp4

rafalpawlow:
Closes #2610

### DIFF
--- a/packages/atlas/src/components/_overlays/Popover/Popover.tsx
+++ b/packages/atlas/src/components/_overlays/Popover/Popover.tsx
@@ -21,6 +21,7 @@ export type PopoverProps = {
   onShow?: () => void
   children?: React.ReactNode
   disabled?: boolean
+  flipEnabled?: boolean
 }
 
 const EXIT_ANIMATION_DURATION = 100
@@ -46,6 +47,7 @@ const _Popover: React.ForwardRefRenderFunction<PopoverImperativeHandle | undefin
     trigger,
     className,
     disabled,
+    flipEnabled = true,
   },
   ref
 ) => {
@@ -90,6 +92,9 @@ const _Popover: React.ForwardRefRenderFunction<PopoverImperativeHandle | undefin
           {children}
         </ContentContainer>
       )}
+      popperOptions={{
+        modifiers: [{ name: 'flip', enabled: flipEnabled }],
+      }}
       placement={placement}
       offset={offset}
     >
--- a/packages/atlas/src/components/_overlays/DialogPopover/DialogPopover.tsx
+++ b/packages/atlas/src/components/_overlays/DialogPopover/DialogPopover.tsx
@@ -13,7 +13,7 @@ const _DialogPopover: React.ForwardRefRenderFunction<PopoverImperativeHandle, Di
   ref
 ) => {
   return (
-    <Popover {...popoverProps} ref={ref}>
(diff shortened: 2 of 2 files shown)


packages/atlas/src/components/_overlays/Popover/Popover.tsx
```diff
@@ -90,6 +92,9 @@ const _Popover: React.ForwardRefRenderFunction<PopoverImperativeHandle | undefin
           {children}
         </ContentContainer>
       )}
+      popperOptions={{
+        modifiers: [{ name: 'flip', enabled: flipEnabled }],
```
drillprop, line 96: I'm wondering, do we actually want this flip behavior enabled in the entire app? I think we should disable this and not worry about passing the `flipEnabled` prop.

drillprop (changes requested):

packages/atlas/src/components/_overlays/Popover/Popover.tsx
rafalpawlow: I'm not sure if this is a good idea, this might be helpful for tooltips

kdembler: What exactly are we discussing here?

drillprop: If you look at Adam's video you'll notice that the popover changes his position when there is not much space for this context menu in the viewport. And I'm guessing this is the default behavior in Popover

https://user-images.githubusercontent.com/35307309/166924214-d2fa225e-3f0c-4f72-8991-7cea15ecb2fb.mp4

My proposition was to just turn off this behavior, and don't add a prop that could control this as Rafał did.

kdembler: Oh, I would say that we should keep this on always. :P It's a good UX, for some popovers there won't be enough space to display them below. The solution isn't to not flip them, but to make sure they remain visible when flipped

rafalpawlow: @kdembler I'm affraid this won't be possible in the case above because of z-indexes.

kdembler: What do you mean? Can't we change those z-indexes?

packages/atlas/src/components/_overlays/Popover/Popover.tsx
```diff
@@ -90,6 +90,7 @@ const _Popover: React.ForwardRefRenderFunction<PopoverImperativeHandle | undefin
           {children}
         </ContentContainer>
       )}
+      appendTo={document.body}
```
rafalpawlow: https://github.com/atomiks/tippyjs-react/issues/207

drillprop (approved):
Looks good to me. @kdembler could you look as well, since you had some concerns about the previous approach? 

Edit: Oh, this will be merged into `dev` so you need to approve anyway 😅

kdembler (approved):

PR #2683, opened 2022-05-13
add twemojis integration

### LINKED ISSUE #2519: Research integration Twemojis
Try integrating [Twemoji](https://twemoji.twitter.com/) into the app, see what possible solutions are

kdembler:
Closes #2519


packages/atlas/src/components/EmojiWrapper/EmojiWrapper.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+        memberAvatarUrl="https://placedog.net/100/100?random=2"
+        memberHandle="johndoe"
+        loading={false}
+        createdAt={new Date(1649761429792)}
+        reactions={[
+          { count: 320, reactionId: 1 },
+          { count: 500, reactionId: 2, active: true },
+          { count: 100, reactionId: 3 },
+        ]}
+      />
```
drillprop: Can we make this story a little prettier?
image
This looks like billboards in Zakopane
image

I'm aware that this is only a storybook, but designers use that, so I think they could be confused about what's going on here. 😅 
Maybe it could be a good idea to create a different template for the Comment component?

packages/atlas/src/components/EmojiWrapper/EmojiWrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children?: React.ReactNode
+}
+
+export const EmojiWrapper = forwardRef<HTMLSpanElement, EmojiWrapperProps>(
+  ({ block, className, children }, outsideRef) => {
+    const ref = useRef<HTMLDivElement>(null)
+
+    // parse the DOM node and replace all emojis with twemojis
+    // no deps array, run on every render
+    useEffect(() => {
```
drillprop, line 19: So, basically for every node where we want to see twemojis, we would need to apply this EmojiiWrapper? 
We can't just wrap `<App>` in EmojiWrapper and see twemojis in the entire app?

drillprop (changes requested):

packages/atlas/src/components/EmojiWrapper/EmojiWrapper.tsx
kdembler: To answer your first question - yes. I didn't wrap the full App since then Twemojis parsing would need to happen whenver any content changes and I'm not sure if that's desired, mostly from performance standpoint. How this works is that the twemoji library will need to traverse the full DOM and see if there are any emojis that should be replaced, I don't want to do that for the full DOM tree on every application render.

drillprop: ok, that makes sense 👍🏻

packages/atlas/src/components/EmojiWrapper/EmojiWrapper.stories.tsx
kdembler: Removed the comment component from the story, changed the text for comment story to include emojis and slightly cleaned up the emoji wrapper story

drillprop (approved):

PR #2718, opened 2022-05-17
Fix issue with not removing processing assets from MyVideos

### LINKED ISSUE #2635: Issues with processing uploads
We need to have some way to either cancel a processing upload if it's been processing for too long or offer the user an option to reupload. We could save the date at which we start processing, and if it's more than a minute or something, we remove it from processing.

To reproduce:
1. Start a big upload so it doesn't finish quickly
2. Refresh the app so you have failed upload
3. Manually modify local storage to include failed upload ID in `processingAssetsId` in uploads store

---

I have seen several reports on Discord from users that struggled with uploads processing forever, not having an option to retry when it's clear the upload failed. Some of the relevant info:

SCREENSHOT

Screenshot 2022-05-11 at 08 59 52

**Usersnap report by**: delranchorecipes@gmail.com

drillprop:
Fix #2635

### DIFF
--- a/packages/atlas/src/providers/uploadsManager/store.ts
+++ b/packages/atlas/src/providers/uploadsManager/store.ts
@@ -9,12 +9,18 @@ type AssetFile = {
   blob: File | Blob
 }
 
+type ProcessingAsset = {
+  // unix timestamp
+  expiresAt: number
+  id: string
+}
+
 type UploadStoreState = {
   uploads: AssetUpload[]
   uploadsStatus: UploadsStatusRecord
   assetsFiles: AssetFile[]
   isSyncing: boolean
-  processingAssetsIds: string[]
+  processingAssets: ProcessingAsset[]
   // store ids of channels that were created as part of the session to ignore them when checking missing assets
   newChannelsIds: string[]
 }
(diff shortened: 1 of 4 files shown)


packages/atlas/src/providers/uploadsManager/store.ts
```diff
@@ -26,11 +32,13 @@ type UploadStoreActions = {
...
   addAssetFile: (assetFile: AssetFile) => void
   setIsSyncing: (isSyncing: boolean) => void
-  removeProcessingAssetId: (contentId: string) => void
-  addProcessingAssetId: (contentId: string) => void
+  removeProcessingAsset: (contentId: string) => void
+  addProcessingAsset: (contentId: string) => void
   addNewChannelId: (channelId: string) => void
 }
 
+const THREE_MINUTES = 1000 * 60 * 3
```
kdembler: Let's call this something like `UPLOAD_PROCESSING_TIMEOUT` and move to global config. I would also change this to one minute I think

kdembler (commented):
Just one comment, otherwise looks good!

kdembler (approved):

PR #2721, opened 2022-05-18
linking to comments

### LINKED ISSUE #2391: Allow linking to a specific comment
## Context

1. User clicks the notification or a hyper link from the comment timestamp
2. We redirect them to specific video page

## Scope:
3. We scroll to make sure the comment from notification is visible on the screen
4. We use highlighted state
- [ ] Both notifications redirect with screen focus to originating comment or reply (https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411:216105)

rafalpawlow:
Closes #2391 

**Design**
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2550%3A230287


packages/atlas/src/api/hooks/comments.ts
```diff
@@ -29,6 +32,21 @@ export const useComments = (
   }
 }
 
+export const useComment = (
+  variables?: GetCommentQueryVariables,
```
drillprop: I guess we could safely take only id as a parameter. No need to pass `variables`. That's how it works at least in `useChannel` and `useVideo`

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -1,61 +1,69 @@
...
+
+  useEffect(() => {
+    if (!commentLink || !commentRef.current || scrolledToComment) {
+      return
+    }
+    const scrollTimeout = setTimeout(() => {
+      commentRef.current?.scrollIntoView({ behavior: 'smooth', block: 'center' })
+      setHighlightedComment(commentLink)
+      setScrolledToComment(true)
+    }, SCROLL_TO_COMMENT_TIMEOUT)
```
drillprop, line 94: Hmm this does work well when you click on the notification and the link leads you to another video, but it's really confusing what's going on when you click several times on comments in the comments section. 
Take a look at this video: 

https://user-images.githubusercontent.com/51168865/169229444-8d2d2e85-21e4-4435-ac7b-0361383f1a12.mov

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -270,6 +267,18 @@ export const CommentsSection: React.FC<CommentsSectionProps> = ({ disabled, vide
         <EmptyFallback title="Be the first to comment" subtitle="Nobody has left a comment under this video yet." />
       )}
       <CommentWrapper>
+        <HightLightedComment
```
drillprop: typo here

packages/atlas/src/views/viewer/VideoView/HightLightedComment.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+import { useComment } from '@/api/hooks'
+import { CommentFieldsFragment, CommentStatus, VideoFieldsFragment } from '@/api/queries'
+import { Comment } from '@/components/_comments/Comment'
+import { ReactionId } from '@/config/reactions'
+import { absoluteRoutes } from '@/config/routes'
+
+import { getCommentReactions } from './utils'
+
+type HightLightedCommentProps = {
```
drillprop: typo

packages/atlas/src/views/viewer/VideoView/HightLightedComment.tsx
```diff
@@ -0,0 +1,103 @@
...
+  reactionPopoverDismissed: boolean
+  authorized: boolean
+  activeMemberId: string | null
+  onCommentReaction: (commentId: string, reactionId: ReactionId) => void
+  onDeleteComment: (comment: CommentFieldsFragment, video: VideoFieldsFragment) => void
+  processingCommentReactionId: string | null
+  video?: VideoFieldsFragment | null
+}
+
+export const HightLightedComment: React.FC<HightLightedCommentProps> = ({
```
drillprop: typo

drillprop (changes requested):

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
drillprop: I think you'll need to talk about this with the designer. Correct me if I'm wrong but I didn't find any stories, that explain how it should work. Youtube for example reloads the page when you click someone's comment in the comment section.  

I'm also fine if we do this as a follow-up.

rafalpawlow: I improved it a bit, let me know what u think

packages/atlas/src/api/hooks/comments.ts
```diff
@@ -29,6 +32,39 @@ export const useComments = (
...
+  memberId: string,
+  videoId: string,
+  opts?: QueryHookOptions<GetCommentQuery, GetCommentQueryVariables>
+) => {
+  const { data, ...rest } = useGetCommentQuery({
+    ...opts,
+    variables: { where: { id }, memberId, videoId },
+  })
+
+  const userCommentReactionsLookup =
```
drillprop: We do this twice, maybe we could create a function which takes comment reactions and convert them into lookup?

packages/atlas/src/views/viewer/VideoView/HighlightedComment.tsx
```diff
@@ -0,0 +1,109 @@
...
+  )
+
+  return (
+    <>
+      {comment && !comment.parentCommentId && !parentCommentLoading ? getComment(comment, highLighted) : null}
+      {parentComment && getComment(parentComment, false)}
+      {parentComment && comment && getComment(comment, highLighted, true)}
+      {(commentLoading || parentCommentLoading) && <Comment type="default" loading />}
+      {parentCommentLoading && <Comment type="default" loading indented />}
+    </>
```
drillprop: let's refactor this a bit and use `if statements`

drillprop (changes requested):

drillprop (changes requested):
Looks good, but I've noticed one thing. When you click the comment in the comment section, all the comments disappear for a split second sometimes. I think this happens pretty consistently when you click the comment that has a parent comment(a comment which is a reply). Can we improve that?

drillprop (approved):

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -226,7 +267,27 @@ export const CommentsSection: React.FC<CommentsSectionProps> = ({ disabled, vide
...
+          ...(commentFromUrl.parentCommentId
+            ? filteredParentCommentReplies
+              ? [filteredParentCommentReplies, commentFromUrl]
+              : []
+            : [commentFromUrl]),
+          ...(comments?.filter(
+            (comment) => ![commentFromUrl.id, commentFromUrl.parentCommentId].includes(comment.id)
+          ) || []),
+        ]
+      : comments
```
kdembler: This is super hard to read, I have no idea what's going on. Maybe we could:
1. Split this into more intermediary steps with named variables or use `if/else`? Not sure, but what we have currently doesn't work great 
2. Extract part of this comments logic to some local hook to encapsulate? (just an idea, up to you)

kdembler (commented):
One observation: when I click a comment to mark it as highlighted, the entire comments section is reloading. Why is that happening? I would expect we could do this without fetching any new data. Maybe something like this in `cache.ts` would help to avoid fetching the comment again?
```tsx
commentByUniqueInput: (existing, { toReference, args }) => {
  return (
    existing ||
    toReference({
      __typename: 'Comment',
      id: args?.where.id,
    })
  )
},
```

kdembler (approved):

PR #2723, opened 2022-05-18
Fetch comment reactions from user separately

### LINKED ISSUE #2719: Fetch comment reactions from user separately
Right now to identify if the user left a reaction under a comment we querying comments left on a particular video and we also add the reactions field to this query. 
```
reactions {
  reactionId
  memberId
}
```
This won't work in the longer term. The` reactions` field will be limited to 50 records by a query and in the result -  it could lead to a bug. For example, it's possible that the particular reaction from a particular member won't be inside these 50 records.

We should probably run another query, which will fetch only comment reactions left under a particular video. Something like:
```
query {
  commentReactions(
  where: { member: { id_eq: "15" }, video: {id_eq:"1"} }
  limit: 1000
) {
  reactionId
  commentId
}
}
```

drillprop:
Fix #2719

### DIFF
--- a/packages/atlas/src/api/hooks/comments.ts
+++ b/packages/atlas/src/api/hooks/comments.ts
@@ -53,16 +53,31 @@ export const useCommentSectionComments = (
 
   const userCommentLookup = data?.userComments && createLookup(data?.userComments)
 
+  const userCommentReactionsLookup =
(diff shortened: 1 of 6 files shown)


kdembler (approved):
Great job!

PR #2725, opened 2022-05-19
Fix button in ending overlay

### LINKED ISSUE #2724: Bug inside video ending overlay play button
From the video ending overlay:

Screenshot 2022-05-18 at 17 57 05

You can see that the icon is offset to the right and not centered

drillprop:
Fix #2724


eldiegod (approved):
all good here 👍

PR #2727, opened 2022-05-20
Handle metaprotocol errors

### LINKED ISSUE #2629: Handling metaprotocol errors/success
Listen to QN updates on metaprotocol extrinsics. See https://github.com/Joystream/joystream/issues/3255#issuecomment-1087929524

drillprop:
Fix #2629

### DIFF
--- a/packages/atlas/src/api/queries/fragments.graphql
+++ b/packages/atlas/src/api/queries/fragments.graphql
@@ -282,6 +282,23 @@ fragment CommentFields on Comment {
   text
   status
   commentcreatedeventcomment {
-    inBlock
+    inExtrinsic
+  }
+}
+
+fragment MetaprotocolTransactionStatusEventFields on MetaprotocolTransactionStatusEvent {
+  inExtrinsic
+  inBlock
+  status {
+    __typename
+    ... on MetaprotocolTransactionErrored {
+      message
+    }
+    ... on MetaprotocolTransactionSuccessful {
+      dummy
+    }
+    ... on MetaprotocolTransactionPending {
+      dummy
+    }
   }
 }
--- a/packages/atlas/src/api/queries/fragments.graphql
+++ b/packages/atlas/src/api/queries/fragments.graphql
@@ -281,6 +281,7 @@ fragment CommentFields on Comment {
   repliesCount
   text
   status
+  # TODO update once QN supports getting ID directly from the status query
   commentcreatedeventcomment {
     inExtrinsic
   }
@@ -293,12 +294,6 @@ fragment MetaprotocolTransactionStatusEventFields on MetaprotocolTransactionStat
     __typename
     ... on MetaprotocolTransactionErrored {
       message
-    }
-    ... on MetaprotocolTransactionSuccessful {
-      dummy
-    }
-    ... on MetaprotocolTransactionPending {
-      dummy
     }
   }
 }
--- a/packages/atlas/src/hooks/useReactionTransactions.ts
+++ b/packages/atlas/src/hooks/useReactionTransactions.ts
@@ -112,9 +112,9 @@ export const useReactionTransactions = () => {
           setCommentInputIsProcessing({ commentInputId, value: false })
 
           const { data } = refetchResult[0]
-          data?.videoCommentsConnection.edges.find(
-            (edge) => edge.node.commentcreatedeventcomment?.[0].inExtrinsic === transactionHash
-          )?.node.id
+          // TODO update once we have replies ready(user comments query ignores replies)
+          data?.userComments.find((comment) => comment.commentcreatedeventcomment?.[0].inExtrinsic === transactionHash)
+            ?.id
         },
         onError: () => {
           setCommentInputIsProcessing({ commentInputId, value: false })
--- a/packages/atlas/src/hooks/useReactionTransactions.ts
+++ b/packages/atlas/src/hooks/useReactionTransactions.ts
@@ -114,6 +114,8 @@ export const useReactionTransactions = () => {
           const { data } = refetchResult[0]
 
           newCommentId = data?.videoCommentsConnection.edges.find(
+            // TODO We shouldn't fetch additional data from the commentcreatedeventcomment
+            // update this once QN supports getting ID directly from the status query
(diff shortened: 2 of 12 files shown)


packages/atlas/src/api/queries/fragments.graphql
```diff
@@ -282,6 +282,23 @@ fragment CommentFields on Comment {
...
+    __typename
+    ... on MetaprotocolTransactionErrored {
+      message
+    }
+    ... on MetaprotocolTransactionSuccessful {
+      dummy
+    }
+    ... on MetaprotocolTransactionPending {
+      dummy
+    }
```
kdembler: I guess we could just skip those? Since type information will be conveyed via `__typename` anyway

packages/atlas/src/hooks/useReactionTransactions.ts
```diff
@@ -107,15 +107,14 @@ export const useReactionTransactions = () => {
             parentCommentId || null,
             proxyCallback(updateStatus)
           ),
-        onTxSync: async ({ block }) => {
+        onTxSync: async ({ transactionHash }) => {
           const refetchResult = await refetchComments()
           setCommentInputIsProcessing({ commentInputId, value: false })
 
-          const newCommentsQueryResult = refetchResult[0]
-          // TODO - We probably shouldn't use inBlock here - it's possible that we could create multiple comments in one block
```
kdembler, line 115: Let's update and keep this TODO - ideally we get the ID directly from the status query, without need to find comment by `commentcreatedeventcomment`

packages/atlas/src/providers/transactionManager/useTransaction.ts
```diff
@@ -100,6 +105,17 @@ export const useTransaction = (): HandleTransactionFn => {
           addBlockAction({ callback: syncCallback, targetBlock: result.block })
         })
 
+        const { data } = await getTransactionStatus({ variables: { transactionHash: result.transactionHash } })
```
kdembler: This needs to happen after `queryNodeSyncPromise` is resolved - this gives us information that QN already processed the block in which our tx was included

kdembler: We should probably add some polling if the data was not found, but with some retry timeout, if possible. I would suggest poll every 1 sec for max 10 tries

kdembler (changes requested):

packages/atlas/src/providers/transactionManager/useTransaction.ts
```diff
@@ -30,13 +32,19 @@ type HandleTransactionFn = <T extends ExtrinsicResult>(opts: HandleTransactionOp
...
+const RETRIES_TIMEOUT = 1000
+const RETRIES = 10
 
 export const useTransaction = (): HandleTransactionFn => {
   const { addBlockAction, setDialogStep, setErrorCode, addPendingSign, removePendingSign } = useTransactionManagerStore(
     (state) => state.actions
   )
   const nodeConnectionStatus = useConnectionStatusStore((state) => state.nodeConnectionStatus)
   const { displaySnackbar } = useSnackbar()
+  const [getTransactionStatus, { refetch }] = useGetMetaprotocolTransactionStatusEventsLazyQuery({
```
kdembler: Can we rename this `refetch` to something more descriptive, maybe `refetchTransactionStatus`?

kdembler: Let's add fetch policy `network-only` here

packages/atlas/src/providers/transactionManager/useTransaction.ts
```diff
@@ -100,26 +108,58 @@ export const useTransaction = (): HandleTransactionFn => {
...
-              setDialogStep(ExtrinsicStatus.Completed)
+        await queryNodeSyncPromise
+
+        if (result.transactionHash) {
+          const { data } = await getTransactionStatus({ variables: { transactionHash: result.transactionHash } })
+          const status = data?.metaprotocolTransactionStatusEvents[0]?.status
+
+          if (!status) {
+            for (let i = 0; i <= RETRIES; i++) {
+              ConsoleLogger.warn(`No transcation status event found - retries: ${i}/${RETRIES}`)
```
kdembler:
```suggestion
              ConsoleLogger.warn(`No transaction status event found - retries: ${i + 1}/${RETRIES}`)
```

packages/atlas/src/providers/transactionManager/useTransaction.ts
```diff
@@ -100,26 +108,58 @@ export const useTransaction = (): HandleTransactionFn => {
...
-            snackbarSuccessMessage &&
-              displaySnackbar({
-                ...snackbarSuccessMessage,
-                iconType: 'success',
-                timeout: TX_COMPLETED_SNACKBAR_TIMEOUT,
-              })
-            resolve(true)
+          }
+
+          if (status?.__typename === 'MetaprotocolTransactionErrored') {
```
kdembler, line 139: This will not work properly if we got `status` during retries - in your `for` loop, `status` is declared as scoped variable so when you check `status` in this line you are accessing the one from L115, even if we got a response during retry

kdembler (changes requested):

packages/atlas/src/providers/transactionManager/useTransaction.ts
```diff
@@ -111,35 +113,43 @@ export const useTransaction = (): HandleTransactionFn => {
...
+              const { data: refetchedData } = await refetchTransactionStatus()
+              const statusAfterRefetch = refetchedData?.metaprotocolTransactionStatusEvents[0]?.status
+
+              if (statusAfterRefetch?.__typename === 'MetaprotocolTransactionErrored') {
+                throw new JoystreamLibError({
+                  name: 'MetaprotocolTransactionError',
+                  message: statusAfterRefetch?.message,
+                  details: result,
+                })
               }
```
kdembler: I don't think we need to duplicate this. You could just change the declaration in L117 to `let status = ...` and later just override `status` instead of creating new variable. Then after the retry block you can do a single `if` catching both cases

kdembler (approved):

Issue #2722, opened 2022-05-18
Fix performance issues in CommentSection


drillprop:
If you go to the video that has many comments(for example`video/1`) you'll notice that the writing in the comment input is very laggy. 
The issue is caused by the Comment component, which re-renders every time you write something in Input. This results in very bad performance.

PR #2729, opened 2022-05-20
Comment components refactor, performance improvements

eldiegod:
Fix: #2722

Change summary:

- Renamed/moved Comment component to InternalComment
- now Comment is a smart container that handles InternalComment + reply input + editing input and passes all the logic and fn handlers down
- CommentThread is now simplified and it serves to show a main Comment and it's replies, is up to discussion if its be needed anymore
- added useComment hook which returns a single comment with it's reactions and replies and accepts a `commented` as parameter
- updated useReactionTransactions hook to not hold any state anymore; such state is stored were it's needed instead mostly in commentSection and Comment components
- CommentSection has been simplified as well and it no longer holds a lot of business logic related to comments

### DIFF
--- a/packages/atlas/src/components/_comments/Comment/Comment.tsx
+++ b/packages/atlas/src/components/_comments/Comment/Comment.tsx
@@ -43,10 +43,11 @@ export const Comment: React.FC<CommentProps> = React.memo(
     const [processingCommentReactionId, setProcessingCommentReactionId] = useState<string | null>(null)
 
     const { activeMemberId, activeMembership, activeAccountId, signIn } = useUser()
-    const { comment, loading: loadingQuery } = useComment(
+    const { comment } = useComment(
       { commentId: commentId ?? '', memberId: activeMemberId ?? undefined },
       {
         skip: !commentId,
+        fetchPolicy: 'cache-only',
       }
     )
     const { isLoadingAsset: isMemberAvatarLoading, url: memberAvatarUrl } = useMemberAvatar(activeMembership)
@@ -196,15 +197,13 @@ export const Comment: React.FC<CommentProps> = React.memo(
       handle: comment.author.handle,
     }))
 
-    const loading = !commentId || loadingQuery
-
-    const userReactionsIds = comment?.userReactions?.map((reaction) => reaction.reactionId)
+    const loading = !commentId
 
     const reactions =
       comment &&
       getCommentReactions({
         commentId: comment?.id,
-        userReactionsIds,
+        userReactionsIds: comment?.userReactions,
         reactionsCount: comment?.reactionsCountByReactionId,
         activeMemberId,
         processingCommentReactionId,
--- a/packages/atlas/src/components/_comments/Comment/Comment.tsx
+++ b/packages/atlas/src/components/_comments/Comment/Comment.tsx
@@ -2,6 +2,7 @@ import React, { useRef, useState } from 'react'
 
 import { useComment } from '@/api/hooks'
 import { CommentFieldsFragment, CommentStatus, VideoFieldsFragment } from '@/api/queries'
+import { DialogModal } from '@/components/_overlays/DialogModal'
 import { ReactionId } from '@/config/reactions'
 import { QUERY_PARAMS, absoluteRoutes } from '@/config/routes'
 import { useDisplaySignInDialog } from '@/hooks/useDisplaySignInDialog'
@@ -15,6 +16,7 @@ import { useUser } from '@/providers/user'
 import { InternalComment } from './InternalComment'
 import { getCommentReactions } from './utils'
 
+import { CommentEditHistory } from '../CommentEditHistory'
 import { CommentInput } from '../CommentInput'
 import { CommentRowProps } from '../CommentRow'
 
@@ -22,26 +24,16 @@ export type CommentProps = {
   commentId?: string
   video?: VideoFieldsFragment | null
   isReplyable?: boolean
-  setOriginalComment?: React.Dispatch<React.SetStateAction<CommentFieldsFragment | null>>
-  setShowEditHistory?: React.Dispatch<React.SetStateAction<boolean>>
   setHighlightedCommentId?: React.Dispatch<React.SetStateAction<string | null>>
   setRepliesOpen?: React.Dispatch<React.SetStateAction<boolean>>
   isRepliesOpen?: boolean
 } & Exclude<CommentRowProps, 'memberAvatarUrl' | 'isMemberAvatarLoading'>
 
 export const Comment: React.FC<CommentProps> = React.memo(
-  ({
-    commentId,
-    video,
-    setOriginalComment,
-    setShowEditHistory,
-    setHighlightedCommentId,
-    setRepliesOpen,
-    isRepliesOpen,
-    isReplyable,
-    ...rest
-  }) => {
+  ({ commentId, video, setHighlightedCommentId, setRepliesOpen, isRepliesOpen, isReplyable, ...rest }) => {
     const replyCommentInputRef = useRef<HTMLTextAreaElement>(null)
+    const [originalComment, setOriginalComment] = useState<CommentFieldsFragment | null>(null)
+    const [showEditHistory, setShowEditHistory] = useState(false)
     const [replyInputOpen, setReplyInputOpen] = useState(false)
     const [editCommentInputIsProcessing, setEditCommentInputIsProcessing] = useState(false)
     const [replyCommentInputIsProcessing, setReplyCommentInputIsProcessing] = useState(false)
@@ -50,9 +42,15 @@ export const Comment: React.FC<CommentProps> = React.memo(
     const [isEditingComment, setIsEditingComment] = useState(false)
     const [processingCommentReactionId, setProcessingCommentReactionId] = useState<string | null>(null)
 
-    const { comment, loading: loadingQuery } = useComment(commentId ?? '', { skip: !commentId })
     const { activeMemberId, activeMembership, activeAccountId, signIn } = useUser()
+    const { comment, loading: loadingQuery } = useComment(
+      { commentId: commentId ?? '', memberId: activeMemberId ?? undefined },
+      { skip: !commentId }
+    )
     const { isLoadingAsset: isMemberAvatarLoading, url: memberAvatarUrl } = useMemberAvatar(activeMembership)
+    const { isLoadingAsset: isCommentMemberAvatarLoading, url: commentMemberAvatarUrl } = useMemberAvatar(
+      comment?.author
+    )
     const commentIdQueryParam = useRouterQuery(QUERY_PARAMS.COMMENT_ID)
     const reactionPopoverDismissed = usePersonalDataStore((state) => state.reactionPopoverDismissed)
     const { openSignInDialog } = useDisplaySignInDialog()
@@ -178,7 +176,19 @@ export const Comment: React.FC<CommentProps> = React.memo(
       setRepliesOpen?.(true)
     }
 
+    const handleOnEditClick = () => {
+      if (comment) {
+        setIsEditingComment(true)
+        setEditCommentInputText?.(comment.text)
+      }
+    }
+
     const handleOpenSignInDialog = () => !activeMemberId && openSignInDialog({ onConfirm: signIn })
+
+    const handleOnEditLabelClick = () => {
+      setShowEditHistory?.(true)
+      comment && setOriginalComment?.(comment)
+    }
 
     const replyAvatars = comment?.replies?.map((comment) => ({
       url:
@@ -188,13 +198,29 @@ export const Comment: React.FC<CommentProps> = React.memo(
 
     const loading = loadingQuery || !commentId
 
-    const userReactions = comment?.reactions
-      .filter((reaction) => reaction.memberId === activeMemberId)
-      .map((reaction) => reaction.reactionId)
+    const userReactionsIds = comment?.userReactions.map((reaction) => reaction.reactionId)
+
+    const reactions =
+      comment &&
+      getCommentReactions({
+        commentId: comment?.id,
+        userReactionsIds,
+        reactionsCount: comment?.reactionsCountByReactionId,
+        activeMemberId,
+        processingCommentReactionId,
+      })
+
+    const commentType =
+      comment && ['DELETED', 'MODERATED'].includes(comment.status)
+        ? 'deleted'
+        : comment && (video?.channel.ownerMember?.id === activeMemberId || comment?.author.id === activeMemberId)
+        ? 'options'
+        : 'default'
 
     if (isEditingComment) {
       return (
         <CommentInput
+          indented={!isReplyable}
           processing={editCommentInputIsProcessing}
           readOnly={!activeMemberId}
           memberHandle={activeMembership?.handle}
@@ -218,7 +244,7 @@ export const Comment: React.FC<CommentProps> = React.memo(
       return (
         <>
           <InternalComment
-            indented={!isReplyable}
+            indented={!!comment?.parentCommentId}
             isCommentFromUrl={commentId === commentIdQueryParam}
             videoId={video?.id}
             commentId={commentId}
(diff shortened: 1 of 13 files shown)


packages/atlas/src/MainLayout.tsx
```diff
@@ -80,14 +81,16 @@ export const MainLayout: React.FC = () => {
...
       <CookiePopover />
-      <Routes>
-        <Route path={absoluteRoutes.embedded.video()} element={<EmbeddedView />} />
-        <Route path={BASE_PATHS.viewer + '/*'} element={<ViewerLayout />} />
-        <Route path={BASE_PATHS.legal + '/*'} element={<LegalLayout />} />
-        <Route path={BASE_PATHS.studio + '/*'} element={<LoadableStudioLayout />} />
-        <Route path={BASE_PATHS.playground + '/*'} element={<LoadablePlaygroundLayout />} />
-      </Routes>
-      <AdminOverlay />
+      <ReactionsTransactionsProvider>
```
rafalpawlow: What's the reason to keep this provider here? All providers live in `App.tsx`

rafalpawlow (changes requested):

packages/atlas/src/MainLayout.tsx
eldiegod: fixed!

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -260,3 +273,138 @@ export const Comment: React.FC<CommentProps> = ({
     </CommentRow>
   )
 }
+
+export type CommentProps = {
+  commentId?: string
+  video?: VideoFieldsFragment | null
+  setCommentInputText?: ({ commentId, comment }: { commentId: string; comment: string | undefined }) => void
+} & CommentRowProps
+export const Comment: React.FC<CommentProps> = React.memo(({ commentId, video, setCommentInputText, ...rest }) => {
```
kdembler: Let's extract this to a separate file

packages/atlas/src/providers/reactionsTransactions/provider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  setShowEditHistory: React.Dispatch<React.SetStateAction<boolean>>
+  isEditingCommentCollection: Set<string>
+  setIsEditingComment: ({ commentId, value }: { commentId: string; value: boolean }) => void
+}
+
+export const ReactionsTransactionsContext = React.createContext<ContextValue | undefined>(undefined)
+ReactionsTransactionsContext.displayName = 'ReactionsTransactionsContext'
+
+export const ReactionsTransactionsProvider: React.FC = ({ children }) => {
+  const [videoReactionProcessing, setVideoReactionProcessing] = useState(false)
```
kdembler: What is this data used for? Let's add a comment please

packages/atlas/src/providers/reactionsTransactions/provider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [processingCommentReactionId, setProcessingCommentReactionId] = useState<string | null>(null)
```
kdembler: What is this data used for? Let's add a comment please

packages/atlas/src/providers/reactionsTransactions/provider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [originalComment, setOriginalComment] = useState<CommentFieldsFragment | null>(null)
```
kdembler: What is this data used for? Seems like this could be kept locally in the `Comment` component?

packages/atlas/src/providers/reactionsTransactions/provider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [highlightedComment, setHighlightedComment] = useState<string | null>(null)
```
kdembler: I think this should still be kept at `CommentsSection` level, it doesn't make much sense for this to be global. Every comment/input can have some event that is emitted when highlight should be changed. Example how `CommentsSection` could look like:
```tsx
const [highlightedComment, setHighlightedComment] = useState<string | null>(null)

// timeout for highlight
useEffect(/* ... */)

return (
  <div>
    <CommentInput onCreated={comment => setHighlightedComment(comment.id)} />
    {comment.map(comment => <Comment {...comment} onEdited={() => setHighlightedComment(comment.id)} />)}
  </div>
)
```

packages/atlas/src/providers/reactionsTransactions/provider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [showEditHistory, setShowEditHistory] = useState(false)
```
kdembler: Seems like this could be kept locally in the `Comment` component?

packages/atlas/src/providers/reactionsTransactions/provider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [commentInputIsProcessingCollection, setCommentInputIsProcessingCollection] = useState(new Set<string>())
```
kdembler: What's a reason to keep this global collection instead of managing this locally in `Comment` state?

packages/atlas/src/providers/reactionsTransactions/provider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [isEditingCommentCollection, setIsEditingCommentCollection] = useState(new Set<string>())
```
kdembler: What's a reason to keep this global collection instead of managing this locally in `Comment` state?

kdembler (changes requested):

packages/atlas/src/providers/reactionsTransactions/provider.tsx
eldiegod: I think is for showing a like or dislike is processing

packages/atlas/src/providers/reactionsTransactions/provider.tsx
eldiegod: `processingCommentReactionId` I think this shows that an emoji reaction is processing? cc @drillprop

kdembler: If that's the case then I guess this data should live on `Comment` level as well?

drillprop (changes requested):
Didn't look at the code, but noticed several issues.
1. Sorting is broken again. I don't know if it's your changes, or maybe Rafał's comment linking PR, but it was fixed recently and it seems that it's broken somehow. 
2. During loading faze I see empty comments: 

https://user-images.githubusercontent.com/51168865/170270532-049204b5-8754-4a71-b11e-c0e07efd782e.mov

3. Reactions are not properly highlighted once you add them:

https://user-images.githubusercontent.com/51168865/170271368-31bde23d-1486-4667-a87c-df85d374c0ef.mov

4. If you reply to a comment input is not cleared
5. If you edit your reply, comment input is not intended. 
6. Needs rebase

rafalpawlow:
@drillprop sorting is working on `reactions` branch so it must be something related to this PR

eldiegod:
> * If you reply to a comment input is not cleared
> 
> * If you edit your reply, comment input is not intended.

@drillprop  I think everything should be fixed, there is the issue with replies not returning an `Id` but I left it just as it was before

drillprop (changes requested):
You didn'f fix: 
> - If you edit your reply, comment input is not intended.

Just to clarify - this is how it worked on the `reactions` branch. Look at the position of the comment input: 

https://user-images.githubusercontent.com/51168865/170318897-fd6fcd91-4d48-4082-85c3-fc5fb34c1c97.mov

Yours:

https://user-images.githubusercontent.com/51168865/170319023-85fc708f-1f05-4939-9fe9-31c0a77bdfba.mov

eldiegod:
Not sure how you are getting that for me is indented locally and on the vercel deploy

imagen

drillprop:
> Not sure how you are getting that for me is indented locally and on the vercel deploy
> 
> imagen

I can reproduce this 100% on vercel deploy. 
Did you **edit your reply**? 
Everything is fine when you **reply to a comment**. The problem occurs **only** when you edit a reply. Check my videos again.

eldiegod:
> > Not sure how you are getting that for me is indented locally and on the vercel deploy
> > imagen
> 
> I can reproduce this 100% on vercel deploy. Did you **edit your reply**? Everything is fine when you **reply to a comment**. The problem occurs **only** when you edit a reply. Check my videos again.

Good catch, I didn't take into account the reply edits

rafalpawlow (changes requested):
I can still see empty comments while loading

https://user-images.githubusercontent.com/13706246/170451054-bc4fe788-9a9f-4076-aa02-72e5ce139376.mov

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -1,336 +1,305 @@
...
-  repliesLoading?: boolean
-  repliesCount?: number
-} & CommentRowProps
+  commentId?: string
+  video?: VideoFieldsFragment | null
+  isReplyable?: boolean
+  setOriginalComment?: React.Dispatch<React.SetStateAction<CommentFieldsFragment | null>>
+  setShowEditHistory?: React.Dispatch<React.SetStateAction<boolean>>
+  setHighlightedCommentId?: React.Dispatch<React.SetStateAction<string | null>>
+  setRepliesOpen?: React.Dispatch<React.SetStateAction<boolean>>
```
kdembler: What do we use these props for? I guess you still keep some state outside of `Comment`? Could you give me an idea of why we need that?

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -1,336 +1,305 @@
...
+    const replyAvatars = comment?.replies?.map((comment) => ({
+      url:
+        comment?.author.metadata.avatar?.__typename === 'AvatarUri' ? comment?.author.metadata.avatar?.avatarUri : '',
+      handle: comment.author.handle,
+    }))
+
+    const loading = loadingQuery || !commentId
+
+    const userReactions = comment?.reactions
+      .filter((reaction) => reaction.memberId === activeMemberId)
```
kdembler: The query should apply this filter already, do we need to `.filter` here as well?

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -1,336 +1,305 @@
...
+            reactions={
+              comment &&
+              getCommentReactions({
+                commentId: comment?.id,
+                userReactions,
+                reactionsCount: comment?.reactionsCountByReactionId,
+                activeMemberId,
+                processingCommentReactionId,
+              })
+            }
```
kdembler: Could we move this to a variable outside of `return` for more clarity in JSX?

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -1,336 +1,305 @@
...
+            memberUrl={comment ? absoluteRoutes.viewer.member(comment.author.handle) : undefined}
+            memberAvatarUrl={
+              comment?.author.metadata.avatar?.__typename === 'AvatarUri'
+                ? comment.author.metadata.avatar?.avatarUri
+                : undefined
+            }
```
kdembler: This should use `useMemberAvatar` for the URL

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -1,336 +1,305 @@
...
+            type={
+              comment && ['DELETED', 'MODERATED'].includes(comment.status)
+                ? 'deleted'
+                : comment &&
+                  (video?.channel.ownerMember?.id === activeMemberId || comment?.author.id === activeMemberId)
+                ? 'options'
+                : 'default'
+            }
```
kdembler: Could we move this to a variable outside of `return` for more clarity in JSX?

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -1,336 +1,305 @@
...
+            onEditClick={() => {
+              if (comment) {
+                setIsEditingComment(true)
+                setEditCommentInputText?.(comment.text)
+              }
+            }}
```
kdembler: Could we move this to a variable outside of `return` for more clarity in JSX?

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -1,336 +1,305 @@
...
+            onDeleteClick={() => video && comment && hadleDeleteComment(comment)}
+            onEditedLabelClick={() => {
+              setShowEditHistory?.(true)
+              comment && setOriginalComment?.(comment)
+            }}
```
kdembler: Could we move this to a variable outside of `return` for more clarity in JSX?

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -434,23 +178,37 @@ export const CommentsSection: React.FC<CommentsSectionProps> = ({ disabled, vide
...
+                highlightedCommentId={highlightedCommentId}
+                indented={!!comment.parentCommentId && comment.id === commentIdQueryParam}
+                setOriginalComment={setOriginalComment}
+                setHighlightedCommentId={setHighlightedCommentId}
+                setShowEditHistory={setShowEditHistory}
+              />
+            ))}
       </CommentWrapper>
       <DialogModal
         size="medium"
```
kdembler: Couldn't this be a part of `Comment` as well? Then we can avoid passing some state around

kdembler (changes requested):
Couple mostly cosmetic comments from me. There are still some issues reported by guys in testing. I think this PR is a very good improvement, I love all the global state deleted, this will likely save us a couple of problems.

Wild idea: what if we got rid of `CommentThread` altogether and made that part of `Comment` as well? Not sure if this is a good idea, let me know what you think

packages/atlas/src/components/_comments/Comment/Comment.tsx
eldiegod: yes, so: 

- setRepliesOpen -> tells the commentThread to show or hide the comment replies
- setHighlightedCommentId -> self explanatory what it does but this state is used in the main commentInput handlers to highlight new comments and Comment's component also use it for new comment replies
~~- setShowEditHistory -> this pretty much toggles the dialog that shows the edit history of a comment~~
~~- setOriginalComment -> also has to do with setting up data for the edit history dialog~~

added the last 2 to Comment

packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
eldiegod: actually yes

packages/atlas/src/components/_comments/Comment/Comment.tsx
eldiegod: yep

drillprop (changes requested):
1. Reactions are still not working correctly. If I go to the video page I don't see my reactions. Only when I click one of them, and the processing is done, I see which one was clicked by me.
2. We're sending too many requests to the Orion. When I go to `video/1` I'm sending  50 requests just to get all the comments. When I react to the comment I'm sending 120 requests more 🤯 . For comparison, previously we sent only two requests to get all the comment data. It looks like something is not working with the cache in this case. I did some investigation, but I didn't find an answer, unfortunately. Maybe you'll find a solution for the problem, Im also fine if we do this as follow up to not extend this PR more.

kdembler:
> Maybe you'll find a solution for the problem, Im also fine if we do this as follow up to not extend this PR more.

I would suggest to go that way. Let's make sure everything works as expected functionality-wise and we will improve network in the next PR

eldiegod:
> I would suggest to go that way. Let's make sure everything works as expected functionality-wise and we will improve network in the next PR

agree I'm checking the reactions issue and will make a follow up for the requests point

eldiegod:
> Reactions are still not working correctly. If I go to the video page I don't see my reactions. Only when I click one of them, and the processing is done, I see which one was clicked by me.

Should be fixed now

drillprop (changes requested):
I still don't see my reactions:
This is what I see  on the `reactions` branch(and what I should see): 
image
On your branch no reactions are active: 
image

I'm almost sure that it worked before in your PR, so check out your commits, maybe you have broken this when you tried to fix something else.

eldiegod:
it works now but there is a cache problem where the `useComment` doesn't find the reactions that `commentSection` just fetched, so for now, there is new requests per comment

eldiegod:
@drillprop I think the cache is working again with `useComment` reactions, maybe even the request problem got fixed along the way

kdembler (approved):

PR #2730, opened 2022-05-20
initial playlists integration

kdembler:
When reviewing this, one piece of information is really important:

> Both videos and playlists are saved on-chain as a single entity (currently called "Video", will be renamed to "Content" later).
They are differentiated at the Query Node level using respective protobuf messages.

So playlists will be also created with `createVideo` extrinsic, use `VideoCreationParameters`, etc. Chain doesn't understand that content anyway, everything is passed via metadata with protobufs.

~Waiting for some things to be fixed in the QN, will update the PR then~


packages/atlas/package.json
```diff
@@ -32,7 +32,7 @@
     "@emotion/react": "^11.7.1",
     "@emotion/styled": "^11.6.0",
     "@hookform/resolvers": "^2.8.8",
-    "@joystream/metadata-protobuf": "2.1.0",
+    "@joystream/metadata-protobuf": "../../../monorepo/metadata-protobuf",
```
kdembler: This needs to be replaced with patched package

drillprop (approved):

PR #2734, opened 2022-05-23
Disable comments integration

### LINKED ISSUE #2532: Integrate enabling / disabling comments with QN
AC

- [ ] connect enabling comments to video form
- [ ] make sure that disabling the comments section does not allow for posting new comments under this video and hides any existing comments. [figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411%3A219280](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411%3A219280)

drillprop:
Fix #2532

~For now you can only enable/disable comments when editing. We cannot do this when a user creates a new video, because to disable comments we need the id of that video(and we don't have the id when the video is not uploaded yet).~


kdembler:
I think Zeeshan will be able to do this change to `VideoMetadata` that we discussed today so maybe worth just putting this on pause?

drillprop:
> I think Zeeshan will be able to do this change to `VideoMetadata` that we discussed today so maybe worth just putting this on pause?

Ok, I set this as a draft for now.

rafalpawlow (approved):

kdembler (approved):

PR #2746, opened 2022-05-25
Comments infinite scroll

### LINKED ISSUE #2454: Infinite Scrolling - Lazy Load
Mechanics of loading additional comments is infinite scroll [Figma>](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2282%3A163649)

rafalpawlow:
Closes #2454

---

You can test it on `/video/2`

### DIFF
--- a/packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
+++ b/packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
@@ -1,4 +1,6 @@
-import React, { useEffect, useRef, useState } from 'react'
+import { NetworkStatus } from '@apollo/client'
+import { debounce } from 'lodash-es'
+import React, { useEffect, useMemo, useRef, useState } from 'react'
 import { useParams } from 'react-router-dom'
 
 import { useComment, useCommentSectionComments } from '@/api/hooks'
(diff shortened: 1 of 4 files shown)


packages/atlas/src/views/viewer/VideoView/CommentsSection.tsx
```diff
@@ -110,8 +119,32 @@ export const CommentsSection: React.FC<CommentsSectionProps> = ({ disabled, vide
     return () => clearTimeout(timeout)
   })
 
+  // increase number of comments when user scrolls to the end of page
```
drillprop, line 102: Unfortunately, this won't work well on smaller viewports, the comment section isn't last section on mobile: 
image
We would have to scroll down to CTA buttons and then scroll up to see new comments.

drillprop (changes requested):
This could be tricky to do, but you didn't handle infinite scrolling for replies. There could be a comment that has 200 replies, we don't want to fetch them all at once.

rafalpawlow:
@drillprop replies fetching will be resolved [here](https://github.com/Joystream/atlas/issues/2757)

drillprop (changes requested):
We had a short call about current issues. Here's what we found:
- When you're on the video page and you switch users, the app breaks
- if you switch to another video and then switch back to the previous video, the number of comments is not correct, and infinite scrolling not working. This is probably cache problem
- The space is awkwardly small below the comments section on mobile
- Sometimes placeholders are blinking(they appear for a split second and disappear and then appear again) when you scroll fast - needs reproduction, I'm not sure how to trigger that. I'm also fine if we ignore this problem or fix it as a followup

drillprop (approved):

Issue #2709, opened 2022-05-17
Update the Textarea component


toiletgranny:
The goal of this task is to update the Textarea component based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7088%3A144013). 

![image](https://user-images.githubusercontent.com/35307309/168807283-e67fbf7b-9fd9-47f0-9064-69ff23f06a0f.png)

PR #2750, opened 2022-05-26
Form field

### LINKED ISSUE #2715: (Re)create the Form field component
- [ ] Make all the updates to the `FormField` component
- [ ] Go through the app and move all the `helperText`s from inputs to `FormField`

---

The goal of this task is to create the Form field component based on the designs and docs, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7056%3A167547). 

> **Changelog (from Figma)**
> _There's no Changelog because this component did not exist previously in the components library in Figma._

image

drillprop:
Fix #2715 
Designs: 
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7056%3A167548
VideoWorkspace
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=3232%3A347471
NFT form
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=3228%3A874494


rafalpawlow (changes requested):
Story for HelperText seems to be broken

drillprop:
> Story for HelperText seems to be broken

I decided to remove it since it's not really needed after my changes.

rafalpawlow (changes requested):
1. Something is wrong with spacing between title and description. In Figma it's 4px but in your code description has 8px padding
Zrzut ekranu 2022-05-30 o 16 09 05


2. Also FormField should have tooltip but I cannot see it in storybook, did I miss something?

drillprop:
> 1. Something is wrong with spacing between title and description. In Figma it's 4px but in your code description has 8px padding
>    Zrzut ekranu 2022-05-30 o 16 09 05
> 2. Also FormField should have tooltip but I cannot see it in storybook, did I miss something?

Both should be fixed.

toiletgranny (changes requested):
Bartek, this is great, as always! Below you'll find mostly questions and ideas you might want to consider. 

1. Let's replace the `(Optional)` label with `(optional)`.
2. Is there any specific reason behind keeping the form field label as `<h6>`? I'm afraid it will not match the on-page heading hierarchy most of the time. Wouldn't it be safer to make it a`<span>`?
3. Was setting the width of the form field description to `85%` intentional? Not that it's bad, I think I like it...
4. What's with the `dense` property? What does it do? Is it a thing of the past, that we still need to keep around for backward compatibility reasons? Asking as it wasn't mentioned anywhere in the designs.
5. `disableErrorAnimation` That's another interesting property. Is it because of accessibility reasons and the `prefers-reduced-motion: reduce` media query? If that's it — wow, that's a nice touch!
6. I'm a bit concerned about setting up `max-height: 24px` on the header-level. I got a feeling it's because of the Information component and the fact it's `32x32px`, right? If this is it, can we maybe wrap the Information component alone in a separate div and handle this over there? My concern is that if we ever change the line-height of the `h300` text style, we would certainly want the header to be of a different height, right? But with the current implementation — it's not going to happen. Another reason to consider wrapping the Information component in a separate div is that now the `4px` left-margin is calculated against the `32x32px` size of the Information component, and not the `24x24px` wrapper it's put within in Figma.
7. Would it make sense or be feasible to wrap child nodes of the FormField component inside a flexbox/grid div, so that we can rely on the `gap` property, rather than `margin-bottom` and `margin-top`? The latter approach, which is used currently, has some well-known drawbacks. For example, if we use the FormField component with a description and without a label (I can see this happening in the future!), we're ending up having an unwanted `4px` of the `margin-top` applied on the FormField description. 
8. Finally, for the switchable variant: I'm not able to focus the Switch component with my keyboard, which ultimately leads to not being able to enable the form field. This could exclude a person like Maria, living with multiple sclerosis, from our community, by not letting her complete the NFT sale form. Is there anything we can do about it on the FormField component level? Or is this something we should tweak on the Switch component level? (cc @rafalpawlow, @kdembler)

drillprop:
> Bartek, this is great, as always! Below you'll find mostly questions and ideas you might want to consider.
> 
> 1. Let's replace the `(Optional)` label with `(optional)`.
> 2. Is there any specific reason behind keeping the form field label as `<h6>`? I'm afraid it will not match the on-page heading hierarchy most of the time. Wouldn't it be safer to make it a`<span>`?
> 3. Was setting the width of the form field description to `85%` intentional? Not that it's bad, I think I like it...
> 4. What's with the `dense` property? What does it do? Is it a thing of the past, that we still need to keep around for backward compatibility reasons? Asking as it wasn't mentioned anywhere in the designs.
> 5. `disableErrorAnimation` That's another interesting property. Is it because of accessibility reasons and the `prefers-reduced-motion: reduce` media query? If that's it — wow, that's a nice touch!
> 6. I'm a bit concerned about setting up `max-height: 24px` on the header-level. I got a feeling it's because of the Information component and the fact it's `32x32px`, right? If this is it, can we maybe wrap the Information component alone in a separate div and handle this over there? My concern is that if we ever change the line-height of the `h300` text style, we would certainly want the header to be of a different height, right? But with the current implementation — it's not going to happen. Another reason to consider wrapping the Information component in a separate div is that now the `4px` left-margin is calculated against the `32x32px` size of the Information component, and not the `24x24px` wrapper it's put within in Figma.
> 7. Would it make sense or be feasible to wrap child nodes of the FormField component inside a flexbox/grid div, so that we can rely on the `gap` property, rather than `margin-bottom` and `margin-top`? The latter approach, which is used currently, has some well-known drawbacks. For example, if we use the FormField component with a description and without a label (I can see this happening in the future!), we're ending up having an unwanted `4px` of the `margin-top` applied on the FormField description.
> 8. Finally, for the switchable variant: I'm not able to focus the Switch component with my keyboard, which ultimately leads to not being able to enable the form field. This could exclude a person like Maria, living with multiple sclerosis, from our community, by not letting her complete the NFT sale form. Is there anything we can do about it on the FormField component level? Or is this something we should tweak on the Switch component level? (cc @rafalpawlow, @kdembler)

1. fixed
2. fixed
3. I don't remember why we set this in the first place. This is a very old code. I also like it, and I would leave it as it is. 
4. I would leave `dense` as well. If you set `dense` to true the top margin will be 24px(not 32px). This could be useful for dialogs: 
image
5. No, unfortunately, this is not because of accessibility reasons :P I've noticed one problem with using the form field with auction date picker, the animation is really glitchy when you select the wrong date. That's why I disabled the error animation just for this input. I guess we could remove the property once we deal with the problem, I don't want to focus on finding the solution for now.   
Edit: 
This is how it works **without** `disableErrorAnimation`


https://user-images.githubusercontent.com/51168865/171601402-139c99fd-334c-4d3a-8dd8-5289e1051c1f.mov


7. fixed
8. fixed
9. fixed

toiletgranny (changes requested):
1. ✅
2. I'm afraid it's not fixed everywhere. Not in the tooltip variant for example. Maybe worth double-checking?
3. ✅
4. Ahh, I see. I'm sensing it might cause us some troubles in the future (like the additional unwanted margin, we see that all the time, right?), and I think for the Dialog example you brought up I would maybe just try using flexbox or grid. But at the end of the day, you guys are the experts here and I trust you!
5. ✅ Ah, I see! Sure, I understand. Thanks for taking the time and bringing this up! ❤️
6. ✅
7. I can see you changed that to grid, but I'm afraid it didn't solve the initial problem. The gap is still present. Is there anything else we can do about it?
CleanShot 2022-06-02 at 16 21 33@2x
8. ✅ Works like a charm. Thank you! ❤️

drillprop:
> 1. ✅
> 2. I'm afraid it's not fixed everywhere. Not in the tooltip variant for example. Maybe worth double-checking?
> 3. ✅
> 4. Ahh, I see. I'm sensing it might cause us some troubles in the future (like the additional unwanted margin, we see that all the time, right?), and I think for the Dialog example you brought up I would maybe just try using flexbox or grid. But at the end of the day, you guys are the experts here and I trust you!
> 5. ✅ Ah, I see! Sure, I understand. Thanks for taking the time and bringing this up! ❤️
> 6. ✅
> 7. I can see you changed that to grid, but I'm afraid it didn't solve the initial problem. The gap is still present. Is there anything else we can do about it?
> 
> CleanShot 2022-06-02 at 16 21 33@2x
> 
> 8. ✅ Works like a charm. Thank you! ❤️

2. Should be fixed.
4. Ok, I removed the margin from FormField and used grid property with a gap in forms instead. No more `dense` property. I hope I didn't break anything 😅 
7. Should be fixed as well.

toiletgranny (approved):
Perfect, everything works! And it looks so simple now! ❤️

packages/atlas/src/components/_inputs/TextArea/TextArea.tsx
```diff
@@ -39,10 +53,46 @@ const TextAreaComponent: React.ForwardRefRenderFunction<HTMLTextAreaElement, Tex
...
+}
+
+export const CharacterCounter: React.FC<CharacterCounterProps> = ({ maxLength, charactersCount }) => {
+  const [charactersWarning, setCharactersWarning] = useState<'warning' | 'error' | null>(null)
+
+  useEffect(() => {
+    if (!maxLength || !charactersCount) {
+      return
+    }
+    const warningLength = maxLength * 0.8
```
rafalpawlow, line 77: What's `0.8`? Let's move it to const

packages/atlas/src/components/_overlays/AdminOverlay/AdminOverlay.styles.ts
```diff
@@ -38,10 +38,10 @@ export const HorizontalSpacedContainer = styled.div`
   }
 `
 
-export const VerticalSpacedContainer = styled.div<{ topMargin?: boolean }>`
+export const VerticalSpacedContainer = styled.div`
   display: flex;
   flex-direction: column;
-  margin-top: ${({ topMargin }) => (topMargin ? '16px' : 0)};
+  margin-top: 16px;
```
rafalpawlow: Use `sizes`

rafalpawlow (changes requested):

rafalpawlow (changes requested):
In switchable variant there are wrong spacings, between switch and label and between description

packages/atlas/src/components/_inputs/TextArea/TextArea.tsx
drillprop: if I'm not mistaken, this is some old code copied from the removed helper text. I just wanted not to break the existing TextArea component. I won't be updating this for now, let's change this in #2709.

drillprop:
> In switchable variant there are wrong spacings, between switch and label and between description

fixed

rafalpawlow (approved):

Issue #2738, opened 2022-05-24
Update the video upload form


toiletgranny:
The goal of this task is to update the entire Video upload form in the Video workspace view so that it's aligned with the new form guidelines.

toiletgranny:
Ok, @kdembler, it's ready to go! The scope might seem large, and it surely got me busy for a while (since updating components in Figma is not as fast as in code), but provided we took care of all form component changes, plus the additional two I'm listing below, prior to opening this task, I would hope for this to be a straightforward task. 

The two additional components that require updating before jumping over to the changes from this ticket are:
- https://github.com/Joystream/atlas/issues/2775
- https://github.com/Joystream/atlas/issues/2776

>  **Changelog (from Figma)**:
> - Recreated video upload [(link)](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/?node-id=308%3A5753) and NFT listing [(link)](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/?node-id=916%3A21173) forms from scratch using the new form components and the new form design guidelines,
> - Recreated user stories for uploading and editing videos [(link)](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/?node-id=2365%3A86185), which includes changes such as:
>       - Visualizations of the form validation,
>       - Rearrangement of a few form fields (specifically around NFT fields),
>       - Mentioning (in the user stories) the new flow of updating video thumbnail in the edit mode.
> - Updated RWD for both the Video workspace and the NFT workspace [(link)](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/?node-id=313%3A4121),
> - Updated the yet-to-be implemented user stories for playlists [(link)](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/?node-id=2365%3A86186) and subtitles [(link)](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/?node-id=2653%3A124732),
> - (Figma-only) Updated user stories for NFTs to capture last-minute changes introduced in the implementation, before the NFT release 
> - (Figma-only) Recreated and simplified layers structure in both the Video & NFT workspaces,
> - (Figma-only) Ditched most local components, as their functionalities are now part of the updated global form components,
> - (Figma-only) Moved the Combo box component to the global web components library in Figma.

![image](https://user-images.githubusercontent.com/35307309/171570135-f4619de3-9650-4290-bbf4-2a63c8b53508.png)

PR #2751, opened 2022-05-26
Rework video form title input

### LINKED ISSUE #2692: Rework video form title input and validation
The title input on video form really blends in with the background and may be easily missed by the user. When you also take into account that validation on this page doesn't let you see any error (the submit button is disabled), the experience can be really confusing. This has been reported by a number of different users (both jsgenesis and community), we should really make some quick fix there to improve this flow

rafalpawlow:
Closes #2692 

---

- [x]  Update the Title input (formerly: Title area) component based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7666%3A157393). The key changes which aim on solving the problem are increasing the brightness of the placeholder text and introducing the Error styling variant, which should be used to direct the user's attention towards this field (if left blank upon form submit attempt).
- [x] Change the Title input component placeholder to "Enter video title" (not on the component level, only in the video upload form).
- [x] Prepopulate the video title with the uploaded video file name while using the following parsing rules:
   - Drop the file extension from the end of a string if matching any of the following: `.webm|.mpg|.mp2|.mpeg|.mpe|.mpv|.ogg|.mp4|.m4p|.m4v|.avi|.wmv|.mov|.qt|.flv|.swf|.avchd|.mkv`,
   - Replace all underscores with spaces,
   - Capitalize the first character of a string.
- [x] Make sure the above-mentioned parsing only takes place once: when fetching the video file name.
- [x] Change the validation strategy to allow submitting the video upload form before filling it out accordingly to the requirements.


kdembler:
@toiletgranny You should take a look at this. I think we certainly need to introduce some kind of message when the assets are not provided. As it is, if I input video, title, etc, but no thumbnail, the submit button will stay enabled but nothing happens when I click it

toiletgranny:
Oh, of course. I have a solution for this in https://github.com/Joystream/atlas/issues/2738, but I forgot we also need to face this problem over here. So my suggestion is to use the Banner component with a custom colored icon and text inside. It either says "Select a video file" or "Select an image file", depending on which step of interacting with the File select component you're on. [Here it is in Figma](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/%F0%9F%9A%A7-Video-workspace?node-id=3145%3A158304) (keep in mind I'm still working on this file).

image

---

@rafalpawlow The implementation is great! There are a few minor things we could improve in regards to the Title input component:
1. It seems like the on-hover tooltip is missing. It should say "Click to edit".
2. Is it an option to get rid of the extra padding around the `textarea` element? 
3. I can see there's this `height: 60px !important;` CSS property that's being added automatically. Not sure where is this value coming from, but if it was supposed to be the line-height of the text style we're using in this component, then I think it should be `56px`.

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -416,18 +444,31 @@ export const VideoForm: React.FC<VideoFormProps> = React.memo(({ onSubmit, setFo
...
+              error={!!errors.title}
+              onFocus={() => setTitleTooltipVisible(false)}
+              onBlur={() => setTitleTooltipVisible(true)}
+              helperText={errors.title && errors.title.message}
             />
           )}
         />
+        {titleTooltipVisible && (
+          <Tooltip text="Click to edit" arrowDisabled placement="top-start" reference={titleInputRef} />
+        )}
```
eldiegod, line 473: Why not just wrap `StyledTitleArea` with the tooltip component?

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -63,6 +66,8 @@ import { StyledSvgWarning, YellowText } from '../VideoWorkspace.style'
 const CUSTOM_LICENSE_CODE = 1000
 const SCROLL_TIMEOUT = 700
 const MINT_NFT_TIMEOUT = 1200
+const MIN_TITLE_LENGTH = 5
```
eldiegod: Wasn't min length 5? I think we weren't meant to change this; @toiletgranny ?

eldiegod (changes requested):
Works good - just a couple small things

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
rafalpawlow: Because of the flag `titleTooltipVisible`. When input is focused, tooltip shouldn't be visible. Do you know easier way to achieve this?

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
rafalpawlow: TBH I took it from figma, but you are right we didn't discuss this. @toiletgranny WDYT?

toiletgranny: Oh, please disregard the min length value in the component's designs. Title input can also be used for other things, such as channel names, and those can have unique length limits. So if the min length was previously set to 5, let's make sure it stays this way and is unaffected by the design change.

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
eldiegod: nop

eldiegod (approved):

kdembler (approved):

Issue #2678, opened 2022-05-13
Update guidelines for dates, time and numbers formatting


toiletgranny:
**AC:**

- [ ] Update formatting functions for numbers (`.` as decimal separator and ` ` (space) as thousands separator)
- [ ] Create `Number` component that will be responsible for displaying all numbers in app (similar to `Text`). Example API: `<Number value={nftStatus.topBidAmount} variant="t100" />`. This component should include a tooltip (that can be disabled) with the full amount on hover. All the instances of `formatNumberShort` etc. in the app will need to be replaced with this component.

### Description

At the meeting with Klaudiusz and Dmitry, we agreed to unify the formatting of all numbers, dates, and times in the app globally to match whatever is the currently selected language of the app. For now, it's only English (UK). 

The goal of this task is to update Notion guidelines on this subject accordingly and then pass the ticket over to the tech team so that we can actually make the change happen.

### Figma link

https://www.notion.so/joystream/Date-time-number-formats-in-Atlas-c5e7c7aec24e4ddfa83104ca4c2e924e

### How urgent this is?

This will block me soon

toiletgranny:
@kdembler, I updated the guidelines in Notion as agreed upon the last Friday. I'm prob not fully aware of how big of a time investment it is for you to make sure that all dates, times, and numbers are aligned with the new guidelines in a systemic way (so that when a new language is added to Atlas, then the formatting is updated automatically) nor if you even want to do it right now. Still, if there's anything else I can do to help, I'm happy to do so. 

One thing I was hoping for us to tackle with this adjustment was adding time-delayed tooltips for all abbreviated numbers in Atlas globally. Let me know if this sounds possible and if so, I'd love to jump on a quick call with a developer assigned to this and discuss the details based on what's possible.

https://joystream.notion.site/Date-time-number-formats-in-Atlas-c5e7c7aec24e4ddfa83104ca4c2e924e

PR #2762, opened 2022-05-30
NumberFormat component with tooltip

eldiegod:
fixes: #2678

### DIFF
--- a/packages/atlas/src/components/NumberFormat/NumberFormat.stories.tsx
+++ b/packages/atlas/src/components/NumberFormat/NumberFormat.stories.tsx
@@ -7,17 +7,24 @@ export default {
   title: 'other/NumberFormat',
   component: NumberFormat,
   args: {
-    value: 100000.123456789,
+    value: 102930,
     format: 'dollar',
-    withToken: false,
-    withTooltip: true,
     variant: 't100-strong',
+    withToken: false,
+  },
+  argTypes: {
+    format: {
+      control: { type: 'radio', options: ['dollar', 'full', 'short'] },
+    },
+    withToken: {
+      control: { type: 'boolean' },
+    },
   },
 } as Meta<NumberFormatProps>
 
-const Template: Story<Omit<NumberFormatProps, 'ref'>> = ({ variant = 't100-strong', ...args }) => (
+const Template: Story<Omit<NumberFormatProps, 'ref'>> = (args) => (
   <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-start', gap: '12px', padding: '200px' }}>
-    <NumberFormat {...args} variant={variant} />
+    <NumberFormat {...args} />
   </div>
 )
 
--- a/packages/atlas/src/components/NumberFormat/NumberFormat.stories.tsx
+++ b/packages/atlas/src/components/NumberFormat/NumberFormat.stories.tsx
@@ -13,7 +13,7 @@ export default {
     withTooltip: true,
     variant: 't100-strong',
   },
-} as Meta
+} as Meta<NumberFormatProps>
 
 const Template: Story<Omit<NumberFormatProps, 'ref'>> = ({ variant = 't100-strong', ...args }) => (
   <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-start', gap: '12px', padding: '200px' }}>
--- a/packages/atlas/src/components/NumberFormat/NumberFormat.stories.tsx
+++ b/packages/atlas/src/components/NumberFormat/NumberFormat.stories.tsx
@@ -1,0 +1,24 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+
+import { NumberFormat, NumberFormatProps } from './NumberFormat'
+
+export default {
+  title: 'other/NumberFormat',
+  component: NumberFormat,
+  args: {
+    value: 100000.123456789,
+    format: 'dollar',
+    withToken: false,
+    withTooltip: true,
+    variant: 't100-strong',
+  },
+} as Meta
+
+const Template: Story<Omit<NumberFormatProps, 'ref'>> = ({ variant = 't100-strong', ...args }) => (
+  <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-start', gap: '12px', padding: '200px' }}>
+    <NumberFormat {...args} variant={variant} />
+  </div>
+)
+
+export const Regular = Template.bind({})
--- a/packages/atlas/src/components/NumberFormat/NumberFormat.tsx
+++ b/packages/atlas/src/components/NumberFormat/NumberFormat.tsx
@@ -3,6 +3,7 @@ import React, { useRef } from 'react'
 import mergeRefs from 'react-merge-refs'
 
 import { Text, TextProps, TextVariant } from '@/components/Text'
+import { formatNumber } from '@/utils/number'
 
 import { Tooltip } from '../Tooltip'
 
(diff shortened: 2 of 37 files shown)


packages/atlas/src/components/NumberFormat/NumberFormat.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+  title: 'other/NumberFormat',
+  component: NumberFormat,
+  args: {
+    value: 100000.123456789,
+    format: 'dollar',
+    withToken: false,
+    withTooltip: true,
+    variant: 't100-strong',
+  },
+} as Meta
```
rafalpawlow:
```suggestion
} as Meta<NumberFormatProps>
```

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Text, TextProps, TextVariant } from '@/components/Text'
+
+import { Tooltip } from '../Tooltip'
+
+export type NumberFormatProps = {
+  value: number
+  format?: 'full' | 'short' | 'dollar'
+  withToken?: boolean
+  withTooltip?: boolean
+  children?: never
```
rafalpawlow, line 15: Shouldn't it be `number | string`?

rafalpawlow (changes requested):
1. Tooltip should work on hover, currently it works on click
2. I'm not sure if this is specified somewhere but I noticed that for dollars you show full value without decimals. Can you confirm if this is a mistake?

toiletgranny:
Hey Diego! I was lucky enough to stumble upon this PR.

The implementation now addresses the main problems, which is great, but I have found a few things you might want to know about:
1. The tooltip component displaying full numbers doesn't seem to be the one we have in our component library. For example, the tooltip we know has no tail, it has slightly rounded corners and a subtle shadow underneath it. [Here's the link to the most up-to-date designs for the Tooltip component.](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6215%3A139642) Can we double-check it?
2. Not all numbers seem to have a tooltip attached to them, like in this example (make sure to switch to the Main Testnet), the Buy now price of 2M has no tooltip. https://atlas-o2aayrzjz-joystream.vercel.app/video/26775
3. While, in fact, we abbreviate large numbers over there, since most of the videos on production have under 99 views right now, it's just quite surprising to see a tooltip for a number such as `14`. Can we set up some threshold for displaying tooltips? Eg. for numbers that are above 999?
   - On the other hand, there are numbers such as `$1.26`, which also come with unnecessary tooltips:
   CleanShot 2022-05-30 at 16 46 15
5. I also noticed a few other, smaller issues, such as unexpected changes to number colors, alignment, or line-height. When combined together, they do make a difference. Below is a short video I recorded when comparing the same page between your deployment and production. Capturing all of those little nuances manually might be cumbersome, even if we were to compare views apples to apples, so I'm thinking if we had any chance to play with some automated visual regression tests here in Atlas, in the past? 

https://user-images.githubusercontent.com/35307309/171014519-adabde99-ea6b-4dfa-aafa-eb4275e3d7ae.mp4

Ps. Next time I'd really appreciate it if you could add me as a reviewer! 🙏

rafalpawlow:
@toiletgranny Diego created this branch from `dev` branch where we don't have new tooltips yet. They will be deployed together with reactions https://atlas-storybook-git-reactions-joystream.vercel.app/?path=/story/other-tooltip--default

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
eldiegod: I don't think so, since this is meant to replace the number formatting functions we always pass numbers in there.

eldiegod:
@toiletgranny 

1. as Rafal mentioned above, we don't have the tooltip changes in this branch yet, but we will eventually.
2. There is a tooltip there; it's just an issue with that particular part of the component that gets rebuilt as we poll the token price very often, and that makes the tooltip not show or show very briefly; I will have to find a patch to fix that
3. I would say we will need to define a bit more clearly the rules for showing or not showing the tooltip. Also, for things like decimals, I think we should show all available decimals in the tooltip even if it's `1.26$` like in the example because in the end is money we are talking about; I think it's worth the info for the user and I have a sense this is common in web3 apps in general, wdyt?
4. I will make some fixes for that and maybe that helps with a good amount but no we don't have any sort of testing atm that I'm aware off

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
rafalpawlow: So it should be `children?: number`, right?

eldiegod: there is `value: number` instead

eldiegod:
> I'm not sure if this is specified somewhere but I noticed that for dollars you show full value without decimals. Can you confirm if this is a mistake?
@rafalpawlow 
1. It works on hover you just have to hover it long enough
2. do you have an example for the dollar thing?

rafalpawlow:
@eldiegod 
1. Right, we seem to have agreed on that. I forgot about that
2.
Zrzut ekranu 2022-05-30 o 20 26 53
Zrzut ekranu 2022-05-30 o 20 27 05

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
rafalpawlow: So why do we even declare `children` prop?

eldiegod:
not sure I understood 100% but yes that looks right, we don't 'short' format dollar amounts, at least until now

rafalpawlow:
That's what I meant 👍

rafalpawlow (changes requested):
One more thing, when you type in storybook `100.5`, you get `100`. I think this is not correct

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
eldiegod: because we expand on `TextProps` and we expect `children` there but I overwrite it so it's not passed by mistake.

rafalpawlow (changes requested):
Looks good, two small things
1. Do we need tooltip for full variant?
Zrzut ekranu 2022-06-3 o 10 46 47

2. Can we improve storybook a little bit? I need to write variant manually, radio would be more convenient
Zrzut ekranu 2022-06-3 o 10 48 52

toiletgranny (changes requested):
As per your ask, Diego, I won't be looking at the number tooltip behavior within the NFT widget. And as such, things look much better already! ❤️ The two things I have are:
1. In the Video tile component, the number of views should be kept together with the `views` label, when it comes to line breaks. After the number change, it's no longer happening. Can we fix that?
CleanShot 2022-06-03 at 12 36 25
2. In regards to in which case scenarios should we be displaying tooltips, I'd be still advocating for doing it only with numbers that are either rounded or have decimal places.
   - In terms of how I, a simple & non-technical person, imagine this could be done, using conditional logic, would be something like this: show tooltips only for numbers that are rounded AND (numbers larger than `999` OR numbers that come with 3 or more decimal places).

eldiegod:
> Looks good, two small things
> 
>     1. Do we need tooltip for full variant?
>        Zrzut ekranu 2022-06-3 o 10 46 47
> 
>     2. Can we improve storybook a little bit? I need to write variant manually, radio would be more convenient
>        Zrzut ekranu 2022-06-3 o 10 48 52

@rafalpawlow fixed that and improved storybook

kdembler (changes requested):
It seems we show too many fractional digits here:

Screenshot 2022-06-08 at 15 27 24

toiletgranny (approved):

rafalpawlow (approved):

kdembler (approved):

PR #2766, opened 2022-05-30
Video details playlist button

### LINKED ISSUE #2124: Video tile details improvements
**Changelog:** 
• Added new "Type" property with values: "Video" and "Playlist" so we are able to support Playlist tiles with that components (added tertiary button with "View playlist details" label)

**Figma link:**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3630%3A112904

rafalpawlow:
Closes #2124 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3630%3A112904

### DIFF
--- a/packages/atlas/src/components/_video/VideoTileDetails/VideoTileDetails.tsx
+++ b/packages/atlas/src/components/_video/VideoTileDetails/VideoTileDetails.tsx
@@ -42,6 +42,7 @@ export type VideoTileDetailsProps = {
   kebabMenuItems?: MenuItemProps[]
   variant?: VideoDetailsVariant
   type?: 'video' | 'playlist'
+  playlistUrl?: string
 }
 
 export const VideoTileDetails: React.FC<VideoTileDetailsProps> = ({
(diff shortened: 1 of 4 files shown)


eldiegod (approved):
Approving, button doesn't seem to match the design but it's because the latest button update is not in this branch

packages/atlas/src/components/_video/VideoTileDetails/VideoTileDetails.tsx
```diff
@@ -98,15 +103,23 @@ export const VideoTileDetails: React.FC<VideoTileDetailsProps> = ({
...
+                      {videoSubTitle
+                        ? videoSubTitle
+                        : createdAt && (
+                            <>
+                              {formatVideoDate(createdAt)} • <Views>{formatVideoViews(views || 0)}</Views>
+                            </>
+                          )}
+                    </Text>
+                  ) : (
+                    <PlaylistButton onClick={onPlaylistDetailsClick} variant="tertiary" size="small" textOnly>
```
kdembler: I think this should be a regular link, not a button

toiletgranny (approved):
If my understanding is correct that the scope of this PR is just adding the "View playlist details" button to the `VideoTileDetails` component, then — no doubt about — this has been accomplished. :D

The button looks different, but I'm assuming this is because of the button update not being in this brunch, as @eldiegod is saying.

kdembler:
Just to make sure, the goal wasn't to introduce a button but rather create a `Playlist` variant of the video tile details component. If the button is all that's needed, I'll leave that up you :D

PR #2772, opened 2022-05-31
clean up network comments network requests, paginate replies

kdembler:
This PR improves upon the number of network requests needed when fetching the comments but also done after some action is taken by the user. It also adds support for pagination inside comment replies. I'm happy to answer any questions


packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -44,7 +56,7 @@ export const Comment: React.FC<CommentProps> = React.memo(
 
     const { activeMemberId, activeMembership, activeAccountId, signIn } = useUser()
     const { comment } = useComment(
-      { commentId: commentId ?? '', memberId: activeMemberId ?? undefined, videoId: video?.id },
+      { commentId: commentId ?? '' },
       {
         skip: !commentId,
         fetchPolicy: 'cache-only',
```
drillprop: actually, do we still need this fetch policy? I commented this line and I didn't notice much difference in terms of the number of requests.

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -251,9 +255,8 @@ export const Comment: React.FC<CommentProps> = React.memo(
             onToggleReplies={() => isReplyable && setRepliesOpen?.((value) => !value)}
             repliesOpen={isReplyable && isRepliesOpen}
             onReplyClick={isReplyable ? handleReplyClick : undefined}
-            replyAvatars={replyAvatars}
             repliesCount={comment?.repliesCount}
-            repliesLoading={!!comment?.repliesCount && !comment?.replies}
+            replyAvatars={[]}
```
drillprop: shouldn't we remove this prop? Or maybe, you wanted to leave it in case something change in the future? 
The same applies to `repliesLoading` I guess

packages/atlas/.storybook/main.ts
```diff
@@ -21,6 +21,7 @@ const config: CustomizedStorybookConfig = {
     }
     storybookConfig.build.target = ''
     const viteConfig = _viteConfig as UserConfig
+    // @ts-ignore complains about flat
```
drillprop, line 24: oh, really? I don't see any error

packages/atlas/src/components/_comments/Comment/InternalComment.tsx
```diff
@@ -54,9 +55,9 @@ export type InternalCommentProps = {
   type: 'default' | 'deleted' | 'options'
   reactions: Omit<ReactionChipProps, 'onReactionClick'>[] | undefined
   reactionPopoverDismissed: boolean | undefined
-  replyAvatars: (AvatarGroupUrlAvatar & { handle: string })[] | undefined
+  replyAvatars?: (AvatarGroupUrlAvatar & { handle: string })[] | undefined
```
drillprop: We could remove `| undefined` here

packages/atlas/src/components/_comments/Comment/InternalComment.tsx
```diff
@@ -54,9 +55,9 @@ export type InternalCommentProps = {
...
   repliesOpen: boolean | undefined
-  repliesLoading: boolean | undefined
+  repliesLoading?: boolean | undefined
```
drillprop: same here

packages/atlas/src/views/viewer/VideoView/CommentThread.tsx
```diff
@@ -31,24 +65,75 @@ export const CommentThread: React.FC<CommentThreadProps> = ({
...
+                commentId={newReply.id}
                 video={video}
                 indented
                 setHighlightedCommentId={setHighlightedCommentId}
+                userReactions={userReactionsLookup ? userReactionsLookup[newReply.id] : undefined}
                 isReplyable={false}
               />
-            )))}
+            ) : (
+              <Comment indented />
```
drillprop: Hmm, maybe it's just me, but this was a little confusing for me. I had to check why we have a blank Comment component without all the props. Can we add a `loading` prop to Comment, just for the sake of clarity?

drillprop (changes requested):
Some small things.

rafalpawlow (changes requested):
I think we need more space here

Zrzut ekranu 2022-06-1 o 10 58 05

packages/atlas/src/components/_comments/Comment/Comment.tsx
kdembler: There shouldn't be a difference if everything works correctly. The difference would be in the case when data wasn't properly fetched outside of the component. But I'll get rid of it

packages/atlas/src/components/_comments/Comment/Comment.tsx
kdembler: Removed `repleisLoading`, but left `replyAvatars`, maybe in the future QN adds some support for this and we can display the avatars

packages/atlas/.storybook/main.ts
kdembler: My editor is complaining:
Screenshot 2022-06-01 at 11 03 53

packages/atlas/src/views/viewer/VideoView/CommentThread.tsx
kdembler: I'm not sure if `loading` prop makes sense for `Comment` but added a comment (😉) here to explain why we display this

kdembler:
All fixed and rebased

rafalpawlow (changes requested):
I found super weird bug. 
1. find thread without replies
2. start adding replies
3. when you add 3rd comment, 2nd disappears (see screenshot, it says 'Hide 3 replies' but there are only 2),
4. the problem is repeated with each next reply

Zrzut ekranu 2022-06-1 o 11 35 47

As suggested in the slack, we could add the correct number of skeletons when loading the replies since we know their number in advance.

packages/atlas/src/components/_comments/Comment/Comment.tsx
rafalpawlow: Maybe let's add some comment here?

rafalpawlow (changes requested):

kdembler:
Sure, added

rafalpawlow (approved):

Issue #2125, opened 2022-02-15
Video thumbnail improvements to support Playlists.


TCzechowski:
**Changelog:**

• Added new "Type" property
• Added new "Playlist" type to support playlists with this component.

**Figma link:**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=3629%3A119374

*Important:
New playlist thumbnail has different hover animation.

kdembler:
@TCzechowski I think yesterday during sync you guys mentioned there is some additional transition of the `X videos` overlay on hover? Is that a thing or maybe I misunderstood?

kdembler:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Atlas-600edbb8799cc700103bc2d7/issues/Joystream/atlas/2125?planning-poker) with ZenHub @drillprop @eldiegod @lcoenengithub @rafalpawlow

PR #2782, opened 2022-06-02
Playlist thumbnail changes

eldiegod:
Fixes: #2125

### DIFF
--- a/packages/atlas/src/components/_nft/NftTile/NftTile.tsx
+++ b/packages/atlas/src/components/_nft/NftTile/NftTile.tsx
@@ -93,6 +93,7 @@ export const NftTile: React.FC<NftTileProps> = ({
   return (
     <Container fullWidth={fullWidth}>
       <VideoThumbnail
+        type="video"
         videoHref={thumbnail?.videoHref}
         onMouseEnter={() => setHovered(true)}
         onMouseLeave={() => setHovered(false)}
--- a/packages/atlas/src/components/_nft/NftTileViewer/NftTileViewer.tsx
+++ b/packages/atlas/src/components/_nft/NftTileViewer/NftTileViewer.tsx
@@ -60,6 +60,7 @@ export const NftTileViewer: React.FC<NftTileViewerProps> = ({ nftId }) => {
       thumbnailUrl: thumbnail.url,
       loading: thumbnail.isLoadingAsset,
       thumbnailAlt: `${nft?.video?.title} video thumbnail`,
+      type: 'video',
     },
     owner: nft?.ownerMember?.id
       ? {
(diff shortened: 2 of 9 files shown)


rafalpawlow (changes requested):
When I hover playlist thumbnail, there is wrong overlay
Zrzut ekranu 2022-06-2 o 12 23 40
Zrzut ekranu 2022-06-2 o 12 23 44

Also you have eslint error

eldiegod:
> When I hover playlist thumbnail, there is wrong overlay Zrzut ekranu 2022-06-2 o 12 23 40 Zrzut ekranu 2022-06-2 o 12 23 44
> 
> Also you have eslint error

@rafalpawlow Not sure how you got to that hover state, how do I reproduce? but maybe is not there anymore

rafalpawlow:
@eldiegod it's still there
1. go to storybook
2. switch type to `playlist`
3. hover video thumbnail

Also, shouldn't we add type prop to the VideoTile itself?

rafalpawlow (changes requested):
I think this slot shouldn't be under overlay
Zrzut ekranu 2022-06-3 o 11 18 16

toiletgranny (changes requested):
This looks so slick! ❤️

1. The hover animation appears to be a bit jagged. If you set the animation speed to 10% in Chrome dev tools, you can observe that the thumbnail jumps up by a few pixels imidietaly after hovering. As a result, the actual transition animation starts from there. Can we fix that? Let me know if you prefer to discuss this over a Slack call.
2. I love how you approached the blue cascading sheets animation. A bit differently than what we did in Figma, but heck yeah — I like your version a lot better. I would suggest, however, dropping the opacity animation (or, in other words, setting the background in the initial state, rather than on hover). Also, maybe we could do the same thing to the width (example: `    width: calc(100% - 48px)`): instead of altering it on the hover state, let's have it applied right from the start.
3. There's something unpleasant about how the initial thumbnail content morphs into the full-covered "Play all videos" state, would you agree? Could we try leaving that part without the animation, just to see if it's any better? Feel free to ping me on Slack when you have it ready, I don't think we need to wait for another PR to evaluate this together.
4. When pressing the thumbnail, the very bottom blue sheet doesn't hide. Can we fix that?

### Semi-related
1. Not sure why haven't we noticed that earlier, but all `Thumbnail` components in Storybook come with the wrong variants of `Pill` components. Instead of being of the Medium size (24px of height) and "Overlay" background, they're used in their "Small" sizes (20px of height), and with the opaque background. Weirdly enough, `Pill` components used in the `VideoTile` are correct. Should we worry about that? Would that be too much of a hustle to fix that in the representation of the `Thumbnail` component?
3. One clever thing we did recently with @drillprop to improve accessibility in Atlas by 1%, was applying the same staying we were using in its `"hover` state, to the `:focus-visible` state of the Switch component. Do you reckon we could do the same with Thumbnail components? For both playlists and the regular types?

toiletgranny (approved):

rafalpawlow (commented):
Everything looks good now, but I would like to check it one more after rebase

rafalpawlow (approved):

PR #2786, opened 2022-06-02
Update Radio component

### LINKED ISSUE #2710: Update the Radio button component
**AC:**
- [ ] Implement all the updates
- [ ] Create the `RadioButtonGroup` component that can manage state of multiple radio buttons and return a single array of values. `<RadioButtonGroup value={categoriesFilter} options={[]} onChange={(values) => void} />`

---

The goal of this task is to update the Radio button component based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7384%3A142876). 

>  **Changelog (from Figma)**
> - Updated fill, border and text colors to use color design tokens,
> - Updated transition animations to use animation design tokens,
> - Added “Radio button group” component for easier and more consistent vertical stacking,
> - Added documentation along with content/usage guidelines,
> - Added Focus state, which inherits its styling from the Hover state.

image

rafalpawlow:
Closes #2710 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7384%3A142876

### DIFF
--- a/packages/atlas/src/components/_inputs/RadioButton/RadioButton.stories.tsx
+++ b/packages/atlas/src/components/_inputs/RadioButton/RadioButton.stories.tsx
@@ -1,9 +1,7 @@
 import { Meta, Story } from '@storybook/react'
-import React, { useState } from 'react'
+import React from 'react'
 
-import { RadioButtonGroup } from '@/components/_inputs/RadioButtonGroup'
-
-import { RadioButton } from './RadioButton'
+import { RadioButton, RadioButtonProps } from './RadioButton'
 
 export default {
   title: 'inputs/RadioButton',
@@ -11,45 +9,19 @@ export default {
   argTypes: {
     className: { table: { disable: true } },
     selectedValue: { table: { disable: true } },
+    selected: { controls: { type: 'boolean' } },
   },
   args: {
     label: 'Hello there',
     helperText: 't100',
     disabled: false,
     error: false,
+    selected: false,
   },
-} as Meta
+} as Meta<RadioButtonProps>
 
 const Template: Story = (args) => {
-  const [selected, setSelected] = useState<string | number>('1')
-  const handleChange: (e: React.ChangeEvent<HTMLInputElement>) => void = (e) => {
-    const element = e.currentTarget
-    setSelected(element.value)
-  }
-
-  const ITEMS = [
-    {
-      label: args.label,
-      helperText: args.helperText,
-      value: '1',
-    },
-    {
-      label: args.label,
-      helperText: args.helperText,
-      value: '2',
-    },
-    {
-      label: args.label,
-      helperText: args.helperText,
-      value: '3',
-    },
-  ]
-
-  return (
-    <div style={{ display: 'grid', gridTemplateColumns: '1fr', gridGap: '12px' }}>
-      <RadioButtonGroup {...args} name="radio-group" options={ITEMS} onChange={handleChange} value={selected} />
-    </div>
-  )
+  return <RadioButton {...args} name="radio-group" value="1" selectedValue={args.selected ? '1' : undefined} />
 }
 
 export const Default = Template.bind({})
--- a/packages/atlas/src/components/_inputs/RadioButton/RadioButton.stories.tsx
+++ b/packages/atlas/src/components/_inputs/RadioButton/RadioButton.stories.tsx
@@ -13,7 +13,7 @@ export default {
   },
   args: {
     label: 'Hello there',
-    helperText: 't100',
+    caption: 't100',
     disabled: false,
     error: false,
     selected: false,
--- a/packages/atlas/src/components/_inputs/RadioButton/RadioButton.stories.tsx
+++ b/packages/atlas/src/components/_inputs/RadioButton/RadioButton.stories.tsx
@@ -1,11 +1,17 @@
 import { Meta, Story } from '@storybook/react'
 import React, { useState } from 'react'
 
+import { RadioButtonGroup } from '@/components/_inputs/RadioButtonGroup'
+
 import { RadioButton } from './RadioButton'
 
 export default {
   title: 'inputs/RadioButton',
   component: RadioButton,
+  argTypes: {
+    className: { table: { disable: true } },
+    selectedValue: { table: { disable: true } },
+  },
   args: {
     label: 'Hello there',
     helperText: 't100',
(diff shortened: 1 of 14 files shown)


toiletgranny (changes requested):
Perfect, same comments as with #2789: 
> 1. The helper text is still `var(--color-core-neutral-300)`, should be `var(--color-text)`,
> 2. This changelog item seems to not be implemented: "Added Focus state, which inherits its styling from the Hover state".  The goal here is to do the same thing that @drillprop has done with the Switch component recently. Ideally, we use the the `:focus-visible`, instead of `:focus`, if that's okay with you.

drillprop (changes requested):
The `:focus-visible` issue is still not fixed. Use`@supports` media query to make this work. 
Basically, if the browser has support for `:focus-visible` use `:focus-visible`. if not use regular `:focus`.  

Check my code in the Switch component. Btw, maybe it's possible to write this cleaner and more readable, but I couldn't figure out the better way.

drillprop:
Crap, I just noticed that you used `:focus-within`. Unfortunately, it's not possible to use `:focus-within` together with `:focus-visible` - https://github.com/WICG/focus-visible/issues/151. You'll need to rebuild HTML a bit.

drillprop:
I just made a small experiment on how it would work with [this package](https://www.npmjs.com/package/what-input). Check the code here https://github.com/drillprop/atlas/commit/993d5ca244718a11f9ad8154eae90b0967661acc

toiletgranny:
I'm afraid we lost the option to select a radio button by clicking on its label.

toiletgranny (approved):

packages/atlas/src/components/_inputs/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,46 @@
...
+export type RadioButtonGroupProps = {
+  options: RadioButtonProps[]
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void
+  value: string | number
+  name?: string
+  disabled?: boolean
+  error?: boolean
+}
+
+export const RadioButtonGroup: React.FC<RadioButtonGroupProps> = ({
```
drillprop: Can we actually make use of this component and see it in action? We could apply this in the VideoForm at least?

packages/atlas/src/components/_inputs/RadioInput/RadioInput.styles.ts
```diff
@@ -1,139 +1,163 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-clip: content-box;
+  padding: 4px;
+  box-sizing: border-box;
+  border: 1px solid ${({ error }) => cVar(error ? 'colorBorderError' : 'colorBorderAlpha')};
+  transition: all ${cVar('animationTransitionFast')};
+
+  ::before {
```
drillprop, line 27: Could you add a comment in the code that describes what `:before` and `:after` is? Just add something like that:
```tsx
/* inner dot */ 
:after
```

packages/atlas/src/components/_inputs/RadioInput/RadioInput.styles.ts
```diff
@@ -1,139 +1,163 @@
...
-  checked?: boolean
-  error?: boolean
-}
+  position: relative;
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-clip: content-box;
+  padding: 4px;
```
drillprop: let's use sizes here

packages/atlas/src/components/_inputs/RadioInput/RadioInput.styles.ts
```diff
@@ -1,139 +1,163 @@
...
 
-    &::before {
-      background-color: ${cVar('colorCoreNeutral800Lighten')};
+  position: absolute;
+  top: 0;
+  left: 0;
+  margin: auto;
+  opacity: 0;
+  cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
+  z-index: 999;
```
drillprop: We could use `zIndex.globalOverlay` here, but on the other hand, Is this value too big?

drillprop (changes requested):

packages/atlas/src/components/_inputs/RadioButtonGroup/RadioButtonGroup.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+    className: { table: { disable: true } },
+    selectedValue: { table: { disable: true } },
+    options: { table: { disable: true } },
+    onChange: { table: { disable: true } },
+    value: { table: { disable: true } },
+    name: { table: { disable: true } },
+  },
+  args: {
+    label: 'Hello there',
+    helperText: 't100',
```
drillprop: https://github.com/Joystream/atlas/pull/2789#discussion_r893545149

packages/atlas/src/components/_inputs/RadioButtonGroup/RadioButtonGroup.tsx
rafalpawlow: This was very good idea 👍

packages/atlas/src/components/_inputs/RadioButton/RadioButton.stories.tsx
```diff
@@ -1,32 +1,27 @@
...
   title: 'inputs/RadioButton',
   component: RadioButton,
+  argTypes: {
+    className: { table: { disable: true } },
+    selectedValue: { table: { disable: true } },
+    selected: { controls: { type: 'boolean' } },
+  },
   args: {
     label: 'Hello there',
     helperText: 't100',
```
drillprop: image
We still have `helperText` control which does nothing. I'm a little surprised why typescript doesn't mark that as an error.

packages/atlas/src/components/_inputs/RadioButton/RadioButton.tsx
```diff
@@ -3,25 +3,19 @@ import React from 'react'
...
+  ({ value, selectedValue, label, caption, className, disabled, error, onChange, ...props }, ref) => {
     return (
-      <RadioAndCheckboxBase
-        disabled={disabled}
-        className={className}
-        label={label}
-        error={error}
-        helperText={helperText}
-      >
+      <RadioAndCheckboxBase disabled={disabled} className={className} label={label} error={error} helperText={caption}>
```
drillprop: Let's also rename helperText to caption here. Is there a reason why we keep that unchanged?

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -590,27 +589,18 @@ export const VideoForm: React.FC<VideoFormProps> = React.memo(({ onSubmit, setFo
...
-                    helperText={errors.isExplicit ? 'Content rating must be selected' : ''}
-                    disabled={videoFieldsLocked}
-                  />
-                </RadioButtonsContainer>
+                <RadioButtonGroup
+                  ref={ref}
+                  options={[
+                    { label: 'All audiences', value: 'false' },
+                    { label: 'Mature', value: 'true' },
+                  ]}
```
drillprop, line 678: I set my video as mature, but nothing happened when I published my changes, and the transaction was completed. Could you check what's wrong?

drillprop (changes requested):

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
drillprop: ok, ignore that. It's not your fault. This should be fixed once we merge https://github.com/Joystream/atlas/pull/2854

kdembler:

drillprop (approved):

PR #2787, opened 2022-06-02
Rework video form title input fixes

rafalpawlow:
Part of #2692 

It turned out we had some problems with latest changes
* category error appears after adding thumbnail
* no possibility to edit video
* no possibility to mint nft from context menu because action bar was hidden


toiletgranny (approved):
All of the issues you listed are now fixed! I haven't found anything new when testing. Thanks a lot!

kdembler (changes requested):
I can remove thumbnail while editing the video and I am not stopped from publishing the changes

kdembler (approved):

Issue #2595, opened 2022-04-29
Introduce thousands separators (space) in input fields


toiletgranny:
The idea captured in the spreadsheet
> Since it's quite likely that we'll be operating on very large numbers in NFT prices (at least at the beginning), shouldn't we consider introducing thousand separators (spaces) in input fields? Currently, it's very difficult to keep track of how many 0s you are entering.

![CleanShot 2022-04-29 at 16 30 48](https://user-images.githubusercontent.com/35307309/165965397-2750e564-4ae1-463b-8739-eb5b5e172422.png)

eldiegod:
Tried different solutions in here but they all presented some problems as there is no native way to format the output of an input without altering it in some way.

Attempted to format the value of the input directly and parse it after change events back into a number but different problems appeared like when a user has it's caret after a formatting space and they try to delete digits, the event registers like they are trying to delete the space but the space is just formatting so we would have to handle change events ourselves which seems like a lot of logic.

Other option would be to render an overlay on top of the Input that shows the formatted value without altering it in the input directly but the main problems I saw is we would not be able to show a caret as the number spacing would differ with the input and also text selection would have the same problem

dmtrjsg:
In the light of denomination being changed in favour of smaller numbers and JOY fractions, this becomes less of a priority if not worth not doing. Closing before becomes more pertinent.

PR #2790, opened 2022-06-03
Input component

### LINKED ISSUE #2707: Update the Input (formerly: Text field) component
**AC:**

- [ ] Create reusable style blocks that will replace `InputBase` component
- [ ] Implement updates to the `Input` component

**OOC:**

- [ ] Formatting of numbers inside an input, captured in #2595

---

The goal of this task is to update the Input (formerly: Text field) component based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7058%3A174881). 

> **Changelog (from Figma)**
> - Renamed the component from “Text field” to “Input”,
> - Updated the overall styling, along with the internal layers structure,
> - Added Prefix & Suffix variants,
> - **(Breaking change!)** Changed the type of icons used in the component from Controls (24px) to Action (16px),
> - Changed the Action variant to include a Button component of a specific type,
> - Added the ability to nest custom components both at the beginning and at the end of the component,
> - Added processing variant,
> - Added Medium size variant (with the original being now called “Large”),
> - Dropped Label and Caption variants (these are now handled by the Form field component),
> - Added documentation along with content/usage guidelines.

⚠️ Please note that the change in the type and size of icons used inside will make the Input component not backward compatible with the earlier usage in that regard. Therefore, my suggestion for the developer taking care of this task would be to reach out to me once the component is implemented and reviewed, schedule a 30-minutes session together, and walk through the app to catch and note down all of the icons that will require updating.

image

---

**⚠️ Absurd number of variants warning!** 
Due to the nature of the component and the characteristics of Figma, as a design tool, the number of variants of this component is very large. As a developer inspecting the designs, you might want to consider duplicating the Figma file over to your Drafts space and playing with the component yourself. [Learn more (2:15 min).](https://youtu.be/66sWUyenQcw)

drillprop:
Fix #2707 
Designs: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7058%3A174881


toiletgranny (changes requested):
@drillprop Marvelous! I especially like how it looks out in the wild, really breathing some fresh air into some of the views in the app. <3

Here's what I found:
1. Both icons and the spinner in the processing state are not dimmed in the Input disabled state.
2. Having set the `type` to `number` in Storybook, I couldn't get the numbers to be formatted [as described in Figma docs](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7058%3A174881) (by the way, please disregard commas as thousand separators, which I already changed to spaces). Was that not possible to implement?
3. The Button component is placed a bit too far away from the right edge of the Input component. If you give it another look in Figma, you'll notice that buttons are actually moved a bit closer to the edge, to accommodate for the additional padding they come with. This is so that when the Button component is not being hovered, its icon is positioned in the same way as if it _was_ just an icon. Happy to jump on a Slack call to discuss this further.
4. I can activate the Input component and then the Button component by tabbing — I respect that. But, could we preserve the focus styling of the input for when the Button component (or any other component) inside is focused? I think this is what `:focus-within` might help us with?
5. Unfortunately, I could not review the styling of prefixes and suffixes, as well as other custom components that could be potentially embedded in the Input (such as a small Pill component, or a 24px Avatar component). Can we have those as props in the Storybook?

On top of that, I went for a quick tour through the app to see the Input component in real-life. Accidentally, I've stumbled upon this issue, where the "License, content rating, published before, marketing, delete video" description was placed next to the button component, instead of underneath it. Not sure if this is because of this PR, but I haven't seen this before. Either way, @kdembler, could it be yet another argument for us to consider introducing automated visual regression tests? 🤔
CleanShot 2022-06-06 at 11 01 32@2x

drillprop:
> 2. Having set the `type` to `number` in Storybook, I couldn't get the numbers to be formatted [as described in Figma docs](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7058%3A174881) (by the way, please disregard commas as thousand separators, which I already changed to spaces). Was that not possible to implement?

I won't be implementing separators in this PR. This will be tackled here https://github.com/Joystream/atlas/issues/2595

drillprop:
> 1. Both icons and the spinner in the processing state are not dimmed in the Input disabled state.

Fixed

> 3. The Button component is placed a bit too far away from the right edge of the Input component. If you give it another look in Figma, you'll notice that buttons are actually moved a bit closer to the edge, to accommodate for the additional padding they come with. This is so that when the Button component is not being hovered, its icon is positioned in the same way as if it _was_ just an icon. Happy to jump on a Slack call to discuss this further.

Should be fixed

> 4. I can activate the Input component and then the Button component by tabbing — I respect that. But, could we preserve the focus styling of the input for when the Button component (or any other component) inside is focused? I think this is what `:focus-within` might help us with?

Fixed

> 5. Unfortunately, I could not review the styling of prefixes and suffixes, as well as other custom components that could be potentially embedded in the Input (such as a small Pill component, or a 24px Avatar component). Can we have those as props in the Storybook?

Fixed. Unfortunately, you won't be able to play around a lot. I had some issues with setting this right in the storybook. I just add a few examples with different elements as suffixes and prefixes. 

> On top of that, I went for a quick tour through the app to see the Input component in real-life. Accidentally, I've stumbled upon this issue, where the "License, content rating, published before, marketing, delete video" description was placed next to the button component, instead of underneath it. Not sure if this is because of this PR, but I haven't seen this before. Either way, @kdembler, could it be yet another argument for us to consider introducing automated visual regression tests? 🤔 CleanShot 2022-06-06 at 11 01 32@2x

We discussed this yesterday. I think this came from the latest form field update.  I'll tackle this in another PR

drillprop:
> We discussed this yesterday. I think this came from the latest form field update. I'll tackle this in another PR

I was wrong actually - I introduced the bug in this PR 😅 . Good catch - should be fixed now.

packages/atlas/src/components/_inputs/Input/Input.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+} as Meta
+
+const Template: Story<InputProps> = (args) => <Input {...args} />
+
+const TemplateWithControlledInput: Story<InputProps> = (args) => {
+  const [value, setValue] = useState('')
+  return <Input {...args} onChange={(e) => setValue(e.currentTarget.value)} value={value} />
+}
+
+const TemplateWithUncontrolledInput: Story<InputProps> = (args) => {
```
kdembler, line 37: Maybe we just make this template the default? (Both are uncontrolled)

packages/atlas/src/components/_inputs/inputs.utils.ts
```diff
@@ -0,0 +1,88 @@
...
+  medium: sizes(4),
+  large: sizes(5),
+} as const
+
+export const verticalPadding = {
+  medium: sizes(2),
+  large: sizes(3),
+} as const
+
+export const getPadding = (size: InputSize, nodeWidthProps?: NodeWidthProps) => {
```
kdembler: `getInputPadding` for consistency?

packages/atlas/src/components/_inputs/inputs.utils.ts
```diff
@@ -0,0 +1,88 @@
...
+    ? `${nodeWidthProps?.rightNodeWidth + sizes(2, true)}px`
+    : horizontalPadding[size]
+
+  return css`
+    padding: ${verticalPadding[size]} ${paddingRight} ${verticalPadding[size]} ${paddingLeft};
+  `
+}
+
+export const getInputPseudoSelectorStyles = ({ error }: { error?: boolean }) => css`
+  ${sharedInputStyles.default};
```
kdembler: This function will also apply the default styles, wich are not pseudo selectors so the name may be slightly misleading. Maybe `getBaseInputStyles`?

packages/atlas/src/components/_inputs/Input/Input.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+} from '../inputs.utils'
+
+type TextInputProps = {
+  error?: boolean
+  inputSize: InputSize
+} & NodeWidthProps
+
+export const TextInput = styled.input<TextInputProps>`
+  width: 100%;
+  padding: ${sizes(3)} ${sizes(5)};
```
kdembler: Won't this be overwritten by `getPadding` anyway?

packages/atlas/src/components/_inputs/Input/Input.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  &[type='number'],
+  &[type='search'] {
+    appearance: textfield;
+  }
+
+  ${({ inputSize, leftNodeWidth, rightNodeWidth }) => getPadding(inputSize, { leftNodeWidth, rightNodeWidth })};
+
+  ${({ error }) => getInputPseudoSelectorStyles({ error })};
```
kdembler: If you wanted to, this could be just `${getInputPseudoSelectorStyles};` Similarly, if you updated the args of `getPadding` to accept a single object, it could be used the same way. But leaving this up to you, maybe it's better to be explicit

packages/atlas/src/components/_inputs/Input/Input.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+  disabled?: boolean
+  isButton?: boolean
+  isFocused?: boolean
+  size: InputSize
+}
+const nodePlacementStyles = ({ left, size, isButton }: NodeContainerProps) =>
+  left
+    ? css`
+        left: 0;
+        padding-left: calc(${horizontalPadding[size]});
```
kdembler: Unneeded `calc` I think

packages/atlas/src/components/_inputs/Input/Input.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+      `
+    : css`
+        right: 0;
+        padding-right: calc(${horizontalPadding[size]} - ${isButton ? sizes(1) : '0px'});
```
kdembler: Maybe better to have the condition outside of `calc` so we don't do `calc(100px - 0px)`?

packages/atlas/src/components/_inputs/Input/Input.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const [dollars, setDollars] = useState<number>()
+  return (
+    <div style={{ display: 'grid', gap: 24 }}>
+      <Input
+        {...args}
+        nodeStart={<Pill label="label" />}
+        nodeEnd={
+          <Text secondary color={cVar('colorTextMuted')} variant="t300">
+            $
+          </Text>
```
kdembler: I'm not sure if this is captured in the Figma docs but when we discussed the suffix/prefix with Adam, he mentioned that it would be good for all those basic text nodes to always use the same styling.

What do you think about this: We could check whether `nodeStart` (or end) is a string and if it is, we could wrap it in this `<Text>` inside the `Input` itself, not requiring the consumers to repeat the style?

Btw, I think you don't need both `secondary` and `color` props

packages/atlas/src/components/_inputs/Input/Input.tsx
```diff
@@ -0,0 +1,126 @@
...
+        onBlur={onBlur}
+        onWheel={handleWheel}
+        onKeyDown={onKeyDown}
+        placeholder={placeholder}
+        type={type}
+        tabIndex={disabled ? -1 : 0}
+        defaultValue={defaultValue}
+      />
+      {nodeStart && (
+        <NodeContainer onClick={() => inputRef.current?.focus()} size={size} ref={nodeLeftRef} left disabled={disabled}>
```
kdembler: We didn't need this `onClick` logic for the old `TextField` and I think it worked. What changed?

kdembler (changes requested):

toiletgranny (approved):

packages/atlas/src/components/_inputs/Input/Input.stories.tsx
drillprop: good idea - added

packages/atlas/src/components/_inputs/Input/Input.tsx
drillprop: The old `Textfield` was wrapped in `label`(`InputBase`). That's why we didn't have to add this logic.

packages/atlas/src/components/_inputs/Input/Input.styles.ts
drillprop: Updated after all

packages/atlas/src/components/_inputs/Input/Input.tsx
```diff
@@ -61,6 +63,24 @@ const InputComponent: React.ForwardRefRenderFunction<HTMLInputElement, InputProp
...
+  ) : (
+    nodeStart
+  )
+  const renderNodeEnd = isNodeEndText ? (
+    <Text variant={size === 'large' ? 't300' : 't200'} color={cVar('colorTextMuted')}>
+      {nodeEnd}
+    </Text>
+  ) : (
+    nodeEnd
+  )
```
kdembler: Let's rename those vars - `render*` suggests it's a function that returns JSX while it's a static JSX instead. Maybe `rendered`?

kdembler (approved):

PR #2807, opened 2022-06-03
fix reactions stepper

### LINKED ISSUE #2792: Reaction stepper fixes
> As is: There's a min-width: 155px set on the Reaction stepper component.
> To be: There should be no min-width, as it makes the nested buttons misaligned horizontally. I'm curious: why was this property set over there in the first place? I'm worried it might have been solving some other CSS issue I'm not aware of.

> 🚨 As is: The Reaction stepper component doesn’t update after leaving a reaction and having the transaction processed successfully. When I refresh the page, the Reaction stepper component displays correct values.
To be: The Reaction stepper should be refreshed the moment the transaction has finished processing

kdembler:
Fixes #2792

### DIFF
--- a/packages/atlas/src/components/_video/ReactionStepper/ReactionButton/ReactionButton.styles.ts
+++ b/packages/atlas/src/components/_video/ReactionStepper/ReactionButton/ReactionButton.styles.ts
@@ -19,6 +19,7 @@ export type ReactionSteppperState = 'liked' | 'disliked' | 'default' | 'loading'
 
 type ReactionsCounterProps = {
   type: 'like' | 'dislike'
+  disabled?: boolean
   state: ReactionSteppperState
 }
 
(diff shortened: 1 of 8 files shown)


packages/atlas/.env
```diff
@@ -9,10 +9,10 @@ VITE_DEVELOPMENT_NODE_URL=wss://atlas-dev.joystream.org/ws-rpc
...
 VITE_DEVELOPMENT_OFFICIAL_JOYSTREAM_CHANNEL_ID=12
 
-VITE_NEXT_ORION_URL=https://atlas-next.joystream.org/orion/graphql
-VITE_NEXT_QUERY_NODE_SUBSCRIPTION_URL=wss://atlas-next.joystream.org/query-node/server/graphql
-VITE_NEXT_NODE_URL=wss://atlas-next.joystream.org/ws-rpc
-VITE_NEXT_FAUCET_URL=https://atlas-next.joystream.org/member-faucet/register
+VITE_NEXT_ORION_URL=https://34.229.168.20.nip.io/orion/graphql
+VITE_NEXT_QUERY_NODE_SUBSCRIPTION_URL=wss://34.229.168.20.nip.io/query-node/server/graphql
+VITE_NEXT_NODE_URL=wss://34.229.168.20.nip.io/ws-rpc
+VITE_NEXT_FAUCET_URL=https://34.229.168.20.nip.io/member-faucet/register
```
drillprop: You pushed too many changes, I guess.

packages/atlas/src/views/playground/Playgrounds/PlaygroundMinimizedTransaction.tsx
```diff
@@ -37,6 +37,7 @@ export const PlaygroundMinimizedTransaction = () => {
       minimized: {
         signErrorMessage: `Your comment to the video ${title} has not been posted.`,
       },
+      snackbarSuccessMessage: {},
```
drillprop: same here

drillprop (changes requested):

packages/atlas/.env
kdembler: I certainly wanted to do this too fast 😓

kdembler: Fixed now!

drillprop (changes requested):
We totally ignored, that this also needs to be fixed 😅 : 
> As is: There's a min-width: 155px set on the Reaction stepper component.
To be: There should be no min-width, as it makes the nested buttons misaligned horizontally. I'm curious: why was this property set over there in the first place? I'm worried it might have been solving some other CSS issue I'm not aware of.`

Also, Adam brought up two more things to fix in the comment

drillprop (approved):

PR #2813, opened 2022-06-03
update tooltip component, fix comment timestamp

### LINKED ISSUE #2794: Comment timestamp improvements
1. Let's add `onClick` handler to timestamp link on the comment and block navigation. This way clicking on it doesn't do anything but the user can still copy the link with right-click. (don't copy on click, just do nothing)
2. Change the `Tooltip` component to have an optional `multiline` property (replace current `oneLine`), which should be default. Update all the instances if necessary 

> As is: The date time tooltip which appears upon hovering over the timestamp, seems to be used in a wrong variant: multiline. This causes the tooltip to be taller and wider (because of the increased padding), that it should be.
To be: The tooltip component, in this scenario, should be used in the non-multiline variant.
To be checked: I think I saw the same issue somewhere else, which makes me think if this is not the Tooltip component implementation being flawed. Perhaps the Multiline variant is now the default one, in which case — it shouldn't be.

kdembler:
Fixes #2794

### DIFF
--- a/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
@@ -35,7 +35,7 @@ const DefaultTooltip: Story<TooltipProps> = (args) => (
   <Tooltip
     {...args}
     showOnCreate
-    customContent={args.customContent && <div style={{ border: '1px solid red' }}>Cusom content</div>}
+    customContent={args.customContent && <div style={{ border: '1px solid red' }}>Custom content</div>}
   />
 )
 
--- a/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
@@ -28,7 +28,6 @@ export default {
     placement: 'bottom-start',
     text: 'Lorem ipsum dolor amet! Lorem ipsum dolor amet!',
     headerText: 'Lorem ipsum',
-    oneLine: false,
   },
 } as Meta<TooltipProps>
 
@@ -36,13 +35,7 @@ const DefaultTooltip: Story<TooltipProps> = (args) => (
   <Tooltip
     {...args}
     showOnCreate
-    customContent={
-      args.customContent && (
-        <div key={0} style={{ marginTop: '12px' }}>
-          Footer content
-        </div>
-      )
-    }
+    customContent={args.customContent && <div style={{ border: '1px solid red' }}>Cusom content</div>}
   />
 )
 
--- a/packages/atlas/src/components/Tooltip/Tooltip.tsx
+++ b/packages/atlas/src/components/Tooltip/Tooltip.tsx
@@ -7,7 +7,7 @@ import { transitions } from '@/styles'
 import {
   IconWrapper,
   StyledSvgAlertsInformative24,
-  StyledTooltip,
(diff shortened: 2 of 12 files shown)


packages/atlas/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -28,21 +28,14 @@ export default {
...
     {...args}
     showOnCreate
-    customContent={
-      args.customContent && (
-        <div key={0} style={{ marginTop: '12px' }}>
-          Footer content
-        </div>
-      )
-    }
+    customContent={args.customContent && <div style={{ border: '1px solid red' }}>Cusom content</div>}
```
rafalpawlow: typo
```suggestion
    customContent={args.customContent && <div style={{ border: '1px solid red' }}>Custom content</div>}
```

rafalpawlow (changes requested):
It seems like `multiline` doesn't do anything except padding. In previous implementation `oneLine` had `max-width: 400px` and it was approved by @toiletgranny 

Zrzut ekranu 2022-06-3 o 18 37 41
Zrzut ekranu 2022-06-3 o 18 37 47

packages/atlas/src/components/Tooltip/Tooltip.tsx
```diff
@@ -43,13 +43,35 @@ export const Tooltip: React.FC<TooltipProps> = ({
   delay,
   customContent,
   showOnCreate,
-  oneLine,
+  multiline,
   className,
 }) => {
   const [isVisible, setIsVisible] = useState(false)
+
+  const content = customContent ? (
```
rafalpawlow, line 51: Wouldn't this to the job?
```suggestion
const content = customContent || (
    <TooltipContent headerText={!!headerText}>
      <TooltipHeader headerText={!!headerText}>
        {icon && (
          <IconWrapper>
            <StyledSvgAlertsInformative24 />
          </IconWrapper>
        )}
        {headerText && <TooltipText variant="h100">{headerText}</TooltipText>}
      </TooltipHeader>
      {text && text.length && (
        <TooltipText withIcon={!!icon} headerText={!!headerText} variant="t100">
          {text}
        </TooltipText>
      )}
    </TooltipContent>
  )
```

rafalpawlow (changes requested):

packages/atlas/src/components/Tooltip/Tooltip.tsx
kdembler: It's the same thing really

kdembler:
> It seems like multiline doesn't do anything except padding. In previous implementation oneLine had max-width: 400px and it was approved by @toiletgranny

Well that seems wrong, there is a section about sizes in Figma and it outlines what max-width the component should have based on variant: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6215%3A144771

toiletgranny:
> In previous implementation `oneLine` had `max-width: 400px` and it was approved by @toiletgranny

Yes, I remember discussing this together, but frankly, I can't recall what was the reason for that. @rafalpawlow, unless you remember it better than I do (in which case let me know), I would say that in the documentation I trust, and therefore, the implementation in this PR is fine by me.

rafalpawlow:
@toiletgranny TBH I don't remember why we decided it...

rafalpawlow (approved):

PR #2816, opened 2022-06-06
Remove comment tweaks

### LINKED ISSUE #2801: Comment remove tweaks
1. As is: The List item component for deleting a comment is labeled as “Remove”.
    To be: “Delete”.
2. As is: A comment that is being deleted is not being grayed out for the time of processing the transaction.
To be: It should be, accordingly to [this user story.](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=4143%3A255341).
3. As is: The delete confirmation Snackbar component doesn’t include a link to the video.
To be: It should, accordingly to [this user story](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2356%3A225990).
4. As is: A deleted comment (with no replies) doesn’t disappear from view once the transaction has been finalized. Instead, it stays there but the contents change to “Comment deleted by the author”.
To be: It should be gone momentarily, accordingly to [this user story](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2464%3A238350).
5. As is: In a deleted comment, which remains in the main thread, the gap between the “Comment deleted by the channel owner/comment author” message and the header part of the Comment body component is unchanged, set to 8px.
To be: It should be set to 12px, just like the gap between the deletion message and the header part of the Comment body componet.
6. As is: When comment is deleted, copy says `Comment deleted by the author` or `Comment deleted by the channel owner`
To be: Let's drop `the`

drillprop:
Fix #2801 

I'm not sure if I fully understand point 5: 
> As is: In a deleted comment, which remains in the main thread, the gap between the “Comment deleted by the channel owner/comment author” message and the header part of the Comment body component is unchanged, set to 8px.
To be: It should be set to 12px, just like the gap between the deletion message and the header part of the Comment body component.

For me, the gap between the text `Comment deleted...` and the header looks correct(12px).
Deleted:
image
Not deleted: 
image

@toiletgranny could you explain which gap you referring to?

### DIFF
--- a/packages/atlas/src/components/_comments/Comment/Comment.styles.ts
+++ b/packages/atlas/src/components/_comments/Comment/Comment.styles.ts
@@ -27,19 +27,27 @@ export const CommentWrapper = styled.div<{ shouldShowKebabButton: boolean }>`
 
   :hover {
     ${KebabMenuIconButton} {
-      opacity: ${({ shouldShowKebabButton: isActive }) => (isActive ? 1 : 0)};
+      opacity: ${({ shouldShowKebabButton }) => (shouldShowKebabButton ? 1 : 0)};
     }
   }
 `
 
-export const CommentHeader = styled.div<{ isDeleted: boolean }>`
+export const CommentArticle = styled.article<{ isDeleted?: boolean }>`
+  display: grid;
+  grid-auto-flow: row;
+  gap: ${({ isDeleted }) => (isDeleted ? sizes(3) : sizes(2))};
+  align-items: start;
+`
+
+export const CommentHeader = styled.header`
   display: flex;
   align-items: center;
   flex-wrap: wrap;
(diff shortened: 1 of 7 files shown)


toiletgranny:
Sorry, @drillprop, I meant the footer, not the header. 
CleanShot 2022-06-06 at 12 30 48@2x

toiletgranny:
Other than this, to me, this PR looks okay, except for the grayed-out state when we're processing the transaction for deleting the comment: could we also disable any clicking/focusing inside, for the time it's grayed out? The purpose of doing this is to prevent the user from adding a reaction to a comment which is being deleted or adding a reply to it.

drillprop:
Ok, both issues should be fixed

rafalpawlow (approved):

PR #2822, opened 2022-06-07
Comment styling tweaks

### LINKED ISSUE #2796: Comments styling improvements
1. As is: The color of the dot separating the username and the timestamp is set to var(--color-background-alpha).
To be: var(--color-background-strong-alpha)
2. As is: On smaller viewports, the “Reply” button is carried over to the next line in the footer of the Comments body component. 🏞 [user-images.githubusercontent.com/35307309/171634903-0a7f0c1a-ce75-4513-afc7-eafa8f3d8fe9.png](https://user-images.githubusercontent.com/35307309/171634903-0a7f0c1a-ce75-4513-afc7-eafa8f3d8fe9.png)
To be: If there's enough space to fit both the reaction buttons and the replies-related elements together, they should be sitting in one line. Otherwise, a lot of vertical space is lost, which is especially harmful on mobile devices.
3. As is: Because of the tricky CSS and playing with a negative margin and additional padding (which, as a reminder, is added there to accommodate for the highlight around the comment), the comment row is no longer 100% width of the parent container. It’s slightly less than that. This is best observed when you add * { outline: solid 1px red; } to the inspect mode in Chrome.
To be: Each Comment row component's width should fill the parent container.
4. Can we offset the Comment body component's footer horizontally by minus 8px? This is so that Reaction chip components appear aligned with the text above. I will update the designs accordingly, but please let me know if we can do that. I believe it's a tiny CSS tweak that makes a ton of difference.
5. As is: Reaction chip components are currently sorted using their original order in the Reaction popover component. Regardless of how many times each reaction has been applied under a given comment.
To be: Reactions under a comment should be sorted from left to right, top to bottom, starting with the most popular reaction.
6. As is: Color of the icon in the new reaction button (the monochromatic smiley face) is derived from the Button component implementation: var(--color-core-neutral-50).
To be: Can we override it with the one which is currently used in the designs for comments? It's var(--color-core-text).
7. As is: Color of the value in the Reaction chip component, in its default and inactive state, is set to var(--color-text-strong).
To be: var(--color-text). The lighter color kicks in only upon hovering or when the Reaction chip component is active.
8. As is: The gap between the reactions part and the replies part in the footer of the Comments body component is set to4px.
To be: It should be set to 12px. It is very important to have this gap between the reactions part and the replies part because if there are no replies under a comment, then the “Reply” button is displayed too closely to the reactions part of the footer. [Please, have a closer look at the layers' structure for this component in Figma.](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2411%3A217091)
9. As is: Color of the label in the Replies chip component is set to var(--color-core-neutral-300) in its default state.
To be: var(--color-text)
10. As is: The upper padding in the Edit history dialog body is set to 12px.
To be: 24px, as it was set up on the Dialog component level. No overrides were applied here, at least not in the Figma designs. Is this an issue with a component, or with its usage?
11. See Adam's comment below

drillprop:
Fix #2796

### DIFF
--- a/packages/atlas/src/components/_comments/Comment/Comment.styles.ts
+++ b/packages/atlas/src/components/_comments/Comment/Comment.styles.ts
@@ -3,7 +3,7 @@ import { Link } from 'react-router-dom'
 
 import { AvatarGroup } from '@/components/Avatar/AvatarGroup'
 import { Text } from '@/components/Text'
-import { Button, TextButton } from '@/components/_buttons/Button'
+import { Button } from '@/components/_buttons/Button'
 import { SvgActionTrash } from '@/components/_icons'
 import { SkeletonLoader } from '@/components/_loaders/SkeletonLoader'
 import { cVar, media, sizes, square } from '@/styles'
@@ -86,8 +86,6 @@ export const DeletedComment = styled(Text)`
 export const CommentFooter = styled.footer<{ isProcessing?: boolean }>`
   pointer-events: ${({ isProcessing }) => (isProcessing ? 'none' : 'unset')};
   touch-action: ${({ isProcessing }) => (isProcessing ? 'none' : 'unset')};
-  position: relative;
-  left: -8px;
 `
 
 export const ReactionsAndPopover = styled.div`
@@ -95,19 +93,14 @@ export const ReactionsAndPopover = styled.div`
   grid-auto-flow: column;
   gap: ${sizes(1)};
   justify-content: start;
+  position: relative;
+  left: -8px;
 `
 
 export const ReactionsWrapper = styled.div`
-  display: grid;
-  grid-template-columns: auto;
-  grid-auto-rows: auto;
-  gap: ${sizes(2)};
-
-  ${media.sm} {
-    align-items: center;
-    grid-template-columns: auto 1fr;
-    gap: ${sizes(1)};
-  }
+  display: flex;
+  flex-wrap: wrap;
+  gap: ${sizes(3)};
 `
 
 export const StyledFooterSkeletonLoader = styled(SkeletonLoader)`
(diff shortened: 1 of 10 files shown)


packages/atlas/src/components/_comments/Comment/Comment.styles.ts
```diff
@@ -88,24 +88,19 @@ export const CommentFooter = styled.footer<{ isProcessing?: boolean }>`
...
-  ${media.sm} {
-    align-items: center;
-    grid-template-columns: auto 1fr;
-    gap: ${sizes(1)};
-  }
+  grid-auto-flow: column;
+  gap: ${sizes(1)};
+  justify-content: start;
+  position: relative;
+  left: -8px;
```
rafalpawlow: use `sizes`

rafalpawlow (changes requested):
Point 3 doesn't seem to be fixed. Also, I'm not able to add reaction where there are no any reactions.

Zrzut ekranu 2022-06-8 o 09 51 32
Zrzut ekranu 2022-06-8 o 09 53 39

rafalpawlow (changes requested):
something is wrong on hover, maybe we should preserve space for reply button? Viewport `534px`
**EDIT: It works, my bad. It's just need to be a touch device**

https://user-images.githubusercontent.com/13706246/172567536-49acb6f8-a2cf-4436-a4ca-30722e61223c.mov

drillprop:
@rafalpawlow everything should be fixed now

rafalpawlow (changes requested):
I think we should improve alignment of replies on mobile
**EDIT: turned out this is fine**
Zrzut ekranu 2022-06-8 o 11 42 32

rafalpawlow (approved):

kdembler (approved):

PR #2833, opened 2022-06-08
Create current block store

### LINKED ISSUE #2781: Isolate re-renders caused by block height updates
With the current implementation, `useJoystream` will force a re-render across all the components that use it every 6 seconds because of updates to block height. This is very sub-optimal because vast majority of those components don't rely on this information and re-render without a need. We probably should extract block updates to a separate hook/provider and only use it where needed

rafalpawlow:
Closes #2781


packages/atlas/src/providers/currentBlock/currentBlock.manager.ts
```diff
@@ -0,0 +1,44 @@
...
+  const { joystream, proxyCallback } = useJoystream()
+  const firstRender = useRef(true)
+
+  // fetch current block from the chain, but only just once
+  useEffect(() => {
+    if (!firstRender.current || !joystream) {
+      return
+    }
+    joystream.getCurrentBlock().then((block) => {
+      setCurrentBlock(block)
```
kdembler: Let's also set the timestamp here, I guess I missed that in the initial implementation

packages/atlas/src/providers/currentBlock/currentBlock.manager.ts
```diff
@@ -0,0 +1,44 @@
+import React, { useEffect, useRef } from 'react'
```
kdembler: Can we rename this to `joystream.manager.ts` and move to `providers/joystream`?

packages/atlas/src/providers/currentBlock/currentBlock.store.ts
```diff
@@ -0,0 +1,28 @@
+import { createStore } from '@/store'
```
kdembler: Can we rename this to `joystream.store.ts` and move to `providers/joystream`?

packages/atlas/src/providers/joystream/joystream.hooks.ts
```diff
@@ -1,6 +1,6 @@
 import { useCallback, useContext } from 'react'
 
-import { JoystreamContext, JoystreamContextValue } from './provider'
+import { JoystreamContext, JoystreamContextValue } from '.'
```
kdembler: This should import from specific file instead of `index.ts`

packages/atlas/src/providers/joystream/joystream.manager.ts
```diff
@@ -0,0 +1,43 @@
+import React, { useEffect, useRef } from 'react'
+
+import { useJoystream, useJoystreamStore } from '.'
```
kdembler: This should import from specific file instead of `index.ts`

kdembler (approved):

Issue #2806, opened 2022-06-03
Optimistic comment reactions


kdembler:
**AC:**
- Optimistically increase reactions counts (and user reactions) once the transaction is signed (reaction processing state applies only when waiting for a signature)
- Modify the way polkadot.js assigns nonces, so that transactions waiting in the pool are taken into account - https://polkadot.js.org/docs/api/cookbook/tx#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce
- Refetch transaction only if something fails

---

> 🚨 It was only once I got a chance to actually write a few comments myself, and attempt to leave a few reactions under other comments when I realized: oh man, @kdembler, you were right! Signing transactions takes forever! I'm deeply concerned about how discouraging this will be for viewers to get involved in social interactions on Atlas. If there's nothing we can do on the technical side, as we've already discussed this a couple of times, I started thinking if there's anything on the surface experience to mitigate this. And that's when I thought of Google, Gmail, and how they're faking sending an email to let you undo it within ~7 seconds after you hit "Send". Guys, could we try doing the same thing here? Here's how I think this could work:
> 1. As a viewer, you hit the Reaction chip component to leave a reaction.
> 2. Immediately after that the Polkadot window appears so that you can sign the transaction. On our side, we're changing the state of the Reaction chip component to processing, and we throw a snackbar letting the viewer know that the ball is on their court.
> 3. So far, points 1. and 2. are the same as with the current implementation. But once the viewer signs the transaction in Polkadot, instead of keeping the Reaction chip component spinning, and the other ones locked, waiting for the transaction to be actually processed... how about we act as if it was it? Done? We show the reaction as applied, the other reaction chip components are unlocked and you, as a viewer, are good to go. Because, when you think about it, what else could we want from you, as a viewer, at this point? What are the chances that once you've signed the transaction it fails? And even if it does, then so what? We still have the error snackbar to let you know about it, right? And the result would be exactly the same as if we were to keep you waiting, right? In return, however, you get an experience that feels much more snappy and responsive. And, most importantly, not discouraging you from doing it again. I don't know... Is it feasible? What am I missing with this?

PR #2835, opened 2022-06-08
add support for multiple transactions at the same time

kdembler:
This is first part of fixes for #2806. While it still doesn't optimistically increase the number of reactions, it adds a support for multiple transactions ongoing at the same time and ability to track progress of each independently. The optimistic part will be done as a followup


drillprop (approved):
I didn't find any issues, works pretty good 👍🏻

PR #2836, opened 2022-06-08
fix description truncating

### LINKED ISSUE #2780: Link overlapping with the description on small screens
Screenshot 2022-06-02 at 11 36 28

Seen here: https://play.joystream.org/video/26800

Only happens on screens below 520px for this specific case although I guess it depends on the exact link.

kdembler:
Fixes #2780


eldiegod (approved):
Works good 👍 

very minor thing, the ellipsis seems to be white while the text is gray

PR #2845, opened 2022-06-09
Update ComboBox & Member badge

### LINKED ISSUE #2776: Update the Combo box & Member badge components
The goal of this task is to update both Combo box & Output pill (formerly known as **Member badge**) components based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7860%3A169362). 

>  **Changelog (from Figma)**
> - Output pill:
>     - Renamed the “Member badge” component to “Output pill” to reflect the intention of using this component only in the context of the Combo box,
>     - Updated its styling (border radius, inner paddings, inner gap max-height, etc.),
>     - Replaced the “X” icon with an icon-only Button component (please note the overridden border-radius),
>     - Updated its properties and variants to better capture the actual usage of the component in the wild (Video & NFT workspaces),
> - Combo box:
>     - Simplified and ditched most properties of the component, because now they are handled through either components which are embedded inside, or the Form field component, for example:
>     - Ditched the loading state, as now there’s a processing variant directly in the Input component,
>     - (Compared to implementation) Ditched label, placeholder, helper text, error, or disabled properties as they can be now achieved through the Form field component,
> - (Figma-only) Rewritten the documentation for both components,
> - (Figma-only) Moved the component outside the Video workspace file into the Web components file to allow global usage.
> 

image

rafalpawlow:
Closes #2776 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7860%3A169362


kdembler:

toiletgranny (changes requested):
@rafalpawlow I'm afraid I don't know how to test it, as I can't seem to be able to add anything to the `ComboBox` component, therefore I can not see the updated `Member badge` component. On top of that, I'm afraid most of the options in Storybook don't do anything with the `ComboBox` component, at least on the surface? 

The one thing I noticed right away was the lack of the "plus" icon, which should be part of the ComboBox component at all times and should not be customizable. :)

rafalpawlow:
@toiletgranny I added new variant with member selector. You can also test it also on nft form

eldiegod (changes requested):
code looks good

In figma design mentions that it should open on click without the need of user input.

Also members story could add some gap between the output pills

toiletgranny:
I wish I could review this component, but the Storybook is still not working. Are we expecting to get it back and running anytime soon?

rafalpawlow:
@toiletgranny Sorry we had problems with storybook recently, now everything should work fine

toiletgranny (changes requested):
Looking great! I love the attention to detail in the ~~Member badge~~ Output pill component and the fact I can use the Combobox component single-handed, with just my keyboard. ❤️ Here's what I found:

1. In the NFT workspace, upon enabling the whitelist in the NFT listing flow, the plus icon appears with a slight delay, causing a layout shift. Not a biggie, but if there's an easy solution to this issue, I'd appreciate it a lot!
2. Not sure if this belongs in this PR, but in the NFT workspace, upon enabling the whitelist in the NFT listing flow, the "Selected:" label is still present preceding Member badge components. Believing that it's an integral part of the Combobox component, that "Selected:" label should not be there, nor in any other instance of this component. The idea is to drop it entirely.
3. Again, in the NFT workspace, upon enabling the whitelist in the NFT listing flow and activating the Combox component **for the first time**, before typing anything into the Input field, I get the "We couldn't find this member" message. Ideally, this is displayed only upon typing something into the Input field.
CleanShot 2022-06-20 at 15 03 10
4. Not sure if this is an issue, or if it's just that PRs are not synced with each other, but a while ago I reported that, for example, the Input component had a wrong bottom border thickness in the active state (among other things). While these issues have been fixed in the Input component PR, I can't see them being addressed here. Why?
5. The Output pill component in the read-only mode has no padding.
6. A nice-to-have: upon removing an Output pill component by hitting the spacebar having focused the "X" button, it would be nice to have the "X" button in the previous (if it exists) instance of the Output pill component selected automatically. This is so that I can quickly clear/adjust my selection using just a spacebar and tab & shift + tab keys.

rafalpawlow:
@toiletgranny 
1. If you don't mind I would like to address this in another issue. This seems to be a problem with Input implementation and I don't want to further delay this PR.
4. I merged changes from `dev` please check if you still have problems with it.

toiletgranny (approved):

eldiegod (changes requested):
Couple of things I found:

Not sure if filtering already selected options should happen inside or outside the component but at least in the storybook example selecting multiple times the same option is possible and could lead to bugs since I can't seem to be able to remove the duplicates manually. Then again not sure, we actually mind this.

Screenshot 2022-06-21 at 15 51 42


The second thing is that when I click on the input to manually select an option without writing, after selecting if I want to click again to select another it stops opening the dropdown

eldiegod (approved):

PR #2848, opened 2022-06-09
Select component

### LINKED ISSUE #2708: Update the Select component
**AC:**

- [ ] Extract `List` component from the `ContextMenu`. This would be just a styled list of `ListItem` components. It needs to support passing arbitrary props to both the container and items, so it can be used with `downshift` inside the `Select`. It should have a `size` prop. Then the `ContextMenu` is just a `List` wrapped in `Popover`
- [ ] Update the `Select` component with all the changes from Figma. We use new `List` component inside to have matching styles.

**OOC:**

- [ ] Support for selecting multiple values

---

The goal of this task is to update the Select component based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7080%3A141520). 

> **Changelog (from Figma)**
> - Updated the overall styling, along with the internal layers structure,
> - **(Breaking change!)** Changed the type of icons used in the component from Controls (24px) to Action (16px),
> - Added an inline label variant,
> - Dropped the Small size variant,
> - Dropped Label and Caption variants (these are now handled by the Form field component),
> - Proposed the use of Context menu for rendering the list of options, along with guidelines on keyboard navigation,
> - Added guidelines for multi-select usage,
> - Added documentation along with content/usage guidelines.

⚠️ Please note that [similarly to the Input component](https://github.com/Joystream/atlas/issues/2707), the change in the type and size of icons used inside will make the Select component not backward compatible with the earlier usage in that regard. Therefore, my suggestion for the developer taking care of this task would be to reach out to me once the component is implemented and reviewed, schedule a 30-minutes session together, and walk through the app to catch and note down all of the icons that will require updating.

image

drillprop:
Fix #2708
Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7080%3A141520

Support for selecting multiple values is out of the scope of this PR


kdembler (changes requested):
It doesn't seem the first point from AC has been completed?

> Extract List component from the ContextMenu. This would be just a styled list of ListItem components. It needs to support passing arbitrary props to both the container and items, so it can be used with downshift inside the Select. It should have a size prop. Then the ContextMenu is just a List wrapped in Popover

kdembler:

toiletgranny (changes requested):
Nice! ❤️

1. **As is:** Distance between the icon and the value is set to `4px`.
**To be:** `8px` in both size variants (just like in the Input component).
2. **As is:** The width of the bottom border is set to `2px` on the default state, and `1px` on focus.
**To be:**  It should be the other way around.
3. **As is:** `ListItem` component size variant is set to `Large`.
**To be:** [As per Figma documentation:](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7479%3A154979) a medium-sized `ListItem` component should be used for the large-sized `Select` component, and a small-sized `ListItem` component should be used for the medium-sized `Select` component.
4. Note, that I was not able to test[ a multi-select variant of the `Select` component](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7483%3A149489). Did we drop it from the scope?

drillprop:
> 4. Note, that I was not able to test[ a multi-select variant of the `Select` component](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7483%3A149489). Did we drop it from the scope?

Yeah: 
> Support for selecting multiple values is out of the scope of this PR

We will tackle this some other time.

drillprop:
> Nice! ❤️
> 
> 1. **As is:** Distance between the icon and the value is set to `4px`.
>    **To be:** `8px` in both size variants (just like in the Input component).
> 2. **As is:** The width of the bottom border is set to `2px` on the default state, and `1px` on focus.
>    **To be:**  It should be the other way around.
> 3. **As is:** `ListItem` component size variant is set to `Large`.
>    **To be:** [As per Figma documentation:](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7479%3A154979) a medium-sized `ListItem` component should be used for the large-sized `Select` component, and a small-sized `ListItem` component should be used for the medium-sized `Select` component.

Everything should be fixed. I also fixed the border issue that you mentioned here: https://github.com/Joystream/atlas/pull/2858#pullrequestreview-1004335298 (point 2)

Just for clarification, we use shared styles for all form components, but I have to update borders also here, since this is a different PR, and the text area PR isn't merged yet.

kdembler (changes requested):
NVM, bad PR :(

toiletgranny (changes requested):
Perfect, thanks, we're almost there! Two more things I found:

1. Now it seems like the placeholder/value is too far away from the left edge of the component when the icon is not present. CleanShot 2022-06-14 at 11 47 29
2. The red bottom border disappears upon hovering when the `error` styling is applied. Worth double-checking with `Input` and `Textarea` components, too, I think.

Once the two above issues are resolved, this is good to go to me! <3 Looks stunning!

packages/atlas/src/components/_inputs/Select/Select.tsx
```diff
@@ -93,78 +83,67 @@ export const _Select = <T extends unknown>(
     onSelectedItemChange: handleItemSelect,
   })
 
+  if (inlineLabel && icon) {
+    ConsoleLogger.warn('Select: inlineLabel and icon are mutually exclusive. icon will be ignored.')
```
eldiegod: good idea, maybe even worth to log as an error

eldiegod (approved):
Looks good, shame Storybook build is still broken

packages/atlas/src/components/List/List.styles.ts
```diff
@@ -1,16 +1,16 @@
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 
-import { cVar, sizes } from '@/styles'
+import { cVar, sizes, zIndex } from '@/styles'
 
-export type ContextMenuSizes = 'small' | 'medium' | 'large'
+import { ListItemSizes } from '../ListItem/ListItem.styles'
```
kdembler: Could we re-export this type in `ListItem/index.ts`? We should avoid multiple imports from "internal" styles files

kdembler (commented):
Also, shouldn't we use pointer cursor when the user hovers over the select?

packages/atlas/src/components/_inputs/Select/Select.tsx
drillprop: Hmm, yeah - good idea. Error is much more prominent

packages/atlas/src/components/List/List.styles.ts
drillprop: fixed

drillprop:
> Also, shouldn't we use pointer cursor when the user hovers over the select?

I asked Adam about cursors and I'm waiting for a response.

toiletgranny (approved):
> Also, shouldn't we use pointer cursor when the user hovers over the select?

Following the behavior of the [native, unstyled `<select>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), I'd recommend not to use the pointer on hover. For select options, I'm fine with using the pointer.

@drillprop To me, it's great! The very last thing, a cherry on top, would be changing the "Select option" placeholder to just say "Select", as per the designs. And that's it! 🚀

kdembler (approved):

Issue #2775, opened 2022-06-02
Update the Action bar component


toiletgranny:
**AC:**
- Clean up weird logic for `isActive`, make this come from props
- Implement changes described below

The goal of this task is to update the Action bar component based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7688%3A149633). 

>  **Changelog (from Figma)**
> - Renamed and changed the contents of all text layers to be content-agnostic and thus allow for broader usage (as opposed to imposing usage only within Video & NFT workspaces),
> - Rearranged and optimized spacing in-between elements to make the overall layout more future-proof,
> - Replaced “New”, “Edit”, “List NFT”, and “Create playlist” variants with a “Content” property with “1 Button” and “2 Buttons” variants,
> - Changed all text & background colors to use color design tokens,
> - Made the presence of the Info and Fee information optional, by tying them to their own respective component properties,
> - Replaced the local “Draft info” component with a string and the Information component,
> - Changed text styles and button sizes in the XXS breakpoint variant,
> - Added an SM breakpoint variant.

![image](https://user-images.githubusercontent.com/35307309/171567527-d0de8672-aaf2-4d59-a9dc-365053f2af20.png)

PR #2857, opened 2022-06-13
Action bar updates

eldiegod:
fixes: #2775

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -3,6 +3,7 @@ module.exports = {
   defaultSeverity: 'warning',
   customSyntax: '@stylelint/postcss-css-in-js',
   rules: {
+    'declaration-block-no-redundant-longhand-properties': null,
     'no-empty-source': null,
     'function-name-case': null,
     'custom-property-empty-line-before': null,
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -3,7 +3,6 @@ module.exports = {
   defaultSeverity: 'warning',
   customSyntax: '@stylelint/postcss-css-in-js',
   rules: {
-    'declaration-block-no-redundant-longhand-properties': null,
     'no-empty-source': null,
     'function-name-case': null,
     'custom-property-empty-line-before': null,
--- a/packages/atlas/src/components/ActionBar/ActionBar.stories.tsx
+++ b/packages/atlas/src/components/ActionBar/ActionBar.stories.tsx
@@ -19,8 +19,7 @@ export default {
         text: 'Drafts system can only store video metadata. Selected files (video, thumbnail) will not be saved as part of the draft.',
       },
     },
-    primaryText: 'Fee: 0.2 Joy',
-    secondaryText: 'Every change to the blockchain requires making a nominal transaction.',
+    fee: 100,
   } as ActionBarProps,
   argTypes: {
     onClick: { table: { disable: true } },
--- a/packages/atlas/src/components/ActionBar/ActionBar.stories.tsx
+++ b/packages/atlas/src/components/ActionBar/ActionBar.stories.tsx
@@ -10,58 +10,24 @@ export default {
     primaryButton: {
       text: 'Create new channel',
     },
+    secondaryButton: {
+      text: 'Cancel',
+    },
+    infoBadge: {
+      text: 'Saving drafts',
+      tooltip: {
+        text: 'Drafts system can only store video metadata. Selected files (video, thumbnail) will not be saved as part of the draft.',
+      },
+    },
     primaryText: 'Fee: 0.2 Joy',
     secondaryText: 'Every change to the blockchain requires making a nominal transaction.',
-    detailsText: 'Video details saved as draft (2 min ago)',
   } as ActionBarProps,
   argTypes: {
-    variant: { table: { disable: true } },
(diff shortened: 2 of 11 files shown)


rafalpawlow (changes requested):
Horizontal padding is wrong, from SM to MD it should be `16px`

rafalpawlow (changes requested):
I cannot see this text in storybook, is it implemented in the component?
Zrzut ekranu 2022-06-14 o 15 08 03

eldiegod:
> I cannot see this text in storybook, is it implemented in the component?



> I cannot see this text in storybook, is it implemented in the component? Zrzut ekranu 2022-06-14 o 15 08 03

@rafalpawlow yes, you just have to be at the right breakpoint (LG)

rafalpawlow:
@eldiegod  Right 🤦. LGTM but let's do the rebase and be sure everything is fine

rafalpawlow (changes requested):
Sorry I found one more thing. Spacing between primary button and badge is wrong, it's `32px` and should be `16px`. I think because of lack of secondary button, gap is doubled.
Zrzut ekranu 2022-06-14 o 22 22 43

Also I'm not sure about padding in badge, as you can see it should be `8px` but in your implementation padding is `16px`
Zrzut ekranu 2022-06-14 o 22 23 40

toiletgranny:
Should I be taking a look at this as well?

eldiegod:
> Should I be taking a look at this as well?

@toiletgranny Ideally, but storybooks in the PRs are sadly not working for the time being 😔

eldiegod:
> Sorry I found one more thing. Spacing between primary button and badge is wrong, it's `32px` and should be `16px`. I think because of lack of secondary button, gap is doubled. Zrzut ekranu 2022-06-14 o 22 22 43
> 
> Also I'm not sure about padding in badge, as you can see it should be `8px` but in your implementation padding is `16px` Zrzut ekranu 2022-06-14 o 22 23 40

@rafalpawlow how did you get that double gap applied? for me when I remove secondary button it just works as expected.

also left the padding of the badge at 12px for md up because it looks a bit better when hovered

eldiegod:
@toiletgranny nvm it works now, feel free to give it a go

rafalpawlow:
@eldiegod this screenshot comes from video workspace

eldiegod:
> @eldiegod this screenshot comes from video workspace

@rafalpawlow ah yes, I didn't notice a tooltip was added to the primary button and it messed the styles.
Spacing should be good now

rafalpawlow (approved):

toiletgranny:
@eldiegod, @kdembler asked me to let you know that we've just introduced a new set of changes to the Action bar component, that Klaudiusz would like to be included in the scope of this PR. All you need to know can be found in the newest changelog entry in Figma, [under this link](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7688%3A149633) (the link points directly to the newest changelog entry).

image

⚠️ Please note that the fee information should be now displayed in all instances of the Action bar component, which includes flows such as the NFT listing (currently, the implementation disregards the designs and omits this information in the Action bar component), or the NFT purchase view.

kdembler (changes requested):
I've noticed that there is no "Cancel" button when editing a video now

toiletgranny (changes requested):
These changes look great! Apart from the missing secondary button in video & NFT workspaces, which @kdembler already pointed out, there's one item in the changelog, which seems to be overlooked, and it is:
> Replaced the local “Draft info” component with a string and the Information component

Basically, the idea is to replace the "Drafts are saved automatically" info, with a standard line of text and an instance of the Information component next to it. Also, please note that the contents of the tooltip, as well as the caption next to it, should now be customizable in different instances of the Action bar component. :)

toiletgranny:
I was unclear with my prev comment, sorry! 😅

The lorem ipsum part is now correct, but, additionally, the ask was to replace this part:
CleanShot 2022-06-21 at 14 42 12
with this:
CleanShot 2022-06-21 at 14 43 52

So, in other words, to replace whatever is currently used for displaying that tooltip and the rectangle around it the "Lorem ipsum (i)" with a raw piece of text and the [Information component](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6215%3A141264).

---

While we're on the subject of the Tooltip component, @kdembler brought my attention to the fact that the current documentation for this component does not match the designs provided in Figma and is in contradiction with its other parts, for which I apologize. 

I was thinking, could we use this opportunity and this PR, to change our approach to defining the tooltip width? For the implementation to remain truthful the initial designs, I would recommend dropping the `max-width`declaration from the tooltip, as a container, and reapplying it to the `<span>` element with the tooltip's body text, with the value of `200px`. Using the CSS class names I found in Chrome dev tools in the above deployment, the CSS changes would literally translate to something as simple as this:
```css
.css-1wdn4wd {
   [...]
    /* max-width: 224px; */
   [...]
}
```

```css
.css-xrxhyo {
    [...]
    max-width: 200px;
}
```

Could we do that? Please, let me know, and if that's a 🟢 from you guys, then I'm updating docs in Figma.

kdembler (approved):

PR #2858, opened 2022-06-13
Textarea component

### LINKED ISSUE #2709: Update the Textarea component

drillprop:
Fix #2709
Design: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7088%3A144014

FYI I updated icons as well


toiletgranny (changes requested):
Nice! I really like the treatment for the drag handler. It looks sleek on Chrome and Safari, while not sacrificing other browsers such as Firefox. Also, based on our Slack discussions, I could not get the resize handler to break on any browser, which is great!

Apart from the above, there are a few other things I spotted, here's the list:

1. **As is:** The width of the bottom border is set to `1px` on focus.
**To be:** `2px` ⚠️ This needs to be addressed in the Input component too, I'm afraid.
2. **As is:** The color of the bottom border is set to `var(--color-border)` on both the default and hover states.
**To be:**  It should be `var(--color-border-alpha)` on default state, and `var(--color-border-strong-alpha)` on the default state. ⚠️ This needs to be addressed in the Input component too, I'm afraid.
3. **As is:** Input padding is set to `8px 16px` in the Medium size variant.
**To be:** `12px 16px`
4. **As is:** Input padding is set to `12px 16px` in the Large size variant.
**To be:** `16px 20px`
5. **As is:** There does not seem to be any text style set in the Input component at all.
**To be:** For both the user value and the placeholder, the text style should be `T200` in the Medium size variant, and `T300` in the Large size variant. Both regular.
6. **As is:** Both values are colored in the counter when the value is in range or out of it.
**To be:** Only the actual char counter should be colored red or strong. The upper limit value should remain gray.

packages/atlas/src/components/_inputs/TextArea/TextArea.styles.ts
```diff
@@ -1,30 +1,39 @@
...
+  ${getInputPadding}
+
+  ::-webkit-scrollbar-corner {
+    background: ${cVar('colorCoreNeutral800Lighten')};
+  }
 
   ::-webkit-resizer {
-    background: url(${resizeIcon}) no-repeat;
+    /* action-resize.svg with updated color */
+    background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 16 16' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M14 7.00003C14 6.59557 13.7564 6.23093 13.3827 6.07615C13.009 5.92137 12.5789 6.00692 12.2929 6.29292L7.29292 11.2929C7.00692 11.5789 6.92137 12.009 7.07615 12.3827C7.23093 12.7564 7.59557 13 8.00003 13H13C13.5523 13 14 12.5523 14 12V7.00003Z' fill='%237B8A95'/%3E%3C/svg%3E");
```
kdembler, line 84: Any specific reason to embed the svg here directly?

toiletgranny: Haha, that's a long story! How about we discuss it over a beer in two weeks? :D (just kidding, I'll let @drillprop do the talking, but let me just add it's some next-level trickery he's doing here to make it work as intended)

drillprop: - I tried to use SVG in the component, but this didn't work well when the scrollbar was visible.
- Importing SVG and placing it in URL as we did before(`background: url(${resizeIcon}) no-repeat;`) didn't work either, because we needed to set different `fill` color.  We would have to add SVG to some folder, manually change its color and then import it. 

That's why this is converted SVG by URL-encoder. I didn't find a better solution to do this.

drillprop:
> 1. **As is:** The width of the bottom border is set to `1px` on focus.
>    **To be:** `2px` ⚠️ This needs to be addressed in the Input component too, I'm afraid.
> 2. **As is:** The color of the bottom border is set to `var(--color-border)` on both the default and hover states.
>    **To be:**  It should be `var(--color-border-alpha)` on default state, and `var(--color-border-strong-alpha)` on the default state. ⚠️ This needs to be addressed in the Input component too, I'm afraid.
> 3. **As is:** Input padding is set to `8px 16px` in the Medium size variant.
>    **To be:** `12px 16px`
> 4. **As is:** Input padding is set to `12px 16px` in the Large size variant.
>    **To be:** `16px 20px`
> 5. **As is:** There does not seem to be any text style set in the Input component at all.
>    **To be:** For both the user value and the placeholder, the text style should be `T200` in the Medium size variant, and `T300` in the Large size variant. Both regular.
> 6. **As is:** Both values are colored in the counter when the value is in range or out of it.
>    **To be:** Only the actual char counter should be colored red or strong. The upper limit value should remain gray.

Everything should be fixed

kdembler (changes requested):
Screenshot 2022-06-14 at 11 08 35

drillprop:
> Screenshot 2022-06-14 at 11 08 35

should be fixed - I decided to not remove the box-shadow, only for textarea l removed box-shadow. Applying CSS border and reducing padding was really problematic - text was not vertically centered, and there was a slight shift when you focused on input(border in focused input has a different thickness - 2px, not 1px).

rafalpawlow (approved):

toiletgranny (approved):
I wasn't able to test all variants of this component in Storybook, due to missing controls that used to be there, but — overall — by the look of the adjustments I was able to test, I'm assuming the rest is also perfectly fine now. Good to go when it comes to me! 🚀

kdembler:
Unfortunately, looks like something is wrong on Firefox with the bottom border. It's few pixels above the bottom edge of the input:
Screenshot 2022-06-20 at 17 11 55

kdembler (approved):

PR #2761, opened 2022-05-28
upgrade to React 18, upgrade dependencies

kdembler:
Need to take a look at:
- [ ] Some Lottie animations are not playing when in `TransactionModal`
- [ ] Make sure autogenerated icons use proper imports

Summary of changes:

- upgraded almost all dependencies to latest versions
- upgraded to React 18, removed all `import React` imports, changed all import to use `import { ... } from 'react'`
- upgraded yarn version
- changed node version for running CI to 16
- made small fixes to `BottomDrawer`, it had some issues with the transitions

**Sorry for the amount of review, this is the only way to do this transition 😅**

PR #2860, opened 2022-06-13
upgrade to React 18, upgrade dependencies

kdembler:
Re-opened #2761

Summary of changes:

- upgraded almost all dependencies to latest versions
- upgraded to React 18, removed all `import React` imports, changed all import to use `import { ... } from 'react'`
- upgraded yarn version
- changed node version for running CI to 16
- made small fixes to `BottomDrawer`, it had some issues with the transitions

**Sorry for the amount of review, this is the only way to do this transition 😅**


packages/atlas/src/components/_inputs/Slider/Slider.tsx
```diff
@@ -24,7 +24,13 @@ export const Slider: React.FC<SliderProps> = ({
   return (
     <StyledRcSlider
       value={value}
-      onChange={onChange}
+      onChange={(value) => {
```
rafalpawlow, line 27: What happened here? Why we didn't need it before?

rafalpawlow (changes requested):
Do you think we could get rid of this warning? I belive this comes from zustand.
Zrzut ekranu 2022-06-14 o 10 19 17

packages/atlas/src/components/_inputs/Slider/Slider.tsx
kdembler: Yeah, previously we received a single value, now it's an array
Screenshot 2022-06-14 at 10 22 46
Screenshot 2022-06-14 at 10 28 22

rafalpawlow (approved):

PR #2862, opened 2022-06-13
Update Text component

### LINKED ISSUE #2783: Update Text component
Currently to add a color to the Text component we need to create new styled component each time.

**AC**
- [ ] replace current color prop (string) with predefined variants, get rid of `secondary` prop which is outdated
- [ ] remove colors from all styled Text components and replace it by new prop
- [ ] remove `h1, h2, h3...` tags in `styledVariants`, set default tag to span.
- [ ] make `as` prop required
- [ ] add `as` prop to every Text component used in the app 🌪️

rafalpawlow:
Closes #2783


packages/atlas/src/components/Text/Text.styles.ts
```diff
@@ -46,72 +74,72 @@ const clampStyles = ({ clampAfterLine }: TextBaseProps) => css`
   overflow: hidden;
 `
 
-const baseStyles = ({ color, secondary, clampAfterLine, margin, align }: TextBaseProps) => css`
-  color: ${color ? color : secondary ? cVar('colorText') : cVar('colorTextStrong')};
+const baseStyles = ({ color = 'strong', clampAfterLine, margin, align }: TextBaseProps) => css`
```
drillprop: I know that changing this will require a lot of work, but It would be much more intuitive for me if the default value were indeed `default` not `strong`.

rafalpawlow: As discussed on atlas dev sync, we decided to not implement this

packages/atlas/src/components/Text/Text.styles.ts
```diff
@@ -3,9 +3,37 @@ import styled from '@emotion/styled'
...
+  muted: cVar('colorTextMuted'),
+  strong: cVar('colorTextStrong'),
+  primary: cVar('colorTextPrimary'),
+  error: cVar('colorTextError'),
+  success: cVar('colorTextSuccess'),
+  caution: cVar('colorTextCaution'),
+  neutral100: cVar('colorCoreNeutral100'),
+  neutral200: cVar('colorCoreNeutral200'),
+  neutral200Lighten: cVar('colorCoreNeutral200Lighten'),
+  neutral500: cVar('colorCoreNeutral500'),
```
drillprop: I don't like the fact that we have a lot of random colors here.
What will happen if a designer adds text with the color that's not on this list of mapped colors? We will need to add the next color to this mappedColors, right? 
I think this should be either:
1. only text colors(`colorTextPrimary`, `colorTextSuccess`, `colorTextMuted` etc.) 
2. all the colors we have.

If we decide to pick the first option, we will need to style all Text components with emotion where we use non-standard colors
If we decide to pick the second option we would need to extract somehow all the colors from variables.

drillprop: What do you guys think about it @kdembler @eldiegod ?

drillprop (changes requested):
There is error in the console when I go to `/video/2` 
image

packages/atlas/src/components/Text/Text.styles.ts
rafalpawlow: @drillprop Maybe we should add some prop with custom color?

kdembler: Maybe just all the colors from tokens then?

drillprop: Prop with custom color? How it would work? 
I'd opt for getting all the colors from tokens. The thing is that it could be problematic with filtering types from `theme` only for colors. This is what `theme` object looks like in `variables.tsx`

```tsx
const theme = {
// color variables here
// animation variables here
// effects variables here
}
```
We could create separate variables for different types of tokens. For example: 
```tsx
const colorVariables = {
...
} 
const effectVariables = {
...
}
// and then merge them
const theme = {
...colorVariables,
...effectVariables
}
```
 If you want - I could refactor the style-dictionary config to make this easier. I did some experimenting yesterday and I'm close to accomplishing that.

kdembler: We can do it with just typescript: https://www.typescriptlang.org/play?#code/MYewdgzgLgBFAWBTAtomBeGBvAUDGA5KADYgBOAGgQFyEEA0ehJ5AmjXY-gRAJYBeiDgQZMeIVAl5gA5gFFiEIbRE4AvjhxQAngAc0AMV7EoiMgGUoAQzJRpMgOq8EAHgDSibRHowACmUQAM14ADxhEENMwABMIGGgyewA+DBgPL3DIxBi4gAMAEix-INC1QulAsxgAfQo1XJgAfjTPONowRAA3My09NABhEFIydLjMIxMzSxs7WSdXAGtPEEC4PpW4JFQfIiHyAiScIA

drillprop (approved):
As for me looks good. Unfortunately, you need to resolve conflicts once more.

Issue #2713, opened 2022-05-17
Update the Option card component


toiletgranny:
**AC:**

- [ ] Implement updates according to changelog (mostly design tokens)
- [ ] Add a `OptionCardRadioGroup` and `OptionCardCheckboxGroup` components, similar to `CheckboxGroup` or `RadioButtonGroup`

---

The goal of this task is to update the Option card component based on the designs, docs, and changelog, which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7382%3A144440). 

>  **Changelog (from Figma)**
> - Updated the overall styling,
> - Changed the label text style from H400 to H300,
> - Renamed the “Destructive” variant to “Error” (in the “Styling” property),
> - Updated borders in Error styling variants to use color design tokens,
> - Changed variants of Checkbox and Radio button components in Error styling variants to use matching styling properties,
> - Added Focus state, which inherits its styling from the Hover state.

![image](https://user-images.githubusercontent.com/35307309/169851337-841d2885-9485-4cfc-a428-5538089ab91c.png)

PR #2864, opened 2022-06-14
Option card

eldiegod:
fixes: #2713

### DIFF
--- a/packages/atlas/src/components/_inputs/OptionCard/OptionCard.styles.ts
+++ b/packages/atlas/src/components/_inputs/OptionCard/OptionCard.styles.ts
@@ -84,6 +84,7 @@ export const OptionCardLabel = styled.label<OptionCardLabelProps>`
     `}
 
   :active {
+    background: ${cVar('colorBackgroundMutedAlpha')};
     box-shadow: inset 0 -1px 0 0 ${getOptionCardBorderColorActive};
   }
 `
--- a/packages/atlas/src/components/_inputs/OptionCard/OptionCard.styles.ts
+++ b/packages/atlas/src/components/_inputs/OptionCard/OptionCard.styles.ts
@@ -58,12 +58,14 @@ export const IconContainer = styled.div<OptionCardLabelProps>`
 `
 
 export const OptionCardLabel = styled.label<OptionCardLabelProps>`
+  --border-height: ${({ checked }) => (checked ? '-2px' : '-1px')};
+
   padding: ${sizes(4)};
   display: flex;
   flex-direction: column;
   cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
-  box-shadow: inset 0 -1px 0 0 ${getOptionCardBorderColor};
-  transition: background-color ${cVar('animationTransitionFast')}, border-color ${cVar('animationTransitionFast')};
+  box-shadow: inset 0 var(--border-height) 0 0 ${getOptionCardBorderColor};
+  transition: all ${cVar('animationTransitionFast')};
   opacity: ${({ disabled }) => (disabled ? 0.5 : 1)};
   background: ${cVar('colorBackgroundMutedAlpha')};
 
@@ -73,7 +75,7 @@ export const OptionCardLabel = styled.label<OptionCardLabelProps>`
       :focus-within,
       :hover {
         background: ${cVar('colorBackgroundAlpha')};
-        box-shadow: inset 0 -1px 0 0 ${getOptionCardBorderColorHover({ checked, error, disabled })};
+        box-shadow: inset 0 var(--border-height) 0 0 ${getOptionCardBorderColorHover({ checked, error, disabled })};
 
         ${IconContainer} {
           * path {
@@ -85,7 +87,7 @@ export const OptionCardLabel = styled.label<OptionCardLabelProps>`
 
   :active {
     background: ${cVar('colorBackgroundMutedAlpha')};
-    box-shadow: inset 0 -1px 0 0 ${getOptionCardBorderColorActive};
+    box-shadow: inset 0 var(--border-height) 0 0 ${getOptionCardBorderColorActive};
   }
 `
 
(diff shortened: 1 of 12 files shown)


packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -387,15 +387,15 @@ export const VideoForm: FC<VideoFormProps> = memo(({ onSubmit, setFormStatus })
                 label="Public"
                 onChange={() => onChange(true)}
                 selectedValue={value?.toString()}
-                helperText="Visible to all"
+                caption="Visible to all"
```
drillprop: Since you created OptionCardGroup we could use it in form and see this component in action.

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/SetUp/SetUp.tsx
```diff
@@ -185,14 +185,14 @@ export const SetUp: FC<SetUpProps> = ({
                   <OptionCardRadio
                     value="open"
                     label="Open auction"
-                    helperText="Pick the winning bid or cancel anytime"
+                    caption="Pick the winning bid or cancel anytime"
```
drillprop: Same as above

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/ListingType/ListingType.tsx
```diff
@@ -26,15 +26,15 @@ export const ListingType: FC<ListingTypeProps> = ({ selectedType, onSelectType }
...
+          caption="Put it on a timed or open auction. See the bids coming."
         />
         <StyledOptionCardRadio
           onChange={() => onSelectType('Fixed price')}
           selectedValue={selectedType}
           value="Fixed price"
           label="Fixed price"
           icon={<SvgActionBuyNow />}
-          helperText="Sell it for a fixed price only. No bids."
+          caption="Sell it for a fixed price only. No bids."
```
drillprop: Same as above, I guess

packages/atlas/src/components/_inputs/OptionCard/OptionCard.styles.ts
```diff
@@ -59,23 +62,29 @@ export const OptionCardLabel = styled.label<OptionCardLabelProps>`
...
 
-  :hover {
-    border: 1px solid ${getOptionCardBorderColorHover};
+  ${({ disabled, checked, error }) =>
+    !disabled &&
+    css`
+      :focus-within,
+      :hover {
+        background: ${cVar('colorBackgroundAlpha')};
+        box-shadow: inset 0 -1px 0 0 ${getOptionCardBorderColorHover({ checked, error, disabled })};
```
drillprop: border thickness for the checked option card is 2px.

drillprop (changes requested):

packages/atlas/src/components/_inputs/OptionCard/OptionCard.styles.ts
drillprop: This is still wrong. Maybe I left a comment in the wrong place, so sorry for the confusion. 

When I hover on the unchecked option card the box-shadow changes to 2px: 
image
In the design it's 1px 
image

packages/atlas/src/components/_inputs/OptionCardGroup/OptionCardGroup.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { OptionCardGroupCheckbox } from './OptionCardGroup'
+
+import { OptionCardGroupProps, OptionCardGroupRadio } from '.'
+
+export default {
+  title: 'inputs/OptionCardGroup',
+  component: OptionCardGroupRadio,
+  args: {
+    label: 'Radio input label',
+    helperText: 'Radio helper text',
```
drillprop:
```suggestion
```
not needed

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -381,24 +380,23 @@ export const VideoForm: FC<VideoFormProps> = memo(({ onSubmit, setFormStatus })
...
-              <OptionCardRadio
-                value="false"
-                label="Unlisted"
-                onChange={() => onChange(false)}
-                selectedValue={value?.toString()}
-                helperText="Visible with link only"
-                disabled={videoFieldsLocked}
-              />
-            </RadioCardButtonsContainer>
+            <OptionCardGroupRadio
```
drillprop, line 386: Not working for me. I tried to edit my video and set it to unlisted and nothing happened once the transaction was done.

drillprop (changes requested):
Also, when you click on the option card, we should change the background of the OptionCard

image

eldiegod:
Fixed video unlisting and the rest of feedback

drillprop (approved):
LGTM!

toiletgranny (commented):
All changes listed in the changelog are implemented beautifully, thanks! 

The one little detail I'd consider is replacing the `<h6>` tag used for rendering the option card's label with a ` <label>` tag, if that's possible in our case. Would that be possible, @drillprop?

drillprop:
> All changes listed in the changelog are implemented beautifully, thanks!
> 
> The one little detail I'd consider is replacing the `<h6>` tag used for rendering the option card's label with a ` <label>` tag, if that's possible in our case. Would that be possible, @drillprop?

But there is a label already on this option card: 
image
Wrapping input in the label is much more comfortable because you don't need to set `for` for label and `id` for inputs. I'd opt not to change that.

But I also agree that we shouldn't use the h6 tag here, I guess this could be just `span`.

toiletgranny (approved):
Oh, haven't noticed that, but — of course — I trust your judgment. And yes, dropping the `<h6>` tag feels like a good idea!

PR #2866, opened 2022-06-15
Geolocation awareness in distributor selection

### LINKED ISSUE #2748: Introduce geolocation awareness in distributor selection
# Background

Currently, Atlas selects distributors randomly, more or less. This means that latency of fetching data, like videos and images, will possibly be quite high, because distributors may be very far away geographically/connectivity-wise from the relevant user.  Fixing this is a relatively cheap undertaking on the application side, and it can substantially improve the UX of Atlas, which he strategic value. It also can help the distributor working group get more realistic load patterns that can help them understand how to manage and maintain the infrastructure under more realistic circumstances than today, where the load is presumably uniform.

# Proposal

Add basic features in Atlas for this kind of selection. I don't think it is advisable to have any sort of user interaction in selecting this at any level, as its never something the user will want to be involved in. Hence there should not be any need for design work here.

rafalpawlow:
Closes #2748


packages/atlas/src/providers/assets/helpers.ts
```diff
@@ -66,3 +69,39 @@ export const testAssetDownload = (url: string, dataObject: StorageDataObjectFiel
...
+  const a =
+    Math.sin(deltaLatitude2 / 2) * Math.sin(deltaLatitude2 / 2) +
+    Math.cos(latitude1) * Math.cos(latitude2) * Math.sin(deltaLongitude / 2) * Math.sin(deltaLongitude / 2)
+  const c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a))
+
+  return Math.floor(R * c) // in metres
+}
+
+export const getUserLocation = async (): Promise<{ latitude: string; longitude: string }> => {
+  const storedLocation = localStorage.getItem('userLocation')
```
kdembler: This should use one of our stores, or if you think none match, a new one. But let's use zustand as well so all local storage is handled via it

packages/atlas-cloudflare-geolocation-worker/.gitignore
```diff
@@ -0,0 +1,10 @@
+/target
```
kdembler: Let's name the package `atlas-geolocation-worker` and rename the other one `atlas-proxy-worker` maybe?

packages/atlas/src/api/queries/__generated__/fragments.generated.tsx
```diff
@@ -1783,6 +1790,12 @@ export const DistributionBucketOperatorFieldFragmentDoc = gql`
     id
     metadata {
       nodeEndpoint
+      nodeLocation {
+        coordinates {
+          latitude
+          longitude
+        }
+      }
```
kdembler, line 1798: Seems only generated files got changed but our input graphql is the same?

packages/atlas/src/providers/assets/helpers.ts
```diff
@@ -66,3 +69,39 @@ export const testAssetDownload = (url: string, dataObject: StorageDataObjectFiel
...
+  const latitude2 = (lat2 * Math.PI) / 180
+  const deltaLatitude2 = ((lat2 - lat1) * Math.PI) / 180
+  const deltaLongitude = ((lng2 - lng1) * Math.PI) / 180
+
+  const a =
+    Math.sin(deltaLatitude2 / 2) * Math.sin(deltaLatitude2 / 2) +
+    Math.cos(latitude1) * Math.cos(latitude2) * Math.sin(deltaLongitude / 2) * Math.sin(deltaLongitude / 2)
+  const c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a))
+
+  return Math.floor(R * c) // in metres
```
kdembler: Could you confirm this package does the same calculation? https://github.com/dcousens/haversine-distance I think it would be good to use some package for this as it's easy to make a mistake in the math and we don't have any tests to even notice

packages/atlas/src/providers/assets/helpers.ts
```diff
@@ -66,3 +69,39 @@ export const testAssetDownload = (url: string, dataObject: StorageDataObjectFiel
...
+export const getUserLocation = async (): Promise<{ latitude: string; longitude: string }> => {
+  const storedLocation = localStorage.getItem('userLocation')
+  const parsedLocation = storedLocation && JSON.parse(storedLocation)
+  const sevenDaysMs = 604800000
+  const now = new Date()
+  if (!parsedLocation || now.getTime() > parsedLocation.expiry) {
+    const userLocation = await axios.get(USER_LOCATION_SERVICE)
+    localStorage.setItem(
+      'userLocation',
+      JSON.stringify({ value: userLocation.data, expiry: now.getTime() + sevenDaysMs })
```
kdembler: For expiry you could just use `add` from `date-fns`, then you don't even need `sevenDaysMs`

kdembler (changes requested):
Also, I get this error when I open the vercel deployment for the first time (without the location fetched already) on Firefox:
Screenshot 2022-06-15 at 13 23 20

packages/atlas/src/providers/assets/operatorsProvider.tsx
```diff
@@ -72,7 +90,19 @@ export const OperatorsContextProvider: FC<PropsWithChildren<unknown>> = ({ child
         // we need to filter operators manually as query node doesn't support filtering this deep
         const operatorsInfos: OperatorInfo[] = bucket.operators
           .filter((operator) => operator.metadata?.nodeEndpoint?.includes('http') && operator.status === 'ACTIVE')
-          .map((operator) => ({ id: operator.id, endpoint: operator.metadata?.nodeEndpoint || '' }))
+          .map((operator) => ({
+            id: operator.id,
+            endpoint: operator.metadata?.nodeEndpoint || '',
+            distance: operator.metadata?.nodeLocation?.coordinates
```
kdembler: We access `operator.metadata?.nodeLocation?.coordinates` 3 times here, can we store in some temporary variable?

packages/atlas/src/providers/assets/operatorsProvider.tsx
```diff
@@ -50,17 +54,31 @@ export const OperatorsContextProvider: FC<PropsWithChildren<unknown>> = ({ child
...
+    let userCoordinates: Coordinates
     const distributionOperatorsPromise = client.query<
       GetDistributionBucketsWithOperatorsQuery,
       GetDistributionBucketsWithOperatorsQueryVariables
     >({
       query: GetDistributionBucketsWithOperatorsDocument,
       fetchPolicy: 'network-only',
     })
+    if (!coordinates || !expiry || now.getTime() > expiry) {
+      const getUserCoordinates = await axios.get(USER_LOCATION_SERVICE)
```
kdembler: The variable is called like it was a function but it's a result of data fetch, maybe `userCoordinatesResponse`?

kdembler: Let's introduce a type with expected result value and use `axios.get<UserCoordiantesData>`

packages/atlas/src/providers/assets/operatorsProvider.tsx
```diff
@@ -72,7 +90,19 @@ export const OperatorsContextProvider: FC<PropsWithChildren<unknown>> = ({ child
         // we need to filter operators manually as query node doesn't support filtering this deep
         const operatorsInfos: OperatorInfo[] = bucket.operators
           .filter((operator) => operator.metadata?.nodeEndpoint?.includes('http') && operator.status === 'ACTIVE')
-          .map((operator) => ({ id: operator.id, endpoint: operator.metadata?.nodeEndpoint || '' }))
+          .map((operator) => ({
+            id: operator.id,
+            endpoint: operator.metadata?.nodeEndpoint || '',
+            distance: operator.metadata?.nodeLocation?.coordinates
+              ? haversine(
+                  { lat: Number(userCoordinates.latitude), lng: Number(userCoordinates.longitude) },
```
kdembler: Won't those values be numbers already?

packages/atlas/src/providers/assets/assetsManager.tsx
```diff
@@ -119,7 +118,7 @@ const sortDistributionOperators = (
   return [
     ...distributionOperators.slice(firstDistributorIndex),
     ...distributionOperators.slice(0, firstDistributorIndex),
-  ]
+  ].sort((a, b) => (a.distance || 0) - (b.distance || 0))
```
kdembler: If we're doing sorting here anyway, all that logic with slicing is unnecessary

kdembler: This will prioritize providers which don't have coordinates set because we use 0 as a fallback, right? Let's adjust the sorting logic so that if `distance` is missing, the provider will be put at the end of the list

kdembler (changes requested):

packages/atlas/src/providers/assets/operatorsProvider.tsx
kdembler: It would be good to handle error here in some way too. You know, if the service was temporary unavailable or something

kdembler (approved):

Issue #2127, opened 2022-02-15
Update file select component to support changes in Playlist: Channel owner


KubaMikolajczyk:
File select, dropzone, and file step components got an update to support the new usage of dropzone in the playlist workspace.
All three components are now documented in a new documentation style and ready to be implemented.

https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5664%3A176055
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5664%3A175508
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-Components?node-id=5664%3A174961

It's highly recommended to first have an intro call with a designer before starting to implement the changes.

kdembler:
Some feedback here: Currently the dropzone component defines only 2 properties, `Type` (Empty or Thumbnail) and `State`. However, it seems there's quite some more possible adjustments to this component. For example, like mentioned in the docs below, copy for specific step or whether the button is primary. I think we should define all possible variants on component-level, so we don't do manual overrides for specific usages. So I would expect to see variants for empty state for both thumbnail and video, and separate variant for playlists.

KubaMikolajczyk:
> Some feedback here: Currently the dropzone component defines only 2 properties, `Type` (Empty or Thumbnail) and `State`. However, it seems there's quite some more possible adjustments to this component. For example, like mentioned in the docs below, copy for specific step or whether the button is primary. I think we should define all possible variants on component-level, so we don't do manual overrides for specific usages. So I would expect to see variants for empty state for both thumbnail and video, and separate variant for playlists.

OK I will make all variants into components 👌

KubaMikolajczyk:
Hey it looks like I shouldn't have to do all of the components as separate instances in Figma - as the only difference occurs in case of empty states for all usages across the app - so I listed all empty states copy here: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=5664%3A175508
Do you think it can resolve this issue?

kdembler:
Well this component needs prop to control what's displayed inside it. For example, the button variant inside changes, so this needs to be controlled somehow. But let's not drag it, we will just do this ourselves in the implementation

kdembler:
[...]

PR #2876, opened 2022-06-17
File Select Updates

eldiegod:
fixes: #2127

### DIFF
--- a/packages/atlas/src/components/_inputs/FileSelect/FileSelect.styles.ts
+++ b/packages/atlas/src/components/_inputs/FileSelect/FileSelect.styles.ts
@@ -3,19 +3,79 @@ import styled from '@emotion/styled'
 import { animated } from 'react-spring'
 
 import { Text } from '@/components/Text'
-import { cVar, media, sizes, transitions } from '@/styles'
+import { cVar, media, sizes } from '@/styles'
 
 type DragAndDropAreaProps = {
   isDragAccept?: boolean
   isFileDialogActive?: boolean
+  fileAccepted?: boolean
 }
 
 type LoadingProp = {
   isLoading?: boolean
 }
 
+export const FileHoverOverlay = styled.div`
+  visibility: hidden;
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  z-index: 1;
+  border: 1px dashed ${cVar('colorCoreBlue500')};
+
+  ::before {
+    z-index: -1;
+    position: absolute;
+    content: '';
+    width: 100%;
+    height: 100%;
+    background: ${cVar('colorBackgroundOverlay')};
+  }
+`
+
+export const FileSelectedOverlay = styled.div`
+  position: absolute;
+  top: 0;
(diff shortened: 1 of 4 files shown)


KubaMikolajczyk:
Reviewing file select - everything looks fine. 
Just 2 things came up - in Storybook, there is an additional margin on the file selected on the left side.
image
And both in the storybook & implementation the file select stays at the end on the blue overlay which does not disappear - the expected behavior would be for the overlay to fade and show only the selected thumbnail.
image
image

No other issues spotted :)

eldiegod:
@KubaMikolajczyk I think the file selected overlay should be as expected now.

P.D: don't mind too much about the margin on the side, it's for presentational purposes in storybook

packages/atlas/src/components/_inputs/FileSelect/FileSelect.styles.ts
```diff
@@ -105,21 +133,19 @@ export const Thumbnail = styled(animated.img, { shouldForwardProp: (prop) => pro
   display: block;
   opacity: ${({ isLoading }) => (isLoading ? 0.1 : 1)};
   transition: opacity 400ms ease-out;
+  padding: 1px;
```
rafalpawlow: In figma thumbnail doesn't have this padding
Zrzut ekranu 2022-06-24 o 11 01 51

rafalpawlow (changes requested):

packages/atlas/src/components/_inputs/FileSelect/FileSelect.styles.ts
```diff
@@ -3,29 +3,87 @@ import styled from '@emotion/styled'
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  z-index: 1;
```
rafalpawlow, line 53: Can we use our `zIndex` variable?

packages/atlas/src/components/_inputs/FileSelect/FileSelect.tsx
```diff
@@ -16,24 +16,22 @@ import { FileType } from '@/types/files'
...
-  SelectedFileInfoBackground,
-  SelectedFileInfoHeading,
-  SelectedFileInfoInnerContainer,
   Thumbnail,
   Title,
 } from './FileSelect.styles'
 
 export type FileSelectProps = {
-  fileType: FileType
+  type: 'video-thumbnail' | 'playlist-thumbnail' | 'video-file'
```
rafalpawlow, line 30: Why do you resign from `FileType`? Can't we extend it to include playlist thumbnail?

rafalpawlow (changes requested):
Two small comments, other than this LGTM

packages/atlas/src/components/_inputs/FileSelect/FileSelect.tsx
eldiegod: Because playlist thumbnail it's not a `fileType` per se, that would make it confusing because there are a lot of differences between each type, and it doesn't just come to what type of file is accepted

rafalpawlow (approved):

PR #2892, opened 2022-06-21
Player settings component

### LINKED ISSUE #1503: Playback speed
### Context
I often want to do this, in particular to see reviews, e.g. of know-your-council. Not urgent, just giving my feedback.

### Scope
* [ ] Control playback speed 
* [ ] Speed setting stored in local storage and applies to next video in playback [applies to all settings]

### Designs
* Link to[ Figma user story](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2127%3A135540)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202474130431850) by [Unito](https://www.unito.io)

### LINKED ISSUE #2389: Settings component for Video Player
Figma link: https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2090%3A120527

**AC:**

* [ ] Create the settings component, make it trigger-able from the icon in the player
* [ ] Include UI of the settings (Speed, ~Subtitles~) without any logic
* [ ] Allow seeing details of a specific setting with animation

**Out of scope:**

* [ ] Shortcuts
* [ ] Any other settings

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202474128564457) by [Unito](https://www.unito.io)

drillprop:
Fix #2389 #1503
Designs: https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2090%3A120608

### DIFF
--- a/packages/atlas/src/components/_overlays/Popover/Popover.tsx
+++ b/packages/atlas/src/components/_overlays/Popover/Popover.tsx
@@ -1,8 +1,8 @@
 import styled from '@emotion/styled'
-import { Placement } from '@popperjs/core'
+import { Boundary, Padding, Placement } from '@popperjs/core'
 import Tippy from '@tippyjs/react/headless'
 import { ForwardRefRenderFunction, PropsWithChildren, ReactNode, forwardRef, useImperativeHandle, useRef } from 'react'
-import { Instance } from 'tippy.js'
+import { Instance, Plugin } from 'tippy.js'
 
 export type PopoverImperativeHandle = {
   hide: () => void
@@ -13,6 +13,8 @@ export type PopoverProps = PropsWithChildren<{
   trigger: ReactNode
   triggerMode?: string
   triggerTarget?: Element | Element[] | null | undefined
+  boundariesElement?: Boundary | null
+  boundariesPadding?: Padding
   placement?: Placement
   offset?: [number, number]
   hideOnClick?: boolean
@@ -33,6 +35,26 @@ const onTrigger = (instance: Instance<unknown>) => {
   })
 }
 
+const hideOnEscPlugin: Plugin = {
+  name: 'hideOnEsc',
+  defaultValue: true,
+  fn({ hide }) {
+    function onKeyDown(event: KeyboardEvent) {
+      if (event.key === 'Escape') {
+        hide()
+      }
+    }
+    return {
+      onShow() {
+        document.addEventListener('keydown', onKeyDown)
+      },
+      onHide() {
+        document.removeEventListener('keydown', onKeyDown)
+      },
+    }
+  },
+}
+
 const _Popover: ForwardRefRenderFunction<PopoverImperativeHandle | undefined, PopoverProps> = (
   {
     hideOnClick = true,
@@ -47,6 +69,8 @@ const _Popover: ForwardRefRenderFunction<PopoverImperativeHandle | undefined, Po
     className,
     disabled,
     flipEnabled = true,
+    boundariesElement,
+    boundariesPadding,
   },
   ref
 ) => {
@@ -55,6 +79,7 @@ const _Popover: ForwardRefRenderFunction<PopoverImperativeHandle | undefined, Po
   useImperativeHandle(ref, () => ({
     hide: () => tippyRef.current?.hide(),
     show: () => tippyRef.current?.show(),
+    unmount: () => tippyRef.current?.unmount(),
   }))
 
   return (
@@ -69,6 +94,7 @@ const _Popover: ForwardRefRenderFunction<PopoverImperativeHandle | undefined, Po
       onCreate={(instance) => {
         tippyRef.current = instance
       }}
+      plugins={[hideOnEscPlugin]}
       onTrigger={onTrigger}
       onShow={(instance) => {
         onTrigger(instance)
@@ -86,18 +112,30 @@ const _Popover: ForwardRefRenderFunction<PopoverImperativeHandle | undefined, Po
           }, EXIT_ANIMATION_DURATION)
         })
       }}
-      render={(attrs) => (
-        <ContentContainer {...attrs} className={className}>
-          {children}
-        </ContentContainer>
-      )}
+      render={(attrs) => {
+        return (
+          <ContentContainer {...attrs} className={className}>
+            {children}
+          </ContentContainer>
+        )
+      }}
       popperOptions={{
-        modifiers: [{ name: 'flip', enabled: flipEnabled }],
+        modifiers: [
+          { name: 'flip', enabled: flipEnabled },
+          {
+            name: 'preventOverflow',
+            enabled: !!boundariesElement,
+            options: {
+              boundary: boundariesElement,
+              padding: boundariesPadding,
+            },
+          },
+        ],
       }}
       placement={placement}
       offset={offset}
     >
-      <TriggerContainer tabIndex={0}>{trigger}</TriggerContainer>
+      <TriggerContainer tabIndex={1}>{trigger}</TriggerContainer>
     </Tippy>
   )
 }
--- a/packages/atlas/src/components/_overlays/Popover/Popover.tsx
+++ b/packages/atlas/src/components/_overlays/Popover/Popover.tsx
@@ -38,8 +38,8 @@ const onTrigger = (instance: Instance<unknown>) => {
 const hideOnEscPlugin: Plugin = {
   name: 'hideOnEsc',
   defaultValue: true,
-  fn({ hide }) {
-    function onKeyDown(event: KeyboardEvent) {
+  fn: ({ hide }) => {
+    const onKeyDown = (event: KeyboardEvent) => {
       if (event.key === 'Escape') {
         hide()
       }
--- a/packages/atlas/src/components/_overlays/Popover/Popover.tsx
+++ b/packages/atlas/src/components/_overlays/Popover/Popover.tsx
@@ -23,6 +23,7 @@ export type PopoverProps = PropsWithChildren<{
   onShow?: () => void
   disabled?: boolean
   flipEnabled?: boolean
+  animation?: boolean
 }>
 
 const EXIT_ANIMATION_DURATION = 100
@@ -70,6 +71,7 @@ const _Popover: ForwardRefRenderFunction<PopoverImperativeHandle | undefined, Po
     disabled,
     flipEnabled = true,
     boundariesElement,
+    animation = true,
     boundariesPadding,
   },
   ref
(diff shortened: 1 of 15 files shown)


kdembler (changes requested):
So nice to see this implemented, community will be so happy 😄 
Couple things from a very quick test that can be seen on this video:

https://user-images.githubusercontent.com/12646744/174808968-d512a6dc-1907-4bcb-baa2-bbd1d4da0f71.mov

1. The settings component overflows the video player and the overflown content is very poorly visible.
2. Because of above, it looks like there's no padding to the right of selected speed value
3. When I have the speed section open and I close player setting and open it back, the speed section is still opened. I would expect for the settings to reset once I close them

rafalpawlow:
Few things
1. Settings are under the video page ui
Zrzut ekranu 2022-06-21 o 15 13 39
2. wrong copy on the tooltip
Zrzut ekranu 2022-06-21 o 15 27 47
3. When I open speed control and close is without choosing any option it doesn't reset it's state after opening again

rafalpawlow (changes requested):
Few things
1. Settings are under the video page ui
Zrzut ekranu 2022-06-21 o 15 13 39
2. wrong copy on the tooltip
Zrzut ekranu 2022-06-21 o 15 27 47
3. When I open speed control and close is without choosing any option it doesn't reset it's state after opening again

rafalpawlow:
Oh, sorry I didn't notice @kdembler  made a review

kdembler:
@rafalpawlow you were requested so it's me that came here without invitation :D

rafalpawlow:
@kdembler No worries 😃

packages/atlas/src/components/_video/Settings/Settings.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+      options: availableQualities.map((q) => ({
+        checked: q === quality,
+        label: `${q}p`,
+        value: q,
+        onSettingClick: (opt) => setQuality(Number(opt.value)),
+      })),
+    },
+  ]
+  return (
+    <>
```
rafalpawlow: Do we need `Fragment` here?

packages/atlas/src/components/_video/Settings/Settings.tsx
```diff
@@ -0,0 +1,132 @@
...
+  onHeaderClick?: () => void
+  isOption?: boolean
+}
+
+export const SettingList: FC<SettingsListProps> = ({ settings, title, onHeaderClick, isOption = false }) => {
+  return (
+    <SettingsWrapper>
+      <SettingsHeader
+        isClickable={isOption}
+        onClick={(e) => {
```
rafalpawlow: Sorry, it's probably just me but I really dislike these meaningless arguments. I may be nitpicking, but I think the code should be as self-describing as possible. If you think I'm exaggerating, let me know 😛, if not can we fix these in all places? 🙏

packages/atlas/src/components/_video/Settings/Settings.tsx
```diff
@@ -0,0 +1,132 @@
...
+        }}
+      >
+        {isOption && <SvgActionChevronL />}
+        <Text variant="h100" secondary>
+          {title}
+        </Text>
+      </SettingsHeader>
+      <SettingsUnorderedList>
+        {settings?.map((setting, idx) => (
+          <SettingsListItem isOption={isOption} key={idx} {...setting} />
```
rafalpawlow: Using `index` as a key is not a good practice. Can we at least add some prefix? ```key={`player-settings-${idx}`}```

rafalpawlow (changes requested):

rafalpawlow (changes requested):
There is double gap between full screen and settings. I guess this is not intentional

Zrzut ekranu 2022-06-21 o 15 42 41

packages/atlas/src/components/_video/Settings/Settings.tsx
kdembler: That's true only if the options can change during runtime, if the options are constant then index should work fine

rafalpawlow: I agree, but I think it is better not to get your eyes used to this solution 😉

rafalpawlow: Now when I'm thinking about it, prefix is not a solution too. NVM, forget about it.

drillprop:
I had a call with Adam. We will need to rebuild this a bit:
- we decided to use ListItem
- settings have the wrong size in the fullscreen mode

drillprop:
Unfortunately, a few things changed while developing this component 😅 The design was very old when I started working on the component and we had to improve it a little bit. 
So to sum this up: 
- The base menu doesn't have the header "Settings". The header will only show in the sub-menu. 
image
- We decided to use ListItems components inside(in the submenu header as well)
- We added an option to disable autoplay(not included in the designs)
- We won't be adding shortcuts yet. At least it's not in the scope of this PR. 
- There is a different approach to the mobile. We will be showing modal instead of popover. ~I didn't implement this yet, I'll do this as a follow-up.~
- I also didn't implement animations. I'm a little concerned that this will be tricky.

New draft designs [here](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=5117%3A264802), full documentation is not there yet, it will be added in the future.
Also, during the review, take a look at the user stories: https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2132%3A135484. Some things are still valid.

kdembler:
@drillprop 2 cents from me: Let's try to not increase the scope too much in here and aim for minimal viable feature. Animations can be def tackled separately, maybe the modal for mobile as well, I'll leave that up to you

drillprop:
> @drillprop 2 cents from me: Let's try to not increase the scope too much in here and aim for minimal viable feature. Animations can be def tackled separately, maybe the modal for mobile as well, I'll leave that up to you

Okay added modal for mobile as well 😅 I'll ignore animations for now

toiletgranny (changes requested):
Bartek, this is delightful! The idea of putting settings dropdown in a dialog on mobile devices was spot-on, so thanks again for the willingness to go the extra mile to do that. I appreciate that a lot when working with you! Tomorrow's beer is on me! 

I found a couple of opportunities to apply mini-adjustments that — as a whole — can make a huge difference (hopefully). 

1. **A nice-to-have:** Would it be possible to prevent the player's UI from disappearing when the settings dropdown is opened?
2. **A nice-to-have:** Would it be possible to disable video player controls tooltips when the settings dropdown is opened? Tooltips interfere with the settings dropdown in browsers not supporting the `backdrop-blur` filter, such as Firefox. That's why I'm asking.
3. **A nice-to-have:** On mobile, despite the lack of header and the close icon in the settings dialog, would it be possible to close the dialog upon touching the backdrop overlay?
4. To make the settings dropdown consistent with the look of the ContextMenu component, I'd suggest applying `var(--radius-small)` border-radius to it.
5. On desktop, ideally, the settings dropdown appears with either (a) no animation, (b) the current animation but with the bottom right corner being the animation-origin, or (c) with a simple `translate: translateY(16px)` and `opacity: 0` transition.
6. On mobile exclusively, ideally, the settings dialog is dismissed upon selecting the desired speed preset or enabling/disabling the autoplay feature. As of now, I need to close the dialog manually after selecting the desired option, which is not ideal.
7. It seems like we're missing a divider above the dialog footer component in mobile settings. **[📷 Screenshot](https://user-images.githubusercontent.com/35307309/175493053-2aa40c19-a71b-465c-b123-027a90ee7b9c.png)**

drillprop:
@toiletgranny I think I fixed all the issues above

packages/atlas/src/components/_overlays/Popover/Popover.tsx
```diff
@@ -106,8 +146,8 @@ const TriggerContainer = styled.div`
   height: max-content;
 `
 
-const ContentContainer = styled.div`
-  transition: 150ms cubic-bezier(0.25, 0.01, 0.25, 1);
+const ContentContainer = styled.div<{ animation?: boolean }>`
+  transition: ${({ animation }) => (animation ? ' 150ms cubic-bezier(0.25, 0.01, 0.25, 1)' : 'unset')};
```
rafalpawlow, line 150: @drillprop @toiletgranny Do you think guys we could define it as design token?

packages/atlas/src/components/_video/VideoPlayer/SettingsButtonWithPopover.tsx
```diff
@@ -0,0 +1,134 @@
...
+  ]
+  return (
+    <span>
+      <Popover
+        placement="top"
+        disabled={mobile}
+        animation={false}
+        ref={popoverRef}
+        boundariesElement={boundariesElement}
+        boundariesPadding={{ right: 16 }}
```
rafalpawlow: Maybe we could use `sizes` helper here?

rafalpawlow (changes requested):
I'm experiencing a strange shifting when I open settings and wait few seconds. Also do you think we could get rid of this scroll? It's not really for anything

https://user-images.githubusercontent.com/13706246/175923085-89e5b8b5-74cb-4496-a310-f847b5934fa3.mov

packages/atlas/src/components/_overlays/Popover/Popover.tsx
drillprop: I think we all need some context here, so let me try to explain. Rafał is referring to the animation of the Popover component(which is mostly used for the context menu) and it has some old value  - `150ms cubic-bezier(0.25, 0.01, 0.25, 1)'`. In this PR I only add a possibility to  **disable** animation for the purpose of using settings popover. So this value is just a default animation transition when it's not **disabled**
@toiletgranny we could change this to our token or we could add another one if you want. But on the other hand, I don't want to increase the scope of this PR any further.

packages/atlas/src/providers/personalData/store.ts
```diff
@@ -15,6 +15,8 @@ export type PersonalDataStoreState = {
   cinematicView: boolean
   cookiesAccepted?: boolean
   reactionPopoverDismissed: boolean
+  playbackRate: number
+  playNext: boolean
```
kdembler: Could we maybe call this `autoPlayNext` or something else? Current name is a bit of mystery

packages/atlas/src/components/_overlays/Popover/Popover.tsx
toiletgranny: @drillprop is right, our intention was not to change the animation here, it was to have an option to disable it. However, since we're here, and in case we won't be visiting this component in a very long while, I guess @rafalpawlow has a good point that it wouldn't hurt to change the hardcoded animation value with our token. The closest one matching these settings would be `var(--animation-transition-fast)`. Would that be a simple thing to do, @drillprop?

kdembler (approved):
LGTM!

toiletgranny (approved):

rafalpawlow (approved):

PR #2897, opened 2022-06-21
Update Dialog crop component

### LINKED ISSUE #2173: Update the Dialog crop component
**AC:**

- [ ] Implement all the design changes mentioned below with this exception: The "Delete" button should be configurable via some prop. If it's not enabled (for example if the user uploads the image for the first time), the button should be hidden, not disabled.

---

There are old "crop & position" dialogs which need to be adjusted to be fully funcionable in new context of setting playlist thumbnails.

rafalpawlow:
Closes #2173 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6341%3A137858

### DIFF
--- a/packages/atlas/src/components/_inputs/MultiFileSelect/MultiFileSelect.tsx
+++ b/packages/atlas/src/components/_inputs/MultiFileSelect/MultiFileSelect.tsx
@@ -181,7 +181,7 @@ export const MultiFileSelect: FC<MultiFileSelectProps> = memo(
 
     const handleReAdjustThumbnail = () => {
       if (files.thumbnail?.originalBlob) {
-        dialogRef.current?.open(files.thumbnail.originalBlob)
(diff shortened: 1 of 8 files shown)


kdembler (changes requested):
There's a bunch of weirdness happening with responsiveness here. This was also broken before but could you maybe take this occasion to try and fix some of those?

https://user-images.githubusercontent.com/12646744/175263407-500ece45-eb27-45c8-9f0c-608f06334f1b.mov

rafalpawlow:
@kdembler OMG, this was such pain in the ass. I did my best, but unfortunately it's not perfect (but better than before). There are a lot of similar issues with resizing on `cropperjs` repo, but unfortunately they are not solved... Let me know what your thoughts are, and if you have any suggestions.

kdembler (approved):

PR #2900, opened 2022-06-23
Get rid of search query

### LINKED ISSUE #2882: Get rid of `search` query usage
- [ ] Instead of using the `search` query we can run separate queries for videos and channels including `where: { title_contains: "xxx" } ` for filtering
Sorting would be something like this `[reactionsCount_Desc, commentsCount_Desc, createdAt_Desc]` 
- [ ] Introduce pagination for search results with list of pages

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202474369687611) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #2882


kdembler (changes requested):
From a quick look at the app:
1. The grid of results is always not full for me - there are a couple of items missing to fill the full page.
2. When I switch page, for a second there's no content in the grid - instead we should display skeleton loaders
3. We fire a request with `title_contains: ""`, let's postpone until user types something
4. We first fire a request for 50 items and then restrict that to 20 or something. Why do we need 50 to begin with?

kdembler:
1. I type in query in the searchbox
2. I wait for short results to load
3. I press enter
4. There are only 3 video results

packages/atlas/src/hooks/useSearchResults.ts
```diff
@@ -43,29 +55,68 @@ export const useSearchResults = ({ searchQuery, limit = 50, videoWhereInput }: S
...
+      first,
+      where: {
+        title_contains: text,
+        avatarPhoto: {
+          isAccepted_eq: true,
+        },
+      },
+    },
+    {
+      fetchPolicy: 'cache-and-network',
```
rafalpawlow: @kdembler  Not sure if this solution is correct, let me know

drillprop (changes requested):
Found another bug, related to cache.
1. search "joystream"
2. In the result you should see lots of results on the first page
3. Click the search input again(just click, don't submit or clear the input)
4. Only three videos are shown on the first page. 

Video: 

https://user-images.githubusercontent.com/51168865/175876557-4babd098-1a0e-4d20-b3b8-bff71251afa1.mov





I also found that the performance of the search input is pretty bad. I need to wait for at least a second to see the text in the box. I'm not sure if this PR introduced that, but on production I didn't notice much performance drop.

https://user-images.githubusercontent.com/51168865/175877572-a0b7f67e-b7fd-4294-bbc0-333380ec9a5f.mov

This is how it works play.joystream.org

https://user-images.githubusercontent.com/51168865/175877922-3e76484c-81ed-426b-9175-8f90c371d5e5.mov

drillprop:
Ok, that's weird. I've noticed that the performance of the input drops only if you have dev tools open. I guess we could ignore this for now(?)

rafalpawlow:
@drillprop god, this cache problem is a nightmare.

kdembler (changes requested):
Because of the `no-cache` policy, all the VideoTiles on the search results page cannot access the data from cache so tens of `GetBasicVideo` requests are fired

kdembler:
Also, if I go to the 2nd page of results, after loading I don't see anything

packages/atlas/src/hooks/useSearchResults.ts
kdembler: I'm not sure what problem you are trying to solve with this (and it's changed since then, right?)

rafalpawlow: I have a problem with `first` parameter. When we fire search query from `SearchBox` for first time, we query videos and channels with `first: 3`. When you press enter, we query videos with `first: 20` but in network tab it stays `3`. There is `first` in `cache.ts` but it seems it doesn't do anything.

kdembler: `first` shouldn't be in cache, we want that to be a single list in the cache. We can sync up on that if you want

drillprop (changes requested):
This doesn't work either. Now I have only three results on the first page

drillprop (approved):
FYI I created an issue to investigate performance drops on Searchbar https://github.com/Joystream/atlas/issues/2922 
Besides, in terms of functionality LGTM!

packages/atlas/src/components/Searchbar/SearchBox/SearchBox.tsx
```diff
@@ -43,7 +43,7 @@ export const SearchBox: FC<SearchBoxProps> = memo(
     onMouseMove,
     hasFocus,
   }) => {
-    const { channels, videos, loading } = useSearchResults({ searchQuery })
+    const { channels, videos, loading } = useSearchResults({ searchQuery, first: 3, fetchPolicy: 'no-cache' })
```
kdembler, line 46: Do we need that fetch policy here?

kdembler (approved):
One comment, other than that looks good to me as well!

PR #2901, opened 2022-06-23
update sign in flow to support multiple wallets

### LINKED ISSUE #2817: Update the Sign in flow and Transaction dialog component to support Talisman
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202474194014638) by [Unito](https://www.unito.io)

kdembler:
Resolves #2817

Designs: https://www.figma.com/file/ymHYOFItrmFaLeBC5Z4tpQ/Sign-in?node-id=1%3A307


toiletgranny (changes requested):
This felt like unpacking a birthday gift! This is fantastic, and the implementation is splendid! 🥰 Here's a couple minor things I have found:

1. Having signed in with Polkadot.js wallet, upon adding new membership, I can see there's a "Back" button. Upon clicking 
the button, I'm redirected to the previous step (wallet selection), where I'm able to select Talisman. Doing so, however, closes the Sign in dialog with no effect, presumably because I haven't disconnected my Polkadot.js wallet first. Instead, in such a scenario, I would suggest either disconnecting the currently selected wallet so that I, as a viewer, can indeed set up an account from a different wallet, OR, dropping the "Back" button entirely in that scenario. 

https://user-images.githubusercontent.com/35307309/175477561-d0c2cb20-9a7c-4f66-915f-a428b4fc153f.mp4

2. In the "Create Joystream membership" step, upon scrolling up/down, while having my cursor placed over the text field, validation is triggered. Why is that? Can we fix that?

https://user-images.githubusercontent.com/35307309/175477580-eec5810f-b576-47b5-9e8c-3a05a7633079.mp4

3. Upon providing an invalid URL in the Avatar URL text field, the avatar component is displayed momentarily, featuring a "broken image". Can we display the Avatar component only if the user-provided URL is valid? 🙏

kdembler:
All should be fixed now

toiletgranny:
Great, thanks! I can confirm all issues I've mentioned are fixed. But now they led me to find a few more. This, I would say, are rather critical.

1. In the "Select account" step, where the app is waiting for the user to create new or allow access for existing wallet accounts to be connected with the site, the copy next to the spinner doesn't mention the wallet name. In the design, the copy says "Waiting for new [wallet name] accounts...". Ideally, the implementation matches the designs in this case. Mentioning the wallet name in the copy had a very specific purpose, which is to reassure the user on which wallet is currently selected and expected to yield new accounts. 
CleanShot 2022-06-24 at 13 24 33
2. Because of the above issue, it was more difficult for me to find this one, which is a bit more severe. Namely, when I select a certain wallet, but then change my mind, go back and select a different one, the app will continue to fetch account data from the initially selected wallet. And it will continue to do so until I refresh the app. Please, see the below recording in which I first select Talisman, then go back and select Polkadot, but the app continues to display Talisman accounts.

https://user-images.githubusercontent.com/35307309/175526491-68c26c16-7b3a-4a4c-9305-b25f79a0e5f3.mp4

packages/atlas/src/components/LottiePlayer/LottiePlayer.tsx
```diff
@@ -1,25 +1,36 @@
...
-      autoplay={play}
-      loop={loop}
-      keepLastFrame
-      src={data}
-      style={size ? { width: size, height: size } : {}}
-      onEvent={(e) => (e === 'complete' ? onComplete?.() : null)}
-      className={className}
-    />
+    <>
+      <Global styles={playerContainerOverrides} />
```
drillprop, line 18: can't we just style the Player?

drillprop (changes requested):
Works pretty well, although found two things: 
1. I had a situation when even though I had talisman membership I still saw this account in `Select Account` list, I think this is happening mostly if you playing around with connecting and disconnecting accounts. Not sure if it's fixable: 

https://user-images.githubusercontent.com/51168865/175537914-31702f2a-bce8-412b-8e39-7aba6cee9aa5.mov

2. If I select wallet and go back and pick another wallet I see accounts from the previous wallet

Edit: AD.1 Oh, and if you look at the left bottom edge of the video, you'll notice that I also got a successful snack bar, but couldn't reproduce this once more.  
AD.2 I guess Adam, already pointed this out

packages/atlas/src/components/LottiePlayer/LottiePlayer.tsx
kdembler: That was my first approach but unfortunately the library passes the `className` to a nested node, the direct container of the animation, and not the outermost container, the one I style with the class name

kdembler:
Both points from @toiletgranny review now fixed. I'm looking into this issue with Talisman accoutns

drillprop (changes requested):
Found one more really minor thing: 

https://user-images.githubusercontent.com/51168865/175541105-ca5231dd-6528-4928-98a7-b19df086c9d3.mov

If I scroll down on the select account step, the next step is also scrolled down. I think we should show the top of the terms of service.

drillprop (approved):

toiletgranny (approved):

PR #2913, opened 2022-06-24
Fix search button in ChannelView

### LINKED ISSUE #2879: Broken channel search button
After some changes on `dev`, the search button in channel view started looking wrong:
Screenshot 2022-06-20 at 13 22 52
 
If needed, ask Adam on how this should work (probably the same as it did before)

drillprop:
Fix #2879

### DIFF
--- a/packages/atlas/src/views/viewer/ChannelView/ChannelSearch.styles.ts
+++ b/packages/atlas/src/views/viewer/ChannelView/ChannelSearch.styles.ts
@@ -11,19 +11,21 @@ export const StyledInput = styled(Input)<TextFieldProps>`
   width: 100%;
 
   /* to align it with SearchIcon button */
-  margin-left: -${sizes(1)};
+  ${media.sm} {
+    margin-left: -${sizes(1)};
+  }
 `
 
 export const SearchContainerForm = styled.form`
   display: flex;
   grid-area: search;
   align-items: center;
-  margin: ${sizes(8)} ${sizes(1)} ${sizes(2)} 0;
+  margin: ${sizes(8)} 0 ${sizes(2)} 0;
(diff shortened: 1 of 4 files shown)


toiletgranny (changes requested):
@drillprop Working as expected, awesome stuff! 

It was only now that I noticed the incorrect position of the Button component in the Small variant of the Input component. Its current distance from the right edge of the component is set to `12px`, but, since it's a small size, it should be set to `8px`. Is this something we could adjust with this PR?

CleanShot 2022-06-24 at 13 56 53@2x


Also, the previous implementation of the search input had a search icon. Could we bring it back here? 🙏 

Other than this, it's all good!

toiletgranny (approved):

packages/atlas/src/views/viewer/ChannelView/ChannelSearch.styles.ts
```diff
@@ -1,57 +1,24 @@
...
-      -webkit-appearance: none;
-    }
-  }
-`
+  width: 100%;
 
-export const SearchButton = styled(Button)`
-  position: absolute;
+  /* to align it with SearchIcon button */
+  margin-left: -${sizes(1)};
```
kdembler: We should probably apply this only on desktop:
Screenshot 2022-06-27 at 09 29 50

kdembler (changes requested):

packages/atlas/src/views/viewer/ChannelView/ChannelSearch.styles.ts
drillprop: fixed

kdembler (approved):

Issue #2847, opened 2022-06-09
Extrinsics signatures update


dmtrjsg:
Will depend on storage details ticket to be done at the same time

https://github.com/Joystream/atlas/issues/2825

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202419733347150) by [Unito](https://www.unito.io)

Issue #2825, opened 2022-06-08
More storage deets for uploading assets extrinsic


dmtrjsg:
App must select buckets that comply with policies, chain does not do it automatically

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202410590044249) by [Unito](https://www.unito.io)

PR #2926, opened 2022-06-28
initial carthage setup, update to new polkadot versions

### LINKED ISSUE #2898: Update to use new Polkadot library
Based on outcome here: https://github.com/Joystream/joystream/issues/3939

kdembler:
Closes #2898, #2847

Scope of changes:
- Upgraded to new @joystream/types which has a lot of breaking changes
- Upgraded all Polkadot libraries to new versions
- Updated signatures for all extrinsics to match Carthage runtime

_Note_: At the moment this is non-testable yet as not all of the software stack has been upgraded to work with Carthage (mostly QN). So for now, let's review the code and we will test it out at a later point


packages/atlas/src/joystream-lib/config.ts
```diff
@@ -1,2 +1,21 @@
...
+  '6': 1,
+  '7': 1,
+  '8': 1,
+  '9': 1,
+  '10': 1,
+  '11': 1,
+  '12': 1,
+}
+
+export const STORAGE_BUCKETS_COUNT = 2
```
kdembler, line 21: This is unused at the moment but will be needed for #2825

drillprop (approved):

PR #2934, opened 2022-06-29
Member dropdown publisher variant

### LINKED ISSUE #2873: Channel accounts
**AC:**

- [ ] Change the publisher variant of member dropdown to display the channel (and its balance)
- [ ] Create the withdraw modal
- [ ] Create the send fund modal 

**Out of scope:**

- [ ] Extrinsic integration
- [ ] Fee amount in the Dialog

---

# Background

In `Carthage` and beyond, all funds for a channel, that is from NFT sales, payouts, etc. will not land in a personal account hte user has in their wallet, but in a _channel account_. Currently we don't show this in Atlas, also, we do not allow the user to withdraw funds from this account to a personal account. The latter is a planned feature for the full `My PAyments` screen, but that will not land for a long time.

In response to a question about this from @toiletgranny I say

> I agree, its actually kind of surprising we did not think of showing the reward account for a channel in the UI more prominently before in the NFT release, there the account which people will consistently see is the membership account, and I would have expected them to be confused about it not increasing when they get NFT revenue. This is actually quite a big problem to not have. Introducing this could actually also create the space to solve the problem that Klaudiusz recently pointed out that there is no way to “claim” NFT revenue which in Carthage and beyond will no longer land in a personal account, but in a channel account.

# Proposal

Introduce a super cheap light weight way to show the channel balance, and allow user to withdraw from this into the user account, which we can ship from `Carthage`, and use until full `My Payments` lands.

**NOTE: It is pivotal to make this a _cheap_ intervention, as the lifetime for this will be a few weeks/months, and Carthage timeline is tight.**

rafalpawlow:
Closes #2873 

**Designs**
Member dropdown: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3201%3A111989
Dialogs: https://www.figma.com/file/uHzbhNX9dP4QtZbIapvZI6/Transfer-dialog?node-id=1%3A307


packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdown.styles.ts
```diff
@@ -119,12 +120,22 @@ export const Divider = styled.div`
   background: ${cVar('colorBorderMutedAlpha')};
 `
 
-export const LearnAboutLink = styled(Text)`
+export const Link = styled(Text)`
```
drillprop: I would not name this Link, it could be confused with react-router Link

packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdown.tsx
```diff
@@ -135,147 +147,188 @@ export const MemberDropdown = forwardRef<HTMLDivElement, MemberDropdownProps>(
...
+      <>
+        <WithdrawDialog
+          avatarUrl={avatarUrl}
+          activeMembership={activeMembership}
+          show={showWithdrawDialog}
+          onExitClick={toggleWithdrawDialog}
+          accountBalance={accountBalance}
+          channelBalance={channelBalance}
+        />
+        <SendDialog show={showSendDialog} onExitClick={toggleSendDialog} accountBalance={accountBalance} />
```
drillprop: Maybe I nitpick a bit here, but I don't like the name. We actually don't know what we send in this dialog based on its name. Couldn't we just rename this to SendTransferDialog?

packages/atlas/src/components/_overlays/SendTransferDialogs/WithdrawDialog.tsx
```diff
@@ -0,0 +1,119 @@
...
+  show,
+  avatarUrl,
+  channelBalance = 0,
+}) => {
+  const [amount, setAmount] = useState<number | null>(null)
+  const { convertToUSD } = useTokenPrice()
+  const convertedAmount = convertToUSD(amount || 0)
+  const [error, setError] = useState<string>()
+
+  const handleWithdraw = () => {
```
drillprop: Let's use `useForm` here. Validation should work on submit and after you submit with wrong values, the mode of validation should change to `onChange`(I think this is the default behavior of react-hook-form).
Right now, after you submit and you focus on the input, the error stays until you click submit again, which is not intuitive in my opinion.

packages/atlas/src/components/_overlays/SendTransferDialogs/SendDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+          variables: { where: { controllerAccount_eq: val } },
+        })
+        setDestinationAccount(memberships.length ? memberships[0] : undefined)
+      } catch (error) {
+        SentryLogger.error('Failed to fetch memberships', 'WhiteListTextField', error)
+      }
+    }, 500)
+  )
+
+  const handleSend = () => {
```
drillprop: same as above

drillprop (changes requested):
You missed joy token icons in inputs and max label above the `Amount to transfer input` 
image

image
~Also shouldn't we show the channel avatar on the member dropdown in the viewer?~ Edit: nvm

packages/atlas/src/components/_overlays/SendTransferDialogs/WithdrawFundsDialog.tsx
```diff
@@ -0,0 +1,149 @@
...
+  avatarUrl,
+  channelBalance = 0,
+}) => {
+  const {
+    handleSubmit,
+    watch,
+    reset,
+    register,
+    formState: { errors },
+  } = useForm<{ amount: number | null }>({ mode: 'onSubmit' })
```
drillprop: We don't need to write `onSubmit` it's default behavior.

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const { convertToUSD } = useTokenPrice()
+  const client = useApolloClient()
+  const {
+    register,
+    reset,
+    handleSubmit,
+    watch,
+    setValue,
+    formState: { errors },
+  } = useForm<{ amount: number | null; account: string | null }>({ mode: 'onSubmit' })
```
drillprop: Same as above

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -0,0 +1,195 @@
...
+              />
+            }
+            placeholder={`${JOY_CURRENCY_TICKER} amount`}
+            error={!!errors.amount}
+          />
+        </FormField>
+        <FormField label="Destination account" error={errors.account?.message}>
+          <Input
+            {...register('account', {
+              validate: {
```
drillprop, line 163: Should we also throw an error if you type the wrong address?

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -0,0 +1,195 @@
...
+                Amount to withdraw
+              </Text>
+              <Button onClick={handleMaxClick} size="medium" variant="tertiary" _textOnly>
+                Max
+              </Button>
+            </LabelFlexWrapper>
+          }
+          error={errors.amount?.message}
+        >
+          <Input
```
drillprop, line 124: I'm wondering - could we use the FormField prop `disableErrorAnimation` to not shake the input when typing(when revalidating onChange) after an unsuccessful submission? I believe that there is something like `submissionCount` or `submitted` in react hook form.  

I mean, I'm fine with leaving this as it is. Just a suggestion

packages/atlas/src/components/_overlays/SendTransferDialogs/SendTransferDialogs.styles.ts
```diff
@@ -0,0 +1,47 @@
+import { css } from '@emotion/react'
+import styled from '@emotion/styled'
+
+import { sizes } from '@/styles'
+
+const verticallyCenteredFlex = css`
+  display: flex;
+  align-items: center;
+`
```
drillprop: This is completely redunant. We could just create: 
```tsx
export const VerticallyCenteredDiv = styled.div`
  display: flex;
  align-items: center;
`
```
and use it in two places

drillprop (changes requested):
1. There is a small shift in MemberDropdown: 
image
I believe I introduced this bug, but can you fix this? You could just add `justify items: start` to the wrapping div.

2. I'm not sure if we should fix this here, but when a member doesn't have a channel and you click the `my channel`, we're going to: `channel/:id?tab=Videos` which is misleading.
image
Shouldn't we just redirect user to `/studio/signin`? There is a call-to-action button `Create first channel` on sign-in view.

drillprop (approved):

PR #2945, opened 2022-06-30
Fees support in components

rafalpawlow:
Closes #2815 

**Designs**
- Designs, docs, and changelog for **comments & reactions** [can be found here](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2140%3A135509),
- Designs, docs, and changelog for the **Action bar component** (a footer we display in _NFT minting / buying / bidding_, _Video Posting / editing/ deleting_, _Updating Membership_, _Channel Creation / Editing_) [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=8477%3A161454),
   - I mentioned this change [in this PR](https://github.com/Joystream/atlas/pull/2857#issuecomment-1160498065), as you requested Klaudiusz.   
- **🆕 NEW** Designs, docs, and changelog for the **Dialog footer component** [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=8638%3A171330). Once this is implemented, we can proceed with the following changes:
   - Add the fee info to the delete video alert dialog, designs for which [can be found here](https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=3171%3A168073),
   - Add the fee info to the remove NFT from sale alert dialog, designs for which [can be found here](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=5076%3A261284),
   - Add the fee info to the remove change NFT price dialiog, designs for which [can be found here](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=4924%3A266241).

PR #2963, opened 2022-07-04
Fee calculation support

### LINKED ISSUE #2911: Fee calculation support
At the moment, Atlas will create a transaction and send it in a single flow. For example, let's take a look at `createVideo` function in `packages/atlas/src/joystream-lib/extrinsics.ts`. It parses the inputs, creates `VideoCreationParameters`, creates the transaction itself (`const tx = `), sends it and finally parses the outputs of the transaction.

To support fee calculation that will need to be split into separate steps. Take a look at this snippet from Pioneer: https://github.com/Joystream/pioneer/blob/ce138cba21be52f7e01343105abd0742345f278c/packages/ui/src/accounts/modals/RecoverBalance/RecoverBalanceModal.tsx#L22-L40 They re-create the transaction on any input change and then call a hook that given the transaction, can estimate the fees that will be used. I think their specific approach can be improved but that's a good starting point.

My suggestion based on the `createVideo` example would be:
1. Split `createVideo` into 2 function `createVideoTx` and `sendCreateVideoTx`:
    1. `createVideoTx` would accept all the inputs and return the `tx` that can be used for estimation.
    2. `sendCreateVideoTx` would accept a `tx` input, send it and parse the outputs
2. We create some hook that will take care of estimating fees. I think this definitely should be debounced so that we don't send a request to the RPC node on every input change. I imagine the API could look something like this: `const { tx, estimatedFee, hasEnoughFunds } = useTx(createVideoTx, [memberId, channelId, inputMetadata, nftInputMetadata, inputAssets])`. This way we could have all the debounce and estimation logic inside that hook, making changes easy in the future. We don't have to do it this way though, I'm open to other approaches
3. We use the estimation hook in every component that sends a transaction

Also see https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-estimate-the-transaction-fees

drillprop:
Fix #2911 
I didn't use hook inside video form and NFT form in the end. There are too many moving parts, and we don't have a fully working playground right now. I'm afraid that I could introduce more bugs. 

To test it out you need to set up a local node.

### DIFF
--- a/packages/atlas/src/components/ActionBar/ActionBar.tsx
+++ b/packages/atlas/src/components/ActionBar/ActionBar.tsx
@@ -28,6 +28,7 @@ type ActionDialogInfoBadge = {
 
 export type ActionBarProps = {
   fee?: number
+  feeLoading?: boolean
   infoBadge?: ActionDialogInfoBadge
   primaryButton: ActionDialogButtonProps
   secondaryButton?: ActionDialogButtonProps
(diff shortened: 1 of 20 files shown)


rafalpawlow:
@drillprop What about fee in comments and reactions?

drillprop:
> @drillprop What about fee in comments and reactions?

There is support to calculate fees in comments and reactions in this PR. We don't have UI yet though, I'll add this once https://github.com/Joystream/atlas/pull/2945 is merged.

kdembler (approved):
Code looks good to me. Testing fully is still blocked by other Carthage pieces so let's merge as-is for now and test later with other parts

PR #2965, opened 2022-07-05
Top overlay in embedded player

### LINKED ISSUE #2935: Top of the embedded player overlay
## Scope
Implement Embedded player top overlay with sharing button 

## AC
- [ ] Embedded version has top of screen overlay
- [ ] Sharing button click summons sharing screen

Out of scope:

- #2936 
- https://github.com/Joystream/atlas/issues/2943

## Design
Screenshot 2022-06-29 at 17 27 28

[Figma link >](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2583%3A232036)

drillprop:
Fix #2935

### DIFF
--- a/packages/atlas/src/components/_video/VideoPlayer/PlayerControlButton.styles.ts
+++ b/packages/atlas/src/components/_video/VideoPlayer/PlayerControlButton.styles.ts
@@ -1,4 +1,4 @@
-import { css } from '@emotion/react'
+import { SerializedStyles, css } from '@emotion/react'
 import styled from '@emotion/styled'
 
 import { Text } from '@/components/Text'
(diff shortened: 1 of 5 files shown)


packages/atlas/src/components/_video/VideoPlayer/PlayerControlButton.styles.ts
```diff
@@ -119,6 +119,14 @@ export const ControlButtonTooltip = styled.div<ControlButtonTooltipProps>`
   padding: 0.5em;
   white-space: nowrap;
   transition: opacity ${transitions.timings.player} ease-in, backdrop-filter ${transitions.timings.player} ease-in;
+
+  ${({ tooltipPosition }) =>
```
rafalpawlow: Can we make some `getTooltipPosition` function which will cover all scenarios?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
```diff
@@ -91,9 +93,51 @@ export const StyledSvgControlsSettingsSolid = styled(SvgControlsSettingsSolid)`
...
   ${defaultIconColor};
 `
+export const StyledSvgControlsShare = styled(SvgControlsShare)`
+  ${defaultIconColor};
+`
 
 export const TRANSITION_DELAY = '50ms'
 
+export const EmbbeddedTopBar = styled.div<{ isFullScreen: boolean }>`
+  font-size: ${sizes(4)};
```
rafalpawlow: Didn't we agree, we use `sizes` only for margins and paddings? Ignore this if I'm wrong

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
```diff
@@ -91,9 +93,51 @@ export const StyledSvgControlsSettingsSolid = styled(SvgControlsSettingsSolid)`
...
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  align-items: center;
+  padding: 0.5em 0.5em 0.5em 1em;
+  gap: 1em;
+  opacity: 0;
+  visibility: hidden;
+  transition: opacity, visibility;
+  transition-delay: ${TRANSITION_DELAY};
+  transition-duration: 200ms;
```
rafalpawlow: Couldn't we use some design token here with similar values?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
```diff
@@ -91,9 +93,51 @@ export const StyledSvgControlsSettingsSolid = styled(SvgControlsSettingsSolid)`
...
+  transition-timing-function: ${transitions.easing};
+  @media (hover: hover) {
+    font-size: ${({ isFullScreen }) => (isFullScreen ? sizes(8) : sizes(4))};
```
rafalpawlow: Same question about `sizes`

rafalpawlow (changes requested):

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
drillprop: These values are quite old. Some of them were passed to me during calls, they were never documented if I'm not mistaken. If I change that we would need to go through all the animations in the player and make sure that everything works as expected. We'd need to update also easings and delays. I don't want to bother designers about this right now. 
For now, let's just keep the same values as we have in the controls overlay until we get proper documentation for the player, wdyt?

rafalpawlow (changes requested):
You missed this margin

Zrzut ekranu 2022-07-7 o 10 54 27

rafalpawlow (approved):

PR #2970, opened 2022-07-06
Upload avatar service

### LINKED ISSUE #2885: Create avatar upload service
Create a simple backend service that can store user avatars. Specific requirements TBD.

rafalpawlow:
Closes #2885


packages/atlas-avatar-service/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+})
+const upload = multer({ storage })
+
+const app = express()
+app.use(cors())
+app.use(`/${DIRECTORY_NAME}`, express.static(path.resolve(UPLOAD_DIRECTORY)))
+
+const PORT = 80
+
+app.post('/upload-avatar', upload.single('file'), (req, res) => {
```
kdembler: I think it's more REST-compatible if the the POST path is the same as GET so `/uploads`

packages/atlas-avatar-service/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  try {
```
kdembler, line 42: I think we should ensure `Content-Type` contains `multipart/form-data`. If it doesn't, it won't be handled by `multer` as I understand

packages/atlas-avatar-service/src/index.ts
kdembler: We should validate that the file is an image and reject if it's not

packages/atlas-avatar-service/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    const fileName = req.file?.filename
+    const oldAvatar = req.body.oldAvatar
+    if (oldAvatar) {
+      fs.unlink(`${UPLOAD_DIRECTORY}/${oldAvatar}`, (error) => {
+        if (error) {
+          console.error(error)
+        }
+      })
+    }
```
kdembler: It's a good idea, but currently this will let anybody remove somebody else's avatar which can be used for abuse. Unless you have some idea for a way to authenticate this, I would suggest to remove the feature

kdembler (changes requested):

packages/atlas-avatar-service/src/index.ts
kdembler: We should also restrict the file size to like 1MB max

packages/atlas-avatar-service/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    callback(null, `${uuid()}${path.extname(file.originalname)}`)
+  },
+})
+const upload = multer({
+  storage,
+  fileFilter: (req, file, callback) => {
+    const extension = path.extname(file.originalname)
+    if (!['.png', '.jpg', '.gif', '.jpeg', '.tiff', '.tif'].includes(extension.toLowerCase())) {
+      return callback(new Error('Only images are allowed'))
+    }
```
kdembler: I think it may be safer to use `file.mimetype` for this instead. Let's allow `image/jpg` `image/jpeg` `image/webp` and `image/avif`

packages/atlas-avatar-service/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const PORT = 80
+
+app.post('/uploads', upload.single('file'), (req, res) => {
+  if (req.headers['content-type'] && !/multipart\/form-data/.test(req.headers['content-type'])) {
+    res.status(500).send('Content-Type header must be set to multipart/form-data')
+    return
+  }
+  try {
+    console.log(req.headers)
```
kdembler: Do we need that?

kdembler (commented):
All looks good, just 2 last points:
1. Let's add a workflow for building Docker in `.github/workflows/docker.yml`
2. Let's add e`x`ecutable flag to `ci/build-docker-avatar-service.sh`

kdembler (approved):

Issue #2943, opened 2022-06-30
Share - regular view


dmtrjsg:
## Scope

- [ ] Copy link is removed and substituted by new share button
- [ ] Share button underneath video player on the video detailed view

![Screenshot 2022-06-30 at 09.38.07.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/2dc2f2d1-51f8-45a3-952a-e75b21e308e4)

Out of scope:

#2935 

## Designs
[Figma> ](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page)

[link TBC]

PR #2972, opened 2022-07-07
Sharing video overlay and dialog

### LINKED ISSUE #2936: Sharing Video + Embedded code snippet
## Scope

Implement the sharing overlay, triggered by the sharing button on the top 
- #2935 

## AC
- [ ] full url displayed
- [ ] embeddable code snippet copied to clipboard
- [ ] Start at timestamp

**Out of scope: ⛔** 
- Social media sharing integrations

## Designs

[LInk to Figma](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2090%3A141867)

Preview:

Screenshot 2022-06-29 at 17 31 57

drillprop:
Fix #2936 #2943 

Designs 
- Flows: 
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2132%3A135542
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2132%3A135550

- Components: 
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2090%3A120607


drillprop:
Draft for now. Since designs are pretty old I had to improvise a bit. I'll show this to @toiletgranny first.

toiletgranny (changes requested):
Hey @drillprop, thanks for putting this together. The designs are indeed quite old now, so I appreciate your extra effort to figure out how it should work with the newest styles and components. ❤️ Here are a couple of ideas from me:

### Implementation issues
1. Dialog title is written in title case; it should be sentence case: `Share Video` -> `Share video`,
2. Please note the extra vertical padding (`16px`) around the social media icons wrapper. It seems to be missing entirely in the implementation. 
    CleanShot 2022-07-11 at 10 34 30@2x
3. The "Start at..." checkbox seems to be missing its outer ring upon hovering. Why is that?
4. Checkbox label says "Starts at", but it should be just "Start at" (no "s").

### Other ideas
Having seen the implementation live, I think we could use more tooltips for icon buttons. Here's what I would suggest doing:
1. Let's add a tooltip for the Copy icon button in the Input: "Copy URL",
2. Let's change the tooltip in the embed code to say "Copy embed code",
3. Upon clicking on either of the buttons, can we please change the tooltip contents, for 2 seconds only, to "Copied to clipboard"?

drillprop:
@toiletgranny 
I think I fixed everything!

toiletgranny (approved):

rafalpawlow (changes requested):
Looks good, one small thing. There is no margin between input and timestamp

Zrzut ekranu 2022-07-12 o 14 50 20

rafalpawlow (approved):

PR #2983, opened 2022-07-08
Use BigNumbers for all token values

### LINKED ISSUE #2899: Use BigNumbers for all token values
Currently in few places in our codebase we keep token values in regular Javascript numbers. For example, we will get `lastSalePrice` from QN as a string and do `Number(lastSalePrice)`. However, after the [upcoming change of token denomination ](https://github.com/Joystream/joystream/issues/3389), all of the current values will be multiplied by 10^10, i.e. 1 `tJOY` will be equal to 10,000,000,000 `HAPI` (internal base unit). This will mean that Javascript number will be able to hold only values up to 900719 tJOY (`Number.MAX_SAFE_INTEGER / 10000000000`) and anything more than that will overflow and bad things will happen. To accomodate for that, instead of using regular numbers, we should use big numbers (most likely from `bn.js`) anywhere token values are used.

eldiegod:
Fix #2899

### DIFF
--- a/packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
+++ b/packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
@@ -101,7 +101,13 @@ export const HistoryItem: FC<HistoryItemProps> = ({ size, member, date, joyAmoun
         <ValueContainer>
           <JoyPlusIcon>
             <JoyTokenIcon size={16} variant="silver" />
-            <NumberFormat as="span" format="short" value={joyAmount} variant={size === 'medium' ? 'h300' : 'h200'} />
+            <NumberFormat
+              as="span"
+              format="short"
+              value={joyAmount}
+              withToken
+              variant={size === 'medium' ? 'h300' : 'h200'}
+            />
           </JoyPlusIcon>
           <SwitchTransition>
             <CSSTransition
--- a/packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
+++ b/packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
@@ -109,20 +109,18 @@ export const HistoryItem: FC<HistoryItemProps> = ({ size, member, date, joyAmoun
               timeout={parseInt(cVar('animationTransitionFast', true))}
               classNames={transitions.names.fade}
             >
-              <span>
-                {dollarValue ? (
-                  <NumberFormat
-                    as="span"
-                    format="dollar"
-                    variant="t100"
-                    color="colorText"
-                    value={dollarValue}
-                    align="end"
-                  />
-                ) : (
-                  '‌'
-                )}
-              </span>
+              {dollarValue ? (
+                <NumberFormat
+                  as="span"
+                  format="dollar"
+                  variant="t100"
+                  color="colorText"
+                  value={dollarValue}
+                  align="end"
+                />
+              ) : (
+                '‌'
+              )}
             </CSSTransition>
           </SwitchTransition>
         </ValueContainer>
--- a/packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
+++ b/packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
@@ -115,18 +115,20 @@ export const HistoryItem: FC<HistoryItemProps> = ({ size, member, date, joyAmoun
               timeout={parseInt(cVar('animationTransitionFast', true))}
               classNames={transitions.names.fade}
             >
-              {dollarValue ? (
-                <NumberFormat
-                  as="span"
-                  format="dollar"
-                  variant="t100"
-                  color="colorText"
-                  value={dollarValue}
-                  align="end"
-                />
-              ) : (
-                '‌'
-              )}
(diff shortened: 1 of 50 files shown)


rafalpawlow (changes requested):
Something is wrong with NFT widget (video/38)
dev:
Zrzut ekranu 2022-07-12 o 13 11 40

PR:
Zrzut ekranu 2022-07-12 o 13 11 35

Also I see worker error in console
Zrzut ekranu 2022-07-12 o 13 13 02

drillprop:
> Also I see worker error in console Zrzut ekranu 2022-07-12 o 13 13 02

This is on `cartage`. I don't think I should fix this in this PR

As for the NFT Widget, I fixed token names - but you'll still see 0 as a price. This is because BN only supports integers. I'm not sure yet if I should change the library, asked @kdembler about it. 

As for the infinite loading - this is happening only on deployment - locally I see the button, I think this is also connected to the error you mentioned(?).

drillprop:
> As for the infinite loading - this is happening only on deployment - locally I see the button, I think this is also connected to the error you mentioned(?).

Yeah, I can confirm that I see skeleton loaders in deployments in this PR https://github.com/Joystream/atlas/pull/2963

rafalpawlow (changes requested):
When I try to start sale I see this when I select Auction

Zrzut ekranu 2022-07-15 o 10 29 11

packages/atlas/src/joystream-lib/lib.ts
```diff
@@ -91,11 +92,11 @@ export class JoystreamLib {
     return number.toNumber()
   }
 
-  async subscribeAccountBalance(accountId: AccountId, callback: (balance: number) => void) {
+  async subscribeAccountBalance(accountId: AccountId, callback: (balance: BN) => void) {
     await this.ensureApi()
 
     const unsubscribe = await this.api.derive.balances.all(accountId, ({ availableBalance }) => {
-      callback(availableBalance.toNumber())
+      callback(availableBalance)
```
kdembler, line 91: Could you also update `getAccountBalance` above? Or just delete it since I don't think it's used?

packages/atlas/src/utils/number.ts
```diff
@@ -6,3 +8,21 @@ const numberFormatter = new Intl.NumberFormat('en-US', { maximumSignificantDigit
...
+  const div = bn.div(conversionBN).toNumber()
+  const mod = bn.mod(conversionBN).toNumber()
+  return div + mod / TEN_BILLIONS
+}
+
+export const TokenNumberToHapiBN = (number: number) => {
+  if (Number.isInteger(number)) {
+    return new BN(number).mul(conversionBN)
+  } else {
+    return new BN(number * TEN_BILLIONS)
```
kdembler: This will potentially overflow, we need a different approach

packages/atlas/src/utils/number.ts
```diff
@@ -6,3 +8,21 @@ const numberFormatter = new Intl.NumberFormat('en-US', { maximumSignificantDigit
 export const formatNumber = (num: number): string => {
   return numberFormatter.format(num).replaceAll(',', ' ')
 }
+
+const TEN_BILLIONS = 10 ** 10
+
+const conversionBN = new BN(TEN_BILLIONS)
+
+export const HapiBNToTokenNumber = (bn: BN) => {
```
kdembler: This and the function below shouldn't be capitalized I think

packages/atlas/src/utils/number.ts
```diff
@@ -6,3 +8,21 @@ const numberFormatter = new Intl.NumberFormat('en-US', { maximumSignificantDigit
 export const formatNumber = (num: number): string => {
   return numberFormatter.format(num).replaceAll(',', ' ')
 }
+
+const TEN_BILLIONS = 10 ** 10
```
kdembler: Let's move this to `@/config/joystream` and call `HAPI_TO_JOY_RATE` or something along the lines

kdembler (approved):

packages/atlas/src/providers/joystream/joystream.hooks.ts
```diff
@@ -1,5 +1,8 @@
+import BN from 'bn.js'
```
rafalpawlow, line 1: Sometimes we use default import of `bn.js` and sometimes named. Can we standardize it?

rafalpawlow (changes requested):

packages/atlas/src/components/_nft/NftWidget/NftHistory.tsx
```diff
@@ -108,18 +109,20 @@ export const HistoryItem: FC<HistoryItemProps> = ({ size, member, date, joyAmoun
...
-                  format="dollar"
-                  variant="t100"
-                  color="colorText"
-                  value={dollarValue}
-                  align="end"
-                />
-              ) : (
-                '‌'
-              )}
+              <span>
```
rafalpawlow: Do we need this span?

rafalpawlow (changes requested):

rafalpawlow (approved):

PR #2984, opened 2022-07-08
Allow passing to prop to ListItem

rafalpawlow:
Fixes: #2952


packages/atlas/src/components/ListItem/ListItem.styles.ts
```diff
@@ -52,6 +52,10 @@ export const Container = styled.div<ContainerProps>`
   cursor: pointer;
   background-color: ${({ highlight }) => (highlight ? cVar('colorBackgroundMutedAlpha') : 'unset')};
 
+  :active {
+    text-decoration: none;
+  }
```
drillprop: I still see underline under links(active and not active): 
image

drillprop (changes requested):

drillprop (approved):

PR #2987, opened 2022-07-11
Fix action bar in video form

### LINKED ISSUE #2950: User needs to click submit button two times to publish changes when editing video
As the title says. The same applies also to the `Cancel` button. Also, I couldn't reproduce this but Adam encounter a problem with validation: 
> After initial unsuccessful validation, the Title input component doesn't turn its styling back to the default one upon providing a valid input. It should.

This could be related. The goal of this task is to make sure that VideoForm working correctly and the validation works as the design requires. 
I think the problem more likely comes from how we managed the state there. We're setting the so-called `formStatus`, and I think this causes most of the issues. Maybe we should rethink how to manage the state in video form? 

Let's try:
- Remove `reference` prop from `Tooltip` and use `Tooltip` as a wrapper always
- Check whether we need `span` wrapper for `children` in `Tooltip`

rafalpawlow:
Fixes #2950

### DIFF
--- a/packages/atlas/src/components/ActionBar/ActionBar.styles.ts
+++ b/packages/atlas/src/components/ActionBar/ActionBar.styles.ts
@@ -3,7 +3,7 @@ import styled from '@emotion/styled'
 
 import { Text } from '@/components/Text'
 import { Button } from '@/components/_buttons/Button'
-import { cVar, media, sizes, transitions } from '@/styles'
+import { cVar, media, sizes, transitions, zIndex } from '@/styles'
 
 import { Information } from '../Information'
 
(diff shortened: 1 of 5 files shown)


packages/atlas/src/components/ActionBar/ActionBar.styles.ts
```diff
@@ -17,7 +17,7 @@ export const ActionBarContainer = styled.div<ActionBarContainerProps>`
   display: grid;
   gap: ${sizes(4)};
   padding: ${sizes(4)};
-  z-index: ${zIndex.transactionBar};
+  z-index: 9999999;
```
drillprop: do we actually need to update `z-index` here?

drillprop (changes requested):
It doesn't solve the issue entirely. I still need to click twice when I edit only the description.

drillprop (approved):
I went through all the pages and everything seems to work fine. The video form works as it should as well. LGTM!

kdembler (changes requested):
It seems all tooltips are broken?

https://user-images.githubusercontent.com/12646744/178738762-9507219e-9b56-4f95-aa7a-f76dcbd011d7.mp4

I also don't see anything if I hover over the description field in video form

rafalpawlow:
@kdembler I'm affraid we need that span inside tooltip component.

> I also don't see anything if I hover over the description field in video form

You mean title, right?

kdembler (approved):

PR #2989, opened 2022-07-12
Isolate embedded player

rafalpawlow:
Closes #2152


packages/atlas/src/hooks/useRedirectMigratedContent.ts
```diff
@@ -29,7 +29,7 @@ export const useRedirectMigratedContent = ({ type }: { type: 'channel' | 'video'
     if (!migratedId) return
 
     if (type === 'embedded-video') {
-      navigate(absoluteRoutes.embedded.video(migratedId))
+      //navigate(absoluteRoutes.embedded.video(migratedId))
```
kdembler: This probably shouldn't be commented

kdembler (changes requested):
I can also see that the polkadot worker bundle is still downloaded when loading the embedded view. Could you look into why that's the case?

kdembler (approved):

PR #2992, opened 2022-07-13
Fix issue with cropped context menu

### LINKED ISSUE #2917: The context menu is cropped on carousel video tiles
image

drillprop:
Fix #2917


rafalpawlow (changes requested):
It seems to me that it is not ok that context menu appears above button, in other video tiles it appears below kebab button. On the other hand I'm not sure if this should be in scope of this PR.

rafalpawlow (approved):

PR #2993, opened 2022-07-13
Improve searchbar performance

### LINKED ISSUE #2922: Fix performance issues in Searchbar component
As the title says. Typing in search input causes a lot of rerenders which results in pretty bad performance.  It's more noticeable when you have your chrome dev tools open.

rafalpawlow:
Fixes #2922 

In Searchbar.tsx the only change was to remove forwardRef, this was completely unnecessary.

### DIFF
--- a/packages/atlas/src/components/Searchbar/Searchbar.tsx
+++ b/packages/atlas/src/components/Searchbar/Searchbar.tsx
@@ -217,6 +217,7 @@ export const Searchbar: FC<SearchbarProps> = ({
               )}
               <StyledForm action=".">
                 <Input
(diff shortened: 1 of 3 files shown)


drillprop (changes requested):
This PR indeed fixes performance problems, but I noticed one issue. 
It seems that the `X` button stopped working correctly. Check the video: 

https://user-images.githubusercontent.com/51168865/178981385-2100694e-1d83-4259-bd82-c4e9d860f602.mov

When you click the button the input isn't cleared. On production, it works as it should.

drillprop (approved):

PR #3003, opened 2022-07-18
Make NFT options in the context menu more consistent

### LINKED ISSUE #2593: Make NFT options in the context menu more consistent across experiences

> We need to review context menu options (and their order) for videos and NFTs in both the Viewer and Publisher experience. Ideally, NFT options, labels, and icons should be the same in both Video tile and NFT tile components in both the Viewer & Publisher experience.
>
> It's counterintuitive that, as an owner, I can not start the sale of my NFT just because I'm looking at it through a Video tile component.

rafalpawlow:
Closes #2593 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7571%3A158370

### DIFF
--- a/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
+++ b/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
@@ -275,7 +275,6 @@ export const NftForm: FC<NftFormProps> = ({ setFormStatus, onSubmit, videoId }) 
     views: video?.views,
     buyNowPrice: watch('buyNowPrice') || 0,
     startingPrice: watch('startingPrice') || 0,
-    contextMenuItems: [],
   }
 
   const stepsContent = [
--- a/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
+++ b/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
@@ -275,6 +275,7 @@ export const NftForm: FC<NftFormProps> = ({ setFormStatus, onSubmit, videoId }) 
     views: video?.views,
     buyNowPrice: watch('buyNowPrice') || 0,
     startingPrice: watch('startingPrice') || 0,
+    contextMenuItems: [],
   }
 
   const stepsContent = [
--- a/packages/atlas/src/components/_nft/NftTile/NftTile.tsx
+++ b/packages/atlas/src/components/_nft/NftTile/NftTile.tsx
@@ -31,7 +31,7 @@ export type NftTileProps = {
   englishTimerState?: EnglishTimerState
   startsAtDate?: Date
   auctionPlannedEndDate?: Date
-  contextMenuItems: ListItemProps[]
+  contextMenuItems?: ListItemProps[]
 }
 
 export const NftTile: FC<NftTileProps> = ({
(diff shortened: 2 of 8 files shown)


toiletgranny (changes requested):
Thanks a lot, @rafalpawlow, for unpacking this uneasy task. It's already so much better, as I can finally perform the same operations on my NFTs without depending on the context. Naturally, there a few bugs that I found:

1. Let's get rid of the "Play in Joysteam" option. 
CleanShot 2022-07-20 at 10 54 12
2. Both min bid and fixed price information are missing from the "Buy now" and "Place bid" options in the Publisher experience.
CleanShot 2022-07-20 at 10 55 01
2. The current price value is wrong in this options. Also, please note that the color of all prices mentioned in ListItem component should be inherited from this component. Currently, they're all white. CleanShot 2022-07-20 at 09 28 33@2x
3. I should not be able to place a new bid in an open auction if I have already placed one. I can either change it, or withdraw it (once the bid lock duration elapses).
CleanShot 2022-07-20 at 10 28 49
4. Having removed my NFT from sale, without refreshing the page, options in the dropdown are displayed as if my NFT was still on sale.
5. If no bids have been placed yet, the "Place bid" option [should include the "Min bid" value](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7583%3A149735).
CleanShot 2022-07-20 at 10 36 38
6. As a non-whitelisted member, I should not be able to bid nor buy an NFT for a fixed price. Currently, the dropdown displays the "Buy now" option, even though it should not be there for me. Please note that the "Place bid" option is indeed hidden.
CleanShot 2022-07-20 at 10 43 05
7. As an NFT owner, having put my NFT on an open auction and having received offers,[ I should be able to review them.](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7583%3A149858) Currently this option is not displayed there. 
CleanShot 2022-07-20 at 10 45 16

rafalpawlow:
@toiletgranny for point 5 I created separate issue https://github.com/Joystream/atlas/issues/3016

toiletgranny (changes requested):
@rafalpawlow Thanks for reviewing it! Actually, when I was going through my list again right now, it seemed like 5) was already fixed, but perhaps there's some additional complexity to it. 

All of the things I found appeared to be fixed or addressed now, so thank you very much! ❤️ 

There is, however, one new thing that I noticed, which is that now the "Review bids" option seems to be displayed for me under all NFTs, including the ones which I have nothing in common with. This, of course, should not be the case; I should be able to access this option only under the NFTs I own.

CleanShot 2022-07-26 at 08 44 45@2x

toiletgranny (approved):

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
```diff
@@ -275,6 +275,7 @@ export const NftForm: FC<NftFormProps> = ({ setFormStatus, onSubmit, videoId })
     views: video?.views,
     buyNowPrice: watch('buyNowPrice') || 0,
     startingPrice: watch('startingPrice') || 0,
+    contextMenuItems: [],
```
drillprop: I've noticed a small bug: 

https://user-images.githubusercontent.com/51168865/180988300-7249507f-fce4-4375-b992-d7b92ef56b1c.mov

drillprop (changes requested):

drillprop (approved):

PR #3007, opened 2022-07-19
Fix issue with not working transaction on production

### LINKED ISSUE #3005: Sending transaction is not working on production build
To reproduce:
1. run yarn atlas:build
2. go to `atlas/packages/atlas/dist` and run `npx serve`
3. If you leave a like to a video you'll get `No joystream instance` in the console
4. Similar thing will happen when you try to do something else like upload a video or update a channel. 
You'll experience the same in our vercel deployments - https://atlas-cukdlawf5-joystream.vercel.app/ 

Probably worker related bug: 
image

drillprop:
Fix #3005

### DIFF
--- a/packages/atlas/vite.config.ts
+++ b/packages/atlas/vite.config.ts
@@ -17,6 +17,10 @@ export default defineConfig({
     setupFiles: ['vitest-setup.ts'],
     globals: true,
   },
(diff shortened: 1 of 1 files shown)


rafalpawlow (approved):
Seems it solves the problem! 🙏

PR #3008, opened 2022-07-19
Withdraw/Send funds extrinsics integration

### LINKED ISSUE #2942: Withdraw/Send founds extrinsic integration
AC 

- [ ] itegrate both, withdraw and send founds dialogs with extrinsic
- [ ] implement error snackbars if needed
- [ ] integrate fee in dialogs

rafalpawlow:
Closes #2942 

Two things
* Transfer founds doesn't work properly because QN doesn't return success status for some reason
* Withdraw doesn't work because we don't have `rewardAccount` on Channel


drillprop (changes requested):
The app does not work in production, once you connect the wallet.

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -74,7 +84,25 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
   )
 
   const handleSendFounds = async () => {
```
drillprop, line 85:
```suggestion
  const handleSendFunds = async () => {
```

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -74,7 +84,25 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
...
+          title: `${formatNumber(
+            data.amount
+          )} ${JOY_CURRENCY_TICKER} (${convertedAmount}) tokens have been sent over to ${data.account.slice(
+            0,
+            ADDRESS_CHARACTERS_LIMIT
+          )}...
+          ${data.account.slice(-ADDRESS_CHARACTERS_LIMIT)} wallet address`,
+        },
+        txFactory: async (updateStatus) =>
+          (await joystream.extrinsics).sendFunds(data.account || '', amount, proxyCallback(updateStatus)),
```
drillprop, line 98: we should use `data.amount` here. Also, this should be multiplied by 10_000_000_000. We're sending tokens and each token is `10_000_000_000` hapi.

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -746,4 +731,28 @@ export class JoystreamLibExtrinsics {
...
     return this.sendMetaprotocolMemberExtrinsic(tx, cb)
   }
+
+  async sendFundsTx(destinationAccount: MemberId, value: BN) {
+    await this.ensureApi()
+
+    return this.api.tx.balances.transfer(destinationAccount, value)
+  }
+
+  async sendFunds(destinationAccount: MemberId, value: BN, cb?: ExtrinsicStatusCallbackFn) {
```
drillprop: I didn't manage to transfer funds to another account. When I checked transactions in https://polkadot.js.org/apps the value is always 0. Something isn't working with passing BN here probably.

drillprop (changes requested):
When I'm trying to transfer the funds, the transaction dialog is stuck on `processing transaction` even if transaction is done.

packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdown.tsx
```diff
@@ -132,6 +132,7 @@ export const MemberDropdown = forwardRef<HTMLDivElement, MemberDropdownProps>(
 
     const selectedChannel = activeMembership?.channels.find((chanel) => chanel.id === channelId)
     const { url: channelAvatarUrl, isLoadingAsset: isChannelAvatarLoading } = useAsset(selectedChannel?.avatarPhoto)
+
     const channelBalance = /* useSubscribeAccountBalance(selectedChannel?.rewardAccount) || */ new BN(0) // TODO: fix for Carthage
```
drillprop: We have rewardAccount now. We should be able to use it.

drillprop (changes requested):
A couple of things: 
- Withdraw balance should be possible now. 
- I just had a call with Adam, and with Send dialog we should be able to send funds to _**_any_**_ account. This means that we could send funds also to the channel account. So, we shouldn't validate if a member exists in `Destination account` input(but we should show the avatar for the member if address points to member account)

drillprop (changes requested):
Also, sending funds doesn't seem to work right now - we always send 0 to a destination account:
image

This is probably the BN issue, let's pass the number to `sendFundsTx`(in tokens) and convert them to Hapi. Check for example `makeNftBidTx` or another method which has a price in parameters.

packages/atlas/codegen.config.yml
```diff
@@ -1,6 +1,6 @@
 overwrite: true
 
-schema: ${VITE_LOCAL_ORION_URL}
+schema: https://atlas-next.joystream.org/orion/graphql
```
drillprop: Let's remove this change. This is being fixed here https://github.com/Joystream/atlas/pull/3029/files

drillprop (changes requested):

drillprop (approved):

kdembler (approved):

PR #3009, opened 2022-07-20
Additional fees support

### LINKED ISSUE #2996: Additional fee support
Add support for additional fees in the system that are not captured at the moment:
1. Channel/Video state bloat bond
2. Data object state bloat bond (paid for each data object)
3. Per MB fee for data object (paid for total data objects size)

Worth keeping in mind: the goal of `expected*BloatBond` is to protect the user from someone maliciously setting those fees to very high values, basically robbing the user. In the current implementation, we don't protect the user because at the time when the tx is created (for example `createChannelTx`), we fetch current values of those params from blockchain, so even if they are set very high, we will allow that without the user knowing. Instead, we should fetch those values once, for example when VideoWorkspace is opened, or even treat them as constants and use `useJoystreamChainState`. Then, we can display those values to the user as part of the fee and also pass them as a parameter to the `createChannelTx` so the value supplied to the extrinsic is always the same as the one user has seen in the fees. Similarly, we should do the same for other bloat bonds and also the per MB fee.

All fees should be added and displayed in relevant action bars.

drillprop:
Fix #2996 
Be aware, that these fees are set to 0 in the runtime config, so you won't be able to properly test it. Check https://github.com/Joystream/joystream/issues/4045

### DIFF
--- a/packages/atlas/src/hooks/useFee.ts
+++ b/packages/atlas/src/hooks/useFee.ts
@@ -1,6 +1,6 @@
 import debouncePromise from 'awesome-debounce-promise'
 import BigNumber from 'bignumber.js'
-import { BN } from 'bn.js'
+import BN from 'bn.js'
 import { isEqual } from 'lodash-es'
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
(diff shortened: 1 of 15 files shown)


packages/atlas/src/hooks/useFee.ts
```diff
@@ -55,8 +82,63 @@ export const useFee = <TFnName extends TxMethodName, TArgs extends Parameters<Jo
...
+  const {
+    chainState: {
+      dataObjectPerMegabyteFee,
+      dataObjectStateBloatBondValue,
+      channelStateBloatBondValue,
+      videoStateBloatBondValue,
+    },
+  } = useJoystream()
+
+  const getFeePerMb = useCallback(
```
kdembler: This name and the property the value is assigned to `feePerMb` is misleading. This  value isn't fee per MB, it's total fee related to assets size, already taking the size into account so no "per MB" anymore

packages/atlas/src/hooks/useFee.ts
```diff
@@ -55,8 +82,63 @@ export const useFee = <TFnName extends TxMethodName, TArgs extends Parameters<Jo
...
+    (assets?: VideoInputAssets | ChannelInputAssets) => {
+      if (!assets) {
+        return
+      }
+      const totalBytes = Object.values(assets)
+        .reduce((a, b) => {
```
kdembler: For clarity, let's have args named `(acc, asset)`

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -190,13 +190,16 @@ export const CreateEditChannelView: FC<CreateEditChannelViewProps> = ({ newChann
   const channelMetadata = createChannelMetadata(watch())
   const channelAssets = createChannelAssets()
 
-  const { fee: updateChannelFee, loading: updateChannelFeeLoading } = useFee(
+  const { channelStateBloatBondValue, dataObjectStateBloatBondValue } = useBloatFeesAndPerMbFees(channelAssets)
```
kdembler: Maybe those could also be returned by `useFee`, what do you think?

packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.hooks.ts
```diff
@@ -38,7 +38,7 @@ export const useHandleVideoWorkspaceSubmit = () => {
   const removeDrafts = useDraftStore((state) => state.actions.removeDrafts)
   const { tabData } = useVideoWorkspaceData()
 
-  const { videoStateBloatBondValue, dataObjectStateBloatBondValue } = useBloatFeesAndPerMbFees()
+  const { videoStateBloatBondValue, dataObjectStateBloatBondValue } = useFee('createVideoCommentTx')
```
kdembler: I think that's supposed to be `createVideoTx`?

packages/atlas/src/views/studio/VideoWorkspace/VideoForm/VideoForm.tsx
```diff
@@ -223,7 +223,7 @@ export const VideoForm: FC<VideoFormProps> = memo(({ onSubmit, setFormStatus })
   const nftMetadata = createNftInputMetadata(getValues())
   const assets = createBasicVideoInputAssetsInfo(getValues('assets'))
 
-  const { videoStateBloatBondValue, dataObjectStateBloatBondValue } = useBloatFeesAndPerMbFees(assets)
+  const { videoStateBloatBondValue, dataObjectStateBloatBondValue } = useFee('createVideoTx')
```
kdembler: We already call `useFee` below, why not extract it from that?

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.tsx
```diff
@@ -190,7 +190,7 @@ export const CreateEditChannelView: FC<CreateEditChannelViewProps> = ({ newChann
   const channelMetadata = createChannelMetadata(watch())
   const channelAssets = createChannelAssets()
 
-  const { channelStateBloatBondValue, dataObjectStateBloatBondValue } = useBloatFeesAndPerMbFees(channelAssets)
+  const { channelStateBloatBondValue, dataObjectStateBloatBondValue } = useFee('createChannelTx')
```
kdembler: Same here, we already have useFee below

kdembler (approved):

PR #3014, opened 2022-07-21
Fix issue with not setting channel id

### LINKED ISSUE #3013: Channel id is not properly set in localstorage when you create new channel

drillprop:
Fix #3013 

This will also fix the issue of not being able to set avatar and cover image on the channel creation


rafalpawlow (approved):

rafalpawlow:
Do you think we could find `channelId` rather than selecting array index?

rafalpawlow (approved):

PR #3019, opened 2022-07-22
Fix issue with crashing app when value is too big

### LINKED ISSUE #3018: Auction input crashes whole app
When you type very hight number e.g in minimum bid, you'll see blank page. I experienced this by entering the number 999999999999999999

drillprop:
Fix #3018

### DIFF
--- a/packages/atlas/src/joystream-lib/lib.ts
+++ b/packages/atlas/src/joystream-lib/lib.ts
@@ -6,6 +6,7 @@ import { proxy } from 'comlink'
 
 import { NFT_PERBILL_PERCENT } from '@/joystream-lib/config'
 import { ConsoleLogger, SentryLogger } from '@/utils/logs'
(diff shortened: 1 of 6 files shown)


rafalpawlow (changes requested):
When you type high minimum bid, it breaks buy now price

https://user-images.githubusercontent.com/13706246/180739334-5e546fdc-8148-4ff6-85ab-5654fdfbce01.mov

drillprop:
> When you type high minimum bid, it breaks buy now price
> 
>  Nagranie.z.ekranu.2022-07-25.o.10.55.31.mov

The same happens on `dev` branch. Should we really focus on that in this PR? Can we do this as a follow-up? 
It's not easy to fix. We would have to convert all the values in inputs to strings, which would require much effort. I think this is the only way to resolve this.

For now, I improved validation a bit - I set maxStartingPrice in the Setup component

rafalpawlow (approved):

PR #3030, opened 2022-07-26
Fix performance issues in VideoForm

### LINKED ISSUE #3024: Potential memory leak after sending few transactions
To reproduce 
1. Go to the studio:
1. Create the first video
2. Create the second video
3. Create the third video - you'll notice that already the app slowing down a bit. You'll notice that the animation in the transaction dialog is laggy a bit and sometimes you need to wait for the Polkadot extension popup a little bit longer.
4. Depending on the computer, on the fourth/fifth video, the app will just freeze, and google chrome should notify you about no response from the tab.  

Edit: You could reproduce this even quicker. Just go to your video, make some changes and click `Publish changes`. When the Polkadot extension asks you for a password click cancel. Do this a few times. 

This doesn't seem to happen to the viewer. Every minimized transaction seems to work fine.

drillprop:
Fix #3024


kdembler (approved):
Good job figuring this out 🥇

PR #3031, opened 2022-07-26
Fix NFT tile preview

### LINKED ISSUE #3027: NFT tile preview is empty when you click start sale from NFT tile's kebab menu
Happened to me only in the viewer: 
1. Go to membership/yourmembership
2. Click on the kebab menu and click put NFT on sale
3. Drawer should come up
4. Nft Tile doesn't contain any data: 
image

rafalpawlow:
Fixes #3027

### DIFF
--- a/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
+++ b/packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
@@ -3,7 +3,7 @@ import { addMilliseconds } from 'date-fns'
 import { FC, useCallback, useEffect, useMemo, useRef } from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
 
-import { useFullVideo } from '@/api/hooks'
+import { useBasicVideo } from '@/api/hooks'
 import { Step, StepProps, getStepVariant } from '@/components/Step'
 import { Text } from '@/components/Text'
 import { SvgActionChevronR } from '@/components/_icons'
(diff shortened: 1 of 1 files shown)


packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
```diff
@@ -93,7 +93,7 @@ export const NftForm: FC<NftFormProps> = ({ setFormStatus, onSubmit, videoId })
     formState: { isValid },
   } = formMethods
 
-  const { video, loading: loadingVideo } = useFullVideo(videoId, { fetchPolicy: 'cache-only' })
+  const { video, loading: loadingVideo } = useFullVideo(videoId, { fetchPolicy: 'cache-first' })
```
drillprop: We could omit his option. `cache-first` is the default cache policy https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy

drillprop (changes requested):

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/NftForm.tsx
kdembler: The question is - do we need a full video here? Maybe this component could work with just a basic video fragments? If yes, then we can even keep the current `fetchPolicy` as the basic video will already be in the cache

drillprop (approved):
LGTM!

PR #3046, opened 2022-07-28
Add royalty split information in NFT form summary

### LINKED ISSUE #2597: Add royalty split information in the listing summary
> When I wanted to start a sale of the NFT, I didn't see a royalty split amounts anywhere. It's probably not a biggie when I am the issuer and first owner, but as a second-hand reseller that's probably an important thing I'd like to know about. Perhaps worth adding it there.

https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=3232%3A210522

rafalpawlow:
Closes #2597 

**Design**
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=3232%3A210522

### DIFF
--- a/packages/atlas/src/api/queries/fragments.graphql
+++ b/packages/atlas/src/api/queries/fragments.graphql
@@ -118,9 +118,6 @@ fragment BasicVideoFields on Video {
   nft {
     id
     creatorRoyalty
-    creatorChannel {
-      title
-    }
   }
 }
 
--- a/packages/atlas/src/api/queries/fragments.graphql
+++ b/packages/atlas/src/api/queries/fragments.graphql
@@ -116,7 +116,11 @@ fragment BasicVideoFields on Video {
     ...StorageDataObjectFields
   }
   nft {
-    id # just to check whether NFT exists
+    id
+    creatorRoyalty
+    creatorChannel {
+      title
+    }
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/atlas/src/api/queries/fragments.graphql
```diff
@@ -116,7 +116,11 @@ fragment BasicVideoFields on Video {
     ...StorageDataObjectFields
   }
   nft {
-    id # just to check whether NFT exists
+    id
+    creatorRoyalty
+    creatorChannel {
+      title
+    }
```
drillprop: I'm not sure if we should fetch more information in the `BasicVideoFields` fragment. Maybe we could create another query or maybe even use `FullVideoFields` in the NFT Form? @kdembler what do you think?

drillprop: Also, we don't need creatorChannel here, we could use just `BasicChannelFields`

kdembler: As Bartosz suggested, we should be able to use the `video.channel` instead of `video.nft.creatorChannel`. Then, I think it's fine to add only `creatorRoyalty`

drillprop (approved):

packages/atlas/src/views/global/NftSaleBottomDrawer/NftForm/AcceptTerms/AcceptTerms.tsx
```diff
@@ -67,6 +69,26 @@ export const AcceptTerms: FC<AcceptTermsProps> = ({ selectedType, formData }) =>
           </DescriptionText>
         </Description>
       </Row>
+      {creatorRoyalty && (
+        <Row>
+          <Title>
+            <TitleText>Royalties</TitleText>
+            <StyledInformation
+              text="By setting royalties you will be entitled to a percentage share in revenue from any future secondary market sale. So if someone sells your work you will get paid."
```
kdembler:
```suggestion
              text="Royalties let the video creator earn revenue from secondary NFT sales"
```

PR #3049, opened 2022-07-29
Follow on-chain bucket assigment policy

### LINKED ISSUE #3041: Follow on-chain bucket assigment policy
In my PR #3029 I used config placed in `@/config/joystream`, namely `NEW_CHANNEL_DISTRIBUTION_BUCKETS_PER_FAMILY` and `NEW_CHANNEL_STORAGE_BUCKETS_COUNT`. That was a mistake, we should use the on-chain policy at all times. So we should remove the config and instead get this data:
CleanShot 2022-07-28 at 13 02 10@2x

rafalpawlow:
Closes #3041

### DIFF
--- a/packages/atlas/src/joystream-lib/lib.ts
+++ b/packages/atlas/src/joystream-lib/lib.ts
@@ -146,4 +146,19 @@ export class JoystreamLib {
       platformFeePercentage: platformFeePercentage.toNumber() / NFT_PERBILL_PERCENT,
     }
   }
+
+  async getDynamicBagCreationPolicies() {
+    await this.ensureApi()
+
+    const { numberOfStorageBuckets, families } = await this.api.query.storage.dynamicBagCreationPolicies('Channel')
+    const transformedFamilies = Object.fromEntries(families)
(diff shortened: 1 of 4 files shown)


packages/atlas/src/joystream-lib/lib.ts
```diff
@@ -146,4 +146,19 @@ export class JoystreamLib {
...
+
+    const { numberOfStorageBuckets, families } = await this.api.query.storage.dynamicBagCreationPolicies('Channel')
+    const transformedFamilies = Object.fromEntries(families)
+    Object.keys(transformedFamilies).forEach((key) => {
+      transformedFamilies[key] = transformedFamilies[key].toNumber()
+    })
+
+    return {
+      numberOfStorageBuckets: numberOfStorageBuckets.toNumber(),
+      families: transformedFamilies,
```
kdembler: Let's call these `storageBucketsCount` and `distributionBucketsCountPerFamily` 🙏

kdembler (approved):

PR #3065, opened 2022-08-02
Fix issue with not loading polkadot worker on firefox

### LINKED ISSUE #3042: Polkadot worker doesn't load on Firefox
Previously the polkadot webworker didn't work in Firefox only in dev, on built version it worked fine. On `carthage` branch it doesn't work at all so we broke something. Let's figure out what that was

drillprop:
Fix #3042 

I'm not too fond of the fact that I had to patch @polkadot/api to make this happen, but I couldn't find a better solution.


.yarn/patches/@polkadot-api-npm-8.9.1-ec90430f8c.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/packageInfo.js b/packageInfo.js
+index a82cfb0d36f703dd8c262a2d6a13178b9ada7d0b..13a9385dd4782e281fb2eb219ad1806e32579c08 100644
+--- a/packageInfo.js
++++ b/packageInfo.js
+@@ -2,8 +2,8 @@
+ // SPDX-License-Identifier: Apache-2.0
+ // Do not edit, auto-generated by @polkadot/dev
+ export const packageInfo = {
+-  name: '@polkadot/api',
+-  path: (import.meta && import.meta.url) ? new URL(import.meta.url).pathname.substring(0, new URL(import.meta.url).pathname.lastIndexOf('/') + 1) : 'auto',
```
drillprop: this code was transpiled to `path: document.currentScript && document.currentScript.src || new URL("polkadot-worker.js",document.baseURI).href...`, that's why we had an error about `document` is not defined: https://github.com/Joystream/atlas/issues/3005

drillprop:
Draft - on second thought, I think it's possible to fix this with vite plugins.

packages/atlas/vite.config.ts
```diff
@@ -26,9 +26,20 @@ export default defineConfig({
...
+    plugins: [
+      {
+        name: 'resolve-import-meta-polkadot',
+        resolveImportMeta(_, { chunkId }) {
+          if (chunkId === 'polkadot-worker.js') {
+            return 'self.location.href'
+          }
+        },
+      },
+    ],
```
kdembler: That's so much better than the previous solution, thanks for taking a second look 😍 

Could you just add a small comment on what this does and what issue it solves?

drillprop: Sure thing!

kdembler (approved):

PR #3066, opened 2022-08-02
Handle kill switch

### LINKED ISSUE #3001: Operator dead man's switch
**Scope:**
- Add a simple `isKilled` flag to Orion's MongoDB
- Add a simple mutation that allows setting that value, given some secret password (should be configured via Orion env variable)
- Add logic in Atlas app load that fetches `isKilled` from Orion and if it's true, it displays a maintenance screen (code taken from `maintenance` branch) and stops further app load
- Add a secret screen in Atlas that lets you switch maintenance mode on (like `/admin` or can be a "Maintenance" tab in admin modal) - there you paste your secret password to authenticate the request

---

# Background
Filtering and policing of uploads is not even in it's infancy by the time we go live, this means an Atlas operator team will be poorly equipped to very quickly respond to noxious or out of policy content displayed in their instance.

# Proposal
Some trivial way for an Atlas instance to be paused or taken down instantly, by one among a small member of trusted - non-technical members of a team. This would be very useful for the Jsgenesis operated instance. Something as simple as `<app>/kill`, and entering a special code - just to avoid fat-finger error, would be sufficient. If you want to revoke access, you just update the code and share with new team.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202622688683484) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #3001


packages/atlas/src/api/hooks/killSwitch.ts
```diff
@@ -0,0 +1,30 @@
...
+      setKillSwitch({
+        variables: { isKilled },
+        context: { headers: { authorization: secret }, ...opts },
+      }),
+    ...rest,
+  }
+}
+
+export const useGetKillSwitch = (opts?: QueryHookOptions<GetKillSwitchQuery, GetKillSwitchQueryVariables>) => {
+  const { data, ...rest } = useGetKillSwitchQuery(opts)
```
kdembler: Let's add some persistence to this data. So that if kill switch was enabled on the last app visit, I will see the maintenance screen by default and not flash of the app loading:
```tsx
const { isLoading, isKilled, wasKilledLastTime } = useMagic()

if (isKilled || (isLoading && wasKilledLastTime)) return <Maintenance />
else return <RegularStuff />
```

kdembler (changes requested):
Generally looks good but let's make sure Maintenance keeps working if some of the requests fails

packages/atlas/src/CommonProviders.tsx
```diff
@@ -13,12 +17,37 @@ export const CommonProviders: FC<PropsWithChildren> = ({ children }) => {
...
+          <Maintenance />
+        </>
+      ) : (
+        children
+      )}
+      {(isKilled || (loading && wasKilledLastTime)) && !loading && !error && (
+        <BrowserRouter>
+          <AdminModal />
+        </BrowserRouter>
+      )}
```
kdembler: I don't like that `AdminModal` is conditionally rendered in two places. Maybe we just remove dependency on the router and place `AdminModal` directly next to `MaintenanceWrapper`?

packages/atlas/src/CommonProviders.tsx
```diff
@@ -13,12 +17,37 @@ export const CommonProviders: FC<PropsWithChildren> = ({ children }) => {
...
+
+  return (
+    <>
+      {isKilled || (error && wasKilledLastTime) || (loading && wasKilledLastTime) ? (
+        <>
+          <Maintenance />
+        </>
+      ) : (
+        children
+      )}
```
kdembler: Let's use regular `if` conditions and `return` statements for this, it's hard to read atm

rafalpawlow: It won't be possible because we need to render `AdminModal` conditionally. I mean it would be possible but readability wouldn't be better. I cleaned up it a little bit, let me know if this is fine for you.

packages/atlas/src/CommonProviders.tsx
```diff
@@ -34,20 +36,13 @@ export const CommonProviders: FC<PropsWithChildren> = ({ children }) => {
 const MaintenanceWrapper: FC<PropsWithChildren> = ({ children }) => {
   const { isKilled, wasKilledLastTime, error, loading } = useGetKillSwitch()
 
+  const shouldRenderMaintenance = isKilled || (error && wasKilledLastTime) || (loading && wasKilledLastTime)
+  const shouldRenderAdminModal = !loading && !error
```
kdembler: What's the reasoning behind this condition? What would be case in which we don't want to show the AdminModal?

rafalpawlow: Sorry, you are right, it was pointless

kdembler (approved):

PR #3069, opened 2022-08-03
carthage refactor and tweaks

### LINKED ISSUE #3063: Carthage finishing touches

kdembler:
**Scope of changes:**

- Tried to use `BN` whenever possible to hold token values. The only exception is when we need token inputs - `react-hook-form` cannot store `BN` in its state so we still do some conversion although much less than before
    - NFT purchase drawer still uses a bunch of numbers because it seemed like refactoring that would be too much effort for now
- Created `TokenInput` component that can be used whenever we need JOY input - it adds an icon prefix and USD conversion suffix automatically. In the future we could potentially also make it return BN’s directly
- `joystream-lib`:
    - Introduced grouping in `extrinsics` by the type of operation
    - Added type helpers so that we define argument types for each extrinsic only once - in `*Tx` function and they are inherited by functions using them
    - All the token values are transferred as strings between the app and Polkadot worker - this way the joystream-lib can accept all the values in Hapis.
    - Improved fetching chain constant by using `queryMulti`
- Moved some helpers (like `hapiBnToTokenNumber` or `calculateAssetsSizeFee`) to `@/joystream-lib/utils` and added unit tests for them.
- Moved `useFee`, `useBloatFeesAndPerMbFees` and `useSubscribeAccountBalance` to `@/providers/joystream/joystream.hooks`
- Updated validation in NFT purchase drawer and NFT sale drawer to follow `onBlur` validation (that’s what we should use everywhere in the app)


drillprop (changes requested):
I'll look at the code and do more testing tomorrow, but I found already that you reintroduced this [bug](https://github.com/Joystream/atlas/issues/3018)
Check the video. 

https://user-images.githubusercontent.com/51168865/182700071-171ba8a4-1733-45c9-8a0c-ed9d94aae40e.mov

packages/atlas/src/components/Fee/Fee.stories.tsx
```diff
@@ -12,18 +13,20 @@ export default {
...
   args: {
     variant: 'h500',
-    amount: 0,
+    amountToken: 0,
     loading: false,
     hideOnMobile: false,
   },
 } as Meta<FeeProps>
 
 const Template: Story<FeeProps> = (args) => (
```
drillprop:
```suggestion
const Template: Story<FeeProps & { amountToken: number }> = (args) => (
```

packages/atlas/src/components/Fee/Fee.stories.tsx
```diff
@@ -12,18 +13,20 @@ export default {
...
   <Wrapper>
-    <Fee {...args} />
+    {/* eslint-disable-next-line @typescript-eslint/no-explicit-any */}
+    <Fee {...args} amount={tokenNumberToHapiBn((args as any).amountToken)} />
```
drillprop:
```suggestion
   <Fee {...args} amount={tokenNumberToHapiBn(args.amountToken)} />
```
I guess this should do the job.

drillprop (commented):
Besides the above, it works well and the code looks good 👍🏻

drillprop (approved):
That's a massive improvement, looks good to me! 🎉

rafalpawlow (approved):
Didn't find any issues 🙌

PR #3077, opened 2022-08-04
Hide dollar signs from Atlas

### LINKED ISSUE #3011: Hide dollar signs from Atlas for Mainnet Launch
## Context

Tjoys <> USD conversion is shown for NFTs while we are on Testnet. It is likely the FIAt onboarding will not be ready on time for the launch, and the listing timeline is being explored at the moment. 

## Scope
- Make the price URL configurable.
- Make sure the app works fine without the conversion rate available (if the URL is missing):
  - NFt tiles
  - nft widget
  - bidding interfaces
  - settling auction
  - notifications

- add tooltip to the JOY price component explaining to users that JOYs can be exchanged to FIAT currencies or swapped to other cryptocurrencies, once listed on exchanges shortly after mainnet, and we are actively working on enabling this

drillprop:
Fix #3011


packages/atlas/svgr.config.js
```diff
@@ -33,10 +34,11 @@ ${variables.imports}
 
 ${variables.interfaces}
 
-const ${variables.componentName} = (${variables.props}) => (
+const ${variables.componentName} = forwardRef((${variables.props}) => (
```
rafalpawlow, line 37: Why do we need it?

rafalpawlow (changes requested):
I'm not actually sure but it seems to me that it should be centered.
Zrzut ekranu 2022-08-5 o 11 13 14

packages/atlas/svgr.config.js
drillprop: Check the code of `JoyTokenIcon`. To use the `reference` prop for the Tooltip I need to forward a ref to the icon.

drillprop:
> I'm not actually sure but it seems to me that it should be centered. Zrzut ekranu 2022-08-5 o 11 13 14

I don't agree, but I'll ask Adam

packages/atlas/svgr.config.js
rafalpawlow: It was hard to find it 😛. Thanks!

drillprop:
@rafalpawlow, the designer agrees with you. Fixed

rafalpawlow (changes requested):
Because of bottom margin it's not centered
Zrzut ekranu 2022-08-5 o 12 15 24
Zrzut ekranu 2022-08-5 o 12 16 13

rafalpawlow (approved):

packages/atlas/src/components/_nft/NftWidget/NftInfoItem.styles.ts
```diff
@@ -5,13 +5,20 @@ import { sizes } from '@/styles'
...
+    disableSecondary
+      ? `
+  'label' min-content
+  'content' min-content / 1fr
+  `
+      : `
+  'label' min-content
+  'content' min-content
+  'secondary' min-content / 1fr
+  `};
```
kdembler: Something's weird is happening here with line breaks. Maybe we could have separate `css` blocks for both cases? This wold improve readability

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
```diff
@@ -67,6 +68,10 @@ export const NumberFormat = forwardRef<HTMLHeadingElement, NumberFormatProps>(
       </StyledText>
     )
 
+    if (format === 'dollar' && !JOY_PRICE_SERVICE_URL) {
+      return null
+    }
```
kdembler: Does this needs to rely on `JOY_PRICE_SERVICE_URL`? I would expect the `convertHapiToUSD` to return null-ish value if the price is not available so we wouldn't need to check here for the URL

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -130,6 +131,7 @@ export const NftWidget: FC<NftWidgetProps> = ({
         <NftInfoItem
           size={size}
           label="Buy now"
+          disableSecondary={!JOY_PRICE_SERVICE_URL}
```
kdembler: Same here, let's not rely on the URL and instead just do `convertHapiToUSD(buyNowPrice)` in a variable and check if that's a truthy value

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -409,6 +411,7 @@ export const NftWidget: FC<NftWidgetProps> = ({
                     />
                   </>
                 }
+                disableSecondary={!JOY_PRICE_SERVICE_URL}
```
kdembler: Same here

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -131,7 +130,7 @@ export const NftWidget: FC<NftWidgetProps> = ({
         <NftInfoItem
           size={size}
           label="Buy now"
-          disableSecondary={!JOY_PRICE_SERVICE_URL}
+          disableSecondary={convertHapiToUSD(buyNowPrice) === null}
```
kdembler: Could we just store result of `convertHapiToUSD(buyNowPrice)` in a variable? We do it 3 separate times now

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
```diff
@@ -411,7 +409,7 @@ export const NftWidget: FC<NftWidgetProps> = ({
                     />
                   </>
                 }
-                disableSecondary={!JOY_PRICE_SERVICE_URL}
+                disableSecondary={convertHapiToUSD(nftStatus.startingPrice) === null}
```
kdembler: Same here

kdembler (approved):

PR #3087, opened 2022-08-05
Ability to filter video content

### LINKED ISSUE #3037: Ability to filter out content as gateway operator
- Create a configuration file that lists blocked:
    - Video IDs
    - Data object IDs
    - (In the future, #3004) Category IDs
- Make the app not show blocked things:
    - For videos, we should have a single function that creates shared filtering paramaters for all video queries. So this should return something like:
```ts
{
  where: {
    isCensored_eq: false,
    isPublic_eq: true,
    // ...
    NOT: [
        id_in: [] // list of IDs from the config,
        category: {
           id_in: [] // list of filtered categories IDs
        }
    ]
  }
}
```
- Make sure NFTs follow the same filtering options as videos

---

Once content is reported as malicious/ not fitting content policy and this flag is captured, we need an ability to filter this content out.  Based on list of video IDs, these would be hidden from gateway.

+  Include separate resource blacklist (URLs & data objects)
 Apply same logic to NFTs wrt content curation as for videos

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202676228106706) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #3037

### DIFF
--- a/packages/atlas/.gitignore
+++ b/packages/atlas/.gitignore
@@ -1,2 +1,1 @@
 figma-import.config.js
-queryFilter.json
--- a/packages/atlas/.gitignore
+++ b/packages/atlas/.gitignore
@@ -1,1 +1,2 @@
 figma-import.config.js
+queryFilter.json
--- a/packages/atlas/queryFilter.json
+++ b/packages/atlas/queryFilter.json
@@ -1,7 +1,0 @@
-{
-  "video": [],
-  "dataObject": {
-    "video": [],
-    "thumbnail": []
-  }
-}
--- a/packages/atlas/queryFilter.json
+++ b/packages/atlas/queryFilter.json
@@ -1,0 +1,7 @@
+{
+  "video": [],
+  "dataObject": {
+    "video": [],
+    "thumbnail": []
+  }
+}
--- a/packages/atlas/src/config/channelFilter.ts
+++ b/packages/atlas/src/config/channelFilter.ts
@@ -1,0 +1,7 @@
+import { ChannelWhereInput } from '@/api/queries'
+
+export const channelFilter: ChannelWhereInput = {
+  isCensored_eq: false,
+  isPublic_eq: true,
+  NOT: [{ id_in: [] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
+}
--- a/packages/atlas/src/config/channelFilter.ts
+++ b/packages/atlas/src/config/channelFilter.ts
@@ -3,5 +3,5 @@ import { ChannelWhereInput } from '@/api/queries'
 export const channelFilter: ChannelWhereInput = {
   isCensored_eq: false,
   isPublic_eq: true,
-  NOT: [{ id_in: ['3'] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
+  NOT: [{ id_in: [] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
 }
--- a/packages/atlas/src/config/channelFilter.ts
+++ b/packages/atlas/src/config/channelFilter.ts
@@ -3,5 +3,5 @@ import { ChannelWhereInput } from '@/api/queries'
 export const channelFilter: ChannelWhereInput = {
   isCensored_eq: false,
   isPublic_eq: true,
-  NOT: [{ id_in: [] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
+  NOT: [{ id_in: ['3'] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
 }
--- a/packages/atlas/src/config/videoFilter.ts
+++ b/packages/atlas/src/config/videoFilter.ts
@@ -1,0 +1,13 @@
+import { VideoWhereInput } from '@/api/queries'
+
+export const videoFilter: VideoWhereInput = {
+  isCensored_eq: false,
+  isPublic_eq: true,
+  thumbnailPhoto: {
+    isAccepted_eq: true,
+  },
+  media: {
+    isAccepted_eq: true,
+  },
+  NOT: [{ id_in: [] }, { thumbnailPhoto: { id_in: [] } }, { media: { id_in: [] } }],
+}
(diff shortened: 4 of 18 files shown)


packages/atlas/.gitignore
```diff
@@ -1 +1,2 @@
 figma-import.config.js
+queryFilter.json
```
drillprop: you added the file to `.gitignore`, but it's not actually ignored. I'm not sure if I understand what's happening here.

packages/atlas/queryFilter.json
```diff
@@ -0,0 +1,7 @@
+{
+  "video": [],
+  "dataObject": {
+    "video": [],
+    "thumbnail": []
+  }
+}
```
drillprop: It's not very clear what you need to pass here. At least to me. Maybe we could rename this file `blacklists.json` and in this file use `videoIds`, `videoDataObjectIds`, `thumbnailDataObjectIds`?

drillprop (changes requested):

drillprop (changes requested):
Looks good, but you didn't add filters for channels as we pointed out during our meeting. I'm fine if you do this as a followup

drillprop (changes requested):
Found this by accident, but I don't see videos with failed uploads anymore in `My videos`. It looks like we overwrite arguments here.

packages/atlas/src/config/channelFilter.ts
```diff
@@ -0,0 +1,7 @@
+import { ChannelWhereInput } from '@/api/queries'
+
+export const channelFilter: ChannelWhereInput = {
+  isCensored_eq: false,
+  isPublic_eq: true,
+  NOT: [{ id_in: [] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
+}
```
drillprop, line 7: I did some testing and is this actually working? For example I added to the list channel id "2"
- I still see videos from this channel in `popular on joystream` and in `top 10 this week`. 
- I still can go to the channel with id "2" and see all the content.

I think we will have to also use these channel filters partly in the video filter to make this right - (`videos where  not channel id_in`). It's also very possible that Orion doesn't apply these filters and we will need to update Orion in that regard, but not super sure about this.

packages/atlas/src/config/videoFilter.ts
```diff
@@ -0,0 +1,13 @@
...
+  isCensored_eq: false,
+  isPublic_eq: true,
+  thumbnailPhoto: {
+    isAccepted_eq: true,
+  },
+  media: {
+    isAccepted_eq: true,
+  },
+  NOT: [{ id_in: [] }, { thumbnailPhoto: { id_in: [] } }, { media: { id_in: [] } }],
+}
```
drillprop, line 22: I still can go to the video which is blocked. Didn't we agree that we should hide this from a user as well?

drillprop (changes requested):

packages/atlas/src/config/channelFilter.ts
rafalpawlow: @drillprop I cannot confirm first point. For the filter below videos `13` and `2` are not showing in sections u mentioned.
```ts
export const videoFilter: VideoWhereInput = {
  isCensored_eq: false,
  isPublic_eq: true,
  thumbnailPhoto: {
    isAccepted_eq: true,
  },
  media: {
    isAccepted_eq: true,
  },
  NOT: [{ id_in: ['13', '2'] }, { thumbnailPhoto: { id_in: [] } }, { media: { id_in: [] } }],
}

```

drillprop: but, you're referring to the video filter. I was referring to the channel filter. If you block the channel, you shouldn't see videos from this channel on the homepage. Am I right?

rafalpawlow: @drillprop  I created a fix, unfortunately due to nesting problem we can only filter by channel id

drillprop (changes requested):
Unfortunately, I still found see some issues. Now if I go to the channel e.g `/channel/1` I'm getting `400`
image

I also think that we could actually block NFTs. They have the same ids as in videos, right? So we could filter them as well, wdyt?

kdembler (changes requested):
This doesn't work fully well with unlisted videos:
1. I cannot see my own unlisted videos in `My videos`
2. I cannot watch unlisted video even if I access it directly with a link

I think in those 2 places you could just override with
```ts
const where = {
  isPublic_eq: undefined
}
```

kdembler:
And I agree with Bartosz's point above, let's use the same IDs that we have for videos to filter out NFTs

drillprop (changes requested):
I'm wondering, should we ignore all the filters in the studio? 
What should see in the studio owner of the blocked channel? 
Now the owner will see: 
image

A similar situation applies to videos. Should we actually use the filters in the studio?

drillprop:
Okay, I have confirmation from Klaudiusz, lets ignore all filters in studio

packages/atlas/src/config/channelFilter.ts
```diff
@@ -0,0 +1,7 @@
+import { ChannelWhereInput } from '@/api/queries'
+
+export const channelFilter: ChannelWhereInput = {
+  isCensored_eq: false,
+  isPublic_eq: true,
+  NOT: [{ id_in: ['3'] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
```
drillprop:
```suggestion
  NOT: [{ id_in: [] }, { avatarPhoto: { id_in: [] } }, { coverPhoto: { id_in: [] } }],
```

drillprop (changes requested):
just one small thing, besides that LGTM

drillprop (approved):

kdembler (changes requested):
1. I still cannot watch unlisted video, even if I go there directly with a link
2. It seems after your changes video title doesn't show up in My uploads. Could you check that out? Doesn't seem to happen on `carthage` to me: 
CleanShot 2022-08-17 at 18 08 25@2x
CleanShot 2022-08-17 at 18 08 19@2x
3. Likewise, it seems that I'm not able to reselect file that failed upload, it always tells me I selected the wrong file. It seems like the video data is missing in my uploads? Why is that happening, can other places be impacted as well?

https://user-images.githubusercontent.com/12646744/185189579-2de5b588-237f-4e71-917a-8b58d9bcff68.mp4

kdembler (approved):

PR #3102, opened 2022-08-10
Video player subtitles

### LINKED ISSUE #2383: Video player subtitles
## Context
As a content Viewer

- [ ] Enable and disable subtitles/cc when watching a video.
- [ ] settings saved in local storage
- [ ] Use some mocked static subtitles just to showcase support

## Out of scope
- Getting the real subtitles for the video from the QN/storage network

## Designs
Figma designs:
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2119%3A135376

rafalpawlow:
Closes #2383

FYI subtitles end at 30 seconds

**Design**
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2119%3A135376

### DIFF
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
@@ -421,9 +421,13 @@ export const Container = styled.div<ContainerProps>`
           text-transform: ${cVar('typographyDesktopT300StrongTextTransform')} !important;
           padding: ${sizes(1)} ${sizes(2)};
           word-break: break-all;
+          line-height: 16px !important;
 
           ${media.sm} {
             font-size: ${({ isFullScreen }) => (isFullScreen ? '32px' : '16px')} !important;
+            padding: ${({ isFullScreen }) =>
+              isFullScreen ? `${sizes(2)} ${sizes(4)}` : `${sizes(1)} ${sizes(2)}`} !important;
+            line-height: 24px !important;
           }
         }
       }
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
@@ -410,17 +410,19 @@ export const Container = styled.div<ContainerProps>`
         transition-delay: ${TRANSITION_DELAY};
         transition-duration: 200ms;
         transition-timing-function: ${transitions.easing};
+        padding: 0 ${sizes(3)};
 
         > div {
           background-color: ${cVar('colorBackgroundOverlay')} !important;
           font: ${cVar('typographyDesktopT300Strong')} !important;
-          font-size: ${({ isFullScreen }) => (isFullScreen ? '32px' : '16px')} !important;
+          font-size: 12px !important;
           letter-spacing: ${cVar('typographyDesktopT300StrongLetterSpacing')} !important;
           text-transform: ${cVar('typographyDesktopT300StrongTextTransform')} !important;
           padding: ${sizes(1)} ${sizes(2)};
+          word-wrap: break-word;
 
           ${media.sm} {
-            font-size: 12px !important;
+            font-size: ${({ isFullScreen }) => (isFullScreen ? '32px' : '16px')} !important;
           }
         }
       }
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
@@ -414,10 +414,14 @@ export const Container = styled.div<ContainerProps>`
         > div {
           background-color: ${cVar('colorBackgroundOverlay')} !important;
           font: ${cVar('typographyDesktopT300Strong')} !important;
-          font-size: ${({ isFullScreen }) => (isFullScreen ? '0.64em' : '16px')} !important;
+          font-size: ${({ isFullScreen }) => (isFullScreen ? '32px' : '16px')} !important;
           letter-spacing: ${cVar('typographyDesktopT300StrongLetterSpacing')} !important;
           text-transform: ${cVar('typographyDesktopT300StrongTextTransform')} !important;
           padding: ${sizes(1)} ${sizes(2)};
+
+          ${media.sm} {
+            font-size: 12px !important;
+          }
         }
       }
     }
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
@@ -28,7 +28,7 @@ import { PlayerControlButton } from './PlayerControlButton'
 import { ControlButton } from './PlayerControlButton.styles'
 
 const DEFAULT_SUBTITLES_POSITION = '0.8em'
-const HOVERED_SUBTITLES_POSITION = '3.5em'
+const HOVERED_SUBTITLES_POSITION = '5.25em'
 
 type ContainerProps = {
   isFullScreen?: boolean
@@ -402,6 +402,7 @@ export const Container = styled.div<ContainerProps>`
   .vjs-text-track-display {
     > div {
       margin: 0 !important;
+      font-size: 16px;
 
       > div {
         inset: unset !important;
@@ -413,7 +414,7 @@ export const Container = styled.div<ContainerProps>`
         > div {
           background-color: ${cVar('colorBackgroundOverlay')} !important;
           font: ${cVar('typographyDesktopT300Strong')} !important;
-          font-size: 0.7em !important;
+          font-size: 1em !important;
           letter-spacing: ${cVar('typographyDesktopT300StrongLetterSpacing')} !important;
           text-transform: ${cVar('typographyDesktopT300StrongTextTransform')} !important;
           padding: ${sizes(1)} ${sizes(2)};
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
@@ -414,7 +414,7 @@ export const Container = styled.div<ContainerProps>`
         > div {
           background-color: ${cVar('colorBackgroundOverlay')} !important;
           font: ${cVar('typographyDesktopT300Strong')} !important;
-          font-size: 1em !important;
+          font-size: ${({ isFullScreen }) => (isFullScreen ? '0.64em' : '16px')} !important;
           letter-spacing: ${cVar('typographyDesktopT300StrongLetterSpacing')} !important;
           text-transform: ${cVar('typographyDesktopT300StrongTextTransform')} !important;
           padding: ${sizes(1)} ${sizes(2)};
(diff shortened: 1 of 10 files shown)


packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
```diff
@@ -395,6 +399,31 @@ export const Container = styled.div<ContainerProps>`
...
   }
 
+  .vjs-text-track-display {
+    > div {
+      margin: 0 !important;
+
+      > div {
+        inset: unset !important;
+        bottom: 0.36em !important;
+        transition: bottom 150ms linear;
```
kdembler: I think we need to adjust this, linear easing doesn't look very nice. Best to ask Adam (since Kuba is away for 2 weeks starting today) to take a look and give feedback

kdembler:
Also seems that subtitles language is not persisted in local storage, only whether subtitles are enabled or not?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
```diff
@@ -395,6 +399,31 @@ export const Container = styled.div<ContainerProps>`
     }
   }
 
+  .vjs-text-track-display {
+    > div {
+      margin: 0 !important;
+
+      > div {
+        inset: unset !important;
+        bottom: 0.36em !important;
```
drillprop: You created variable above(DEFAULT_SUBTITLES_POSITION), let's use it. 
Also why this is 0.36? I think you could set font size here to 16px and set the bottom to just 1em, right?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
drillprop: I think we're using mostly `transition.easing` in the player. Sometimes we also apply small delays. You could try and check how it works.

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
drillprop: I'm also getting the impression that subtitles are too low. Are you sure, that's 16px?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
```diff
@@ -556,6 +616,29 @@ const VideoPlayerComponent: ForwardRefRenderFunction<HTMLVideoElement, VideoPlay
...
+    for (let i = 0; i < tracks.length; i++) {
+      const track = tracks[i]
+
+      // Find the English captions track and mark it as "showing".
+      if (track.language === selectedTrack.language) {
+        track.mode = 'showing'
+      } else {
+        track.mode = 'hidden'
+      }
+    }
```
drillprop, line 661: can't we use forEach here?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
```diff
@@ -481,6 +506,41 @@ const VideoPlayerComponent: ForwardRefRenderFunction<HTMLVideoElement, VideoPlay
...
+      if (!captionsEnabled) {
+        track.mode = 'hidden'
+      } else {
+        if (activeTrack && activeTrack.language === track.language) {
+          track.mode = 'showing'
+        } else {
+          track.mode = 'hidden'
+        }
+      }
+    }
```
drillprop, line 561: same as above

packages/atlas/src/components/_video/VideoPlayer/utils.ts
```diff
@@ -113,6 +114,9 @@ export const hotkeysHandler = (
       }
       return
     case 'KeyC':
+      toggleCaptions()
```
drillprop, line 117: playerInstance doesn't have a method for toggling subtitles? Asking out of curiosity?

drillprop (changes requested):

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
kdembler: Probably `.map` will work even better?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
kdembler: Probably `.map` will work even better?

packages/atlas/src/components/_video/VideoPlayer/utils.ts
rafalpawlow: No, you have to do it manually

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
rafalpawlow: @kdembler @drillprop unfortunately `forEach` and `map` doesn't exists for this data. Dunno why to be honest

Zrzut ekranu 2022-08-12 o 15 28 26
Zrzut ekranu 2022-08-12 o 15 29 23

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
rafalpawlow: More than sure :P
Zrzut ekranu 2022-08-12 o 15 32 42

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
kdembler: Oh, it's actually not an array: https://developer.mozilla.org/en-US/docs/Web/API/TextTrackList

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
```diff
@@ -414,7 +414,7 @@ export const Container = styled.div<ContainerProps>`
         > div {
           background-color: ${cVar('colorBackgroundOverlay')} !important;
           font: ${cVar('typographyDesktopT300Strong')} !important;
-          font-size: 1em !important;
+          font-size: ${({ isFullScreen }) => (isFullScreen ? '0.64em' : '16px')} !important;
```
drillprop: subtitles are way too big on fullscreen. I think we should stick to 32px in this case. 

You also didn't handle the size for mobile phones. The size of the subtitles should be 12px. Not sure about fullscreen(android users), but I would keep 12px as well. 

Btw, do these subtitles work on ios on fullscreen?

drillprop (changes requested):

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
rafalpawlow: @drillprop yup, they work properly on ios

rafalpawlow: However they don't have proper font-family, but I'm affraid we can't do much about it

toiletgranny (changes requested):
Subtitles, finally! ❤️ I was able to turn them on and off, and I was super glad to see my language selection being preserved across videos. Also, the mechanism of picking a language works flawlessly! 🥰 Here's what I found:

1. I couldn't check how this remembering mechanism performs in less ideal circumstances, such as, for example when I pick French subtitles in one video and then open up another one that doesn't come with this particular language.
2. I also wasn't able to check if subtitles wrap properly if too long.
3. On mobile devices, ideally, selecting a different language closes the settings dialog instantly, in the same manner as it's currently done with playback speed controls.
4. It seems like we're missing an "Off" list item in the "Subtitles/CC" video settings section. This should allow me to disable subtitles.
5. The "CC" icon wasn't really meant to be displayed in the mobile version of the player, letting users enable/disable/change subtitles only from the settings menu, but I guess since we already have it, let's stick with it!
6. Lastly, the look of the subtitles themselves is somewhat off compared to the designs. On desktop, the font size is too small (`12px` instead of `16px`), on mobile there's a large border radius applied around the subtitles wrapper, etc.

rafalpawlow:
@toiletgranny
1. I've implemented it but frankly I don't really have a way to show you because we have mocked data. I tested it by changing values in the localStorage, but I think we will be able to do the final testing only with the query node integration.
2. We can't control it anyway 😛

kdembler:
> 2. We can't control it anyway 😛

I'm sure we can and we should test it :D

kdembler:
It seems there is horizontal padding only at the start and at the end of the subtitles. Margin left next to "Everyone" and margin right next to the last word and no horizontal padding at all in the middle lines:

CleanShot 2022-08-18 at 13 48 52@2x

Can we do something about it?

rafalpawlow:
One thing which comes to my mind is to change `display` property to `block`. WDYT @kdembler @toiletgranny?
FYI @kdembler longer subtitles lives in Russian language
Zrzut ekranu 2022-08-18 o 14 04 33

toiletgranny:
> One thing which comes to my mind is to change `display` property to `block`. WDYT @kdembler @toiletgranny? FYI @kdembler longer subtitles lives in Russian language Zrzut ekranu 2022-08-18 o 14 04 33

That would be very nice! If we could also keep the block as wide as the longest line of text (rather than having it always be 100% wide), that would make the implementation fully aligned with the designs in that regard [(design preview)](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=2090%3A141975)! 🥰

rafalpawlow:
@toiletgranny done!

toiletgranny:
@rafalpawlow Works and looks like a charm! One last thing from me would be enabling/disabling subtitles via the Settings dropdown: 
- **As is:** Having subtitles turned off, when I go into the Settings dropdown and select a language, subtitles are not being enabled. It's just the language selection that's being saved.
- **To be:** Selecting a language from the Settings dropdown, while subtitles are turned off, should turn them on.

kdembler (approved):
LGTM, but let's get approval from @toiletgranny and @drillprop as well

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.styles.ts
```diff
@@ -395,6 +399,41 @@ export const Container = styled.div<ContainerProps>`
...
+        padding: 0 ${sizes(3)};
+
+        > div {
+          display: inline-block !important;
+          background-color: ${cVar('colorBackgroundOverlay')} !important;
+          font: ${cVar('typographyDesktopT300Strong')} !important;
+          font-size: 12px !important;
+          letter-spacing: ${cVar('typographyDesktopT300StrongLetterSpacing')} !important;
+          text-transform: ${cVar('typographyDesktopT300StrongTextTransform')} !important;
+          padding: ${sizes(1)} ${sizes(2)};
```
drillprop, line 422: I think we would need to do some changes to the padding: 
This is a fullscreen desktop: 
image

drillprop: Correct me if I'm wrong, but line height also seems wrong on mobile: 
image

drillprop (changes requested):

drillprop (approved):

toiletgranny (approved):

PR #3103, opened 2022-08-10
Fix crashing embedded player

rafalpawlow:
@drillprop found out that when ending overlay shows up in embedded player, app crashes. It was caused by `ChannelLink` which uses `useConfirmationModal`. I added `ConfirmationModalProvider` to `CommonProviders`.

### DIFF
--- a/packages/atlas/src/App.tsx
+++ b/packages/atlas/src/App.tsx
@@ -1,7 +1,6 @@
 import { AnalyticsManager } from '@/AnalyticsManager'
 import { CommonProviders } from '@/CommonProviders'
 import { SignInModal } from '@/components/_auth/SignInModal'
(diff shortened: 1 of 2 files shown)


drillprop (changes requested):
Still crashes when I switch to mobile view and click play. iI looks like we're missing the Overlay provider as well? 
image

rafalpawlow:
@drillprop This is fixed on dev branch 
https://github.com/Joystream/atlas/blob/dev/packages/atlas/src/CommonProviders.tsx#L20

kdembler:
Let's change base to `dev`

rafalpawlow:
@kdembler done

kdembler (approved):
LGTM, let's get approval from @drillprop as well

drillprop (approved):

PR #3110, opened 2022-08-11
Withdraw bid confirmation dialog

### LINKED ISSUE #3072: Show confirmation dialog before withdrawing NFT bid
This will also let us display a fee for the user.

- Implement https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=5731%3A263684
- Add fee info to the dialog

rafalpawlow:
Closes #3072 

**Design**
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=5731%3A263684


packages/atlas/src/hooks/useNftTransactions.tsx
```diff
@@ -4,18 +4,27 @@ import { useCallback } from 'react'
...
   const { joystream, proxyCallback } = useJoystream()
   const handleTransaction = useTransaction()
   const [openModal, closeModal] = useConfirmationModal()
   const client = useApolloClient()
+  const { loading: withdrawFeeLoading, updateFullFeeHandler: updateWithdrawFee } = useFee(
+    'cancelNftBidTx',
+    undefined,
+    undefined,
+    true
+  )
```
kdembler: This will cause every user of `useNftTransaction` to calculate fee for this action. That means every NftTile and every VideoTile will calculate this fee (independently). That may not be a good idea. I think we should just create separate dialog component (like `ChangePriceDialog`) that can handle fee calculation and keep only the extrinsic calls in this file

kdembler (changes requested):

packages/atlas/src/hooks/useNftState.ts
```diff
@@ -1,19 +1,50 @@
...
+
   const auction = nft?.transactionalStatusAuction || null
+  const auctionId = auction && auction.id
+  const unwithdrawnUserBids = userBids?.filter(
+    (bid) =>
+      bid.auction.auctionType.__typename === 'AuctionTypeOpen' &&
+      (auction || bid.auction.id !== auctionId) &&
+      bid.auction.winningMemberId !== memberId
+  )
+  const bidFromPreviousAuction = unwithdrawnUserBids?.[0]
```
kdembler: Why is this happening in this hook now? Doesn't seem like data about unwithdrawn bids from previous auctions is needed for every nft/video tile, we still want to keep that in NftWidget only

kdembler (changes requested):

packages/atlas/src/hooks/useNftState.ts
rafalpawlow: What about withdrawing past bids from context menu?

kdembler: I'm pretty sure so far we only allowed withdrawing bids from context menu if the bid was made in the current auction. Is that not correct?

packages/atlas/src/components/_overlays/WithdrawBidDialog/WithdrawBidDialog.tsx
```diff
@@ -0,0 +1,73 @@
+import BN from 'bn.js'
+import { Dispatch, FC, SetStateAction } from 'react'
+
+import { Fee } from '@/components/Fee'
+import { NumberFormat } from '@/components/NumberFormat'
+import { AlertDialogModal } from '@/components/_overlays/AlertDialogModal'
+import { useFee } from '@/providers/joystream'
+import { formatDateTime } from '@/utils/time'
+
+type MemberId = string | null
```
kdembler: We already have a type called `MemberId` in `@joystream-lib/types`, maybe it's better to avoid confusion? You could import the type and in line 20 just do `MemberId | null`. Same for `NftId` below, we have `VideoId`

kdembler (approved):
Other than one comment, approving 👍

packages/atlas/src/components/_overlays/WithdrawBidDialog/WithdrawBidDialog.tsx
rafalpawlow: @kdembler done

PR #3113, opened 2022-08-11
Fee display improvements

### LINKED ISSUE #3108: Fee display improvements
- In `VideoWorkspace`, when fee is unknown and still loading, don't show "<0.01", just show "0.00". (We probably want to change the fee/action bar component)
- `NftWorkspace` doesn't show fee until I make a first change in the form, even though the defaults should be enough to compute a fee (if I just use defaults I will not see a fee)
- Same situation as above in edit membership view. Let's show the fee to update with current values right away
- In `CommentInput`, don't show "<0.01" when the fee isn't known. Use "0.00" for loading and load the base fee (for empty comment) right away
- Show fee in NFT remove from sale confirmation dialog
- Show fee in NFT accept bid dialog

rafalpawlow:
Closes #3108 

- [x] In `VideoWorkspace`, when fee is unknown and still loading, don't show "<0.01", just show "0.00". (We probably want to change the fee/action bar component)
- [x] `NftWorkspace` doesn't show fee until I make a first change in the form, even though the defaults should be enough to compute a fee (if I just use defaults I will not see a fee)
- [x] Same situation as above in edit membership view. Let's show the fee to update with current values right away
- [x] In `CommentInput`, don't show "<0.01" when the fee isn't known. Use "0.00" for loading and load the base fee (for empty comment) right away
- [x] Show fee in NFT remove from sale confirmation dialog
- [x] Show fee in NFT accept bid dialog


kdembler (changes requested):
Just one note - in both `AcceptBidDialog` and `CommentInput` let's load the fee right away - if no bid is selected we can use any for bid estimation. For comment - let's estimate fee using empty string right away. Other than that, looks good

kdembler (changes requested):
Hmm it seems with the new implementation comment fee will be loaded even before the user opens the input, right? In that case we should delay it until the input is actually opened so that we avoid a lot of unnecessary loading if there's a lot of comments

kdembler (approved):

PR #3127, opened 2022-08-17
Upload avatar implementation, validation improvements

### LINKED ISSUE #2884: Avatar Image Can be Uploaded
## Context
Ensure less friction and better experience for all users, since avatar image via URL come with many problems. Atlas team is building a simple back end service to support this.

## Scope
* [ ] Update the membership creation and editing flows in favour of uploaded avatars - https://github.com/Joystream/atlas/issues/1366
* [ ] Check the Avatar component in Figma and make sure the component is up to date with the design

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202474259593866) by [Unito](https://www.unito.io)

drillprop:
Fix #2884


kdembler:
I think the default state for editable avatar is not correct - it should show the overlay even without hover: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=9113%3A161999

drillprop:
I set this as a draft. Updating a member is not working, I think I messed up smth during the avatar update.

packages/atlas/src/components/Avatar/Avatar.styles.ts
```diff
@@ -22,14 +22,50 @@ type ContainerProps = {
...
-  size: Omit<AvatarSize, 'default'>
-}
+export const IconAndOverlayWrapper = styled.div`
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  border-radius: 100%;
+  z-index: 3;
```
rafalpawlow: Can we use `z-index` from const? I noticed that you are hardcoding z-indexes in several places in this file.

rafalpawlow (changes requested):

packages/atlas/src/components/Avatar/Avatar.styles.ts
drillprop: I changed that to `1`, and removed the z-indexes where it was redundant. I would also not opt to use `${zIndex.someproperty}` in this case, because this particular z-Index matters only locally in the context of the Avatar component itself.

rafalpawlow (changes requested):
I cannot delete avatar and publish changes
Zrzut ekranu 2022-08-24 o 13 56 58

rafalpawlow (changes requested):
Sorry, one more thing. I cannot delete or edit avatar photo when editing channel. When I click existing avatar I need to select file.

drillprop:
> Sorry, one more thing. I cannot delete or edit avatar photo when editing channel. When I click existing avatar I need to select file.

But this task was about editing and creating an avatar for **membership** not for the channel, right? This is the channel issue and it is also present in `dev` branch. I didn't change anything here. I created issue to fix this problem https://github.com/Joystream/atlas/issues/3173

rafalpawlow (approved):

rafalpawlow (approved):

packages/atlas/src/views/viewer/EditMembershipView/EditMembershipView.tsx
```diff
@@ -78,17 +126,25 @@ export const EditMembershipView: FC = () => {
...
     if (!joystream || !activeMembership) {
       return
     }
 
     const success = await handleTransaction({
       txFactory: async (updateStatus) => {
-        const memberInputMetadata: MemberInputMetadata = createMemberInputMetadata(formData)
+        let fileUrl = ''
+        if (data.avatar?.blob && dirtyFields.avatar) {
+          fileUrl = await uploadAvatarImage(data.avatar.blob)
```
kdembler: I think we should handle failure a bit better here. Currently if the avatar service is unavailable, it will look like the transaction failed and nothing in the console points to the avatar service being the issue:
CleanShot 2022-08-25 at 15 45 42@2x

I think ideally we would show some snackbar mentioning that it's the avatar that couldn't be uploaded so that users know that they can get rid of avatar and proceed anyway. The same applies to new member creation

packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalMembershipStep.tsx
```diff
@@ -75,62 +67,116 @@ export const SignInModalMembershipStep: FC<SignInModalMembershipStepProps> = ({
...
+      darkBackground
       title="Create Joystream membership"
-      subtitle="Tell us more about yourself."
+      backgroundImage={watch('avatar')?.url || undefined}
+      subtitle={
+        <>
+          To get the full Atlas experience, you need a free Joystream blockchain membership.
+          <Text as="p" variant="t100" color="inherit">
+            <Anchor href={JOYSTREAM_URL} target="_blank">
+              Learn about joystream &rarr;
```
kdembler:
```suggestion
              Learn about Joystream &rarr;
```

kdembler (approved):

Issue #3084, opened 2022-08-05
Incorrect NFT purchase view implementation + lack of support for the insufficient funds case scenario


toiletgranny:
The implementation of the NFT purchase view in the Buy now mode doesn't match [the designs](https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-purchase-view?node-id=1024%3A114229). However, the designs aren't that good either, as they don't provide any guidelines on handling the insufficient funds case scenario properly.

The goal of this task would be to revisit the NFT purchase view designs, address the insufficient funds case, and then create a detailed list of implementation issues to be passed over to engineers.

![play joystream org_video_27232](https://user-images.githubusercontent.com/35307309/183050909-210bcdfc-82e5-4daf-b05f-dac888a8e730.png)

cc @kdembler @dmtrjsg

kdembler:
@toiletgranny You may want to take a look at Carthage implementation of this, as we've done some improvements here: https://atlas-git-carthage-joystream.vercel.app/

kdembler:
Will be handled in https://github.com/Joystream/atlas/issues/3070

PR #3143, opened 2022-08-18
NFT purchase improvements

### LINKED ISSUE #3070: NFT purchase improvements (price + fee)
**Scope:**

- [ ] Include price breakdown in Buy now flow ("you will pay" and "transacion fee")
- [ ] Show loading state for "You will pay" when the fee is being fetched
- [ ] Ensure fee loads properly for both buy now and place bid
- [ ] Ensure price + fee is properly validated

---

I think there is a little discrepancy between what we show in the `Place bid` view and what we show in the `Buy now` view. 
Yes - there is a fee in the action bar, but I think we should be more clear about how much the user needs to actually pay
Place bid view:
image
Buy now view
image

rafalpawlow:
Closes #3070 

**Design**
https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-purchase-view?node-id=1036%3A102823

### DIFF
--- a/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
+++ b/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
@@ -583,7 +583,7 @@ export const NftPurchaseBottomDrawer: FC = () => {
                 />
               )}
             </Row>
-            {(bid || canBuyNow) && (
+            {bid || canBuyNow ? (
               <>
                 <Row>
                   <Text as="span" variant="t100" color="colorText">
@@ -620,6 +620,12 @@ export const NftPurchaseBottomDrawer: FC = () => {
                   )}
                 </Row>
               </>
+            ) : (
+              <Row>
+                <Text as="span" variant="t100" color="colorText">
+                  You need to fill out the amount first
+                </Text>
+              </Row>
             )}
             {type === 'open_auction' && bidLockingTime && (
               <Messages>
--- a/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
+++ b/packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
@@ -103,6 +103,7 @@ export const NftPurchaseBottomDrawer: FC = () => {
     }
   }, [isBuyNow, isEnglishAuction, isOpenAuction])
 
+  const bid = watch('bid')
   const auctionBuyNowPrice = isAuction ? hapiBnToTokenNumber(nftStatus.buyNowPrice || new BN(0)) : 0
   const bidLockingTime = isAuction && nftStatus.bidLockingTime && convertBlocksToDuration(nftStatus.bidLockingTime)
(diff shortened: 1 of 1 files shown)


kdembler (changes requested):
There's a split second after you input the bid that the bid in summary is shown before the loading triggers:

https://user-images.githubusercontent.com/12646744/185422009-d5b99f07-de12-492c-be6a-8718f8971b39.mp4

packages/atlas/src/views/global/NftPurchaseBottomDrawer/NftPurchaseBottomDrawer.tsx
```diff
@@ -587,27 +583,42 @@ export const NftPurchaseBottomDrawer: FC = () => {
...
+                  makeBidFeeLoading ? (
+                    <SkeletonLoader width={80} height={16} />
+                  ) : (
+                    <NumberFormat as="span" value={transactionFee} withToken variant="t100" color="colorText" />
+                  )}
                 </Row>
                 <Row>
                   <Text as="span" variant="h500" color="colorText">
                     You will pay
                   </Text>
```
drillprop, line 603: In the design there is a short text: "You need to fill out the amount first" when the input is empty. You didn't include that. 
https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-purchase-view?node-id=1036%3A102849

drillprop (changes requested):
I still don't like the flow and how validation works there, but not sure if the designs are up to date. 
1. When I click buy now and I don't have money. I still see the big call to action button "buy nft", which encourages me as a user to click: 
image
2.  In the inputs I know that at some point we decided that we will do validation on blur.  But should we actually do it that way in this case? Maybe we could use `onChange` mode in this case.  I'm still confused about what's happening to be honest

Can you consult this with a designer? I mean, I'm fine if we leave it as it is and if we decide to change that we could do this as a follow-up.

toiletgranny:
@drillprop That's a fair point. The UX here is far from ideal, and the designs did not really provide a solid solution for that case. That's why I created [this issue](https://github.com/Joystream/atlas/issues/3084#issue-1329710328), and now that I think of it, @kdembler, I think we should reopen it and let me take a closer look and think about it. I'd love to give you guys some ideas right now, but I really have to get back to Pioneer (cc @dmtrjsg)...

If you agree, I'd leave this implementation in its current state (in the PR) and open up a new PR once we know how to handle this case properly.

drillprop:
> @drillprop That's a fair point. The UX here is far from ideal, and the designs did not really provide a solid solution for that case. That's why I created [this issue](https://github.com/Joystream/atlas/issues/3084#issue-1329710328), and now that I think of it, @kdembler, I think we should reopen it and let me take a closer look and think about it. I'd love to give you guys some ideas right now, but I really have to get back to Pioneer (cc @dmtrjsg)...
> 
> If you agree, I'd leave this implementation in its current state (in the PR) and open up a new PR once we know how to handle this case properly.

Thanks for your input, I'd opt not to reopen your issue. Perhaps we could create another issue (third one😅) and rethink validation strategies in these particular cases - `buy now nft` and `placing bid`. And yeah, I'm also fine with doing this some other time. @kdembler wdyt?

drillprop (approved):
LGTM

kdembler (approved):

PR #3159, opened 2022-08-22
Show fee in delete video dialog

### LINKED ISSUE #3133: Show fee in delete video dialog
As in the title

rafalpawlow:
Closes #3133 

@kdembler @drillprop 
I had a bit of trouble firing off `useFee` on the first render and while looking for a solution I thought we could add `useFee` to `AlertDialogModal` since it is used there very often. Maybe instead of creating separate components for dialogs, we could fire off fee this way? The disadvantage is that I had to move `UserProvider` to `CommonProviders`. Let me know what you think.


kdembler (approved):
I think we could now also get rid of `WithdrawBidDialog` with this new approach?

In general, I like it, let's get approval from @drillprop as well

drillprop (approved):
I don't have a better solution to how we could improve this, so it's also okay for me. LGTM as well.

PR #3160, opened 2022-08-22
Withdraw channel payment flow

### LINKED ISSUE #3054: Withdraw Channel Payment Flow
## Context
Originating Epic

* https://github.com/Joystream/atlas/issues/2401

## Design
Link to Figma > [LINK](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=1547%3A19347)

## Scope
* [ ] Add "Max" button next to the withdraw input > [implemented here](https://www.figma.com/file/uHzbhNX9dP4QtZbIapvZI6/Transfer-dialog?node-id=1013%3A5857)
* [ ] Take fees into account when doing validation for available amount
* [ ] Take fees into account when user clicks "Max" button
* [ ] Make sure all the other validation works exactly as in the design
* [ ] Make sure the transfer dialog and withdraw dialog correctly set value **with decimals** in token input once you click the max button. Currently, this is not working in the transfer dialog

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1202701540991054) by [Unito](https://www.unito.io)

drillprop:
Fix #3054

Designs: 
https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=1602%3A25902
https://www.figma.com/file/uHzbhNX9dP4QtZbIapvZI6/Transfer-dialog?node-id=336%3A2174

### DIFF
--- a/packages/atlas/src/components/_inputs/FormField/FormField.tsx
+++ b/packages/atlas/src/components/_inputs/FormField/FormField.tsx
@@ -10,7 +10,6 @@ import {
   FormFieldHeader,
   FormFieldTitleWrapper,
   FormFieldWrapper,
-  Label,
   StyledInformation,
   StyledSvgActionWarning,
   SwitchLabel,
@@ -29,8 +28,8 @@ type WithSwitchProps =
     }
 
 export type FormFieldProps = PropsWithChildren<{
-  // We use ReactNode only in exceptional circumstances! In most cases you should use regular string.
-  label?: ReactNode
+  headerNode?: ReactNode
+  label?: string
   optional?: boolean
   error?: string
   disableErrorAnimation?: boolean
@@ -50,6 +49,7 @@ export const FormField = memo(
         className,
         optional,
         switchProps,
+        headerNode,
         tooltip,
         error,
         disableErrorAnimation,
@@ -85,7 +85,7 @@ export const FormField = memo(
         <FormFieldWrapper className={className} ref={ref}>
           {(label || description) && (
             <FormFieldHeader switchable={switchable}>
-              {label && (
+              {(label || headerNode) && (
                 <FormFieldTitleWrapper>
                   {switchable ? (
                     <SwitchLabel>
@@ -99,15 +99,11 @@ export const FormField = memo(
                       )}
                     </SwitchLabel>
                   ) : (
-                    <Label onClick={handleFocusOnClick}>
-                      {typeof label === 'string' ? (
-                        <Text as="span" variant="h300">
-                          {label}
-                        </Text>
-                      ) : (
-                        label
-                      )}
-                    </Label>
+                    <label onClick={handleFocusOnClick}>
+                      <Text as="span" variant="h300">
+                        {label}
+                      </Text>
+                    </label>
                   )}
                   {optional && (
                     <Text as="span" variant="t200" color="colorText" margin={{ left: 1 }}>
@@ -115,6 +111,7 @@ export const FormField = memo(
                     </Text>
                   )}
                   {tooltip && <StyledInformation {...tooltip} />}
+                  {headerNode && headerNode}
                 </FormFieldTitleWrapper>
               )}
               {description &&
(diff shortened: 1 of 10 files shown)


rafalpawlow (changes requested):
`Take fees into account when user clicks "Max" button`

When I click `MAX` I got 7 in input but according to this AC, it should be 6.94. Also it's not properly validated, I'm able to send the transaction from the video below.

https://user-images.githubusercontent.com/13706246/186150512-c3bc5d00-df37-4bfc-9b21-7c624fcf4182.mov

rafalpawlow:
Updated video with failed transaction

https://user-images.githubusercontent.com/13706246/186152030-da8ad2f9-5854-4da3-99c6-7637e82415a6.mov

rafalpawlow (changes requested):
When I have 0 balance on destination account, nothing happens when I click `MAX`

https://user-images.githubusercontent.com/13706246/186155792-745ec7de-1b1c-409a-802d-90e337a13173.mov

drillprop:
Ok, did some improvements: 
- Based on the slack discussion I subtracted existential deposit from the **channel balance**. So basically, in the member dropdown, you'll see only available tokens. I also added a small comment, that this will change in the future and we will use the channel state bloat bond.  
- We don't subtract fees when we click max in send funds dialogs. When you click max input is fulfilled with all the balance that you have. This was consulted with Adam. We added a small description in the input form field, that the transaction fee will be deducted from the amount in the input. And we will do this under the hood when the user clicks max. 
- When we click the max button, the value in the input is rounded **down** to the two decimal places. This causes some confusion, but I have no idea how to do this better. For example, I have a channel balance that equals 49,4998333334 tokens(494998333334 hapi). In the member dropdown, this will be displayed as 49.5 because NumberFormat for some reason is rounding this value up. If I click the max button the value in the input will equal to `49,49` This creates some inconsistency. Check the video:

https://user-images.githubusercontent.com/51168865/186354935-3ebf4639-8a92-4ecf-bec1-456e998259a5.mov

I could round the input value up as we did in the NumberFormat, but this value will be incorrect, because you don't have that amount of tokens, and validation will fail. So I decided to leave this as I implemented. Feel free if you have a better solution to do this. 

Also, there is another problem If I have `49,4998333334` and send `49,49` there will be always some hapi that's left in the wallet - You'll have `0,0098333334`. So basically you can't withdraw all the money that you have, there will be always some pennies left.

packages/atlas/src/components/_inputs/FormField/FormField.tsx
```diff
@@ -99,22 +99,19 @@ export const FormField = memo(
...
+                      </Text>
+                    </label>
                   )}
                   {optional && (
                     <Text as="span" variant="t200" color="colorText" margin={{ left: 1 }}>
                       (optional)
                     </Text>
                   )}
                   {tooltip && <StyledInformation {...tooltip} />}
+                  {headerNode && headerNode}
```
kdembler: This is equivalent to just `headerNode`. If you want to be safe from `0` for example, you would need to do `headerNode ? headerNode : null`

packages/atlas/src/providers/joystream/joystream.hooks.ts
```diff
@@ -155,15 +161,31 @@ export const useSubscribeAccountBalance = (controllerAccount?: string | null) =>
...
         controllerAccount || activeMembership.controllerAccount,
-        proxyCallback((balance) => setAccountBalance(new BN(balance)))
+        proxyCallback((balance) => {
+          if (opts?.isRewardAccount) {
+            // substract existential deposit from channel balance
+            // TODO in future existentialDeposit will be replaced with channel state bloat bond
+            const rewardBalance = new BN(balance).sub(chainState.existentialDeposit)
+            setAccountBalance(rewardBalance)
+            return
+          }
```
kdembler: I think first you need to confirm that the account actually has any funds. Because (at least currently, without the channel state bloat bond being in the account) new channels will have initial balance of 0 (the account doesn't really "exist" until it gets it first income). So for new channels the balance we display will actually be negative (0 - existential deposit):
CleanShot 2022-08-25 at 16 26 36@2x

(-0 because of rounding most likely)

kdembler (changes requested):

kdembler (approved):

PR #3168, opened 2022-08-24
Improve video and channel filtering

### LINKED ISSUE #3167: In video workspace sometimes thumbnail and video data is not correctly set on MultiFileSelect
On the `carthage` and the `subtitles` branches, we're getting some errors in the console once the video workspace is added. The animation is not fluent and sometimes thumbnails and videos are not loading at all.

The problem was probably introduced in #3087. 

https://user-images.githubusercontent.com/51168865/186414862-d8f94d3f-4f4a-4f67-bf4b-40765f0a942a.mov

rafalpawlow:
Fixes #3167


packages/atlas/src/config/channelFilter.ts
```diff
@@ -1,7 +1,24 @@
 import { ChannelWhereInput } from '@/api/queries'
 
+type ChannelFilter = string[]
+
+const channelFilterIds: ChannelFilter = []
+const avatarPhotoFilterIds: ChannelFilter = []
+const coverPhotoFilterIds: ChannelFilter = []
```
kdembler: Haven't noticed this before but we have those separate lists for filtered avatars, covers, medias and thumbnails. Let's use a single list instead that just contains banned asset IDs and pass that list as a filter to all NOT asset filters.

Maybe it would be easier if we actually had all filters in a single `contentFilter` file?

kdembler (changes requested):
1. There are still some unnecessary `NOT` filters applied:
CleanShot 2022-08-25 at 18 51 56@2x
CleanShot 2022-08-25 at 18 52 16@2x
2. This wasn't introduced by this PR, but I think it may have been introduced by your previous filtering PR. Even if not, let's try to fix here. Both in viewer and studio there's a lot of unnecessary requests being fired. In viewer, separate `GetNfts` query is fired for every NFT tile and in studio, separate `GetFullVideos` is fired for ever publisher video tile

packages/atlas/src/api/hooks/video.ts
```diff
@@ -44,6 +47,20 @@ export const useFullVideo = (
...
+) => {
+  const { data, ...queryRest } = useGetFullVideoQuery({
+    ...opts,
+    variables: { where: { id } },
+  })
+  return {
+    video: data?.videoByUniqueInput,
+    ...queryRest,
+  }
+}
```
kdembler: I don't like that we're using this type of hook only for videos, we're inconsistent when it comes to nfts and channels. We should use the same pattern of hook names.

What if we only keep the old hook (`useFulLVideo`) and add an optional param to it like `cached?: boolean`? Based on that, we either run the `GetFullVideos` to get the full and filtered data from the network, or if true, we run `GetFullVideo` with `cache-only` fetch policy? What do you think?

rafalpawlow: Sounds good, however I don't think we need to define `fetch-policy`, it's `cache-first` by default, right?

kdembler: Sorry, I meant `cache-only`

kdembler (changes requested):
1. Do the same trick with `cached` flag for:
- NFTs
- Channels
- Basic videos
2. Add issue for video view always fetching data from network

kdembler:
When visiting video directly from a link I get "Video not found": https://atlas-2fpbuwtni-joystream.vercel.app/video/35

kdembler (approved):

Issue #3118, opened 2022-08-12
Initial subtitles integration


kdembler:
Initial extrinsics integration and a playground

PR #3169, opened 2022-08-24
add subtitles publishing integration

### LINKED ISSUE #3043: Integration for adding subtitles
Integrate `VideoWorkspace` subtitles support:
- Supply subtitles in the video metadata
- Start subtitles upload
- Show subtitles for existing videos
- Allow downloading subtitles from existing videos
- Allow editing subtitles

kdembler:
**Left to do:** (possibly as a followup)
- Allow downloading already published subtitle (via "Download file" context menu option)
- Improve subtitle validation in VideoForm:
    - Disallow publishing without selecting a subtitle file
    - Disallow selecting subtitle language if subtitles for that language have already been published
- Verify that code I removed from `UploadStatusGroup.tsx` wasn't needed. Confirm that you can still re-upload file from a new browser.

Apart from that everything should be working.

Please reach out to ask any questions or suggest any improvements.

Closes #3043 and #3118


packages/atlas/src/components/_inputs/SubtitlesComboBox/SubtitlesCombobox.tsx
```diff
@@ -2,51 +2,53 @@ import styled from '@emotion/styled'
...
   disabled: boolean
+  displayName: string
 }
 
 type SubtitlesComboboxProps = {
-  subtitlesArray: Subtitles[] | null
-  languages: string[]
-  onLanguageAdd: (language: Subtitles) => void
-  onLanguageDelete: (language: Subtitles) => void
+  subtitlesArray: SubtitleInput[] | null
```
drillprop: Hmm, maybe I nitpicking here, but the singular word "subtitle" means: 
1. the title that comes after the main title
2. one line of a movie `subtitles`, I guess?


To avoid some misunderstanding, I think we should always use the plural noun `subtitles`. So in my opinion we should change this to **SubtitlesInput**. The same applies to `SubtitleBox`(I named it wrong when I was working on subtitles UI).
WDYT?

packages/atlas/src/providers/videoWorkspace/hooks.ts
```diff
@@ -34,6 +34,19 @@ export const useVideoWorkspaceData = () => {
     onError: (error) => SentryLogger.error('Failed to fetch video', 'useVideoWorkspaceData', error),
   })
 
+  const subtitlesArray = useMemo(
+    () =>
+      video?.subtitles.map(
+        (s) =>
+          ({
+            languageIso: s.language.iso,
+            type: s.type === 'closed-captions' ? 'closed-captions' : 'subtitles',
```
drillprop, line 43: hmm, can't we just use `s.type` here?

packages/atlas/src/utils/size.ts
```diff
@@ -1,4 +1,4 @@
-const units = ['bytes', 'KB', 'MB', 'GB']
+const units = [' bytes', 'KB', 'MB', 'GB']
```
drillprop, line 1: Is this white space intentional?

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
```diff
@@ -135,11 +135,16 @@ export const UploadStatus: FC<UploadStatusProps> = ({ isLast = false, asset, siz
...
+      [asset.type === 'video' ? 'video/*' : asset.type === 'subtitle' ? 'text/vtt' : 'image/*']: [],
     },
   })
 
-  const fileTypeText = isVideo ? 'Video file' : `${asset.type.charAt(0).toUpperCase() + asset.type.slice(1)} image`
+  const fileTypeText =
+    asset.type === 'video'
+      ? 'Video file'
+      : asset.type === 'subtitle'
+      ? 'Subtitle'
```
drillprop:
```suggestion
      ? 'Subtitles'
```
I would also change `asset.type` as I mentioned in one of my comment above

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
```diff
@@ -271,7 +276,13 @@ export const UploadStatus: FC<UploadStatusProps> = ({ isLast = false, asset, siz
...
           <FileInfo size={size}>
             <FileInfoType warning={isReconnecting && size === 'compact'}>
-              {isVideo ? <SvgActionVideoFile /> : <SvgActionImageFile />}
+              {asset.type === 'video' ? (
+                <SvgActionVideoFile />
+              ) : asset.type === 'subtitle' ? (
+                <SvgActionClosedCaptions />
+              ) : (
+                <SvgActionImageFile />
+              )}
```
drillprop, line 287: image
I'm not sure where we setting these values in the code, but I guess, this should be somewhere in this file. Shouldn't we put subtitles size on the next column(the third one)? Below sizes of the video file and image file?
**Suggestion** - maybe in the second column we could show the language or language iso?

drillprop (changes requested):
Functionality wise works well 👍🏻 , added a few suggestions.

packages/atlas/src/providers/videoWorkspace/hooks.ts
kdembler: Not really, because `s.type` is of type `string`, I need to create additional Typescript guards so we limit the type to `'subtitles' | 'closed-captions'`

packages/atlas/src/utils/size.ts
kdembler: Yep, because when we format the size, `10KB` looks good, but `10bytes` is a bit weird

packages/atlas/src/providers/videoWorkspace/hooks.ts
drillprop: What about adding `... as const`? Shouldn't do the job?

drillprop: nvm, you're right. This will also not work. We're getting this from graphql.

packages/atlas/src/components/_inputs/SubtitlesComboBox/SubtitlesCombobox.tsx
kdembler: Done, renamed all instances

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
kdembler: Done

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
kdembler: Good call, added subtitle language in the second column

drillprop (approved):
Looks good to me 👍🏻 but let's maybe wait for final approval from @rafalpawlow

rafalpawlow (changes requested):
Something is not right, I added 4 different languages but I get doubled `ru` after publish
Zrzut ekranu 2022-08-25 o 20 23 28
Zrzut ekranu 2022-08-25 o 20 25 18

kdembler:
@rafalpawlow This was an initial upload with subtitles or video edit? I can not reproduce this issue, added a bunch of different languages and everything seems to look fine. If you can reproduce, please provide more details.

rafalpawlow (approved):

PR #3174, opened 2022-08-25
Video player subtitles integration

### LINKED ISSUE #3044: Video player subtitles integration
Once the pure UI/player work is done, let's use actual subtitles from QN/storage network:

- Add "immediate" mode to asset `useAsset` - just construct the URL without trying to download the asset
- Show the list of available subtitles based on QN data
- Use subtitles from the storage network

rafalpawlow:
Closes #3044

### DIFF
--- a/packages/atlas/src/providers/assets/assets.hooks.ts
+++ b/packages/atlas/src/providers/assets/assets.hooks.ts
@@ -1,7 +1,8 @@
-import { useCallback, useEffect, useMemo, useState } from 'react'
+import { useEffect, useMemo } from 'react'
+import shallow from 'zustand/shallow'
 
 import { BasicMembershipFieldsFragment, StorageDataObjectFieldsFragment, SubtitlesFieldsFragment } from '@/api/queries'
-import { LANGUAGES_LOOKUP } from '@/config/languages'
+import { createLookup } from '@/utils/data'
 
 import { ResolvedAsset, useAssetStore } from './assets.store'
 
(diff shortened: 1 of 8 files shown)


packages/atlas/src/providers/assets/assets.store.ts
```diff
@@ -1,29 +1,37 @@
...
+type ContentId = string
+export type ResolvedAsset = {
   url?: string | null
   blob?: File | Blob | null
 }
 
 type AssetStoreState = {
   assets: Record<string, ResolvedAsset> // mapping of content ID to resolved assets
   pendingAssets: Record<string, StorageDataObjectFieldsFragment> // list of content IDs pending resolution
+  immediatePendingAssets: Record<string, StorageDataObjectFieldsFragment>
```
kdembler: Do we need a separate list for this type of assets? I think it will be easier if we just keep a single `pendingAssets` list, just include an optional boolean flag for `immediate`.

Btw, I think `immediate` may be a bit undescriptive of what it really does. Maybe instead we call this something like `skipAssetTest` or something along the lines? That should give a better understanding of what effect it has? WDYT?

packages/atlas/src/providers/assets/assets.hooks.ts
```diff
@@ -39,3 +40,74 @@ export const useMemberAvatar = (member?: BasicMembershipFieldsFragment | null):
 
   return { url: null, isLoadingAsset: true }
 }
+
+export const useSubtitlesAssets = (subtitles?: SubtitlesFieldsFragment[] | null) => {
```
kdembler, line 45: I think this hook is overly complicated, it's very hard for me to understand what is happening.

There are also some issues:
1. You use full `assets` and `pendingAssets` state. This hook will force a re-render whenever something in those objects changes which may be quite often. We need to be careful to not reduce performance of `VideoView`
2. This is placed inside the assets provider but also contains some logic about display of subtitles (CC, language, etc). We should separate concerns of resolving assets and creating lists of subtitles.

Take a look at my PR to your branch: https://github.com/rafalpawlow/atlas/pull/9 I simplified this hook with couple of scoped store accessors and single useEffect. I managed to reduce the number of VideoView renders by around 10 (it's still stupidly high, but that's because of other unoptimized code). Please review, I think it works properly.

kdembler (changes requested):

kdembler (approved):

PR #3185, opened 2022-08-29
Fix unsupported file type dialog

### LINKED ISSUE #2910: Unsupported file type dialog problem
When you try to upload video file as thumbnail image, an `unsupported file type` dialog appears, which is fine. But then it keeps popping up with various actions unrelated to this.

1. Go to Video workspace
2. Upload video file
3. Drag and drop video file to thumbnail zone
4. Cancel unsupported file type dialog 
5. Try to upload video file again

https://user-images.githubusercontent.com/13706246/175499101-a76a2038-2801-4e35-89aa-13b5c0ceff0e.mov

rafalpawlow:
Fixes #2910

### DIFF
--- a/packages/atlas/src/components/_inputs/FileSelect/FileSelect.tsx
+++ b/packages/atlas/src/components/_inputs/FileSelect/FileSelect.tsx
@@ -1,5 +1,5 @@
 import beazierEasing from 'bezier-easing'
-import { FC, MouseEvent, useCallback, useEffect } from 'react'
+import { FC, MouseEvent, useCallback, useEffect, useRef } from 'react'
 import { DropzoneOptions, FileRejection, useDropzone } from 'react-dropzone'
 import { useTransition } from 'react-spring'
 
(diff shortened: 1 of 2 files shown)


packages/atlas/src/components/_inputs/FileSelect/FileSelect.tsx
```diff
@@ -37,9 +36,9 @@ export type FileSelectProps = {
   isFileLoading?: boolean
   onReAdjustThumbnail?: () => void
   onDropRejected?: (fileRejections: FileRejection[]) => void
-  onError?: (error: string | null, fileType: FileType) => void
   error?: string | null
   maxSize?: number
+  setError: Dispatch<SetStateAction<string | null>>
```
drillprop: I would keep the `onError` name. Also `Dispatch<SetStateAction<string | null>>` pretty much the same as `(error: string | null) => void`, and the latter is more readable to me to be honest.

drillprop (changes requested):
This PR fixes the problem but I found another issue. When I pick a video that has some incorrect metadata the warning dialog I showing to me twice. I can send you the file to test it out if you want. 

https://user-images.githubusercontent.com/51168865/187686591-5e0b8b86-2d3a-432d-8ed9-4df4f487968c.mov

I'm fine if you do this as followup

kdembler:
> I'm fine if you do this as followup

Let's fix it here

drillprop (approved):
LGTM

PR #3187, opened 2022-08-31
Update member dropdown

### LINKED ISSUE #2556: Introduce new switchable member dropdown
## Scope

- [ ] Implement changes to member dropdown, allowing to switch between member and channel
- [ ] Add animation for above
- [ ] Add animation for switching channel in the new flow
- [ ] Display balance breakdown in a tooltip

---

## Context
Faucet gives spendable 10 tJoys while new membership gets 100 of locked funds (spending only on transaction fees allowed).
Also channel balance is not visible to the viewer 

## Scope
* [ ] Locked due to faucet constraints (other locks tbc) Details on what is possible to do with locked funds (Invitation lock, for all JOY that are received from faucet or when other member invites you). https://github.com/Joystream/atlas/issues/3090
* [ ] Display channel balance in viewers top nav
* [ ] Suggested to show both member and channel balance at the same time (not sum it up, but show next to each other)
* [ ] https://github.com/Joystream/atlas/issues/3085

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202736674734968) by [Unito](https://www.unito.io)

drillprop:
Fix #2556 
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3201%3A111990

### DIFF
--- a/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
+++ b/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
@@ -100,8 +100,8 @@ export const AddAvatar = styled.div`
 
 export const StyledIconWrapper = styled(IconWrapper)`
   position: absolute;
-  right: -8px;
-  bottom: -8px;
+  right: -${sizes(2)};
+  bottom: -${sizes(2)};
 `
 
 type FixedSizeContainerProps = {
@@ -118,7 +118,7 @@ export const UserBalance = styled.div`
   grid-auto-flow: column;
   grid-auto-columns: max-content;
   align-items: center;
-  gap: 5px;
+  gap: ${sizes(1)};
   margin-top: ${sizes(2)};
 `
 
--- a/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
+++ b/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
@@ -30,6 +30,7 @@ export const BlurredBG = styled.div<{ memberUrl?: string | null; channelUrl?: st
     background-repeat: no-repeat;
     background-position: center;
     content: '';
+    background-color: ${cVar('colorCoreNeutral500')};
   }
 
   &::before {
@@ -46,7 +47,7 @@ export const BlurredBG = styled.div<{ memberUrl?: string | null; channelUrl?: st
 export const Filter = styled.div`
   position: absolute;
   backdrop-filter: blur(32px);
-  background: ${cVar('colorBackgroundOverlay')};
+  background-color: ${cVar('colorBackgroundOverlay')};
   width: 280px;
   height: 100%;
 `
@@ -78,6 +79,14 @@ export const StyledAvatar = styled(Avatar)<{ isDisabled: boolean }>`
   );
   opacity: ${({ isDisabled }) => (isDisabled ? 0.25 : 1)};
   transition: opacity ${cVar('animationTransitionMedium')};
+
+  :hover {
+    opacity: 1;
+  }
+
+  :active {
+    opacity: 0.75;
+  }
 `
 
 export const AddAvatar = styled.div`
--- a/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
+++ b/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
@@ -67,10 +67,6 @@ export const AvatarsGroupContainer = styled.div`
   margin-bottom: ${sizes(6)};
 `
 
-export const AvatarWrapper = styled.div`
-  position: relative;
-`
-
 export const StyledAvatar = styled(Avatar)<{ isDisabled: boolean }>`
   width: 40px;
   height: 40px;
@@ -79,9 +75,16 @@ export const StyledAvatar = styled(Avatar)<{ isDisabled: boolean }>`
   );
   opacity: ${({ isDisabled }) => (isDisabled ? 0.25 : 1)};
   transition: opacity ${cVar('animationTransitionMedium')};
+`
+
+export const AvatarWrapper = styled.div`
+  position: relative;
+  cursor: pointer;
 
   :hover {
-    opacity: 1;
+    ${StyledAvatar} {
+      opacity: 1;
+    }
   }
 
   :active {
--- a/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
+++ b/packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
@@ -114,7 +114,7 @@ export const FixedSizeContainer = styled.div<FixedSizeContainerProps>`
 `
 
 export const UserBalance = styled.div`
-  display: grid;
+  display: inline-grid;
   grid-auto-flow: column;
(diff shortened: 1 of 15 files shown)


kdembler:
Can't test right now because of a failed build but when checking it out yesterday I wasn't able to initialize funds transfer, the link would always open dialog for withdrawing funds from channel, even if I was in the "member mode"

drillprop:
> Can't test right now because of a failed build but when checking it out yesterday I wasn't able to initialize funds transfer, the link would always open dialog for withdrawing funds from channel, even if I was in the "member mode"

fixed

packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
```diff
@@ -0,0 +1,173 @@
...
+  height: 100%;
+  background-color: ${cVar('colorBackgroundAlpha')};
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`
+
+export const StyledIconWrapper = styled(IconWrapper)`
+  position: absolute;
+  right: -8px;
```
rafalpawlow: I think we should use `sizes` here

packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.styles.ts
```diff
@@ -0,0 +1,173 @@
...
+  height: ${({ height }) => (height ? (typeof height === 'number' ? `${height}px` : height) : 'unset')};
+  width: ${({ width }) => (width ? (typeof width === 'number' ? `${width}px` : width) : 'unset')};
+`
+
+export const UserBalance = styled.div`
+  display: grid;
+  grid-auto-flow: column;
+  grid-auto-columns: max-content;
+  align-items: center;
+  gap: 5px;
```
rafalpawlow: I know this is not your change, but it should be `4px`, let's also change it to `sizes`

rafalpawlow (changes requested):
Works pretty well, few small issues

- [x] this icon seems to have wrong size
Zrzut ekranu 2022-09-1 o 12 23 26
Zrzut ekranu 2022-09-1 o 12 23 43

- [x] tooltip has wrong width
Zrzut ekranu 2022-09-1 o 12 39 29
Zrzut ekranu 2022-09-1 o 12 40 31

- [x] we agreed to add `interactive` flag to the tooltip I mentioned above
- [x] when I touch tooltip on mobile, it fires links and buttons beneath (not sure if we can do something about it)
- [ ] do you think we could add some offset to this tooltip like it's on the design? Currently, it's a bit of a meld with dropdown for me

Zrzut ekranu 2022-09-1 o 12 56 29

drillprop:
> * [ ]  do you think we could add some offset to this tooltip like it's on the design? Currently, it's a bit of a meld with dropdown for me

What do you mean? The offset is set to 8px by default

rafalpawlow:
@drillprop Now it is ok, but before the tooltip was exactly in the middle of the dropdown. Have you changed anything in this regard?

rafalpawlow:
@drillprop One more question. Shouldn't we show channel variant as default in studio?

drillprop:
> @drillprop Now it is ok, but before the tooltip was exactly in the middle of the dropdown. Have you changed anything in this regard?

We had this offset set previously as well, but yeah I did a small change.
The tooltip is appended to MemberDropdown container now, not the HTML element which keeps the account balance inside https://github.com/Joystream/atlas/pull/3187/commits/1df6e12b99c2e1fe7156dd0475d03d1539fb9efd
(check the `containerRef` prop)
> @drillprop One more question. Shouldn't we show channel variant as default in studio?

Yeah, I guess you're right. I'll change that.

toiletgranny (changes requested):
@drillprop It felt soo good to take this component for a spin! The animations you implemented are buttery smooth! 🥰

1. **As is:** Account balance tooltip appears upon hovering over this entire area. This causes accidental tooltip triggers. 
    
**To be:** Ideally, the tooltip is triggered when hovering specifically over the numerical value and the JOY icon.

2. **As is:** In the total balance tooltip, JOY token amounts are displayed using the `--typography-desktop-t200` CSS variable. Also, the the contents of each row in the breakdown is not centered vertically.
**To be:** JOY token amounts should be displayed using the `--typography-desktop-t200-strong` CSS variable, and there should be `align-items: center` applied on all rows in the breakdown (make sure the "[lock-icon] Invitation lock" label remains centered vertically after applying this change).

4. **As is:** I can't seem to be able to scroll down the contents of the member dropdown.
    
**To be:** I should be able to do so.

5. **As is:** When in Studio, the Member dropdown component opens up in the Member mode by default.
**To be:** When in Studio, the Member dropdown component should open up in the Channel mode by default.

6. **As is:** Having just one channel created, I see the "Switch channel" list item in the Channel mode of the Member dropdown component. 
    
**To be:** As long as I have just one channel, I should see the "Add new channel" list item instead of the "Switch channel" [(documentation reference)](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3201%3A114961).

7. All transition animations seem to fire with a slight delay, and rather than animating from 0% to 100%, they go from ~25% to 100% only. This is best inspected if recorded on the screen and then played in slow motion. Happy to jump on a call to discuss details. 

8. **As is:** Hovering over the little icon indicators in the "Member dropdown switcher" component triggers tooltip visibility but does not lighten up the avatar and does nothing upon pressing.
   
   **To be:** Ideally, hover interactions should always reflect what can be clicked on. In this case, I'd opt for either not displaying the tooltip at all when hovering over the little icon in the corner, OR letting the user click on the entire area of the "Member dropdown switcher" instead of just the Avatar component itself, as it's implemented now. In the latter case, the Avatar should lighten up upon hovering over the little icon in the corner too.

9. **As is:** Having created a new channel from the "Switch channel" -> "Add new channel" menu (which I already pointed out is not correct if I have only one channel), opening the Member dropdown again I can see it still in the "Switch channel" mode (it should be in the "home" mode), and its height doesn't reflect the new list contents.
   
   **To be:** Opening the Member dropdown component should always default to the "home" screen of either the Channel mode or Member mode, depending on the experience. 

10. **As is:** Changing channel in the Viewer experience doesn't do anything. 
    **To be:** It should set the active channel to the one I clicked on. This will be important once we implement the YouTube Partner Program landing page.

toiletgranny:
@drillprop Thanks a lot! I can see that points 2 and 4 are not 100% addressed yet, so let me clarify:
- For the 2nd point, I meant applying this bolder text style token to all JOY token values on the right-hand side. Oh, and since we're here, is there a way to disable the above tooltip now that we have a richer one below? It feels very odd to have two tooltips for the same thing. 😅 
CleanShot 2022-09-02 at 15 52 04
   - Can we also make sure the lock icon and the "Invitation lock" label remain vertically centered in their surrounding container? 🙏
- For the 4th point, we're halfway there, but what I meant is that the Member dropdown should _always_ default to either the Member or Channel mode, based on the experience. This means that if I open up the Member dropdown in the Viewer experience, change its mode to Channel, close it and then reopen — ideally, I see it in the Member mode again because of the fact I'm still in the Viewer experience. Hope it makes sense! :D

drillprop:
> Can we also make sure the lock icon and the "Invitation lock" label remain vertically centered in their surrounding container?

It's already centered. I don't think I can center this lock more :P 

Besides that, everything should be fixed.

rafalpawlow (approved):

PR #3191, opened 2022-08-31
Fix issue with not displaying correct error in Transaction modal

drillprop:


packages/atlas/src/providers/transactions/transactions.hooks.ts
```diff
@@ -250,10 +235,14 @@ export const useTransaction = (): HandleTransactionFn => {
 
         if (extrinsicFailed) {
           // extract error code from error message
-          const errorCode = Object.keys(ErrorCode).find((key) => error.message.includes(key)) as ErrorCode | undefined
```
drillprop, line 253: I had to change that because we were getting the wrong error code. For example, if someone got an error `InsufficientBalanceForChannelCreation`, we were displaying an error for `InsufficientBalance`. This is because the `includes` method was finding the word `InsufficientBalance`  first and returning that.

kdembler (approved):

PR #3193, opened 2022-08-31
Fix issue with not being able to remove the member avatar

### LINKED ISSUE #3192: User can't remove uploaded member avatar
as the title says

drillprop:
Also:
- fix an issue with changing form to dirty, when you focus and blur on about text area
- ~improve calculating fees. Now, when the form is not dirty you'll see 0 fees.~
- fix issue with infinitely loading avatars

Fix #3192

### DIFF
--- a/packages/atlas/src/providers/assets/assets.hooks.ts
+++ b/packages/atlas/src/providers/assets/assets.hooks.ts
@@ -38,7 +38,7 @@ export const useMemberAvatar = (member?: BasicMembershipFieldsFragment | null): 
   } else if (avatar?.__typename === 'AvatarObject') {
     return avatarAsset
   }
-  // if avatar is `undefined` it means that avatar is not loaded yes, If it's `null` it means that it's not set
+  // if avatar is `undefined` it means that avatar is not loaded yet, If it's `null` it means that it's not set
   return { url: null, isLoadingAsset: avatar === null ? false : true }
 }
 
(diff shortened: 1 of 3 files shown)


packages/atlas/src/providers/assets/assets.hooks.ts
```diff
@@ -38,8 +38,8 @@ export const useMemberAvatar = (member?: BasicMembershipFieldsFragment | null):
   } else if (avatar?.__typename === 'AvatarObject') {
     return avatarAsset
   }
-
-  return { url: null, isLoadingAsset: true }
+  // if avatar is `undefined` it means that avatar is not loaded yes, If it's `null` it means that it's not set
```
rafalpawlow: typo 🙂
```suggestion
  // if avatar is `undefined` it means that avatar is not loaded yet, If it's `null` it means that it's not set
```

rafalpawlow (approved):

packages/atlas/src/providers/joystream/joystream.hooks.ts
```diff
@@ -254,6 +254,8 @@ export const useFee = <TFnName extends TxMethodName, TArgs extends Parameters<Jo
     () =>
       debouncePromise(async (args?: TArgs) => {
         if (!args || !accountId) {
+          setFullFee(new BN(0))
+          setLoading(false)
```
kdembler: Why was this change needed? I did a quick test with commenting and this changes the fee display behavior:

https://user-images.githubusercontent.com/12646744/188065724-96084c9b-26b4-400b-a143-95dc98cd8269.mp4

Previously the fee would stay the same when signing the tx. Maybe this actually a problem in the CommentInput, I'm not sure, but we should check.

kdembler (changes requested):

packages/atlas/src/providers/joystream/joystream.hooks.ts
kdembler: Also if I reject the signature, the fee stays 0

kdembler (approved):

PR #3197, opened 2022-09-01
Fix selected default subtitles

rafalpawlow:


packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
```diff
@@ -562,11 +571,15 @@ const VideoPlayerComponent: ForwardRefRenderFunction<HTMLVideoElement, VideoPlay
...
-      setActiveTrack(findDefaultLanguage || availableTextTracks[0])
-    }
   }, [activeTrack, availableTextTracks, captionsEnabled, captionsLanguage, player, storedLanguageExists])
 
+  useMountEffect(() => {
+    if (!captionsEnabled || !availableTextTracks) {
+      return
+    }
+    setActiveTrack(findDefaultLanguage)
+  })
```
kdembler: Won't `availableTextTracks` be always empty (empty array) at first render? Does this condition make sense then?

packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
```diff
@@ -562,11 +571,15 @@ const VideoPlayerComponent: ForwardRefRenderFunction<HTMLVideoElement, VideoPlay
         }
       }
     }
-    if (!activeTrack) {
-      setActiveTrack(findDefaultLanguage || availableTextTracks[0])
-    }
   }, [activeTrack, availableTextTracks, captionsEnabled, captionsLanguage, player, storedLanguageExists])
 
+  useMountEffect(() => {
+    if (!captionsEnabled) {
```
kdembler: Well, you removed the condition from here but this doesn't really solve the issue. `findDefaultLanguage` has the same exact condition:
```ts
    if (!availableTextTracks) {
      return
    }
```
Is this mount effect needed at all? Based on the code I really doubt it has any effect, since like I said, `availableTextTracks` will most likely be always empty on first render

kdembler (changes requested):

kdembler (approved):

PR #3198, opened 2022-09-01
Fix issue with not displaying error, show correct balance in member dropdown

drillprop:


packages/atlas/src/providers/joystream/joystream.hooks.ts
```diff
@@ -165,9 +165,7 @@ export const useSubscribeAccountBalance = (
         proxyCallback(({ availableBalance, lockedBalance }) => {
           setLockedAccountBalance(new BN(lockedBalance))
           if (opts?.isRewardAccount) {
-            // substract existential deposit from channel balance
-            // TODO in future existentialDeposit will be replaced with channel state bloat bond
-            const rewardBalance = new BN(availableBalance).sub(chainState.existentialDeposit)
+            const rewardBalance = new BN(availableBalance).sub(chainState.channelStateBloatBondValue)
```
kdembler: While this will work most of the time, this is actually a bit more nuanced than that. The issue is that here we subtract the current channel bloat bond, but for true balance we would need to subtract the channel bloat bond that you paid when you created the channel. Check this example:
1. The bloat bond is 1 JOY
2. I create a channel
3. My true balance is `X - 1 JOY`
4. Content lead changes the bloat bond to 0.5 JOY
5. My true balance is still `X - 1 JOY,` not `X - 0.5 JOY` as this code would produce

You can see the relevant runtime code here: https://github.com/Joystream/joystream/blob/carthage/runtime-modules/content/src/lib.rs#L3662 You can see that the code access `channel.channel_state_bloat_bond.amount` which is the value of bloat bond persisted at channel creation.

Now, to get this accurate value at the moment we would need to run an RPC query to runtime node to query the channel and extract the value out of that. You can see I asked on Slack (https://joystreamworkspace.slack.com/archives/C039QJHMM38/p1662096030742769) whether we could expose this via QN to make our life simpler. For now, let's merge as it is, but please just add a TODO comment here to capture above

kdembler (approved):

packages/atlas/src/providers/joystream/joystream.hooks.ts
drillprop: ok, didn't think this through. Added a comment in `joystream.hooks.ts`

packages/atlas/src/providers/joystream/joystream.hooks.ts
```diff
@@ -142,7 +142,7 @@ export const useBucketsConfigForNewChannel = () => {
 }
 
 type UseSubscribeAccountBalanceOpts = {
-  isRewardAccount: true
+  channelStateBloatBond?: string
```
kdembler: I think we should accept BN here

kdembler (approved):

PR #3217, opened 2022-09-06
YPP landing page - hero section

### LINKED ISSUE #2975: YT LP hero  section
## Context
- [YPP programme ](https://github.com/Joystream/atlas/issues/2973)

## Scope

- [ ] Shown for  guest users and members that have not signed up to YPP with their selected Joystream channel
- [ ] CTA Authorise with Youtube
- [ ] ☝️ Image of Joystream slides in the same way when you scroll down as Joystream.org images

⚠️ participation in the programme will be on a channel basis, so one membership can participate multiple times with diff channels. meaning: when signed in member hits YPP page on atlas, we always show start programme, but when a channel is selected from the drop down, we validate if that channel is already in the programme. 

For registered channels this page will display dasboard instead of LP 

out of scope:
- #2974 
- [ ] Top Creators Carousel
- [ ] Referred by

## Design Preview

[LInk to Figma>](https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1604%3A85511) 

Screenshot 2022-08-18 at 20 17 58

rafalpawlow:
Closes #2975 

**Design**
https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1789%3A110713


packages/atlas/src/views/global/YppLandingView/YppLandingView.styles.ts
```diff
@@ -0,0 +1,50 @@
+import { css } from '@emotion/react'
+import styled from '@emotion/styled'
+
+import { media, sizes, zIndex } from '@/styles'
+
+export const Wrapper = styled.div`
+  padding-top: ${sizes(24)};
```
drillprop: Why not margin? The gap is different on smaller sizes - ${sizes(16})

packages/atlas/src/views/global/YppLandingView/YppLandingView.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { BackImage, FrontImage, HeroWrapper, Wrapper } from './YppLandingView.styles'
+
+export const YppLandingView: FC = () => {
+  const mdMatch = useMediaMatch('md')
+  const smMatch = useMediaMatch('sm')
+  const frontImagePosition = useScrollPosition(smMatch ? 0.2 : 0.1)
+
+  return (
+    <Wrapper>
```
drillprop: Why do we need another wrapper here? 
We could use just style LimitedWidthContainer

packages/atlas/src/views/global/YppLandingView/YppLandingView.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <LimitedWidthContainer>
+        <LayoutGrid>
+          <GridItem colSpan={{ base: 12, sm: 8, lg: 6 }} colStart={{ sm: 3, lg: 4 }}>
```
drillprop: a would pass here `as="header"`, you would need to make a small change to GridItem though. 
Line 45
```
const filteredProps = ['colStart', 'colSpan', 'rowStart', 'rowSpan', 'as']

File: packages/atlas/src/views/global/YppLandingView/YppLandingView.tsx [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+            <Button size="large" icon={<SvgActionChevronR />} iconPlacement="right">
+              Sign up now
+            </Button>
+            <Text as="p" variant="t100" color="colorText" margin={{ top: 2 }}>
+              It takes 3 minutes and is 100% free.
+            </Text>
+          </GridItem>
+        </LayoutGrid>
+      </LimitedWidthContainer>
+      <HeroWrapper>
```
drillprop: Let's rename it to HeroImageWrapper? It contains only images, so I think this will be more suitable name.

packages/atlas/src/views/global/YppLandingView/YppLandingView.styles.ts
```diff
@@ -0,0 +1,50 @@
...
+  ${media.lg} {
+    max-width: 1152px;
+  }
+`
+
+export const BackImage = styled.img`
+  ${commonHeroImagesStyles};
+
+  transform: translateX(-50%);
+  width: 84%;
```
drillprop: I don't like how we position all of these images. How did you measure that? This 84%? I'm a little concerned, that it could be hard to manage in the future. This could be written much simpler. 

I would just, use the approach presented by Adam in Figma. Just export images with transparent parts, set their widths to 100%, and paste them.  We wouldn't need to set max-width separately for both of these images. WDYT? What are the benefits of your approach?

drillprop (changes requested):

packages/atlas/src/views/global/YppLandingView/YppLandingView.styles.ts
rafalpawlow: No, no you are 100% right. I just forgot to export it after our discussion

kdembler:
Assets on desktop need to be of much higher quality, currently both joystream and YT screenshots are very blurred

packages/atlas/src/views/global/YppLandingView/YppLandingView.tsx
```diff
@@ -3,52 +3,51 @@ import { FC } from 'react'
...
+          <Button size="large" icon={<SvgActionChevronR />} iconPlacement="right">
+            Sign up now
+          </Button>
+          <Text as="p" variant="t100" color="colorText" margin={{ top: 2 }}>
+            It takes 3 minutes and is 100% free.
+          </Text>
+        </GridItem>
+      </LayoutGrid>
+      <HeroImageWrapper>
+        <BackImage srcSet={`${yt576} 576w, ${yt864} 864w, ${yt1152} 1152w, ${yt2304} 2304w`} alt="Hero back" />
```
kdembler, line 48: Maybe we could use `<picture>` tag with multiple `<source>` inside to improve readability? WDYT?

rafalpawlow: Correct me if I'm wrong but I think `srcSet` is more accurate here + we would have manually specify `media` attribute. According to mdn:

> Common use cases for <picture>:
> 
> * Art direction. Cropping or modifying images for different media conditions (for example, loading a simpler version of an image which has too many details, on smaller displays).
> * Offering alternative image formats, for cases where certain formats are not supported.

drillprop (approved):
LGTM

packages/atlas/src/views/global/YppLandingView/YppLandingView.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { SvgActionChevronR } from '@/components/_icons'
+import { useMediaMatch } from '@/hooks/useMediaMatch'
+import useScrollPosition from '@/hooks/useScrollPosition'
+
+import { BackImage, FrontImage, HeroImageWrapper, StyledLimitedWidthContainer } from './YppLandingView.styles'
+
+export const YppLandingView: FC = () => {
+  const mdMatch = useMediaMatch('md')
+  const smMatch = useMediaMatch('sm')
+  const frontImagePosition = useScrollPosition(smMatch ? 0.2 : 0.1)
```
kdembler: This is very inefficient way to do parallax - you are listening to global scroll event and changing state based on that. As a result, when scrolling, this view will be rendered hundreds of times because of all the state changes. To alleviate you would need following changes:
1) Isolating re-renders of parallax images
2) Throttling scroll event
3) Probably a bunch of other things. Take a look at the performance section of the lib you mentioned the other time: https://parallax-controller.damnthat.tv/docs/performance There's a bunch of things they do to optimize.

I think our own JS-based solution will not work well, so let's do one of the following:
a) we do pure CSS based parallax, without using JS to track positions (not sure if fully possible)
b) we use an optimized library that does the heavy lifting for us

kdembler (changes requested):

packages/atlas/src/views/global/YppLandingView/YppLandingView.tsx
```diff
@@ -13,41 +14,46 @@ import { Text } from '@/components/Text'
...
-      <HeroImageWrapper>
-        <BackImage srcSet={`${yt576} 576w, ${yt864} 864w, ${yt1152} 1152w, ${yt2304} 2304w`} alt="Hero back" />
-        <FrontImage
-          srcSet={`${hero576} 576w, ${hero864} 864w, ${hero1152} 1152w, ${hero2304} 2304w`}
-          alt="Hero front"
-          parallaxPosition={frontImagePosition}
-        />
-      </HeroImageWrapper>
-    </StyledLimitedWidthContainer>
+    <ParallaxProvider>
```
kdembler: You have already added `ParallaxProvider` to `App.tsx`, do we need it here as well?

packages/atlas/src/views/global/YppLandingView/YppLandingView.styles.ts
```diff
@@ -6,6 +6,7 @@ import { media, sizes } from '@/styles'
 export const StyledLimitedWidthContainer = styled(LimitedWidthContainer)`
   margin-top: ${sizes(16)};
   text-align: center;
+  height: 3000px;
```
kdembler: Why are we setting this?

kdembler (changes requested):

kdembler (approved):

drillprop (approved):

PR #3227, opened 2022-09-08
upgrade deps, use Vite v3

kdembler:


rafalpawlow (changes requested):
I see this error in Webstorm
https://stackoverflow.com/questions/73509984/eslint-typeerror-this-liboptions-parse-is-not-a-function

Zrzut ekranu 2022-09-8 o 13 50 05

kdembler:
It's a Webstorm issue, but okay, pinned eslint to 8.22 for now

drillprop (changes requested):
The error `document is not defined` is back. With this error, we won't be able to send transactions at all on production/deployment.
image

I don't remember how I fixed this before, I think I did some patching, but not super sure.

drillprop:
https://github.com/Joystream/atlas/pull/3065 I fixed this here

kdembler:
Fixed

rafalpawlow (approved):

drillprop (approved):

PR #3235, opened 2022-09-09
Allow customizing app name

### LINKED ISSUE #3220: Allow customizing app name
Currently all the product names ("Joystream" on master/dev and "Atlas" on carthage) are hardcoded into specific components and messages. Let's introduce a variable in the app config letting an operator to easily update the product name to something else

drillprop:
Fix #3220 
FYI - I added a plugin in vite which should read variables in HTML, it was pretty easy to replace

I wasn't sure if I should update atlas-meta-server, but I did this anyway. For some reason, I couldn't read APP_NAME from .env in the  `packages/atlas-meta-server/src/tags.ts` - I was getting `process` is undefined(I guess this is because this code is running on the browser(?)).


packages/atlas-meta-server/src/tags.ts
```diff
@@ -3,7 +3,8 @@ import { formatISODuration } from 'date-fns'
 import { BasicChannelFieldsFragment, BasicVideoFieldsFragment } from './api/__generated__/sdk'
 import { joinUrlFragments } from './utils'
 
-const BASE_ATLAS_URL = 'https://play.joystream.org'
+const BASE_APP_URL = 'https://play.joystream.org'
+const APP_NAME = 'Atlas'
```
kdembler: Not sure what the issue you've seen here was but this code doesn't run in the browser, it's all a Node backend. Could you try again to use the env variable here? Feel free to ping me on Slack if it's problematic, I can help

packages/atlas-meta-server/src/tags.ts
```diff
@@ -19,12 +20,12 @@ export type MetaTags = Record<string, string | number>
 
 const commonMetaTags = {
   'twitter:site': '@JoystreamDAO',
```
kdembler: If we're going through this exercise let's make this configurable via env as well

packages/atlas-meta-server/src/tags.ts
```diff
@@ -3,7 +3,8 @@ import { formatISODuration } from 'date-fns'
 import { BasicChannelFieldsFragment, BasicVideoFieldsFragment } from './api/__generated__/sdk'
 import { joinUrlFragments } from './utils'
 
-const BASE_ATLAS_URL = 'https://play.joystream.org'
+const BASE_APP_URL = 'https://play.joystream.org'
```
kdembler: If we're going through this exercise let's make this configurable via env as well

kdembler (changes requested):
Other than requested changes, looks good

packages/atlas-meta-server/src/config.ts
```diff
@@ -1,21 +1,13 @@
...
 
-const APP_NAME = process.env.APP_NAME as string
-if (!APP_NAME) {
-  // eslint-disable-next-line no-console
-  console.error('Missing required APP_NAME env variable')
-  process.exit(1)
-}
+const GRAPHQL_URL = getEnvVariable('GRAPHQL_URL', true)
+const APP_NAME = getEnvVariable('APP_NAME', true)
+const TWITTER_ID = getEnvVariable('TWIITER_ID')
```
kdembler: Typo in the string - should be `TWITTER_ID`

drillprop:
```suggestion
const TWITTER_ID = getEnvVariable('TWITTER_ID')
```

packages/atlas-meta-server/.env
```diff
@@ -1,3 +1,5 @@
 # This file is committed. Do not store secrets here
 
 GRAPHQL_URL=https://orion.joystream.org/graphql
+APP_NAME=Atlas
+TWITTER_ID=@JoystreamDAO
```
drillprop:
```suggestion
TWITTER_ID=@JoystreamDAO

```

packages/atlas/src/config/env.ts
```diff
@@ -9,6 +9,7 @@ export const AVATAR_SERVICE_URL = readEnv('ASSETS_SERVICE_URL', true, true)
 
 export const SENTRY_DSN = readEnv('SENTRY_DSN', false)
 
+export const TWIITER_ID = readEnv('TWITTER_ID', false, true)
```
kdembler: Typo in the variable name, should be `TWITTER_ID`

packages/atlas-meta-server/src/tags.ts
kdembler: @drillprop This wasn't resolved

kdembler (approved):

PR #3253, opened 2022-09-14
Fix studio redirects

### LINKED ISSUE #3246: Fix Studio redirects
When you reload while being on Studio URL, the app always redirects you to `studio/videos` and does some weird redirect loop. We should fix that

rafalpawlow:
Fixes #3246

I noticed that this line https://github.com/Joystream/atlas/pull/3253/files#diff-3f09a3eb15223392d4e71639fbf35349926f439f12cf86a2f9b4e776370da3e0R52 was causing a lot of unnecessary rerenders of `StudioLayout` and it seems it's not needed there. I don't see any issues after removing it, but make sure I'm not wrong.

### DIFF
--- a/packages/atlas/src/providers/user/user.provider.tsx
+++ b/packages/atlas/src/providers/user/user.provider.tsx
@@ -25,6 +25,7 @@ export const UserProvider: FC<PropsWithChildren> = ({ children }) => {
   const {
     memberships: currentMemberships,
     previousData,
+    loading: membershipsLoading,
     refetch,
     error,
   } = useMemberships(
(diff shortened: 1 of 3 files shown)


drillprop (approved):
FYI I did some testing, and it works for me. You have my approval. Since we had problems with these redirects in the past could you @kdembler also test this carefully?

kdembler (approved):
Really curious why I wrote this `let accounts = walletAccounts` but don't remember 😅. Tested switching members and switching wallets, everything seems to be working nicely. Good job with the investigation!

PR #3267, opened 2022-09-16
Transfer/Withdraw dialogs fixes

### LINKED ISSUE #3206: Transfer/Withdraw dialogs PR review
2. **As is:** Hovering over the Avatar component in the Destination account input field doesn't trigger a tooltip with a member handle.
   **To be:** It should, accordingly to the designs [(link)](https://www.figma.com/file/uHzbhNX9dP4QtZbIapvZI6/Transfer-dialog?node-id=1013%3A5566).
3. ~**As is:** (*) Having clicked on the "Max" button next to the "Amount to transfer" input field, the inserted value comes with just one decimal number, and a comma is used as a decimal separator.
**To be:** To be aligned with our global number formatting rules, a dot should be used as the decimal separator, and — ideally — there are two decimal numbers displayed, not just one.~
4. ~**As is:** (*) Having clicked "Send," the Transaction dialog appears on top of the Transfer dialog.
**To be:** The Transaction dialog should be displayed in place (instead) of the Transfer dialog, not above it.~
5. ~(*) Having signed the transaction in the Polkadot extension, the Transaction dialog is stuck at the "Propagating changes" step. It doesn't close, and I don't get any success snackbar notification, even though the tokens have indeed been transferred as I can see it in another browser tab.~
6. **As is:** (*) Size of the JOY token icon in the "Amount to transfer" input field is `24px`. 
**To be:** Any icon used inside of the `Input` component must always be used in the `16px` size. [(link)](https://www.figma.com/file/uHzbhNX9dP4QtZbIapvZI6/Transfer-dialog?node-id=1013%3A5720).
7. **As is:** Validation error message displayed upon providing an invalid wallet address format says "Invalid destination account format.".
**To be:** Accordingly to our form validation copy guidelines and the designs for this very dialog, the copy should say "Enter a valid Polkadot wallet address." (or "Joystream").

---
(*) — This comment applies to both Transfer and Withdraw dialogs.

rafalpawlow:
Closes #3206

### DIFF
--- a/packages/atlas/src/components/_inputs/TokenInput/TokenInput.tsx
+++ b/packages/atlas/src/components/_inputs/TokenInput/TokenInput.tsx
@@ -37,7 +37,7 @@ const _TokenInput: ForwardRefRenderFunction<HTMLInputElement, TokenInputProps> =
       {...rest}
       ref={ref}
       type="number"
-      nodeStart={<JoyTokenIcon variant="gray" size={24} />}
+      nodeStart={<JoyTokenIcon variant="gray" size={16} />}
       nodeEnd={
         !!valueInUSD && (
           <NumberFormat as="span" variant="t300" format="dollar" color="colorTextMuted" value={valueInUSD} />
--- a/packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
+++ b/packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
@@ -94,6 +94,7 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
         return
       }
       handleTransaction({
+        disableQNSync: true,
         snackbarSuccessMessage: {
           title: `${formatNumber(data.amount)} ${JOY_CURRENCY_TICKER} ${
             convertedAmount === null ? '' : `$(${formatNumber(convertedAmount || 0)})`
@@ -206,7 +207,7 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
                 },
                 wrongAddress: (value) => {
                   if (value && value.length < ADDRESS_LENGTH) {
-                    return 'Invalid destination account format.'
+                    return 'Enter a valid Polkadot wallet address.'
                   }
                   return true
                 },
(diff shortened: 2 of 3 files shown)


kdembler (changes requested):
I don't think this has been resolved from the original issue:

> As is: (*) Having clicked on the "Max" button next to the "Amount to transfer" input field, the inserted value comes with just one decimal number, and a comma is used as a decimal separator.
To be: To be aligned with our global number formatting rules, a dot should be used as the decimal separator, and — ideally — there are two decimal numbers displayed, not just one.

> As is: (*) Size of the JOY token icon in the "Amount to transfer" input field is 24px.
To be: Any icon used inside of the Input component must always be used in the 16px size. [(link)](https://www.figma.com/file/uHzbhNX9dP4QtZbIapvZI6/Transfer-dialog?node-id=1013%3A5720).

rafalpawlow:
As we discussed with @kdembler we will not able to accomplish this because it's default Chrome behaviour

>  As is: (*) Having clicked on the "Max" button next to the "Amount to transfer" input field, the inserted value comes with just one decimal number, and a comma is used as a decimal separator.
> To be: To be aligned with our global number formatting rules, a dot should be used as the decimal separator, and — ideally — there are two decimal numbers displayed, not just one.

packages/atlas/src/components/_inputs/TokenInput/TokenInput.tsx
```diff
@@ -37,7 +37,7 @@ const _TokenInput: ForwardRefRenderFunction<HTMLInputElement, TokenInputProps> =
       {...rest}
       ref={ref}
       type="number"
-      nodeStart={<JoyTokenIcon variant="gray" size={24} />}
+      nodeStart={<JoyTokenIcon variant="gray" size={16} />}
```
kdembler, line 40: This isn't desired in every place that uses `TokenInput`. For example, in NFT purchase view, the icon has 24 px

rafalpawlow: It seems that it isn't correct anymore

`(Breaking change!) Changed the type of icons used in the component from Controls (24px) to Action (16px),`
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=7058%3A174881

kdembler: You are right, thanks for pointing out

kdembler (approved):

PR #3288, opened 2022-09-29
Comprehensive category filtering

### LINKED ISSUE #3004: Comprehensive category filtering
**Scope:**
- Add config for whitelisted/blacklisted categories (whitelist takes precedence)
- Add filtering based on that config (see #3037)
- Filter categories displayed in Discover view

# Background
We will soon start having different instances of Atlas, with distinct branding. It would be good if there was some way we could allow these instances to brand them selves a little bit more strongly out of the gate in terms of the content they surfaced as well. This is the simplest way they can start to differentiate in a way which does rely on building any new product features, but instead at the content level. A second reason this came to mid was simply by observing how hopelessly bad our current category selection is. I have already inquired about addressing this through this issue https://github.com/Joystream/atlas/issues/1659.

# Proposal
Make it possible to configure each Atlas instance to only surface content associated with videos from a predefined set of _video categories_. By screening, I here mean hiding everywhere:

* Player: More from x/next video, **also block playback of video not with suitable category**.
* featured content
* search results
* channel view
* NFT marketplace
* discovery: Here also allow custom preview assets for each category, as suggested by https://github.com/Joystream/atlas/issues/1659. Arguably, the discovery section only needs to really show a subset of the full set of categories
* New/Channels ...
* Notifications
* etc.

# Question
* Are there any parts of the product such filtering is difficult or not feasible because of missing Hydra filtering on video categories
* Are there any unintended edge cases or weird problems by trying to segment out only specific kinds of content in Atlas.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202625706080903) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #3004


packages/atlas/src/config/categories.tsx
```diff
@@ -1,113 +1,140 @@
...
-  '14': {
-    id: '14',
-    icon: <SvgCategoriesScienceAndTechnology />,
+export type VideoCategoryData = {
+  id: string
+  icon: string
+  color: string
+  coverImg: string
+  name: string
+  videoCategories: VideoId[]
```
kdembler: This is supposed to be list of category IDs, not video IDs

packages/atlas/src/config/categories.tsx
```diff
@@ -1,113 +1,140 @@
...
+import { VideoId } from '@/joystream-lib/types'
 
-export type VideoCategoryData = { id: string; icon: ReactNode; color: string; coverImg: string }
-export const videoCategories: Record<string, VideoCategoryData> = {
-  '14': {
-    id: '14',
-    icon: <SvgCategoriesScienceAndTechnology />,
+export type VideoCategoryData = {
+  id: string
+  icon: string
```
kdembler: Let's make this `iconUrl` 🙏

packages/atlas/src/config/categories.tsx
```diff
@@ -1,113 +1,140 @@
...
+  color: string
+  coverImg: string
```
kdembler: `coverImgUrl` 🙏

packages/atlas/src/config/categories.tsx
```diff
@@ -1,113 +1,140 @@
...
+    name: 'Fifteenth category',
+    icon: 'https://dl.dropboxusercontent.com/s/amjwibevfqav8n4/categories-news-and-politics.svg?dl=0',
     color: '#48F0B3',
     coverImg: 'https://eu-central-1.linodeobjects.com/atlas-assets/category-images/news-and-politics.webp',
+    videoCategories: ['63-2'],
   },
-} as const
+]
+
+export const findDisplayCategory = (id?: string | null) => displayCategories.find((category) => category.id === id)
```
kdembler: I think we could just create a lookup once and then avoid the `.find()`?

```ts
const displayCategoriesLookup = createLookup(displayCategories)
const category = displayCategoriesLookup[id]
```

packages/atlas/src/config/categories.tsx
```diff
@@ -1,113 +1,140 @@
...
-  SvgCategoriesTravelAndEvents,
-} from '@/components/_icons'
+import { VideoId } from '@/joystream-lib/types'
 
-export type VideoCategoryData = { id: string; icon: ReactNode; color: string; coverImg: string }
-export const videoCategories: Record<string, VideoCategoryData> = {
-  '14': {
-    id: '14',
-    icon: <SvgCategoriesScienceAndTechnology />,
+export type VideoCategoryData = {
```
kdembler: Maybe just `DisplayCategory`? We should try to differentiate between different types of categories to make it clear which ones we're using where

packages/atlas/src/hooks/useVideosInCategories.ts
```diff
@@ -0,0 +1,14 @@
+import { useCategories } from '@/api/hooks/categories'
+import { VideoCategoryData } from '@/config/categories'
+
+export const useVideosInCategories = () => {
+  const { categories } = useCategories()
```
kdembler: Can we rename this hook (and maybe the associated file) to `useQnCategories` so it's clear this isn't about display categories?

packages/atlas/src/hooks/useVideosInCategories.ts
```diff
@@ -0,0 +1,14 @@
+import { useCategories } from '@/api/hooks/categories'
+import { VideoCategoryData } from '@/config/categories'
+
+export const useVideosInCategories = () => {
```
kdembler: Let's rename this hook, it's not clear what it does. `useVideosCountForCategory`?

And maybe it accepts category ID right away and returns memoized count instead of a function?

packages/atlas/src/components/FiltersBar/FiltersBar.tsx
```diff
@@ -96,7 +96,7 @@ export const FiltersBar: FC<ReturnType<typeof useFiltersBar> & FiltersBarProps>
   const datePopoverRef = useRef<PopoverImperativeHandle>(null)
   const lengthPopoverRef = useRef<PopoverImperativeHandle>(null)
   const othersPopoverRef = useRef<PopoverImperativeHandle>(null)
-  const { categories } = useCategories()
+  const { categories } = useQnCategories()
```
kdembler: Pending response from Adam on Slack whether the filtering shouldn't use display categories instead

packages/atlas/src/config/categories.tsx
```diff
@@ -1,113 +1,146 @@
...
+    videoCategories: ['63-2'],
   },
-} as const
+]
+
+export const findDisplayCategory = (id?: string | null) => {
+  const displayCategoriesLookup = createLookup(displayCategories)
+  if (!id) return null
+  return displayCategoriesLookup[id]
+}
```
kdembler: Why do you create lookup inside a function? This isn't very performant - the list of categories is constant so we can create the lookup only once, as top-level const in this file and just import it directly when needed.

packages/atlas/src/providers/videoWorkspace/hooks.ts
```diff
@@ -45,7 +45,7 @@ export const useVideoWorkspaceData = () => {
       },
     }
   )
-  const { categories } = useCategories()
+  const { categories } = useQnCategories()
```
kdembler: We discussed this with Bedeho yesterday - what kind of categories we should use during publishing. We decided that the select in video workspace should show display categories. Also, in category config, for each display category, aside list of `videoCategories`, we define `defaultVideoCategory` which will be used when user tries to publish video under this display category.

packages/atlas/src/providers/videoWorkspace/provider.tsx
```diff
@@ -24,7 +24,7 @@ export const VideoWorkspaceProvider: FC<PropsWithChildren> = ({ children }) => {
   }, [])
 
   // trigger categories fetch so that they are available once the video workspace is opened
-  useCategories(undefined, {
+  useQnCategories(undefined, {
     context: { delay: 1500 },
     onError: (error) => SentryLogger.error('Failed to fetch video categories', 'VideoWorkspaceProvider', error),
   })
```
kdembler: This probably won't be needed anymore once we use display categories in video workspace

packages/atlas/src/views/viewer/CategoryView/CategoryView.tsx
```diff
@@ -28,25 +31,22 @@ export const CategoryView = () => {
   const mdBreakpointMatch = useMediaMatch('md')
   const { id = '' } = useParams()
 
-  const { categories, totalVideosCount, loading } = useCategories()
+  const { totalVideosCount, loading } = useQnCategories()
```
kdembler: This `totalVideosCount` won't be the correct value for us - it will be total count of all videos, but what we should be interested in is the total count of videos that could be displayed in the app - so taking into account only whitelisted categories. I think we should remove `totalVideosCount` from `useQnCategories` (My guess is that once you implement changes to video workspace, `useQnCategories` won't be used anywhere in the app, outside some hooks). Instead, you could add `totalVideosCount` to `useVideoCategoriesWithCounter`

packages/atlas/src/hooks/useVideoCategoriesWithCounter.ts
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react'
+
+import { useQnCategories } from '@/api/hooks/categories'
+import { displayCategories } from '@/config/categories'
+
+export const useVideoCategoriesWithCounter = () => {
```
kdembler: I think we could rename this to `useVideoDisplayCategoriesWithCounter` so it's clear those are display categories?

packages/atlas/src/views/viewer/CategoryView/CategoryView.tsx
```diff
@@ -28,25 +31,22 @@ export const CategoryView = () => {
...
-    [id, mappedVideoCategories]
+    [id, videoCategories]
   )
-  const currentCategory = mappedVideoCategories?.find((category) => category.id === id)
+  const currentCategory = findDisplayCategory(id)
 
   const headTags = useHeadTags(currentCategory?.name)
 
-  const { categoriesFeaturedVideos } = useCategoriesFeaturedVideos(id)
+  const { categoriesFeaturedVideos } = useCategoriesFeaturedVideos(sample(currentCategory?.videoCategories) || '')
```
kdembler: `useCategoriesFeaturedVideos` should just use display category IDs going forward

packages/atlas/src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -28,19 +29,29 @@ export const DiscoverView: FC = () => {
...
 
   const featuredVideoCategoryCardsData = useMemo(() => {
+    if (!categories) {
+      return [null, null, null]
+    }
+    const usedCategories: string[] = []
     const _featuredVideoCategoryCardsData =
       categories
-        ?.map((category) => {
+        .map(function (category) {
```
kdembler: Why not an arrow function?

packages/atlas/src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -18,8 +19,8 @@ import {
 } from './DiscoverView.styles'
 
 export const DiscoverView: FC = () => {
-  const { categories, totalVideosCount, loading } = useCategories()
-  const mappedVideoCategories = useCategoriesMatch()
+  const { categories, totalVideosCount, loading } = useQnCategories()
+  const videoCategories = useVideoCategoriesWithCounter()
```
kdembler: This should be called `displayCategories`. In this component we have both `categories` and `videoCategories`, that's super misleading

kdembler: Btw, why do we even need QN categories here? I would expect that display categories will be enough?

kdembler (changes requested):

packages/atlas/src/views/viewer/CategoryView/CategoryView.tsx
rafalpawlow: At first, I wasn't sure what do you mean but are you sure about this? `useCategoriesFeaturedVideos` is an api hook

packages/atlas/src/config/categories.ts
```diff
@@ -0,0 +1,19 @@
...
+export type DisplayCategory = RawConfig['content']['categories'][0]
+
+export const displayCategories = atlasConfig.content.categories
+
+export const displayCategoriesLookup = createLookup(displayCategories)
+
+export const findDisplayCategory = (id?: string | null) => {
+  if (!id) return null
+  return displayCategoriesLookup[id]
+}
```
kdembler: Why does this function still exist? If we export `displayCategoriesLookup`, then anyone can just do `displayCategoriesLookup[id]` themselves. What's the purpose of this function?

packages/atlas/src/hooks/useVideoDisplayCategoriesWithCounter.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (!categories) {
+      return
+    }
+    const filteredCategories = categories?.filter((category) => allUniqueVideoCategories.includes(category.id))
+    return filteredCategories.reduce((acc, cur) => acc + cur.activeVideosCounter, 0)
+  }, [categories])
+
+  const displayCategoriesWithCounter = useMemo(
+    () =>
+      displayCategories.map((category) => ({
```
kdembler: Let's rename the parameter to `displayCategory`

packages/atlas/src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -29,9 +27,9 @@ export const DiscoverView: FC = () => {
 
   const featuredVideoCategoryCardsData = useMemo(() => {
     const _featuredVideoCategoryCardsData =
-      categories
-        ?.map((category) => {
-          const video = categoriesFeaturedVideos?.[category.id]?.categoryFeaturedVideos.find(
+      displayCategoriesWithCounter
+        .map((category) => {
+          const video = categoriesFeaturedVideos?.[category.defaultVideoCategory]?.categoryFeaturedVideos.find(
```
kdembler: This isn't correct - like I mentioned previously, video featuring will not use QN categories but display categories IDs. So to get featured videos for a display category you should just do `categoriesFeaturedVideos[displayCategory.id]`

packages/atlas/src/views/viewer/DiscoverView/DiscoverView.tsx
```diff
@@ -29,9 +27,9 @@ export const DiscoverView: FC = () => {
 
   const featuredVideoCategoryCardsData = useMemo(() => {
     const _featuredVideoCategoryCardsData =
-      categories
-        ?.map((category) => {
-          const video = categoriesFeaturedVideos?.[category.id]?.categoryFeaturedVideos.find(
+      displayCategoriesWithCounter
+        .map((category) => {
```
kdembler: Let's call this parameter `displayCategory`

packages/atlas/src/views/viewer/VideoView/MoreVideos.tsx
```diff
@@ -31,14 +32,22 @@ export const MoreVideos: FC<MoreVideosProps> = ({
   videoId,
   type,
 }) => {
-  const where = type === 'channel' ? { channel: { id_eq: channelId } } : { category: { id_eq: categoryId } }
+  const videoCategories = displayCategories?.find((category) => category.id === categoryId)?.videoCategories
```
kdembler: We have `displayCategoriesLookup`, let's use that instead of `.find()`

kdembler (changes requested):
Sorry it's dragging out, but I've mentioned this change is gonna be complex 😅 Good news, it's all shaping up and even with the couple of comments I'm adding now, in general the change seems to work mostly properly, so we're close.

1. Something is wrong with the category's video count pie chat. There are only 2 categories here and both have 3 videos so both should be filled in half:
CleanShot 2022-10-10 at 08 17 36@2x
2. Also:
CleanShot 2022-10-10 at 08 29 00@2x

packages/atlas/src/config/categories.ts
rafalpawlow: I left it to not have to do a ternary condition everywhere if there is no id. `const category = id ? displayCategoriesLookup[id] : null`. But ok, I'll remove it

kdembler (approved):

Issue #3105, opened 2022-08-10
WIP: Atlas Operator Guide


bedeho:
WIP.

* rebrand
* QN + RPC nodes?
* categories
* faucet
* avatar service
* captcha configuration
* content reporting
* social previews
* URL of Pioneer instance to use, e.g. in `learn more` on membership page.
* setting DMCA policy, terms, privacy
* disable stuff: YPP?
* analytics hooks?
* price feed
* deeper changes
* Orion: which instance? use shared one run by content lead for now as recommended?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202778996006812) by [Unito](https://www.unito.io)

PR #3293, opened 2022-10-03
move all the app config into single YAML file

kdembler:
Part of #3105

Now the only config you need to provide is this single YAML file and URLs/secrets passed via env variables.

Very open to feedback and questions on this one, want to make this work as best as possible


packages/atlas/atlas.config.yml
```diff
@@ -0,0 +1,160 @@
...
+  blockedVideoIds: [] # Array of video IDs that should be blocked from being displayed in the app
+  blockedChannelIds: [] # Array of channel IDs that should be blocked from being displayed in the app
+
+  categories: # List of display categories to be used in the app. Those are local to the application and are related to metaprotocol categories via 'videoCategories' array.
+    - id: '1' # Local category ID
+      name: 'Music' # Display name of the category
+      color: '#D92E61' # Color used in the UI for this category
+      iconUrl: '...' # URL for category icon SVG
+      coverImgUrl: '...' # URL for category cover image
+      videoCategories: ['1', '2'] # List of metaprotocol category IDs that should be included in this display category
```
kdembler, line 70: This will get updated after work in #3288 is done

drillprop (approved):
Looks good to me 👍🏻

rafalpawlow (approved):

PR #3300, opened 2022-10-04
Fix issues with updating channel

### LINKED ISSUE #3224: Cover and description didn't change after the channel was updated(carthage local)
The problem was noticed when running Atlas locally from the carthage branch

Cover and description didn't change after the channel was updated

Successful transaction: Withdraw, Withdraw, Deposit, Withdraw, VoucherChanged, VoucherChanged, Transfer, DataObjectsUpdated, ChannelUpdated, ExtrinsicSuccess
on network: wss://atlas-next.joystream.org/ws-rpc
in block: 0x0bf80efcbac254bf85623ca6a509cd160110409562741f7c62bb68dec8e6af1a
extrinsic hash: 0x3363e891dc5259e77388f5eafd942cae97e7348d388219b989003a57c4d4d6e3
more details at: https://polkadot.js.org/apps/?rpc=wss://atlas-next.joystream.org/ws-rpc#/explorer/query/0x0bf80efcbac254bf85623ca6a509cd160110409562741f7c62bb68dec8e6af1a

Updated after page reloading

drillprop:
Fix #3224 
While working on this task I found that updating a channel is pretty buggy right now. When testing, make sure that you manage: 
- create a new channel and set an avatar and cover
- update an existing channel without an avatar and cover
- update an existing channel with an avatar and cover
- update only one asset on an existing channel
- switch between channels while the edit channel form is open(check if the content in the form is updated correctly)
- update only one asset on an existing channel - either avatar or cover, and once this asset is uploaded and processed, try to update the channel again.

Just FYI, deleting assets won't be covered in this PR. It's still not working fully and we have a separate issue for this.

### DIFF
--- a/packages/atlas/src/providers/assets/assets.hooks.ts
+++ b/packages/atlas/src/providers/assets/assets.hooks.ts
@@ -29,7 +29,9 @@ export const useAsset = (dataObject?: StorageDataObjectFieldsFragment | null) =>
 }
 
 export const useRawAsset = (contentId: string | null) => {
-  return useAssetStore((state) => (contentId ? state.assets[contentId] : null))
+  return useAssetStore((state) => {
+    return contentId ? state.assets[contentId] : null
+  })
 }
 
 export const useRawAssetResolver = () => {
--- a/packages/atlas/src/providers/assets/assets.hooks.ts
+++ b/packages/atlas/src/providers/assets/assets.hooks.ts
@@ -29,9 +29,7 @@ export const useAsset = (dataObject?: StorageDataObjectFieldsFragment | null) =>
 }
 
 export const useRawAsset = (contentId: string | null) => {
-  return useAssetStore((state) => {
-    return contentId ? state.assets[contentId] : null
-  })
+  return useAssetStore((state) => (contentId ? state.assets[contentId] : null))
 }
 
 export const useRawAssetResolver = () => {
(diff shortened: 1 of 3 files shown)


packages/atlas/src/providers/assets/assets.hooks.ts
```diff
@@ -29,7 +29,9 @@ export const useAsset = (dataObject?: StorageDataObjectFieldsFragment | null) =>
 }
 
 export const useRawAsset = (contentId: string | null) => {
-  return useAssetStore((state) => (contentId ? state.assets[contentId] : null))
+  return useAssetStore((state) => {
+    return contentId ? state.assets[contentId] : null
```
rafalpawlow: Why did you make this change?

rafalpawlow (changes requested):
1. All scenarios work pretty solid however, I have one doubt. After I upload an avatar I cannot update another one, first I need to delete the previously updated one. Not sure if this is expected behavior. Check the video below: 

https://user-images.githubusercontent.com/13706246/195081777-db276fe1-4e3d-46cb-b6d2-9a1fbfaf1100.mov

2. Do you think we could try to get rid of this as part of this PR? It appears after the channel is created and looks worrying
Zrzut ekranu 2022-10-11 o 13 34 22

rafalpawlow:
After a discussion with @drillprop we agreed that 1st point is expected behavior.

drillprop:
> 2. Do you think we could try to get rid of this as part of this PR? It appears after the channel is created and looks worrying
>    Zrzut ekranu 2022-10-11 o 13 34 22

Fixed, I don't think we should worry about this warning.  The `NewAccount` event
is marked as a system event for some reason, I updated the if statement in `parseExtrinsicEvents`

rafalpawlow (approved):

PR #3301, opened 2022-10-04
Better error handling on filtered out videos

### LINKED ISSUE #3213: Better error handling on filtered out videos
## Context

Atlas instances will filter out videos that are not within the category space of a given instance, and this means that even if you provie a valid video ID to the player, you should not get playback. We should have a better explainer of why the video cannot be played than simply "missing video" or whatever the default message is when a video literally does not exist.

Additionally, content censored by DAO should be presumed to be the worst offenses, it basically stuff the DAO is considering potentially not even willing to provide any infrastructure for regardless of what jurisdiction to which end consumer is subject. It is on the cusp of getting removed. Ironically, currently the infrastructure just entirely ignores this signal, but this will get fixed: https://github.com/Joystream/joystream/issues/4326. As you will read in that issue, infra level provisions will be made to allow certain key stakeholders ultimately relevant to the adjudication process to consume it, but noone else. Atlas must follow this approach.

## Scope 
- [ ] Update the design of the unavailable video page based on the designs added [here](https://github.com/Joystream/atlas/issues/3213#issuecomment-1252319327)
- [ ] Copy should go ""This video was deleted by creator, moderated by the DAO content curation team or not included to be viewed by the application operators."

rafalpawlow:
Closes #3213 

~I blocked video `5` for tests, will remove it before merge~

**Design**
https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=6176%3A332457

### DIFF
--- a/packages/atlas/src/views/viewer/VideoView/VideoView.tsx
+++ b/packages/atlas/src/views/viewer/VideoView/VideoView.tsx
@@ -267,10 +267,10 @@ export const VideoView: FC = () => {
     <GridItem colSpan={{ xxs: 12, md: 4 }}>
       {videoIsBlocked
         ? mdMatch && (
-            <div>
+            <>
               {!cinematicView && <BlockedVideoPlaceholder />}
               <BlockedVideoGradientPlaceholder />
-            </div>
+            </>
           )
         : !!nftWidgetProps && (
             <NftWidget
--- a/packages/atlas/src/views/viewer/VideoView/VideoView.tsx
+++ b/packages/atlas/src/views/viewer/VideoView/VideoView.tsx
@@ -6,7 +6,6 @@ import { useInView } from 'react-intersection-observer'
 import { useParams } from 'react-router-dom'
 
 import { useAddVideoView, useFullVideo } from '@/api/hooks/video'
-import { EmptyFallback } from '@/components/EmptyFallback'
 import { GridItem, LayoutGrid } from '@/components/LayoutGrid'
 import { LimitedWidthContainer } from '@/components/LimitedWidthContainer'
 import { NumberFormat } from '@/components/NumberFormat'
@@ -26,7 +25,6 @@ import { filteredVideoIds } from '@/config/contentFilter'
 import { CTA_MAP } from '@/config/cta'
 import { APP_NAME, BASE_APP_URL, TWITTER_ID } from '@/config/env'
 import { LANGUAGES_LOOKUP } from '@/config/languages'
-import { absoluteRoutes } from '@/config/routes'
 import { useDisplaySignInDialog } from '@/hooks/useDisplaySignInDialog'
 import { useHeadTags } from '@/hooks/useHeadTags'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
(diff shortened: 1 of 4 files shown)


toiletgranny (changes requested):
Looking great, as always! As finishing touches, for the "Video unavailable", I have a few suggestions that I hope would not break the Video player component:
1. Remove `padding: 0px 12px` altogether,
2. Don't set `height`,
3. Don't set `overflow`,
4. Set `aspect-ratio` to `16 / 9`,
5. Changing the color of the body copy from `var(--color-core-neutral-300)` to `var(--color-text-default)` (this one was a mistake in designs, apologies).

For the secondary column:
1. I'd suggest dropping the `<div>` that's direct child of `.css-d7meg8`,
2. I'd suggest setting the `height` parameter on the `.css-o7pm0s` to `calc(100% - 160px)`.

Here's the end result:

https://user-images.githubusercontent.com/35307309/193830968-ef95e814-9875-4db7-87cc-5398169aa06c.mp4

rafalpawlow:
@toiletgranny we decided to not use `aspect-ratio` (even for video it's percentage padding) due to poor browsers coverage.

toiletgranny:
Ah, sure, of course! Hopefully, the things I suggested still work with the padding hack. If not, or if it requires a significant amount of tweaking and writing spaghetti code, then my only suggestion would be to set `512px` of `max-height` on the player component so that it doesn't grow into a square on wider viewports.

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -135,6 +139,7 @@ export const VideoView: FC = () => {
   const categoryId = video?.category?.id
   const numberOfLikes = video?.reactions.filter(({ reaction }) => reaction === 'LIKE').length
   const numberOfDislikes = video?.reactions.filter(({ reaction }) => reaction === 'UNLIKE').length
+  const videoIsBlocked = !!(id && filteredVideoIds.includes(id))
```
kdembler: This was mentioned in the original issue (https://github.com/Joystream/atlas/issues/3213), this new screen should be shown whenever the video in unavailable, for whatever reason. With your current code this new screen will be shown only if the video ID is blocked, but following cases would be ignored:
- Video is outside of app's category list
- Video was marked as censored by curators
- Video just doesn't exist
This new view should be used in place of `EmptyFallback` used previously

kdembler (changes requested):

toiletgranny (approved):

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -139,7 +136,7 @@ export const VideoView: FC = () => {
   const categoryId = video?.category?.id
   const numberOfLikes = video?.reactions.filter(({ reaction }) => reaction === 'LIKE').length
   const numberOfDislikes = video?.reactions.filter(({ reaction }) => reaction === 'UNLIKE').length
-  const videoIsBlocked = !!(id && filteredVideoIds.includes(id))
+  const videoNotAvailable = !!(id && filteredVideoIds.includes(id)) || (!loading && !video)
```
kdembler: What's the point of `!!(id && filteredVideoIds.includes(id))`? If the video is blocked, the second clause will still capture that?

kdembler (approved):

drillprop (approved):

PR #3311, opened 2022-10-05
Optional member sign up captcha

### LINKED ISSUE #3302: Make captcha during member sign up optional
Currently Captcha is always a part of the member creation. However, some faucets may not require captcha and some Atlas operator may not want to use it. We should make the Captcha optional. We can make it visible based on whether captcha key env variable is set

rafalpawlow:
Closes #3302

### DIFF
--- a/packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalMembershipStep.tsx
+++ b/packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalMembershipStep.tsx
@@ -190,26 +190,28 @@ export const SignInModalMembershipStep: FC<SignInModalMembershipStepProps> = ({
               }}
             />
           </FormField>
-          <Controller
-            control={control}
-            name="captchaToken"
-            render={({ field: { onChange }, fieldState: { error } }) => (
-              <FormField error={error?.message}>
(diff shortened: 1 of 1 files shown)


packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalMembershipStep.tsx
```diff
@@ -189,26 +189,28 @@ export const SignInModalMembershipStep: FC<SignInModalMembershipStepProps> = ({
...
-                    trigger('captchaToken')
-                  }}
-                />
-              </FormField>
-            )}
-            rules={{
-              required: { value: !!HCAPTCHA_SITE_KEY, message: "Verify that you're not a robot." },
-            }}
-          />
+          {HCAPTCHA_SITE_KEY && !!HCAPTCHA_SITE_KEY.length && (
```
drillprop: isn't just `HCAPTCHA_SITE_KEY &&` enough?

drillprop (changes requested):
Just a suggestion. As part of this PR can we also disable the captcha in development or in localhost? This message is pretty annoying: 
image

drillprop (approved):

PR #3335, opened 2022-10-11
YPP Landing Page review fixes

### LINKED ISSUE #3271: YPP LP PR review
1. **As is:**`width` and `height` attributes are not specified on `` elements.
**To be:** They should be, so that layout shifts do not occur when the page is loading (this actually happened to me) and so that the SEO score of the page is higher.
2. **As is:** `` elements do not have `alt` attributes.
**To be:** They should have. Let me know if you need help with those, happy to schedule a 15-30 minute call and do it together.
3. **As is:** ` <button>` elements don't have accessible names.
**To be:** Ideally, and again — for SEO and accessibility purposes — they have. I believe this involves changing the component. Happy to discuss details over a call.
4. **As is:** The fade gradient in Benefit card components goes down to `var(--color-core-neutral-900)` color.
**To be:** It should be `var(--color-core-neutral-800)`.
CleanShot 2022-09-21 at 15 09 00
5. **As is:** (Parallax) Different parts of product illustrations move in different directions and, sometimes, on different axes.
**To be:** All parts of all product illustrations, ideally, move with the same speed, the same direction, and the along the same axis (vertically). This referees to the below illustrations:
     - "Upload your videos and own them like you mean it",
     - "New ways to engage and get support from your audience".
6. **As is:** In the illustration for the "Monetize your videos with NFT sales and royalties" section, the NFT card component moves at the highest speed, the dropdown on top of it moves slightly slower, and the cursor above it all stays in place.
**To be:** Ideally, the cursor moves the fastest, then there should be the dropdown, and — finally — the NFT card component, being the closest to the imaginary _ground_, should move the slowest of these three. [Please, see the 1:1 benchmark here, in the "Monetise your videos with NFT sales and royalties" section.](https://careful-day-509189.framer.app/)
7. **As is:** All elements scrolled horizontally are clipped within boundaries of their parent containers.
**To be:** They shouldn't be. 
CleanShot 2022-09-21 at 15 19 03
8. I'm afraid the image compression is a bit too heavy for the kind of imagery we're using on this page. This is especially noticeable on gradient fades, like in the below example. Could we experiment with the compression intensity or type?
image
9. ✨ **Nice-to-have** ✨ If we have time allowance for this, I'd be super happy to jump on a quick call with whoever would be eager to add (as a cherry on top) simple on-scroll CSS entrance animations here and there, possibly using very simple and lightweight ready-made solutions such as the [AOS](https://michalsnik.github.io/aos/) or [Animate.css](https://animate.style/) libraries. Let me now!

rafalpawlow:
Closes #3271


toiletgranny (changes requested):
WOW, just WOW! I can't believe how much more alive the page feels now, this is so great! And Lighthouse's improved performance, SEO, and accessibility scores speak for themselves! ❤️

CleanShot 2022-10-12 at 09 05 16@2x

I can see that this little image in the background is still problematic, though. Even Lighthouse complains about its large size (<800 KB) and the lack of width and height attributes. I thought that perhaps I could help with the first one, by providing you with an exported and compressed image that's still @2x scale, under 300 KB, and has no compression issues. Please, see if the below image will work. It's a JPEG, as GitHub won't let me upload WebP, but I'm sure you can convert the image to WebP yourself (without compressing it).

illustration-video-nfts-l1

drillprop:
> illustration-video-nfts-l1

@toiletgranny I'm not sure if this will work actually. This particular image was semi-transparent(I think that's why it so heavy). 
image

If we use your image, the background will be different if I'm not mistaken

Edit: This is how it will look like:
image

packages/atlas/src/views/global/YppLandingView/YppCardsSection.styles.ts
```diff
@@ -40,6 +40,7 @@ export const CardImage = styled.img<{ absolute?: boolean; dropShadow?: boolean }
   width: 100%;
   min-width: 480px;
   max-width: 640px;
+  height: auto;
```
drillprop, line 43: Why do we need to set `height: auto`. This is the default value.

packages/atlas/src/views/global/YppLandingView/YppHero.styles.ts
```diff
@@ -15,13 +15,17 @@ export const HeroImageWrapper = styled.div`
 
 export const FrontImage = styled.img`
   width: 100%;
+  max-width: 100%;
+  height: auto;
```
drillprop, line 19: same question

packages/atlas/src/views/global/YppLandingView/YppHero.styles.ts
```diff
@@ -15,13 +15,17 @@ export const HeroImageWrapper = styled.div`
...
   display: block;
   ${imageShadow}
 `
 
 export const BackImage = styled.img`
   position: absolute;
   width: 100%;
+  max-width: 100%;
+  height: auto;
```
drillprop, line 28: same question

packages/atlas/src/views/global/YppLandingView/YppThreeStepsSection.styles.ts
```diff
@@ -65,6 +71,7 @@ export const StepCardNumber = styled.span`
 export const StepCardImg = styled.img`
   display: block;
   max-width: 100%;
+  height: auto;
```
drillprop, line 70: same question

packages/atlas/src/components/_buttons/Button/Button.tsx
```diff
@@ -64,10 +66,11 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>((props, ref) =>
       disabled={disabled}
       aria-disabled={disabled}
       variant={variant}
-      iconOnly={!!iconOnly}
+      iconOnly={iconOnly}
       textOnly={!!_textOnly}
       size={size}
       data-badge={badge}
+      aria-label={ariaLabel || (typeof children === 'string' ? children : undefined)}
```
drillprop, line 73: Correct me if I'm wrong, but I don't think we need to set up `arial-label` if we already have text inside the button. Button with a text inside is already accessible.

packages/atlas/src/views/global/YppLandingView/YppLandingView.styles.ts
```diff
@@ -18,6 +18,14 @@ export const imageShadow = css`
     drop-shadow(0 2.1945px 6.6501px rgb(0 0 0 / 0.0889)); */
 `
 
+export const Wrapper = styled.div`
+  [data-aos] {
+    &[data-aos][data-aos-easing='atlas-easing'] {
+      transition-timing-function: cubic-bezier(0.09, 0.43, 0.3, 0.99);
```
drillprop, line 24: How did you come up with this value? Is it Adam's suggestion?

packages/atlas/src/views/global/YppLandingView/YppThreeStepsSection.styles.ts
```diff
@@ -76,3 +83,14 @@ export const StepCardFade = styled.div`
   width: 100%;
   bottom: 0;
 `
+
+export const StyledBackgroundContainer = styled(BackgroundContainer)`
+  padding: 0;
+`
+
+export const StyledLimitedWidthContainer = styled(_StyledLimitedWidthContainer)`
```
drillprop: This could result in importing the wrong component by accident in this case. I would name this differently.

drillprop (changes requested):

toiletgranny:
@drillprop Ah, of course, you're right. Here's the same image with the correct background color!
illustration-video-nfts-l1

packages/atlas/src/views/global/YppLandingView/YppCardsSection.styles.ts
rafalpawlow: It's not when we set it as HTML attribute https://github.com/Joystream/atlas/pull/3335/files#diff-fb3a251bc61562a12fc7ffb4834ee80ed8ce21f63ef798ec792da761abfb3dd3R86

packages/atlas/src/components/_buttons/Button/Button.tsx
rafalpawlow: The problem is that the direct child of a button is `div` and it turns out that it needs to be a string to be accessible, otherwise we need to set `aria-label`.

packages/atlas/src/views/global/YppLandingView/YppLandingView.styles.ts
rafalpawlow: Yes, this comes from Adam. @toiletgranny maybe we should add it to design tokens?

packages/atlas/src/components/_buttons/Button/Button.tsx
toiletgranny: That's right, but also not true in our case, as we have a `<span>` element inside of the `<button>`. That's why Google Lighthouse yielded this as not accessible.

packages/atlas/src/views/global/YppLandingView/YppLandingView.styles.ts
toiletgranny: @rafalpawlow I really don't think we'll be using this easing value anywhere else, anytime soon. It's a really custom thing that works just for this LP. Would it make it any easier for you guys if this was a design token? I believe it's best if we keep them as simple as possible.

drillprop: Ok, I agree with @toiletgranny, I would opt to not add this to design tokens, not every value needs to be a token

packages/atlas/src/views/global/YppLandingView/YppThreeStepsSection.styles.ts
rafalpawlow: These styles are actually unused, I'll remove them

packages/atlas/src/components/_buttons/Button/Button.tsx
drillprop: I'm not sure why Lighthouse doesn't like it, because it's correct to have nested elements with text. And it should have an accessible name. Try to run the lighthouse here: https://acrobaticregulareyestrain.drillprop.repl.co/ 

Anyway, let's keep `aria-label` here.

drillprop (approved):

toiletgranny (approved):

PR #3336, opened 2022-10-11
Handle Substrate and Joystream addresses in transfer modal

### LINKED ISSUE #3256: Substrate vs Joystream address format UX upd
## Scope

- [ ] Update the designs of the Transfer Modal as per [this comment](https://github.com/Joystream/atlas/issues/3256#issuecomment-1268447352)
- [ ] Update the validation to accept both Substrate generic and Joystream Specific addresses.
- [ ] Convert the address to Joystream format before sending an extrinsic

Designs for the Transfer dialog have been created around the assumption that JOY tokens can be transferred from one Polkadot wallet to another; therefore, it's the Polkadot wallet address that's an identifier of a destination account. [(link)](https://www.figma.com/file/uHzbhNX9dP4QtZbIapvZI6/Transfer-dialog?node-id=1013%3A5720) In the implementation, to my surprise, the "Polkadot wallet address" has been swapped with "Joystream wallet address".

Now, "Joystream wallet address" is a concept to which I, as an Atlas user, haven't been properly introduced to yet, meaning (a) I don't know what's my Joystream wallet address or (b) where to find it.

Unless changing "Polkadot wallet address" to "Joystream wallet address" isn't trivial (no one talked to me about it) and is necessary for the Transfer dialog to work, I believe we must create an issue to think about onboarding Atlas users with the concept of Joystream wallet addresses, it's relation with Polkadot wallet addresses and how does transfering funds work in general. And for that to happen, I'd need to be onboarded with that knowledge myself, too.

@dmtrjsg @kdembler If you guys agree with the above, I'd appreciate pointing me to further reading on the subject, if it's available (I tried looking it up in our Handbook, but I haven't found anything), or arranging a meeting where I can have someone explain this to me.

rafalpawlow:
Closes #3256

### DIFF
--- a/packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
+++ b/packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
@@ -25,12 +25,14 @@ import { hapiBnToTokenNumber, tokenNumberToHapiBn } from '@/joystream-lib/utils'
 import { useMemberAvatar } from '@/providers/assets/assets.hooks'
 import { useFee, useJoystream, useTokenPrice } from '@/providers/joystream/joystream.hooks'
 import { useTransaction } from '@/providers/transactions/transactions.hooks'
+import { formatJoystreamAddress, shortenAddress } from '@/utils/address'
 import { SentryLogger } from '@/utils/logs'
 import { formatNumber } from '@/utils/number'
 
 import { FormFieldsWrapper, PriceWrapper, StyledMaxButton, VerticallyCenteredDiv } from './SendTransferDialogs.styles'
 
-const ADDRESS_LENGTH = 49
+const POLKADOT_ADDRESS_LENGTH = 48
+const JOYSTREAM_FIRST_CHAR = 'j4'
 const ADDRESS_CHARACTERS_LIMIT = 4
 
 type SendFundsDialogProps = {
@@ -88,6 +90,9 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
     }, 500)
   )
 
+  const formatAddress = (address: string) =>
+    address.toLowerCase().slice(0, 2) === JOYSTREAM_FIRST_CHAR ? address : formatJoystreamAddress(address)
+
   const handleSendFounds = async () => {
     const handler = await handleSubmit((data) => {
       if (!joystream || !data.account || !data.amount) {
@@ -98,13 +103,12 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
         snackbarSuccessMessage: {
           title: `${formatNumber(data.amount)} ${atlasConfig.joystream.tokenTicker} ${
             convertedAmount === null ? '' : `$(${formatNumber(convertedAmount || 0)})`
-          } tokens have been sent over to ${data.account.slice(0, ADDRESS_CHARACTERS_LIMIT)}...
-          ${data.account.slice(-ADDRESS_CHARACTERS_LIMIT)} wallet address`,
+          } tokens have been sent over to ${shortenAddress(data.account, ADDRESS_CHARACTERS_LIMIT)} wallet address`,
         },
         txFactory: async (updateStatus) => {
           const amount = amountBN.add(fullFee).gte(accountBalance) ? amountBN.sub(fullFee) : amountBN
           return (await joystream.extrinsics).sendFunds(
-            data.account || '',
+            formatAddress(data.account || ''),
             amount.toString(),
             proxyCallback(updateStatus)
           )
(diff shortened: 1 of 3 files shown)


packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -88,6 +90,9 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
     }, 500)
   )
 
+  const formatAddress = (address: string) =>
+    address.toLowerCase().slice(0, 2) === JOYSTREAM_FIRST_CHAR ? address : formatJoystreamAddress(address)
```
kdembler: Let's skip this condition and always use `formatJoystreamAddress`

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -25,12 +25,14 @@ import { hapiBnToTokenNumber, tokenNumberToHapiBn } from '@/joystream-lib/utils'
...
 import { useTransaction } from '@/providers/transactions/transactions.hooks'
+import { formatJoystreamAddress, shortenAddress } from '@/utils/address'
 import { SentryLogger } from '@/utils/logs'
 import { formatNumber } from '@/utils/number'
 
 import { FormFieldsWrapper, PriceWrapper, StyledMaxButton, VerticallyCenteredDiv } from './SendTransferDialogs.styles'
 
-const ADDRESS_LENGTH = 49
+const POLKADOT_ADDRESS_LENGTH = 48
+const JOYSTREAM_FIRST_CHAR = 'j4'
```
kdembler: I'm not comfortable with hardcoding this here - this prefix depends on `JOYSTREAM_SS58_PREFIX` which is exported from `@/joystream-lib/config`. It would be good to keep this config centralized.

Having said that, we will need to know those first chars because they are used in the tooltip copy. However, I think we can easily get those based on that config const. For example:

```ts
const EXAMPLE_ADDRESS = '5Dbstm8wPgrKAwHeMe8xxqxDXyFmP3jyzYdmsiiwTdCdt9iU'
const formattedExampleAddress = formatJoystreamAddress(EXAMPLE_ADDRESS)
const joystreamAddressPrefix = formattedExampleAddress.slice(0, 2)
```

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -195,7 +199,14 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({ onExitClick, account
             }}
           />
         </FormField>
-        <FormField label="Destination account" error={errors.account?.message}>
+        <FormField
+          label="Destination account"
+          error={errors.account?.message}
+          tooltip={{
+            text: 'Any Polkadot wallet address format is supported, but if you’re transferring tokens over to another Joystream member, we recommend using the Joystream wallet address format, which starts with “J4”.',
```
kdembler: Let's use computed `joystreamAddressPrefix` here instead of hardcoded value. Also, I think we should use regular qutoes `""`, instead of `“”`, that's what we do in the rest of the app

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -25,12 +25,14 @@ import { hapiBnToTokenNumber, tokenNumberToHapiBn } from '@/joystream-lib/utils'
...
 import { useFee, useJoystream, useTokenPrice } from '@/providers/joystream/joystream.hooks'
 import { useTransaction } from '@/providers/transactions/transactions.hooks'
+import { formatJoystreamAddress, shortenAddress } from '@/utils/address'
 import { SentryLogger } from '@/utils/logs'
 import { formatNumber } from '@/utils/number'
 
 import { FormFieldsWrapper, PriceWrapper, StyledMaxButton, VerticallyCenteredDiv } from './SendTransferDialogs.styles'
 
-const ADDRESS_LENGTH = 49
+const POLKADOT_ADDRESS_LENGTH = 48
```
kdembler: Where does this value come from? I think some formats are shorter than this, my Kusama address uses 47 chars

kdembler (changes requested):

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
rafalpawlow: Do you think [this validator](https://github.com/Joystream/atlas/pull/3336/files#diff-f952c324408bcd44638e0d7bb9e3812efc268849074fc0af406df08d15e19f4aR13) is fine? I found this in [Polkadot docs](https://polkadot.js.org/docs/util-crypto/examples/validate-address/)

kdembler (approved):

packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
kdembler: Nice find!

PR #3341, opened 2022-10-17
Welcome studio view update and few sign in flow copy updates

### LINKED ISSUE #3092: Improve "Welcome to Studio" screen
On [signin screen](https://play.joystream.org/studio/signin), which now will be destination from main screen, we use old illustrations which I think are far inferior to what we now have on the landing page. Can we replace?

Screenshot 2022-08-08 at 13 06 52

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203003530215165) by [Unito](https://www.unito.io)

drillprop:
Fix #3092 

figma: https://www.figma.com/file/ymHYOFItrmFaLeBC5Z4tpQ/Sign-in-flows?node-id=513%3A18750


rafalpawlow (changes requested):
1. This margin is wrong, according to figma it should be `32px`
Zrzut ekranu 2022-10-17 o 12 38 21

2. Something is wrong with position of app screenshots on mobile
**Yours**
Zrzut ekranu 2022-10-17 o 12 37 26
**Figma**
Zrzut ekranu 2022-10-17 o 12 42 11

3. Links section has no margin
Zrzut ekranu 2022-10-17 o 12 42 44

toiletgranny (changes requested):
1. **As is:** The Content frame wrapper has a `32px` of padding on all sides, including the LG breakpoint.
**To be:** On the LG breakpoint and higher, the padding should be set to `32px 80px`.
   CleanShot 2022-10-17 at 13 49 03
2. **As is:** The grid system seems to be a missing margin on all breakpoints. 
**To be:** The grid system should have `32px` of margin on `>= 1024px` viewport sizes, and `16px` on `< 1024px` viewport sizes, [as specified here.](https://www.figma.com/file/Cc3VDoK6qglJ617ChA2EMr/Foundation?node-id=716%3A62089)
CleanShot 2022-10-17 at 13 59 30
3. Apart from the links section having no padding, as @rafalpawlow has pointed out, it seems like screenshots also don't have padding on mobile.
CleanShot 2022-10-17 at 13 54 21
4. I realize this might be outside the scope of this PR (or is it not?), as it's been an issue for quite a while now, but the logo positioning on the top navbar is just very off compared to the designs. 
CleanShot 2022-10-17 at 13 56 54
5. **As is:** The image compression is way too heavy for the kind of imagery we're using here. Images are pixelated and blurry, and the colors are off.
**To be:** I remember pointing out the same thing when reviewing the initial YPP LP implementation. Can we use the same compression settings that have been ultimately used over there? If not, I'm happy to jump on a call to investigate and solve this together.
6. **As is:** Images are displayed at a `@1x` scale only, making them extra blurry on higher PPI displays, such as MacBooks. Also, they don't have specified `width` and `height` attributes.
**To be:**: Higher-resolution versions of images should be available for higher PPI displays, and both images should have `width` and `height` attributes set to `720` and `450` respectively.

---

Also, I updated all product screenshots earlier today to feature content better suited for Gleev's content focus.

rafalpawlow (changes requested):
1. This section has wrong margin-top, it's `40px` and it should be `48px`
Zrzut ekranu 2022-10-19 o 14 40 55

2. These screenshots on mobile are aligned and should be offset. Also I would add some padding at the bottom, there is no space when you scroll to the bottom.
Zrzut ekranu 2022-10-19 o 14 39 51

packages/atlas/src/views/studio/StudioWelcomeView/StudioWelcomeView.styles.ts
```diff
@@ -1,34 +1,81 @@
...
+
+export const LEFT_ANIMATION_MD = 'left-animation-md'
+
+export const OverflowHiddenContainer = styled.div`
+  margin: 0 calc(-1 * var(--size-global-horizontal-padding)) 0;
+  overflow: hidden;
+
+  [data-aos] {
+    &[data-aos][data-aos-easing='atlas-easing'] {
+      transition-timing-function: cubic-bezier(0.09, 0.43, 0.3, 0.99);
```
rafalpawlow: I wonder if we should add this to `GlobalStyles` since we will be using it in several places.

packages/atlas/src/views/studio/StudioWelcomeView/StudioWelcomeView.styles.ts
```diff
@@ -48,24 +95,200 @@ export const StyledSignInIllustrationSVG = styled(SvgSigninIllustration)`
...
+    overflow: hidden;
+    padding: ${sizes(3)};
+
+    ::before,
+    ::after {
+      position: absolute;
+      content: '';
+      background: linear-gradient(
+        90deg,
+        #000 0%,
```
rafalpawlow: Figma is broken 😅. I bet this is simple gradient from black to transparency. @toiletgranny do you think we could just do `background: linear-gradient(90deg, rgba(0,0,0,0) 0%, rgba(0,0,0,1) 100%);`?

drillprop:
> 2\. These screenshots on mobile are aligned and should be offset. Also I would add some padding at the bottom, there is no space when you scroll to the bottom.

What do you mean, here? They are aligned with the grid, I thought this is how it should work here: 
image
image

The first image sticks to the edge of the first column, second sticks to the edge of the last column.

packages/atlas/src/views/studio/StudioWelcomeView/StudioWelcomeView.styles.ts
```diff
@@ -48,24 +95,200 @@ export const StyledSignInIllustrationSVG = styled(SvgSigninIllustration)`
...
+        rgb(0 0 0 / 0.784) 46.66%,
+        rgb(0 0 0 / 0.7037) 55.41%,
+        rgb(0 0 0 / 0.6056) 64.39%,
+        rgb(0 0 0 / 0.488) 73.47%,
+        rgb(0 0 0 / 0.349) 82.52%,
+        rgb(0 0 0 / 0.187) 91.41%,
+        rgb(0 0 0 / 0) 100%
+      );
+      width: 24px;
+      z-index: 1;
```
rafalpawlow, line 113: Can we use some value from `zIndex` const?

rafalpawlow (changes requested):

packages/atlas/src/views/studio/StudioWelcomeView/StudioWelcomeView.styles.ts
drillprop: I don't agree. Not everything should use const, this is used only locally - to put a shadow upfront

rafalpawlow:
@drillprop yeah you are 100% right, but on this particular resolution (iPhone 12 pro) they do not go beyond the screen and I am not sure if this is ok. @toiletgranny ?
Zrzut ekranu 2022-10-19 o 14 59 51

drillprop:
> 1. This section has wrong margin-top, it's `40px` and it should be `48px`

Because it was added to these `spans`. So the gap should be correct(40 + 8 = 48). I added this margin to these spans because we can't use the gap in flex, so I applied margin-top to every span in case of flex wrapping. 
image

packages/atlas/src/views/studio/StudioWelcomeView/StudioWelcomeView.styles.ts
drillprop: yeah, I was about to do that but forgot in the end. I'll add this to `GlobalStyles`

packages/atlas/src/views/studio/StudioWelcomeView/StudioWelcomeView.styles.ts
toiletgranny: What is this gradient? 😅 Where is it used?

rafalpawlow: Zrzut ekranu 2022-10-19 o 15 18 26

toiletgranny (changes requested):
Awesome improvement! :feelsgood: (by the way, that's an interesting emoji for a `:feelsgood:`) Apart from the things @rafalpawlow had already pointed out, I got two minor details:

1. **As is:** In the "Ready to create your channel?" variant of the Studio welcome screen, which — as a reminder — is a displayed when I have my membership connected and haven't created a channel yet, the CTA in the top navigation says "Set up membership", while the other one (in the main column) says "Create first channel". And has an icon.
**To be:** The CTA in the main column should be "Create channel," and there should be no icon. For the Top navigation, we should be displaying the active member avatar and the notification icon, not the button. Link to the [Figma designs](https://www.figma.com/file/ymHYOFItrmFaLeBC5Z4tpQ/Sign-in-flows?node-id=508%3A59710).
2. It's great to see 2x resolution images on my retina MacBook Pro display <3, but the 1x images appear to be unchanged compared to the last deployment I looked at. 

---

> @drillprop yeah you are 100% right, but on this particular resolution (iPhone 12 pro) they do not go beyond the screen and I am not sure if this is ok. @toiletgranny ?

@rafalpawlow This is fine by me. But if you, @drillprop, also think this looks odd on iPhones, then feel free to change the size of both images to `480x300` on the smallest breakpoint (currently this size kicks in on the XS breakpoint `425px`).

packages/atlas/src/views/studio/StudioWelcomeView/StudioWelcomeView.styles.ts
toiletgranny: Oh, that's a spectacular gradient, indeed. :D Yeah, we can totally use simpler version if you ask me.

toiletgranny (changes requested):

rafalpawlow (approved):

toiletgranny (approved):

PR #3342, opened 2022-10-17
Mobile devices onboarding flow

### LINKED ISSUE #2903: Mobile user agent onboarding flow
## Context
Mobile wallets are not supported in Polkadot

## Scope
* [ ] Adjust copy update for onboarding flow if user agent is mobile 
>(e.g. Before Polkadot JS signers support mobile, you can look but to do smth use desktop)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202496816781150) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #2903 

@toiletgranny I didn't apply your suggestion with detecting if app is installed, it just didn't work correctly. Without going into details, the disadvantage of my solution is that instead of opening a store, we open a https://polkawallet.io/
@drillprop please make sure the app opens correctly on android if it is installed on the phone

**Design**
https://www.figma.com/file/ymHYOFItrmFaLeBC5Z4tpQ/Sign-in-flows?node-id=1910%3A31505


toiletgranny (changes requested):
1. **As is:** The copy in the Sign in dialog (the Mobile variant), is exactly the same, no matter if the dialog is triggered by tapping on the "Connect wallet" button or by trying to perform an action that requires a wallet connection.
**To be:** Upon displaying the Sign in dialog (the Mobile variant), the copy should be different, [as documented here.](https://www.figma.com/file/ymHYOFItrmFaLeBC5Z4tpQ/Sign-in-flows?node-id=1910%3A31745)
2. **As is:** Upon tapping on the "Connect anyway" button in the Sign in dialog (the Mobile variant), having brought it up as a result of trying to perform an action that requires a wallet connection, the Sign in dialog is dismissed, and nothing else happens.
**To be:** Connect wallet dialog should be displayed instead.
3. **As is:** For some reason, the padding in the footer of the Sign in dialog (the Mobile variant) is larger than it's supposed to be. There's an additional 8px of margin at the top. This might be a component issue, but I noticed it just now.
**To be:** Of course, there should be no unnecessary margin at the top. On a related note, @kdembler, this is yet another instance of an abandoned margin left in the implementation that I found recently. Couldn't we switch to using `gap` in Flexbox, knowing that resetting margins is an issue?
4. **As is:** In the Sign in dialog (the Mobile variant), there's no padding around the text.
**To be:** In the Sign in dialog (the Mobile variant), an 8px of additional padding is applied around the text. This padding is applied only up until the SM breakpoint.
5. **As is:** Illustration proportions Sign in dialog (the Mobile variant) remain unchanged thorough breakpoints.
**To be:** Illustration proportions change across breakpoints. For example, on the smallest breakpoint, the illustration of the computer is larger compared to the patterns in the background. The logic behind it is similar as in the Desktop variant of the Sign in dialog.
6. **As is:** Polkawallet icon is blurred on higher PPI displays.
**To be:** There should be a higher resolution version of the Polkawallet icon served on higher PPI displays. Please note, that unlike icons of other wallets, which we're getting from Talisman, the Polkawallet icon is a raster image inside (this is because we simply do not have a vector version of it).
7. **As is:** On desktop, the header for the first step of the Connect wallet dialog is "Select wallet false".
**To be:** It should be just "Select wallet". 
   CleanShot 2022-10-17 at 16 10 11@2x
8. **As is:** The `ListItem` component, used in the Connect wallet dialog, in the Active state, does not have any background to it.
**To be:** The `ListItem` component in the Connect wallet dialog had this custom treatment of having a `Background/Default/Muted/Alpha` background applied to it in its Active state. We seem to have lost this visual cue some time ago, and while this is not a direct result of changes made in this PR, could we try to bring this back, as it's something we haven't caught before?
9. **As is:** When trying to sign in using the Polkawallet app, in the Connect wallet dialog, the "Polkawallet" name is displayed using all lowercase letters.
**To be:** Why is that? Could we have the "Polkawallet" name displayed with the capital "P" letter? And, doing so, to make sure that it won't affect how other wallet names are displayed (such as "SubWallet")?

toiletgranny (approved):

packages/atlas/src/components/ListItem/ListItem.styles.ts
```diff
@@ -51,7 +57,8 @@ export const Container = styled('div', { shouldForwardProp: isPropValid })<Conta
   align-items: center;
   user-select: none;
   cursor: pointer;
-  background-color: ${({ highlight }) => (highlight ? cVar('colorBackgroundMutedAlpha') : 'unset')};
+  background-color: ${({ highlight, highlightWhenActive }) =>
```
drillprop: Can we have only `highlight` here? Why do we need another prop? We could pass on the component level just `highlight || highlightWhenActive`? Am I right?

rafalpawlow: I agree 👍

drillprop (approved):

Issue #3348, opened 2022-10-18
Skip Sentry init if DSN is not provided


kdembler:
Like in the title

PR #3351, opened 2022-10-18
add env interpolation to atlas.config.yml

### LINKED ISSUE #3332: Add env vars interpolation to atlas.config.yml
So that users can use single config file but reference values passed via env variables. For example:
```yaml
  faucet:
    hcaptchaSiteKey: %VITE_APP_HCAPTCHA_SITE_KEY%
```

Then, let's also move hcaptcha config to atlas.config.yml

kdembler:
Closes #3332 and #3348


drillprop (approved):
LGTM

PR #3353, opened 2022-10-18
add Atlas operator guide

kdembler:
Adds a guide for running your own Atlas instance, configuring according to your needs, and running any additional services Atlas depends on.

Preview link: https://github.com/kdembler/atlas/blob/operator-guide/docs/operator-guide.md
Config file link: https://github.com/kdembler/atlas/blob/operator-guide/packages/atlas/atlas.config.yml

**Left to do:**

- [x] Figure out how to customize `Terms of Service` and `DMCA Policy`

**Any feedback will be highly appreciated**


docs/operator-guide.md
```diff
@@ -0,0 +1,191 @@
...
+**Tip:** Atlas provides built-in "admin tools" panel that can be useful when running your own instance. You can access it in the app by pressing Ctrl+Shift+D.
+
+### Prerequisites
+
+- Basic knowledge of how to use the command line, how to operate Docker, and how to host a web application
+- Domain name (e.g. `example.com`) to make your gateway publicly available
+- Server to deploy required services (e.g. VPS, dedicated server, or cloud instance)
+- Some kind of web server (e.g. Nginx) or PaaS (e.g. Vercel) to host the frontend app
+
+### Running Atlas
```
drillprop, line 50: Maybe I'm nitpicking here, but I'm not sure if this section should be first. By reading this article, I'm getting the impression, that you need to **first** run the atlas, and **then** configure it(without building it again). I know that this guide is aimed at rather to technical people, and most people will figure this out, but on the other hand, someone could not get it.  Maybe we should add at least just one sentence in `Configuring Atlas`, that you need to build Atlas after configuring the Atlas? WDYT?

kdembler: Good call, I will mention this

kdembler: Added in c867be77039f1015bee43840e54e29ed4dc20aa4

README.md
```diff
@@ -50,6 +50,8 @@ If you are looking for a good place to start, check out the [good first issue la
 
 ### Quickstart
 
+If you are interested in running your own instance of Atlas, please take a look at our [operator guide](docs/operator-guide.md).
```
bedeho: I think running as operators, as distinct from working on the code base, is such a different, and arguably even more relevant, use-case to the typical reader, that it deserves its own section.

bedeho (changes requested):
Looks pretty complete! more than enough to get stared with

README.md
kdembler: Sure, elevated to separate section

PR #3359, opened 2022-10-21
Channel payouts POC

drillprop:


kdembler:
Let's create a new branch `payments` from `dev` and target this PR at that

packages/atlas/src/api/queries/fragments.graphql
```diff
@@ -10,6 +10,7 @@ fragment BasicChannelFields on Channel {
   follows
   rewardAccount
   channelStateBloatBond
+  cumulativeRewardClaimed
```
kdembler: Do we need this on `BasicChannel`? Seems like this should be useful only in contexts where we already have the full channel?

packages/atlas/src/api/queries/storage.graphql
```diff
@@ -43,3 +43,17 @@ query GetBasicStorageBuckets {
...
+      operatorStatus_json: { isTypeOf_eq: "StorageBucketOperatorStatusActive" }
+      bags_some: { id_eq: $bagId }
+      operatorMetadata: { nodeEndpoint_contains: "http" }
+    }
+  ) {
+    operatorMetadata {
+      nodeEndpoint
+    }
+  }
+}
```
kdembler: This data shouldn't be fetched from storage provider, but from a distribution provider. Seems no new queries should be needed, we should already have bag->provider mapping with out current assets provider

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -465,7 +469,105 @@ export class JoystreamLibExtrinsics {
...
+  ) => {
+    const payoutProof = await channelPayoutProof(
+      'URL',
+      `${nodeEndpoint}/api/v1/files/${payloadDataObjectId}`,
+      Number(channelId)
+    )
+
+    const cashout = new BN(payoutProof.cumulativeRewardEarned).sub(new BN(cumulativeRewardEarned || 0))
+    return cashout.toString()
+  }
```
kdembler: This isn't an extrinsic actually, maybe we should move it to `utils` or something?

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -465,7 +469,105 @@ export class JoystreamLibExtrinsics {
...
+    payloadDataObjectId: string,
+    commitment: string
+  ) => {
+    await this.ensureApi()
 
+    const payoutProof = await channelPayoutProof(
+      'URL',
+      `${nodeEndpoint}/api/v1/files/${payloadDataObjectId}`,
+      Number(channelId)
+    )
```
kdembler: Could we reuse `getClaimableReward()` here?

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -465,7 +469,105 @@ export class JoystreamLibExtrinsics {
...
+
+    const maxCashoutAllowed = await this.api.query.content.maxCashoutAllowed()
+    const minCashoutAllowed = await this.api.query.content.minCashoutAllowed()
+    const cashout = new BN(payoutProof.cumulativeRewardEarned).sub(new BN(cumulativeRewardEarned || 0))
+    const isPayoutProofVerified = verifyChannelPayoutProof(payoutProof) === commitment
+
+    if (maxCashoutAllowed.lt(cashout)) {
+      throw new JoystreamLibError({
+        name: 'FailedError',
+        message: `Channel cashout amount is too low to be claimed`,
```
kdembler: I think in this case it's `too high` actually? Maybe we could switch the condition around and make it `cashout.gt(maxCashoutAllowed)`? That's a bit easier to read IMO

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -465,7 +469,105 @@ export class JoystreamLibExtrinsics {
...
+      reason: u8aToHex(Buffer.from(payoutProof.reason, 'hex')),
+    })
+    const merkleBranch: PalletCommonMerkleTreeProofElementRecord[] = []
+    payoutProof.merkleBranch.forEach((m) => {
+      const proofElement = createType('PalletCommonMerkleTreeProofElementRecord', {
+        hash_: u8aToHex(Buffer.from(m.hash, 'hex')),
+        side: m.side ? { Right: null } : { Left: null },
+      })
+      merkleBranch.push(proofElement)
+    })
```
kdembler: I think this code was included in `@joystream/js` after all so we may not need it in our codebase. Can you check?

packages/atlas/vite.config.ts
```diff
@@ -52,6 +53,15 @@ export default defineConfig({
...
     }),
+    {
+      ...inject({
+        include: ['node_modules/**/*.js*'],
+        modules: {
+          Buffer: ['buffer', 'Buffer'],
+        },
+      }),
+      enforce: 'post',
+    },
```
kdembler, line 65: Why is this needed? I was pretty sure the `Buffer` polyfill worked properly before. Maybe with the ESM `@joystream/js` version it's not needed anymore?

packages/atlas/vite.config.ts
```diff
@@ -71,7 +81,7 @@ export default defineConfig({
     },
   },
   optimizeDeps: {
-    include: ['blake3/browser-async', 'multihashes'],
+    include: ['blake3/browser-async', 'multihashes', 'buffer'],
```
kdembler, line 85: Same here

packages/atlas/vite.config.ts
drillprop: I'm still getting `Buffer is not defined` when I run `claimReward` unfortunately. I think I need this, unfortunately

packages/atlas/src/joystream-lib/extrinsics.ts
drillprop: I can't move this to `utils.ts` because tests are failing when I did this, even though I didn't test or import that particular function. I don't understand why. Here are logs from the terminal. 
image
I visited https://github.com/vitest-dev/vitest/issues/1191 but didn't find any reasonable solution. 

In the end, I created `channelPayouts.ts` and moved `getClaimableReward` there

packages/atlas/src/joystream-lib/channelPayouts.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const getClaimableReward = async (
+  channelId: string,
+  cumulativeRewardEarned: StringifiedNumber | null,
+  nodeEndpoint: string,
+  payloadDataObjectId: string
+) => {
+  const payoutProof = await channelPayoutProof(
+    'URL',
+    `${nodeEndpoint}/api/v1/assets/${payloadDataObjectId}`,
```
kdembler: This should ideally use `createDistributionOperatorDataObjectUrl` utility (or some modified version of it) so that:
a) we use asset path from config
b) we don't reimplement the same URL join

packages/atlas/vite.config.ts
```diff
@@ -71,7 +82,7 @@ export default defineConfig({
     },
   },
   optimizeDeps: {
-    include: ['blake3/browser-async', 'multihashes'],
+    include: ['blake3/browser-async', 'multihashes', 'buffer', 'crypto-js'],
```
kdembler: We can get rid of `crypto-js`, it's not used after my patch

kdembler (approved):
Additionally to comments, could you change the default env on this branch to `next` so that we don't have to manually switch on new deployments?

Other than that, seems to work great, amazing job. Feel free to merge once comments are resolved 👍

PR #3360, opened 2022-10-21
Improve language selects

### LINKED ISSUE #3257: Improve language selects
## Context

Review and extend the list of supported languages

## Scope

- [x] Update the languages list in the app
- [ ] Replace language select in VideoForm with Combobox
- [ ] Implement the new list separator
- [ ] Introduce separation into "Popular" and "All" languages

Out of scope: ⛔ 
- [ ] 🆕  ~Update the language select component for uploading subtitles to support multiple languages.~ not required 

🔗 Spreadsheet [LINK > ](https://docs.google.com/spreadsheets/d/1DS8nyJWLQpXWKrX9q697VHj2snV4rUwTVkWN3PlKEOE/edit?usp=sharing)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202987367684496) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #3257 

**Design**
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=9696%3A171951
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=1945%3A109809


packages/atlas/atlas.config.yml
```diff
@@ -174,6 +174,7 @@ content:
       name: Vietnamese
     - isoCode: yo
       name: Yoruba
+  popularLanguages: ['ar', 'zh', 'en', 'fr', 'de', 'ja', 'ms', 'pt', 'ru', 'st']
```
drillprop: I'm not sure if this list should be sorted alphabetically. I would display the english language first, then probably russian(since we have a lot of russian speaking users)

drillprop (changes requested):
I'll look at the code later, but can we add a Combobox variant with popular choices to the storybook?

packages/atlas/atlas.config.yml
rafalpawlow: I'm not sure about this, this is how it was ordered here: https://docs.google.com/spreadsheets/d/1DS8nyJWLQpXWKrX9q697VHj2snV4rUwTVkWN3PlKEOE/edit#gid=0

Let's confirm this with @dmtrjsg

packages/atlas/src/components/ListItem/ListItemSeparator.tsx
```diff
@@ -0,0 +1,10 @@
+import { SeparatorWrapper } from '@/components/ListItem/ListItem.styles'
+
+export const ListItemSeparator = () => {
+  return (
+    <SeparatorWrapper>
+      <div />
+      <div />
```
drillprop: This is confusing for me, I would opt for creating styled component for each div and style them separately.

packages/atlas/src/components/_inputs/VideoLanguageCombobox/VideoLanguageCombobox.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export default {
+  title: 'inputs/VideoLanguageCombobox',
+  component: VideoLanguageCombobox,
+  args: {
+    onSelectedItemChange: () => null,
+  },
+} as Meta<VideoLanguageComboboxProps>
+
+const Template: Story<VideoLanguageComboboxProps> = (args) => <VideoLanguageCombobox {...args} />
```
drillprop: I thought you will add the variant in the Combobox stories, something like "Combobox with separator", or "Combobox with suggested options". 
IMO It would be more helpful if we have that story in the Combobox component. It would mean, that adding suggested options is actually possible with the Combobox component. 

Also when the value is not available, I don't see information about a not found item
Combobox
image
VideoLangaugeCombobox
image

Not sure if it's a bug or you just need to provide some props to make this work.

packages/atlas/src/components/ListItem/ListItem.tsx
```diff
@@ -58,50 +60,65 @@ export const ListItem = forwardRef<HTMLDivElement, ListItemProps>(
...
   ) => {
     const [hoverRef, isHovering] = useHover<HTMLDivElement>()
     const linkProps = getLinkPropsFromTo(to)
+
+    const labelColor = useMemo(() => {
+      if (destructive) {
+        return 'colorTextError'
+      }
+      if (isSeparator) {
+        return 'colorText'
```
drillprop: It should be `colorTextMuted` If I'm not mistaken. 
image

drillprop (changes requested):

packages/atlas/atlas.config.yml
rafalpawlow: We agreed that we will sort them alphabetically except for English which will be in first place

drillprop (changes requested):
Found two bugs: 
1.

https://user-images.githubusercontent.com/51168865/197769937-ed2375bd-42e9-4e41-bc41-35ad0d83d320.mov

Does doubled key cause it?

2. It's possible to submit with blank language Combobox. To reproduce: pick some language, remove language in Combobox and click publish changes.

rafalpawlow:
> It's possible to submit with blank language Combobox. To reproduce: pick some language, remove language in Combobox and click publish changes.

@drillprop I disagree it's a bug, combobox is not a regular text field, what you select from the list is sent to the form, not what you type in the input. I think that instead of doing some wonky stuff to handle it, we should display information about what language was selected as we do for subtitles or whitelisting. @toiletgranny Could you take a look at it?

packages/atlas/atlas.config.yml
```diff
@@ -126,7 +126,7 @@ content:
       name: Filipino
     - isoCode: fi
       name: Finnish
-    - isoCode: de
+    - isoCode: el
```
drillprop, line 129: You fixed the issue with the German language, but the problem with selecting Portuguese still exists.

drillprop (changes requested):

drillprop:
> @drillprop I disagree it's a bug, combobox is not a regular text field, what you select from the list is sent to the form, not what you type in the input. I think that instead of doing some wonky stuff to handle it, we should display information about what language was selected as we do for subtitles or whitelisting. @toiletgranny Could you take a look at it?

Hmm, after second thought maybe we shouldn't use the combobox at all to select language video? Just use the regular select component. We'll be still able to navigate with the keyboard to select language. Also, I don't think using combobox in this particular place was requested by the designers, it was requested by Dima, if I remember correctly. 

@toiletgranny we need your input here.

toiletgranny:
There is a couple of problems with using the Combobox component here, both UX and UI-wise. In a nutshell, Combobox !== Select component with typing functionality on top of it. If we want to have a Select component with type-to-filter capabilities, it requires updating the Select component and therefore thinking of and addressing all of the edge cases, such as the one found by @drillprop above.

For this reason, I'd recommend reverting this back to the Select component. Let's keep in mind that its current implementation allows for pressing letters on keyboards to focus on list items with these letters in their names.

@dmtrjsg @kdembler **In the future, I'd really appreciate it if you guys let us know about these kinds of changes so that, at the very least, we have an opportunity to update the designs accordingly.**

drillprop (approved):

PR #3363, opened 2022-10-24
Provide better errors from faucet

### LINKED ISSUE #3172: Provide better error messages for failed faucet requests
## Scope

- [ ] implement custom messages in the popup notiffs as per [this comment ](https://github.com/Joystream/atlas/issues/3172#issuecomment-1268525367)
- [ ] keep generic error message for all cases not defined in point 1

## Context

**Describe the bug**

Membership creation with Talisman failing.

**To Reproduce**

1. Go to https://play.joystream.org/studio/
1. Click on `Connect wallet`
1. Select Talisman wallet
1. Select account
1. Accept terms of service
1. Type in member handle ‘talisman’
1. Click create membership
1. Wait ~25 seconds on creating membership screen
1. See error “Something went wrong. There was a problem with creating your membership. Please try again later.”

**Screenshots**

OS: MacOS Monterey 12.2.1
Browser: Chrome

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202882495509820) by [Unito](https://www.unito.io)

drillprop:
Fix #3172

To make this PR work, this PR needs to be merged(and deployed) first https://github.com/Joystream/membership-faucet/pull/23

### DIFF
--- a/packages/atlas/src/components/_auth/SignInModal/SignInModal.tsx
+++ b/packages/atlas/src/components/_auth/SignInModal/SignInModal.tsx
@@ -1,4 +1,4 @@
-import axios, { AxiosError } from 'axios'
+import axios from 'axios'
 import { BN } from 'bn.js'
 import { FC, useCallback, useEffect, useRef, useState } from 'react'
 import shallow from 'zustand/shallow'
@@ -14,6 +14,7 @@ import { useSnackbar } from '@/providers/snackbars'
 import { useTransactionManagerStore } from '@/providers/transactions/transactions.store'
 import { useUser } from '@/providers/user/user.hooks'
 import { useUserStore } from '@/providers/user/user.store'
+import { isAxiosError } from '@/utils/error'
 import { uploadAvatarImage } from '@/utils/image'
 import { ConsoleLogger, SentryLogger } from '@/utils/logs'
 
@@ -37,6 +38,7 @@ export const SignInModal: FC = () => {
   const [hasNavigatedBack, setHasNavigatedBack] = useState(false)
   const { joystream } = useJoystream()
   const dialogContentRef = useRef<HTMLDivElement>(null)
+  const [previouslyFailedData, setPreviouslyFailedData] = useState<MemberFormData | null>()
 
   const { displaySnackbar } = useSnackbar()
   const { walletStatus, refetchUserMemberships, setActiveUser, isLoggedIn } = useUser()
@@ -129,6 +131,7 @@ export const SignInModal: FC = () => {
         }
         const { block } = await createNewMember(selectedAddress, data)
         addBlockAction({ targetBlock: block, callback })
+        setPreviouslyFailedData(null)
       } catch (error) {
         if (error.name === 'UploadAvatarServiceError') {
           displaySnackbar({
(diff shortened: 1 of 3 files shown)


packages/atlas/src/components/_auth/SignInModal/SignInModal.tsx
```diff
@@ -140,15 +143,56 @@ export const SignInModal: FC = () => {
...
+                iconType: 'error',
+              })
+              break
+          }
+        } else {
+          displaySnackbar({
+            title: 'Something went wrong',
+            description: `There was a problem with creating your membership. Please try again later.`,
+            iconType: 'error',
+          })
```
kdembler: Let's try to avoid duplicating like this. You can extract `errorCode` inline and then you can handle everything with just a switch. This also has a benefit of explicitly including `errorCode` in the Sentry log. I think when we pass only the full error, the `errorCode` is too deeply nested and gets dropped from the Sentry event.
```tsx
const errorCode = isAxiosError<NewMemberResponse>(error) ? error.response?.data?.error : null

SentryLogger.error('Failed to create a membership', 'SignInModal', error, { error: { errorCode } })

switch (errorCode) {
  case 'TooManyRequestsPerIp':
    displaySnackbar({
      title: 'You reached a membership limit',
      description:
        'Your membership could not be created as you already created one recently from the same IP address. Try again in 2 days.',
      iconType: 'error',
    })
    break
  case 'TooManyRequests':
    displaySnackbar({
      title: 'Our system is overloaded',
      description:
        'Your membership could not be created as our system is undergoing a heavy traffic. Please, try again in a little while.',
      iconType: 'error',
    })
    break
  case 'OnlyNewAccountsCanBeUsedForScreenedMembers':
    displaySnackbar({
      title: 'This account is not new',
      description:
        'Your membership could not be created as the selected wallet account has either made some transactions in the past or has some funds already on it. Please, try again using a fresh wallet account. ',
      iconType: 'error',
    })
    break

  default:
    displaySnackbar({
      title: 'Something went wrong',
      description: `There was a problem with creating your membership. Please try again later.${
        errorCode ? ` Error code: ${errorCode}` : ''
      }`,
      iconType: 'error',
    })
    break
}

goToPreviousStep()
setPreviouslyFailedData(data)
return
```

packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalMembershipStep.tsx
```diff
@@ -40,7 +42,8 @@ export const SignInModalMembershipStep: FC<SignInModalMembershipStepProps> = ({
     watch,
     control,
     formState: { errors, isSubmitting },
-  } = useForm<MemberFormData>({ reValidateMode: 'onSubmit' })
+  } = useForm<MemberFormData>({ reValidateMode: 'onSubmit', defaultValues: previouslyFailedData || undefined })
```
kdembler, line 45: Nice 👍

packages/atlas/src/components/_auth/SignInModal/SignInModal.tsx
```diff
@@ -140,15 +143,56 @@ export const SignInModal: FC = () => {
           return
```
kdembler, line 147: This case will miss setting failed data and also Sentry log :(
1. Maybe `data` could be passed to `goToPreviousStep` and `setPreviouslyFailedData` could be called there? This way we will not have to call both functions manually in all places. Or just added here
2. Could you add a Sentry log here? 🙏 `Failed to upload member avatar` **OR** we actually stick that log directly in `uploadAvatarImage()` and then we also get same log if edited upload fails.

packages/atlas/src/components/_auth/SignInModal/SignInModal.tsx
```diff
@@ -227,9 +282,10 @@ export const SignInModal: FC = () => {
...
   )
 }
+type FaucetErrorType = 'TooManyRequestsPerIp' | 'TooManyRequests' | 'OnlyNewAccountsCanBeUsedForScreenedMembers'
 
 type NewMemberResponse = {
   memberId: MemberId
   block: number
-  error?: string
+  error?: FaucetErrorType
 }
```
kdembler: Could you extract `type NewMemberErrorResponse` to not assume `memberId` and `block` are available on error?

kdembler (commented):
Good job on this, also on the faucet side. Works well, left some small comments

kdembler (approved):

PR #3382, opened 2022-10-31
fix subtitles edit and missing data in VideoForm

### LINKED ISSUE #3375: Can't edit video, when I open VideoWorkspace, the form is cleared
I think we had this bug before. Basically, I can't edit certain videos, because the form in VideoWorkspace is cleared out.  Most of them work, but some of them don't. 
Unfortunately, I don't know the exact steps to reproduce this but happened to me twice. 
I'll do more testing and try to find the exact steps to reproduce. 

https://user-images.githubusercontent.com/51168865/198523532-98957957-8307-418e-a8e0-c81259c5660e.mov

kdembler:
Fixes #3375

The issue was again with loading of subtitles when all of them had the asset missing. Because the metadata was defined, but missing assets would not get resolved, the form though that the data was still loading. While investigating I found a couple more issues with subtitles edit/removal, all should be fixed now.


rafalpawlow (approved):

drillprop (approved):
Looks good to me as well!

PR #3383, opened 2022-10-31
Better handling of failed uploads

### LINKED ISSUE #3130: Poor handling of failed uploads
### Problem
If you publish a video and some of its assets fail to upload, you will see this in `My uploads`:

Then, user can reconnect the file and try to upload again. However, if they don't have the file available or some other problem occured, they may see the following modal:

There are some issues with the above:

1. We don't offer an option to just close the modal and don't take any action. You are forced to either edit the video or try selecting the file again. This can be frustrating to the user. Because of the current implementation, if you click Esc, it its equal to clicking the secondary button in the dialog and results in opening the video workspace.
1. The "Edit video" option actually doesn't work - it just opens an empty video workspace and you cannot edit the video in question - implementation issue
1. I'm unsure if the "Edit video" option makes sense if we are talking about media files. Even if it worked, we currently don't allow editing the media file inside Atlas, only the thumbnail. So you still wouldn't be able to fix the issue

### Proposal
I think we need to refine this flow as uploading assets will be one of the most fundamental actions creators will take on the platform and we need to make sure that everything works smoothly. Fixing first point is probably quite easy with some basic design treatment and improving the implementation. Second point is purely technical. When it comes to the third one, I suggest we allow editing the media file if its upload has failed. Maybe it's also worth taking a more holistic look at the issue, maybe there are some other improvements we could do.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202882567635505) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #3130 

**Design**
https://www.figma.com/file/Pe1Y0nrl4mpHWs7AkuQOk5/My-uploads-page?node-id=1133%3A11900

### DIFF
--- a/packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
+++ b/packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
@@ -32,6 +32,11 @@ import {
 } from './ImageCropModal.styles'
 import { CropperImageType, useCropper } from './cropper'
 
+type FeeData = {
+  methodName: TxMethodName
+  args?: Parameters<JoystreamLibExtrinsics[TxMethodName]>
+}
+
 export type ImageCropModalProps = {
   imageType: CropperImageType
   onConfirm: (
@@ -43,20 +48,15 @@ export type ImageCropModalProps = {
   ) => void
   onDelete?: () => void
   onError?: (error: Error) => void
+  fee?: FeeData
 } & Pick<DialogModalProps, 'onExitClick'>
-
-type FeeData = {
-  methodName: TxMethodName
-  args?: Parameters<JoystreamLibExtrinsics[TxMethodName]>
-}
 
 export type ImageCropModalImperativeHandle = {
   open: (file?: File | Blob | null, cropData?: ImageCropData, edit?: boolean) => void
-  setFee: (fee?: FeeData) => void
 }
 
 const ImageCropModalComponent: ForwardRefRenderFunction<ImageCropModalImperativeHandle, ImageCropModalProps> = (
-  { imageType, onConfirm, onDelete, onError },
+  { imageType, onConfirm, onDelete, onError, fee },
   ref
 ) => {
   const [showModal, setShowModal] = useState(false)
@@ -65,7 +65,6 @@ const ImageCropModalComponent: ForwardRefRenderFunction<ImageCropModalImperative
   const [editedImageHref, setEditedImageHref] = useState<string | null>(null)
   const [cropData, setCropData] = useState<ImageCropData | null>(null)
   const [editMode, setEditMode] = useState(false)
-  const [fee, setFee] = useState<FeeData>()
   const [originalBlob, setOriginalBlob] = useState<File | Blob | null>(null)
   const { fullFee, loading } = useFee(fee?.methodName, fee?.args && showModal ? fee.args : undefined)
   const { currentZoom, zoomRange, zoomStep, handleZoomChange, cropImage } = useCropper({
@@ -98,9 +97,6 @@ const ImageCropModalComponent: ForwardRefRenderFunction<ImageCropModalImperative
         if (cropData) setCropData(cropData)
         inputRef.current?.click()
       }
-    },
-    setFee: (fee?: FeeData) => {
-      setFee(fee)
     },
   }))
 
--- a/packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
+++ b/packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
@@ -184,14 +184,16 @@ const ImageCropModalComponent: ForwardRefRenderFunction<ImageCropModalImperative
         secondaryButton={{ text: 'Cancel', onClick: resetModal }}
         additionalActionsNodeMobilePosition="bottom"
         additionalActionsNode={
-          <>
-            {fee && <Fee loading={loading} variant="h200" amount={fullFee || new BN(0)} />}
-            {editMode && onDelete && (
+          fee ? (
+            <Fee loading={loading} variant="h200" amount={fullFee || new BN(0)} />
+          ) : (
+            editMode &&
+            onDelete && (
               <Button onClick={handleDeleteClick} variant="destructive-secondary" icon={<SvgActionTrash />}>
                 Delete
               </Button>
-            )}
-          </>
+            )
+          )
         }
         onExitClick={resetModal}
         dividers
--- a/packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
+++ b/packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
@@ -1,3 +1,4 @@
+import BN from 'bn.js'
 import {
   ForwardRefRenderFunction,
   forwardRef,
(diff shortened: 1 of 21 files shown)


drillprop (changes requested):
I didn't look at the code, but I see some issues when I was trying to re-upload a cover image file. Everything seems to be fine with a video file. Did you test this as well on images?

So when I re-uploaded the new image in my uploads view I see: 
image
As you may see, there is no title, and when I click "Try again" I'm getting an error

In addition to the above, when I go to "My videos" I see two tiles with information about broken assets: 
image
Again, one doesn't have a title and it shouldn't be here. We should have only one with the name `sdsdsa`

kdembler:
@rafalpawlow @drillprop Since I won't be available till Monday, let's make @drillprop the primary reviewer here. However, I will want to take a look on Monday before we merge as well

packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
```diff
@@ -54,7 +65,9 @@ const ImageCropModalComponent: ForwardRefRenderFunction<ImageCropModalImperative
   const [editedImageHref, setEditedImageHref] = useState<string | null>(null)
   const [cropData, setCropData] = useState<ImageCropData | null>(null)
   const [editMode, setEditMode] = useState(false)
+  const [fee, setFee] = useState<FeeData>()
   const [originalBlob, setOriginalBlob] = useState<File | Blob | null>(null)
+  const { fullFee, loading } = useFee(fee?.methodName, fee?.args && showModal ? fee.args : undefined)
```
drillprop, line 69: I don't like the fact that we're getting fees inside ImageCropModalComponent. Can we just pass the fee to the component via props? Is it possible?  Why do we actually need this here in the first place?

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.hooks.ts
```diff
@@ -0,0 +1,244 @@
...
+type CreateEditChannelData = {
+  newChannel: boolean
+  metadata: ChannelInputMetadata
+  assets: ChannelAssets<ImageAsset>
+  avatarAsset: ResolvedAsset | null
+  coverAsset: ResolvedAsset | null
+  channel?: FullChannelFieldsFragment
+}
+
+type FormMethods = {
```
drillprop: I don't like that we need to pass these methods like this. Maybe we could pass some methods like `onUploadAvatar` or `onUploadCover`(maybe you could do something like `onUploadAssets` instead of these two) and then `onTxSync` instead? And all the react-hook-form calls do on the component level? Wdyt?

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
```diff
@@ -38,73 +48,224 @@ import {
...
 type UploadStatusProps = {
   isLast?: boolean
   asset: AssetUpload
-  size?: UploadStatusGroupSize
+  size?: 'compact' | 'large'
 }
 
+const FILE_NAME_LENGTH_LIMIT = 24
+
 export const UploadStatus: FC<UploadStatusProps> = ({ isLast = false, asset, size }) => {
```
drillprop, line 54: Ok, this file is super hard to read, every function has at least 100 lines and nested ifs. I'm a little worried that this will be very bug-prone, and hard to manage. Is there any room for improvement here? Could we make some effort to clean this up a bit?

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
```diff
@@ -38,73 +48,224 @@ import {
...
+              removeAssetFromUploads(asset.id)
+            },
+          },
+          secondaryButton: {
+            text: 'Cancel',
+            onClick: () => closeConfirmationModal(),
+          },
+        })
       } else {
         startFileUpload(
```
drillprop: What will happen if we pass just `asset` here?

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
```diff
@@ -159,17 +312,161 @@ export const UploadStatus: FC<UploadStatusProps> = ({ isLast = false, asset, siz
           id: asset.parentObject.id,
         },
         type: asset.type,
+        name: asset.name,
```
drillprop: same here

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.tsx
```diff
@@ -159,17 +312,161 @@ export const UploadStatus: FC<UploadStatusProps> = ({ isLast = false, asset, siz
...
+          secondaryButton: {
+            text: 'Cancel',
+            onClick: () => closeConfirmationModal(),
+          },
+        })
+      } else {
+        await handleUpdate()
+      }
     } else {
       startFileUpload(
```
drillprop: same here

drillprop (changes requested):
I see some layout shift, when there are two files in the upload status.  
image
Could be caused by the long name, but not sure

packages/atlas/src/components/_overlays/ImageCropModal/ImageCropModal.tsx
rafalpawlow: It's related to this view https://www.figma.com/file/Pe1Y0nrl4mpHWs7AkuQOk5/My-uploads-page?node-id=1115%3A7968

kdembler:
Just did a quick test of this and after selecting different video file and uploading it, my old video wasn't modified but actually a new empty video was created with the new video file. That's not desired, we should only update the existing videos from my uploads

kdembler:
Also if my thumbnail upload fails and then I select the same file, I still get the "Different file was selected". Seems the crop isn't automatic for files that we have crop data saved for, so it produces a different image

kdembler:
Needs rebase and change of target to `dev`

drillprop (changes requested):
Unfortunately found another issue. 
1. Have two failed assets(thumbnail and video). 
image
2. Click select and select a video
3. Cancel transaction in polkadot
4. Video file disappear from upload status 
image

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.hooks.ts
```diff
@@ -0,0 +1,280 @@
...
+        title: 'Continue with a different file?',
+        description: `This file (${
+          name.length > FILE_NAME_LENGTH_LIMIT ? shortenString(name, 16, 8) : name
+        }) is different from the one you selected before. To upload it, you’ll need to sign a blockchain transaction to confirm editing your video. Are you sure you want to continue?`,
+        type: 'warning',
+        primaryButton: {
+          text: 'Continue',
+          onClick: async () => {
+            closeConfirmationModal()
+            await cb()
```
drillprop: Is this `async/await` redundant here?

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.hooks.ts
```diff
@@ -0,0 +1,280 @@
...
+      openConfirmationModal({
+        fee,
+        title: 'Upload different file?',
+        description: `This file (${
+          name.length > FILE_NAME_LENGTH_LIMIT ? shortenString(name, 16, 8) : name
+        }) is different from the one you selected before. To upload it, you’ll need to sign a blockchain transaction to confirm editing your video. Are you sure you want to continue?`,
+        type: 'warning',
+        primaryButton: {
+          text: 'Upload file',
+          onClick: async () => {
```
drillprop: same here?

packages/atlas/src/views/studio/MyUploadsView/UploadStatus/UploadStatus.styles.ts
```diff
@@ -123,24 +119,16 @@ type FileInfoProps = {
...
 
 export const FileInfo = styled.div<FileInfoProps>`
-  display: grid;
   align-items: center;
   width: 100%;
   max-width: 600px;
-  margin-left: ${sizes(2)};
+  margin-left: ${sizes(4)};
   color: ${cVar('colorCoreNeutral300')};
+  row-gap: ${sizes(1)};
```
drillprop: Does this row-gap work here? You removed `display: grid`

drillprop (changes requested):

drillprop:
Another unexpected thing happened to me: 
image
I was re-uploading the video file for `Sample-5s`, and a new file jumped to another upload group.

drillprop (approved):
LGTM!

kdembler (approved):
Seems to work like a charm. Great job @rafalpawlow!

PR #3407, opened 2022-11-03
Figma export improvements

### LINKED ISSUE #3378: I can't generate icons from figma
I'm getting an error during saving icons. Tried a few times and with different Figma token

```
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^
AxiosError: connect ETIMEDOUT
```
To reproduce: 
1. run `atlas:svgs:icons`

drillprop:
Fix #3378 

Didn't fix the issue entirely, also everything seems to work fine for me today for some reason. Maybe the problem was on the Figma API side. Anyway, I added some improvements:  
- now it's possible to import only selected icons
- slightly better error handling 
- we're clearing svgs directory instead of removing it
- small adjustments
- generated new icons


rafalpawlow (approved):
Works really nice! 🙌

PR #3411, opened 2022-11-04
Validate user has enough funds for tx

### LINKED ISSUE #3222: Validate user has enough funds for tx
**AC:**

- [ ] Every time we would send a transaction, validate that the user has enough funds for transaction

---

Currently if the user doesn't have enough funds to pay for transaction fee, we will still let them initialize the tx and they will end up with a failed transaction and "not enough funds" message. We should check whether user has enough funds before we let them sign a transaction and display a message in the UI. What this needs to take into account, is the fact that different types of funds can be paid for different types of fees. So even though the free balance of a user may be 0, they can still send a trasaction using locked funds

Designs for minimized transactions already done:
> :warning: I wasn't able to test whether error snackbar components have been implemented to handle insufficient funds to cover transaction fees. kdembler I'm afraid you'll need to verify this yourself. It's the third bullet point [in this changelog](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=5739%3A263821).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202987390279959) by [Unito](https://www.unito.io)

rafalpawlow:
Closes #3222

### DIFF
--- a/packages/atlas/src/components/ActionBar/ActionBar.tsx
+++ b/packages/atlas/src/components/ActionBar/ActionBar.tsx
@@ -6,8 +6,8 @@ import { Fee } from '@/components/Fee'
 import { Text } from '@/components/Text'
 import { TooltipProps } from '@/components/Tooltip'
 import { ButtonProps } from '@/components/_buttons/Button'
+import { useFeeLoadingState } from '@/hooks/useFeeLoadingState'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
-import { useWaitForFee } from '@/hooks/useWaitForFee'
 import { transitions } from '@/styles'
 
 import {
@@ -42,7 +42,7 @@ export type ActionBarProps = {
 export const ActionBar = forwardRef<HTMLDivElement, ActionBarProps>(
   ({ fee, feeLoading, isActive = true, className, primaryButton, secondaryButton, infoBadge }, ref) => {
     const smMatch = useMediaMatch('sm')
-    const { actionHandler, loadingState } = useWaitForFee(!!feeLoading, fee, primaryButton.onClick)
+    const { signTransactionHandler, loadingState } = useFeeLoadingState(!!feeLoading, fee, primaryButton.onClick)
 
     return (
       <ActionBarContainer ref={ref} className={className} isActive={isActive}>
@@ -71,7 +71,7 @@ export const ActionBar = forwardRef<HTMLDivElement, ActionBarProps>(
         <ActionButtonPrimary
           {...primaryButton}
           disabled={primaryButton.disabled || loadingState}
-          onClick={actionHandler}
+          onClick={signTransactionHandler}
           secondaryButtonExists={!!secondaryButton}
           size={smMatch ? 'large' : 'medium'}
           type="submit"
--- a/packages/atlas/src/components/ActionBar/ActionBar.tsx
+++ b/packages/atlas/src/components/ActionBar/ActionBar.tsx
@@ -6,7 +6,7 @@ import { Fee } from '@/components/Fee'
 import { Text } from '@/components/Text'
 import { TooltipProps } from '@/components/Tooltip'
 import { ButtonProps } from '@/components/_buttons/Button'
-import { useFeeLoadingState } from '@/hooks/useFeeLoadingState'
+import { useHasEnoughBalance } from '@/hooks/useHasEnoughBalance'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
 import { transitions } from '@/styles'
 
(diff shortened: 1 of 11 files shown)


packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -199,13 +199,19 @@ export const VideoView: FC = () => {
         return false
       } else if (video?.id) {
         setVideoReactionProcessing(true)
-        const reacted = await likeOrDislikeVideo(video.id, reaction, video.title)
+        let fee
+        if (!reactionFee) {
+          fee = await getReactionFee([memberId || '', video?.id, reaction])
+        } else {
+          fee = reactionFee
+        }
```
drillprop: Is it the same as: 
```ts
const fee = reactionFee || (await getReactionFee([memberId || '', video?.id, reaction]))
```
?

packages/atlas/src/components/ActionBar/ActionBar.tsx
```diff
@@ -40,6 +42,7 @@ export type ActionBarProps = {
 export const ActionBar = forwardRef<HTMLDivElement, ActionBarProps>(
   ({ fee, feeLoading, isActive = true, className, primaryButton, secondaryButton, infoBadge }, ref) => {
     const smMatch = useMediaMatch('sm')
+    const { actionHandler, loadingState } = useWaitForFee(!!feeLoading, fee, primaryButton.onClick)
```
drillprop:
```suggestion
    const { actionHandler, loadingState } = useWaitForFee(feeLoading, fee, primaryButton.onClick)
```
I assume `feeLoading` is a boolean

packages/atlas/src/components/_comments/Comment/Comment.tsx
```diff
@@ -170,7 +174,15 @@ export const Comment: FC<CommentProps> = memo(
...
         setProcessingReactionsIds((previous) => [...previous, reactionId])
-        await reactToComment(commentId, video?.id || '', reactionId, comment?.author.handle || '')
+        let fee
+        if (!reactionFee) {
+          fee = await getReactToVideoCommentFee(
+            memberId && comment?.id ? [memberId, comment.id, reactionId] : undefined
+          )
+        } else {
+          fee = reactionFee
+        }
```
drillprop: Again, is it the same as: 
```ts
const fee = reactionFee || (await getSomeTxFee(argshere))
```
?

packages/atlas/src/hooks/useWaitForFee.ts
```diff
@@ -0,0 +1,47 @@
...
+  }, [callback, displaySnackbar, fee, totalBalance])
+
+  useEffect(() => {
+    if (loadingState && !feeLoading && totalBalanceLoaded) {
+      callbackHandler?.()
+      setLoadingState(false)
+    }
+  }, [callback, loadingState, feeLoading, totalBalanceLoaded, totalBalance, displaySnackbar, callbackHandler])
+
+  const actionHandler = useCallback(() => {
```
drillprop: this is really unintuitive name. If you see something like that in the component like `Button onClick={actionHandler}` this means nothing to you basically. Can we change the function name to something more associated with `fee`?

packages/atlas/src/hooks/useWaitForFee.ts
```diff
@@ -0,0 +1,47 @@
+import BN from 'bn.js'
+import { useCallback, useEffect, useState } from 'react'
+
+import { useSubscribeAccountBalance } from '@/providers/joystream/joystream.hooks'
+import { useSnackbar } from '@/providers/snackbars'
+
+export const useWaitForFee = (feeLoading: boolean, fee?: BN, callback?: () => void) => {
```
drillprop: Maybe that's just me and I'm nitpicking here, but the name suggests something different, I would expect that I got fee from this hook. 

Also - I'm not sure if we talk about this during our call, but is it possible to combine this hook with the `useFee` hook? It looks like we're passing `fee` and `feeLoading`, which also feels counter-intuitive. I'm suggesting that we could get a fee from this hook as well. Wdyt?

drillprop (changes requested):

packages/atlas/src/components/ActionBar/ActionBar.tsx
rafalpawlow: But it may be undefined here

packages/atlas/src/hooks/useWaitForFee.ts
rafalpawlow: > Also - I'm not sure if we talk about this during our call, but is it possible to combine this hook with the useFee hook? It looks like we're passing fee and feeLoading, which also feels counter-intuitive. I'm suggesting that we could get a fee from this hook as well. Wdyt?

I don't think it's a good idea. `ActionBar` is a good example, now waiting for a fee is handled only there, after your change we would have to handle it in every place where `ActionBar` is used. Let's call each other if something is unclear.

kdembler:
Needs rebase and change of target to `dev`

drillprop (approved):

PR #3419, opened 2022-11-07
Update maintenance copy, remove apple-touch-icon link

drillprop:


packages/atlas/src/Maintenance.tsx
```diff
@@ -2,22 +2,37 @@ import styled from '@emotion/styled'
...
+    <StyledSvgEmptyStateIllustration width={240} height={152} />
     <Header as="h1" variant="h500">
-      We’ll be right back!
+      We will be up & running soon
     </Header>
     <Text as="p" variant="t200" color="colorText">
-      {atlasConfig.general.appName} is currently undergoing maintenance work <br /> in order to ensure the best features{' '}
-      <br /> and performance for our users. <br /> <br />
-      We should be back shortly. <br /> See you soon!
+      ${atlasConfig.general.appName} is currently unavailable due to the network upgrade, new infrastructure release
```
kdembler: You don't need this leading `$`, it's only needed inside backticks string template (`)

kdembler: Currently it will be included in the final string ("$Atlas")

drillprop: yeah, that was my omission, fixed.

kdembler (approved):

PR #3430, opened 2022-11-09
Validate user has enough funds for tx

rafalpawlow:
Because of the bug in this condition `if (feeLoading || totalBalanceLoaded)` https://github.com/Joystream/atlas/pull/3430/files#diff-0a360fad9db35e67a16a7df5e43d47aabeafaf49f153d79581760bbb0d426f07R36 you weren't able to process the transaction because `loadingState` was set to infinity when `totalBalanceLoaded` was true.

### DIFF
--- a/packages/atlas/src/components/ActionBar/ActionBar.tsx
+++ b/packages/atlas/src/components/ActionBar/ActionBar.tsx
@@ -6,8 +6,8 @@ import { Fee } from '@/components/Fee'
 import { Text } from '@/components/Text'
 import { TooltipProps } from '@/components/Tooltip'
 import { ButtonProps } from '@/components/_buttons/Button'
+import { useFeeLoadingState } from '@/hooks/useFeeLoadingState'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
-import { useWaitForFee } from '@/hooks/useWaitForFee'
 import { transitions } from '@/styles'
 
 import {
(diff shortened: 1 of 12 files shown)


kdembler (changes requested):
1. If I try to create channel, seems the handler is called 2 times and I get a message about multiple transactions waiting for signature. The first time I try to submit, I actually get 2 signature requests, later I only get one, but still see the message about outstanding transactions.

https://user-images.githubusercontent.com/12646744/200847574-02702fdd-a147-4696-83f7-fbfc329dd154.mp4

2. If channel create validation fails (too long description) I still get "Not enough funds" snackbar even though the form should not be submitted at all:

https://user-images.githubusercontent.com/12646744/200847902-3e8e6aa4-75d1-4654-97cc-622161ff38bf.mp4

3. If I don't have enough funds to create a channel, I get a snackbar about funds, but the transactions still proceeds:

https://user-images.githubusercontent.com/12646744/200848745-a6b5711a-a7d2-444a-8e55-6d4f20fbb15f.mp4

kdembler (changes requested):
Seems the channel create validation still doesn't work properly.

In this recording I don't have enough funds to create a channel but I don't get any error message and can proceed:

https://user-images.githubusercontent.com/12646744/201646069-4f0b7883-e1e0-4e21-9bad-d6144b9a52ea.mp4

kdembler (approved):

PR #3437, opened 2022-11-15
Ypp integration

### LINKED ISSUE #3154: Integrate YPP authorization with YPP backend
**Scope:**

- [ ] Do pre-validation once user comes back from Google (send only `code`) so that we can show them failed criteria earlier
- [ ] Send all the data to YPP backend (google auth `code`, email, referrer ID, etc)
- [ ] Handle success
- [ ] Handle failure

drillprop:
Fix #3154 
Designs: https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1774%3A111432&t=5GTv7uxnMbB4DHrC-4

~I'm not sure if it's possible to implement [this user story](https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1818%3A90535&t=5GTv7uxnMbB4DHrC-4) for now. I think we should get some error backend in this case, right?~

### DIFF
--- a/packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
+++ b/packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
@@ -70,7 +70,9 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
   const hasMoreThanOneChannel = unSyncedChannels && unSyncedChannels.length > 1
   const [finalFormData, setFinalFormData] = useState<FinalFormData | null>(null)
   const selectedChannelId = useYppStore((store) => store.selectedChannelId)
+  const referrerId = useYppStore((store) => store.referrerId)
   const setSelectedChannelId = useYppStore((store) => store.actions.setSelectedChannelId)
+  const setReferrerId = useYppStore((store) => store.actions.setReferrerId)
 
   const smMatch = useMediaMatch('sm')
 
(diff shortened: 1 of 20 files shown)


packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -145,18 +339,28 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({ currentS
...
           additionalSubtitleNode: (
-            <AdditionalSubtitle>
-              <Text variant="h400" as="span">
+            <AdditionalSubtitleWrapper>
+              <Text variant={smMatch ? 'h400' : 'h300'} as="h3" margin={{ bottom: 4 }}>
+                Transaction fee
+              </Text>
+              <Text variant="t200" as="p" color="colorText" margin={{ bottom: 6 }}>
+                Applying for the program requires requires a blockchain transaction, which comes with a fee of{' '}
+                <NumberFormat as="span" variant="t200" value={updateChannelFee} withToken /> . Transaction fees are
```
kdembler: Unnecessary space before the dot

```suggestion
                <NumberFormat as="span" variant="t200" value={updateChannelFee} withToken />. Transaction fees are
```

kdembler (commented):
Found some small stuff but otherwise looks good I think

1. My user register call to YPP backend failed with the below data. The error wasn't handled in any way by the UI and I got back to "Requirements step". We need to show some error message in that case. I think this happened because of mismatch between frontend and backend Google Console ID
```json
{"statusCode":400,"message":"Could not get User's access token using authorization code Error: unauthorized_client","error":"Bad Request"}
```
2. We probably should clear channel requirements errors when the user switches channel. Currently if one of my channels fails validation and I immediately try another one, it will "inherit" previous channel's errors

https://user-images.githubusercontent.com/12646744/203379758-34945e2b-897e-4c94-934c-fda8dba52d78.mp4

kdembler:
Re 2 from my review: actually the error shouldn't be associated with Joystream channel but YouTube channel so maybe we just clear it when the users clicks "Back" instead?

PR #3452, opened 2022-11-24
Ypp studio views

### LINKED ISSUE #3428: Adapt UI based on YPP channel status
**Scope:**

- [ ] Fetch channel's YPP status when user enters Studio
- [ ] If the user is YPP participant, show YPP dashboard in Studio, otherwise show YPP LP
- [ ] Show real data in YPP dashboard (channel tier)
- [ ] Implement additional states for YPP LP, according to this Figma story - https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1774%3A111432
- [ ] Automatically trigger new channel creation after users logs in with YPP LP
- [ ] Show "Connect with YouTube?" modal after users creates JS channel

drillprop:
Fixes #3428

### DIFF
--- a/packages/atlas/src/providers/transactions/transactions.manager.tsx
+++ b/packages/atlas/src/providers/transactions/transactions.manager.tsx
@@ -7,6 +7,7 @@ import { ExtrinsicStatus } from '@/joystream-lib/types'
 import { usePersonalDataStore } from '@/providers/personalData'
 import { useSnackbar } from '@/providers/snackbars'
 import { useUserStore } from '@/providers/user/user.store'
+import { useYppStore } from '@/providers/ypp/ypp.store'
 import { SentryLogger } from '@/utils/logs'
 
 import { METAPROTOCOL_SNACKBAR_ID } from './transactions.config'
@@ -19,6 +20,8 @@ export const TransactionsManager: FC = () => {
     showFirstMintDialog,
     actions: { removeOldBlockActions, setShowFistMintDialog, removeTransaction },
   } = useTransactionManagerStore((state) => state)
+  const shouldContinueYppFlow = useYppStore((state) => state.shouldContinueYppFlow)
+  const { setShouldContinueYppFlow, setShowConnectToYoutubeDialog } = useYppStore((state) => state.actions)
 
   const updateDismissedMessages = usePersonalDataStore((state) => state.actions.updateDismissedMessages)
   const userWalletName = useUserStore((state) => state.wallet?.title)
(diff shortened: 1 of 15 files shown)


packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -294,6 +300,27 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
 
   const authorizationStep = useMemo(() => {
     switch (currentStep) {
+      case 'connect-with-youtube':
+        return {
+          headerIcon: (
+            <LogosWrapper>
+              <StyledSvgLogoYoutube />
+              <StyledSvgControlsConnect />
+              {/* todo use AppLogo component here */}
```
kdembler: Do we need this comment still?

packages/atlas/src/providers/transactions/transactions.manager.tsx
```diff
@@ -97,13 +100,23 @@ export const TransactionsManager: FC = () => {
...
   }
 
+  const handleTransactionModalClose = () => {
+    if (firstNonMinimizedTransaction) {
+      removeTransaction(firstNonMinimizedTransaction.id)
+    }
+    if (shouldContinueYppFlow) {
+      setShouldContinueYppFlow(false)
+      setShowConnectToYoutubeDialog(true)
+    }
```
kdembler: Do we really need to make TransactionManager aware of YPP stuff? That seems like unnecessary coupling tbh. Couldn't we stick that logic as a callback after creating channel? I think that's the only time we want to display this youtube dialog?

In general, I'm not sure why we need all this additional state, couldn't we just show the youtube dialog always when the user creates a channel?

drillprop: > In general, I'm not sure why we need all this additional state, couldn't we just show the youtube dialog always when the user creates a channel? 

Hmm, yeah, that's a good idea. But how would I open this dialog, without redirecting to YPP landing view? That means I should always redirect a user to the Ypp landing view when he creates a channel? I'm confused how this flow should looks like. 

> Do we really need to make TransactionManager aware of YPP stuff? That seems like unnecessary coupling tbh. Couldn't we stick that logic as a callback after creating a channel? I think that's the only time we want to display this youtube dialog?

So I put a logic here because I want to show dialog **after** transaction modal is closed. Without this, the user will experience two opened modals for two seconds, and the transaction dialog will close itself behind the "Connect to youtube dialog". This is not a great experience I think.  I guess I could move this to YppLandingView and just add the same delay, as we have in TransactionModal: 
```
  useEffect(() => {
    if (status === ExtrinsicStatus.Completed) {
      setTimeout(() => {
        onClose()
      }, 2000)
    }
  }, [status, onClose])
```

WDYT? Or maybe I should just ignore this delay and don't bother about two dialogs opened at once?

drillprop: hmm, I guess this `Connect to youtube` dialog could be outside the Ypp authorization modal and the flow will look like this: 
1. User creates the channel. 
2. After the transaction is done. We wait two seconds for the closing transaction modal and then we're opening the `Connect to youtube` dialog 
3. User agrees to connect to youtube, and gets redirected to `studio/ypp`(this part I really don't like). Immediately after that first step of the authorization modal is showing up
4. User continues the flow. 

Let me know what you think. I would need to do a small refactor, and I need to know that we're on the same page, before I start doing that.

kdembler: yes, that was my thinking, I don't think `Connect to youtube` needs to be part of ypp sign up flow

kdembler (approved):
LGTM

Issue #3055, opened 2022-08-01
Payout Transactions: Totals Earned and Withdrawn Widgets


dmtrjsg:
## Context
Originating Epic

* https://github.com/Joystream/atlas/issues/2401

## Design
Link to Figma > [LINK](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=1552%3A80425)

![](https://user-images.githubusercontent.com/93924525/182211656-e05db10e-1893-4609-89d4-b7ad83122a8a.png)

## Scope
* [ ] Create a component to host widgets below
* [ ] Total Earned (this probably comes from the payout data file)
* [ ] Total WIthdrawn (this should be an aggregate of events from QN **OR** query with RPC)
* [ ] Fetch the data in the view (if available)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1202701571107659) by [Unito](https://www.unito.io)

PR #3471, opened 2022-12-01
🤑  My payments ➡️ Transactions

### LINKED ISSUE #3056: Payouts: Fetch historical data
## Context
Originating Epic

* https://github.com/Joystream/atlas/issues/2401

## Design
Link to Figma > [LINK](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=1552%3A80425)

## Scope

- [ ] Fetch the data from QN and display in the table (done in https://github.com/joystream/atlas/issues/3417)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1202701622036430) by [Unito](https://www.unito.io)

WRadoslaw:
fixes #3056 #3055


packages/atlas/src/views/studio/MyPaymentsView/PaymentsTransactionTab/PaymentTransactionsTab.styles.ts
```diff
@@ -0,0 +1,13 @@
+import styled from '@emotion/styled'
+
+import { media, sizes } from '@/styles'
+
+export const TilesWrapper = styled.div`
+  display: grid;
+  gap: ${sizes(4)};
+  margin-bottom: ${sizes(4)};
```
drillprop, line 8: I just noticed that the margin and the gap are different on `md`. It should be 24px not `16 px`.  I did the same mistake in `Overview tab`. Could you fix it there as well? 
https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments-page?node-id=1505%3A17428&t=gGHo1JzBlZspPmLd-4

packages/atlas/src/views/studio/MyPaymentsView/PaymentsTransactionTab/PaymentTransactionsTab.tsx
```diff
@@ -0,0 +1,62 @@
...
+          text={formatNumber(hapiBnToTokenNumber(paymentHistoryOverview.totalEarned))}
+          loading={loading}
+          caption={
+            <StyledNumberFormat
+              variant="t100"
+              as="p"
+              format="dollar"
+              value={convertHapiToUSD(paymentHistoryOverview.totalEarned) ?? 0}
+              margin={{ top: 1 }}
+            />
```
drillprop: As far as I know, we won't be supporting converting joy to USD for some time. The designs are quite old for these views. Let's remove the caption for now.

packages/atlas/src/views/studio/MyPaymentsView/PaymentsTransactionTab/PaymentTransactionsTab.tsx
```diff
@@ -0,0 +1,62 @@
...
+          loading={loading}
+          text={formatNumber(hapiBnToTokenNumber(paymentHistoryOverview.totalWithdrawn))}
+          caption={
+            <StyledNumberFormat
+              variant="t100"
+              as="p"
+              format="dollar"
+              value={convertHapiToUSD(paymentHistoryOverview.totalWithdrawn) ?? 0}
+              margin={{ top: 1 }}
+            />
```
drillprop: same here

packages/atlas/src/views/studio/MyPaymentsView/PaymentsTransactionTab/PaymentTransactionsTab.tsx
```diff
@@ -0,0 +1,62 @@
...
+          }
+          icon={<SvgJoyTokenMonochrome24 />}
+        />
+      </TilesWrapper>
+      <TablePaymentsHistory data={paymentData} />
```
drillprop: The same applies to table

packages/atlas/src/views/studio/MyPaymentsView/PaymentsTransactionTab/utils.ts
```diff
@@ -0,0 +1,41 @@
+import BN from 'bn.js'
```
drillprop: I would name this file `PaymentTransactionTab.utils.ts`. I know that we still have some files called just `utils`, but we decided one day, that if something is "component specific" we're naming this `ComponentName.utils`. We already have `Comment.utils` or `NftForm.utils`

drillprop (changes requested):
Works great! Found only a few minor things

packages/atlas/src/components/TablePaymentsHistory/TablePaymentsHistory.tsx
```diff
@@ -38,13 +37,16 @@ const COLUMNS: TableProps['columns'] = [
 
 type PaymentType = 'nft-sale' | 'nft-royalty' | 'claimed-reward' | 'withdrawal' | 'ypp-reward'
 
+export interface PaymentHistory {
```
kdembler: We're using `type`s instead of `interface`s

packages/atlas/src/components/TablePaymentsHistory/TablePaymentsHistory.tsx
```diff
@@ -58,7 +60,7 @@ export const TablePaymentsHistory: FC<TablePaymentsHistoryProps> = ({ data }) =>
       })),
     [data]
   )
-  return <Table title="History" columns={COLUMNS} data={mappedData} />
+  return <Table title="History" columns={COLUMNS} data={mappedData} minWidth="750px" />
```
kdembler: Why do we need to pass width as a prop?
1. Table should probably just take full width and the rest should be handled by  the layout
2. If point above is not true for some reason, we probably should allow passing `className` to `Table` and just provide this style via styled components

packages/atlas/src/views/studio/MyPaymentsView/MyPaymentsView.tsx
```diff
@@ -8,6 +8,7 @@ import { useHeadTags } from '@/hooks/useHeadTags'
 import { useMountEffect } from '@/hooks/useMountEffect'
 import { useConfirmationModal } from '@/providers/confirmationModal'
 import { usePersonalDataStore } from '@/providers/personalData'
+import { PaymentTransactionsTab } from '@/views/studio/MyPaymentsView/PaymentsTransactionTab'
```
kdembler: This should be relative import

packages/atlas/src/views/studio/MyPaymentsView/PaymentsTransactionTab/PaymentTransactionsTab.hooks.ts
```diff
@@ -0,0 +1,71 @@
+import { useEffect, useState } from 'react'
+
+import { GetFullChannelsQuery, useGetChannelPaymentEventsQuery } from '@/api/queries/__generated__/channels.generated'
+import { PaymentHistory } from '@/components/TablePaymentsHistory'
+import { useJoystream } from '@/providers/joystream/joystream.hooks'
+import { mapEventToPaymentHistoryFactory } from '@/views/studio/MyPaymentsView/PaymentsTransactionTab/PaymentTransactionsTab.utils'
```
kdembler: This should be relative import

packages/atlas/src/components/TablePaymentsHistory/TablePaymentsHistory.tsx
WRadoslaw: Good point, thanks!

drillprop (changes requested):
Unfortunately, found another issue. 
image
These sales were done on a different channel(that belongs to the same member), so those shouldn't be displayed on my channel. 

The problem is, that due to query node limitations, we can't pass channel id to for example `nftBoughtEvents`. Maybe we should get the channel id from the query: 
```
query {
    nftBoughtEvents(where: {ownerMember: {id_eq: $ownerMemberId} }) {
    inBlock
    video {
      channelId
    }
    createdAt
    price
  }
}
```
 and filter this on the client? I don't know, I don't see a clean solution for this.

WRadoslaw:
> Unfortunately, found another issue. image These sales were done on a different channel(that belongs to the same member), so those shouldn't be displayed on my channel.
> 
> The problem is, that due to query node limitations, we can't pass channel id to for example `nftBoughtEvents`. Maybe we should get the channel id from the query:
> 
> ```
> query {
>     nftBoughtEvents(where: {ownerMember: {id_eq: $ownerMemberId} }) {
>     inBlock
>     video {
>       channelId
>     }
>     createdAt
>     price
>   }
> }
> ```
> 
> and filter this on the client? I don't know, I don't see a clean solution for this.

Nice catch! I don't see another solution either, since Hydra doesn't allow that deeply nested parameter.

WRadoslaw:
@drillprop Should be good now 🤞

WRadoslaw:
@drillprop I've reverted the changes as discussed on planning call, so this should be good to go

drillprop (approved):

drillprop:
Created #3505 #3504 as followup

PR #3472, opened 2022-12-01
🦅  Expose privacy policy to config

### LINKED ISSUE #3470: Create Own Privacy Policy Page for Atlas codebase/ Gleev + Update Terms of Service
## Context

1️⃣ 
Currently Privacy policy is linked to joystream.org master policy (link from attached img). This was okay when Atlas was hosted on the sub-domain but with launch of Gleev the GDPR requirements ask for each APP to have its own clearly accessible Privacy policy. Moreover external App operators may choose to modify/ have their own privacy policy. 

Screenshot 2022-12-01 at 14 57 53

Suggested explicit link to be added to the bottom of the extended top nav menu:

Screenshot 2022-12-01 at 15 05 00

2️⃣  Update respective parts of Terms of Service for Gleev to focus on Gleev (replace Atlas mentions with dynamic values)

Screenshot 2022-12-01 at 15 06 45

## Scope
1️⃣  Privacy Policy
- [ ] Create Privacy policy page for Gleev gleev.xyz/privacy-policy
- [ ] Add link to Privacy Policy to the bottom of the menu
- [ ] Copy content from joystream.org/privacy-policy/
- [ ] ⚠️ Make Privacy policy easily customisable for External App operators

2️⃣  Terms of Service
- [ ] Update "Atlas" to [app name] "Gleev" for JSG run instance
- [ ] Update hosted at "atlas.joystream.org" to  [app dmain] and "gleev.xyz" for JSG run instance
- [ ] ⚠️  hese have to be easily configurable for External App Operators

WRadoslaw:
partially fixes #3470 

⚠️ Still requires changes on gleev fork ⚠️


drillprop (changes requested):
Looks good, but after a quick chat on the discord, we decided to add markdown support.

kdembler:
@WRadoslaw Let's mention this change in `CHANGELOG.md` 🙏

packages/atlas/src/views/legal/PrivacyPolicy.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import { atlasConfig } from '@/config'
+import { useHeadTags } from '@/hooks/useHeadTags'
+
+export const PrivacyPolicyView: FC = () => {
+  const headTags = useHeadTags('Privacy Policy')
+  return (
+    <>
+      {headTags}
+      <ReactMarkdown>{atlasConfig.legal.privacyPolicy}</ReactMarkdown>
```
kdembler: Let's add some wrapper component for `ReactMarkdown` to use within the app

kdembler (changes requested):
There's still an open question on what we do with markdown styling. I see 2 ways forward:

1. We resolve this in this PR, so make sure that rendered styles are matching the designs
2. We use default styling for now and create a followup issue to update styling later in the future

Since this styling is only about those legal views, it's not the biggest priority. Leaving the choice up to you though

packages/atlas/src/components/MarkdownPreview/MarkdownPreview.tsx
```diff
@@ -0,0 +1,18 @@
+import { Global } from '@emotion/react'
+import { FC } from 'react'
+import ReactMarkdown from 'react-markdown'
+
+import { MarkdownGlobalStyles } from '@/components/MarkdownPreview/MarkdownPreview.styles'
```
kdembler: Let's make this import relative 🙏

packages/atlas/src/components/MarkdownPreview/MarkdownPreview.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+type MarkdownPreviewProps = {
+  markdown: string
+}
+
+export const MarkdownPreview: FC<MarkdownPreviewProps> = ({ markdown }) => {
+  return (
+    <>
+      <ReactMarkdown className="markdown-preview">{markdown}</ReactMarkdown>
+      <Global styles={MarkdownGlobalStyles} />
```
kdembler: I think we could just do this instead?

```tsx
const StyledReactMarkdown = styled(ReactMarkdown)`
& h1 {
  // ...
}
`
```

This has 2 benefits:
1. We don't need to keep track of `markdown-preview` class name (we could make typo for example), since it will be handled by emotion
2. We don't need to inject global styles

kdembler (changes requested):

packages/atlas/src/components/MarkdownPreview/MarkdownPreview.tsx
WRadoslaw: Unfortunately, `emotion` doesn't accept `ReactMarkdown` in `styled`

WRadoslaw: To elaborate it works but requires `@ts-ignore` comment

drillprop (commented):
I did some testing and I think we could use react-markdown in a different way. Instead of creating some global styles, which could be bug-prone, we could make use of react-markdown prop called `components`: 

```tsx
  <ReactMarkdown
        components={{
          h1: ({ children }) => (
            <Text as="h1" variant="h400" margin={{ bottom: 4 }}>
              {children}
            </Text>
          ),
          p: ({ children }) => {
            return (
              <Text as="p" variant="t300" color="colorText" margin={{ bottom: 4 }}>
                {children}
              </Text>
            )
          },
        }}
      >
        {markdown}
      </ReactMarkdown>
```
The advantages of this approach would be 
- we will always use our predefined `Text` component.
- no need to create global styles
- we could replace the HTML tag of the element(for example h1 to h2, to not have multiple h1 tags on the site)
- in the designs, we're applying different margins based on if the element is last or first in the document.  Check [here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=10059%3A184916&t=AmtBK4WmcUpfoLtL-4). We could actually figure out which tag element is last or first, by reading `index` and `sibilingCount` parameter like this:
```tsx 
  <ReactMarkdown
        includeElementIndex
        components={{
          p: ({ children, siblingCount, index }) => {
            const isLastElement = siblingCount === index + 1
            return (
              <Text as="p" variant="t300" color="colorText" margin={{ bottom: 4 }}>
                {children}
              </Text>
            )
          },
        }}
      >
        {markdown}
      </ReactMarkdown>
```

Wdyt @kdembler @WRadoslaw ? 
I will also be fine if we keep what we have now. We could create an issue and refactor this later. As far as I know, we will need this `Markdown container` later when we start implementing CRT tokens, so we will need to update this anyway.

WRadoslaw:
I have no preference tbh. It can be styled by emotion to avoid using global styles and margins can be added there as well. About bug prone, I can't think of any danger here (Pioneers markdown is styled by CSS since the beginning and I think there wasn't any problem with it).

kdembler:
I like @drillprop approach more tbh, let's go with that

kdembler (approved):
LGTM, wanna take a look as well @drillprop

drillprop (approved):

drillprop (approved):
Looks good to me as well, thanks for refactoring this 👍🏻

Issue #3123, opened 2022-08-16
Improve social preview for Atlas mainpage and YPP landing page


kdembler:
### Scope

- [x] Make sure generated social previews for channel and video page match designs
- [x] Update meta tags in static `index.html` to match new homepage open graph designs
- [ ] Add social preview for `/ypp` route (probably via atlas-meta-server)

---

![](https://user-images.githubusercontent.com/12646744/190399944-851877af-8c84-4862-bf5d-c3701005e3ff.png)

I think we should improve social preview for Atlas mainpage and YPP landing pages when shared externally for marketing on social media.. It would be nice to add some kind of an image to those previews and potentially have separate description for YPP landing page (worth discussing)

➕ 

- [ ] Joystream.org 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203003511974838) by [Unito](https://www.unito.io)

toiletgranny:
@kdembler, @dmtrjsg Open graph images for Atlas and YPP landing page can be found here:
- [Atlas (in general)](https://www.figma.com/file/Cc3VDoK6qglJ617ChA2EMr/Foundation?node-id=1513%3A10143),
- [YouTube Partner Program](https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=2568%3A80850).

Please, let me know if you have any thoughts and would like to experiment with different visuals. But before you export the images or update other OG meta tags, allow me to put this issue into the "On hold" column while I take care of other issues involving Atlas branding (https://github.com/Joystream/atlas/issues/3276 and https://github.com/Joystream/marketing/issues/293). Once we have a new logo for Gleev and Vinto, I'll swap logos in the above images.

![image](https://user-images.githubusercontent.com/35307309/193283942-0e184d1b-9536-4f74-a5f0-7cd856a60fa2.png)

---

FYI, I've also gone ahead and made Figma open graph templates for us to use going forward. From now on, you guys can expect having `🌐 Open graph` page in future Figma files, whenever there's an opportunity for having a custom preview and description (like with the YPP LP). Based on the current implementation of OG previews for the video and channel pages, I've updated Figma files to reflect the current state.

- [Video page](https://www.figma.com/file/sMydmZJlzhK91FTIPOd7D2/Video-page?node-id=6281%3A296705),
[...]

PR #3485, opened 2022-12-06
improve static OpenGraph tags generation

kdembler:
Part of #3123

**Scope of changes:**
- Added a couple of additional OpenGraph tags, including an `og:image` for nice previews
- All of the OpenGraph tags are now also statically applied to `index.html` file so that linking to it will offer a richer preview, without any additional generation
- Added a Vite plugin so that `index.html` is populated with meta tags automatically during build. This way we can just edit code in atlas-meta-server and proper tags will always be applied to HTML, without need of manual sync

---

Before
CleanShot 2022-12-06 at 13 12 34@2x

After
CleanShot 2022-12-06 at 13 12 14@2x


drillprop (approved):
LGTM!

PR #3486, opened 2022-12-06
improve atlas-meta-server tag generation

kdembler:
Part of #3123

**Scope of changes:**
- `atlas-meta-server` will no longer return an empty `index.html` with only `<meta>` tags. After this change when generating a preview, meta-server will fetch latest `index.html` from the app URL and populate it with meta tags. This way, previews generated with the server still contain valid `index.html` with the full application
- `atlas-meta-server` will no longer require explicitly passing `APP_NAME`, `GRAPHQL_URL` (Orion URL) and `TWITTER_ID` env variables. Instead, those values are extracted from the fetched `index.html`. This way meta-server won't require an update any time those values change
- On startup, meta-server will fetch `index.html` to make sure it contains all the necessary info and quit if it doesn't. It will also test connection to Orion


WRadoslaw (approved):
lgtm

drillprop (approved):
LGTM as well
I'm wondering why git thinks that there were so many line changes and not so many deletions. We already had an introspection schema, right?

kdembler:
> I'm wondering why git thinks that there were so many line changes and not so many deletions. We already had an introspection schema, right?

It hasn't been updated in a while though. We've updated graphql-codegen since last time, maybe it produces more output now

PR #3489, opened 2022-12-06
💳 Support for mobile Subwallet 

### LINKED ISSUE #3431: Add support for mobile SubWallet wallet
Currently our mobile version of wallet connect flow only supports Polkawallet. Recently SubWallet reached out to us that their mobile wallet is in beta and they would love to see us support it. Since we weren't very happy with Polkawallet, we should definitely check it out, as it can be much improved. Let's add support for it to Atlas.

---

Beside adding support for SubWallet, there is an additional issue - as it is, Atlas will only allow you to use "whitelisted" mobile wallet. What I mean is that if I currently use mobile SubWallet to go to Atlas, it won't be detected and I won't be able to use it. We should make sure that users are free to use whatever they want, without us adding any explicit support. So in the modal below, all the available wallets should be listed instead

image

WRadoslaw:
fixes #3431

### DIFF
--- a/packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalWalletStep.tsx
+++ b/packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalWalletStep.tsx
@@ -6,6 +6,7 @@ import { IconWrapper } from '@/components/IconWrapper'
 import { Loader } from '@/components/_loaders/Loader'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
 import { useMountEffect } from '@/hooks/useMountEffect'
+import { UnknownWallet } from '@/providers/user/user.helpers'
 import { useUser } from '@/providers/user/user.hooks'
 import { useUserStore } from '@/providers/user/user.store'
 import { isMobile } from '@/utils/browser'
@@ -26,7 +27,6 @@ const MOBILE_SUPPORTED_WALLETS = {
       alt: 'Polkawallet logo',
     },
   },
-  'subwallet-js': {},
 }
 
 export const SignInModalWalletStep: FC<SignInStepProps> = ({
@@ -45,7 +45,12 @@ export const SignInModalWalletStep: FC<SignInStepProps> = ({
   const wallets = useMemo(() => {
     const unsortedWallets = getWalletsList()
     if (isMobileDevice) {
-      const supportedWalletsNames = Object.keys(MOBILE_SUPPORTED_WALLETS)
+      const supportedWalletsNames = [
+        'polkawallet',
+        ...unsortedWallets
+          .filter((wallet) => wallet instanceof UnknownWallet || wallet.extensionName === 'subwallet-js')
+          .map((wallet) => wallet.extensionName),
+      ]
 
       return supportedWalletsNames.map((walletName) => {
         const possiblyInstalledWallet = unsortedWallets.find((wallet) => wallet.extensionName === walletName)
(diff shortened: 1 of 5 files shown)


kdembler (commented):
Let's also show any unknown wallets so we don't need to add explicit support

packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalWalletStep.tsx
```diff
@@ -35,27 +43,34 @@ export const SignInModalWalletStep: FC<SignInStepProps> = ({
...
-      return [
-        {
-          title: 'Polkawallet',
-          extensionName: POLKAWALLET,
+      const allMoblieWallets = new Set([
+        'polkawallet',
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ...Object.keys((window as any).injectedWeb3 || {}),
+        ...unsortedWallets
+          .filter((wallet) => wallet.extensionName !== 'talisman')
```
kdembler: Why are we filtering Talisman here?

kdembler (changes requested):
I think we should try to improve this list:

image

1. On the screenshot above, I don't have any extension available, why does `polkadot-js` show up on that list?
2. If the injected `polkadot-js` is a duplicate of something else already on the list, how could we identify that and remove the duplicate? Maybe there's some additional data on the signer object that we could use for matching?
3. The "Recommended" label currently gets applied to all entries in the list, I think the safest bet would be to apply it to SubWallet only (but only on mobile)

packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalWalletStep.tsx
WRadoslaw: Talisman doesn't have mobile wallet, but it would be shown as a Recommended choice. All because of using @talisman-connect/wallets to get "supported wallets", this list is only adjusted for the standard web user.

WRadoslaw:
> I think we should try to improve this list:
> 
> image
> 
> 1. On the screenshot above, I don't have any extension available, why does `polkadot-js` show up on that list?
> 2. If the injected `polkadot-js` is a duplicate of something else already on the list, how could we identify that and remove the duplicate? Maybe there's some additional data on the signer object that we could use for matching?
> 3. The "Recommended" label currently gets applied to all entries in the list, I think the safest bet would be to apply it to SubWallet only (but only on mobile)

Ad 1. This one is a bug - I will just filter it out same as Talisman
Ad 2. Not really from `injectedWeb3` we receive only the name and version. Although most of the time version between duplicates will be the same, there is possibility that another wallet will have same version and it will be treated as duplicate
Ad 3. Gotcha

I've added improvements, hopefully sufficient.

packages/atlas/src/components/_auth/SignInModal/SignInSteps/SignInModalWalletStep/SignInModalWalletStep.tsx
```diff
@@ -1,24 +1,32 @@
+import { Wallet } from '@talismn/connect-wallets'
 import { FC, useCallback, useEffect, useMemo, useState } from 'react'
 
 import { SvgActionNewTab, SvgAlertsError24, SvgAlertsInformative24, SvgLogoPolkadot } from '@/assets/icons'
-import polkaWalletLogo from '@/assets/images/polkawallet-logo.webp'
 import { IconWrapper } from '@/components/IconWrapper'
+import {
+  MOBILE_SUPPORTED_WALLETS,
+  walletSort,
+} from '@/components/_auth/SignInModal/SignInSteps/SignInModalWalletStep/SignInModalWalletStep.utils'
```
kdembler: Let's use relative import here

kdembler (approved):

PR #3512, opened 2022-12-09
🌞 Suggestions on YPP modal details

### LINKED ISSUE #3497: Channel autosuggest in the YPP auth flow

WRadoslaw:
fixes #3497


packages/atlas/src/components/_inputs/InputAutocomplete/InputAutocomplete.tsx
```diff
@@ -0,0 +1,89 @@
...
+  selectedItem,
+  perfectMatcher,
+  ...comboBoxProps
+}: InputAutocompleteProps<Q, V, R>) => {
+  const [result, setResult] = useState<Q | null>(null)
+  const [inputValue, setInputValue] = useState<string>('')
+  const [isLoading, setLoading] = useState(false)
+  const client = useApolloClient()
+
+  const debouncedFetchChannels = useRef(
```
drillprop: if this supposes to be generic, we need to change the name of the function

packages/atlas/src/components/_inputs/InputAutocomplete/InputAutocomplete.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+      try {
+        setLoading(true)
+        const data = await client.query<Q, V>({
+          query: documentQuery,
+          variables: queryVariablesFactory(val),
+        })
+        setResult(data.data)
+      } catch (e) {
+        SentryLogger.error('Failed to fetch channels', 'InputAutocomplete', e)
```
drillprop: Same applies here

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationSteps/YppAuthorizationDetailsFormStep/YppAuthorizationDetailsFormStep.tsx
```diff
@@ -143,38 +86,51 @@ export const YppAuthorizationDetailsFormStep: FC = () => {
...
-      </FormField>
+      <Controller
+        name="referrerChannelTitle"
+        control={control}
+        render={({ field: { onChange, value } }) => (
+          <FormField
+            optional
+            label="Referrer"
+            description="Enter the title of the Joystream channel which recommended the program to you."
+            error={errors.referrerChannelTitle?.message}
```
drillprop, line 105: if I as a user ignore the message `We couldn't find any result. Please check if the spelling is correct.`, and then I click continue I won't get any error. I could even go to the next step. I think the user shouldn't be able complete this step in this case. Previously it worked like that at least. 
image

packages/atlas/src/components/_inputs/InputAutocomplete/InputAutocomplete.tsx
```diff
@@ -0,0 +1,89 @@
...
+      onSelectedItemChange={onItemSelect}
+      onInputValueChange={(value) => {
+        setLoading(true)
+        setInputValue(value ?? '')
+        selectedItem && clearSelection()
+        debouncedFetchChannels.current(value)
+      }}
+    />
+  )
+}
```
drillprop, line 90: Can we get rid of this empty rectangle underneath the input?
image
I think it should be an easy fix - we could pass 
```
          <ListWrapper {...getMenuProps()} isOpen={isOpen && items.length > 0}>
```
Just make sure that it doesn't break anything.

drillprop (changes requested):

drillprop (approved):

PR #3520, opened 2022-12-12
Create video tile variation for synced video

### LINKED ISSUE #3448: Create video tile variation for "uploading in progress" state when done by Auto-sync
## Context

Autosync service will result in states of the app where multiple videos are fetched from newly joined creator YT channels, but upload is still in progress. Since the state will not be defined for the App this will result in "Video upload failed" variation of tile and toaster notification, which is not a great first experience for the creators signed up. 

We want to handle this case.

## Scope
- [x] Create video tile variation for the video upload in progress. Add a label/ badge that it is done via Auto-sync service from YPP programme
- [x] Add toaster notification variation "Video upload in progress"

drillprop:
partly fixes #3448 
Designs: https://www.figma.com/file/LF7LOMhOA46bRcYHWlw1Bq/My-videos-page?node-id=1801%3A46731&t=YVNPqpH7zO4jpepg-4
I only added a tile variant. If you want to test it visually go to `VideoTilePublisher.tsx` and replace line 94 with 
```tsx
    const uploadVideoStatus = useUploadsStore((state) => ({
      ...state.uploadsStatus[video?.media?.id || ''],
      lastStatus: 'yt-sync',
    }))
```
and go to `/studio/videos`

I didn't add the integration because it's currently impossible to get consistent information on which uploaded videos are actually synced from youtube and which are failed uploads.

### DIFF
--- a/packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
+++ b/packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
@@ -6,6 +6,7 @@ import { CSSTransition } from 'react-transition-group'
 import { getNftStatus } from '@/api/hooks/nfts'
 import { useFullVideo } from '@/api/hooks/video'
 import {
+  SvgActionDownload,
   SvgActionEdit,
   SvgActionHide,
   SvgActionReupload,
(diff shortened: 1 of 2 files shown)


kdembler:
@drillprop I think we will also need to handle this in uploads manager, otherwise the user will get a snackbar about upload that didn't finish. Unless that's part of some other issue

drillprop:
> @drillprop I think we will also need to handle this in uploads manager, otherwise the user will get a snackbar about upload that didn't finish. Unless that's part of some other issue

Let's deal with this once https://github.com/Joystream/youtube-synch/issues/88 is done. Created the issue https://github.com/Joystream/atlas/issues/3526

WRadoslaw (approved):
LGTM

PR #3522, opened 2022-12-12
update dependencies

kdembler:
**Scope of changes:**
- Updated `@joystream/types` to v0.20.5
- Updated to Vite v4
- Updated to Storybook v7
  - This simplifies our SB configuration
  - All the component props are now analyzed properly
  - Because of the above, removed a bunch of unnecessary `argTypes` entries from stories
- Updated all the other dependencies to latest versions
  - `zustand` is the only exception since it needs some migration, we can do that as a followup
  - Some new major versions needed some minor changes. Affected components: `<Datepicker />` and `<Combobox />`


packages/atlas/src/components/_inputs/ComboBox/ComboBox.tsx
```diff
@@ -102,19 +101,19 @@ export const ComboBox = <T extends unknown>(props: ComboBoxProps<T>) => {
...
-          {...getInputProps({ ref: textFieldRef })}
-          nodeEnd={processing && inputValue && <Loader variant="small" />}
-          nodeStart={<StyledSvgActionPlus />}
-          onFocus={(event) => {
-            textFieldProps?.onFocus?.(event)
-          }}
-          onClick={toggleMenu}
-        />
-      </div>
+      {/*<div {...getComboboxProps()}>*/}
```
drillprop: is this supposed to be removed?

packages/atlas/src/components/_inputs/ComboBox/ComboBox.tsx
```diff
@@ -102,19 +101,19 @@ export const ComboBox = <T extends unknown>(props: ComboBoxProps<T>) => {
...
+        error={error || !!noItemsFound}
+        {...getInputProps({ ref: textFieldRef })}
+        nodeEnd={processing && inputValue && <Loader variant="small" />}
+        nodeStart={<StyledSvgActionPlus />}
+        onFocus={(event) => {
+          textFieldProps?.onFocus?.(event)
+        }}
+        onClick={toggleMenu}
+      />
+      {/*</div>*/}
```
drillprop: same here?

packages/atlas/.storybook/preview.jsx
```diff
@@ -32,6 +32,16 @@ export const decorators = [StylesWrapperDecorator]
...
+
+export const argTypes = {
+  to: { table: { disable: true } },
+  className: { table: { disable: true } },
+  onClick: { table: { disable: true } },
+  onChange: { table: { disable: true } },
+  as: { table: { disable: true } },
+  icon: { table: { disable: true } },
+  children: { table: { disable: true } },
+}
```
drillprop, line 47: These props will be disabled in all storybook components? That's nice!

drillprop (approved):
Just one thing, I think you overlooked one comment in the code. Besides that, looks good 👍🏻

packages/atlas/.storybook/preview.jsx
kdembler: Yep!

packages/atlas/src/components/_inputs/ComboBox/ComboBox.tsx
kdembler: Yes, removed

packages/atlas/src/components/_inputs/ComboBox/ComboBox.tsx
kdembler: Removed

PR #3559, opened 2022-12-19
fix payments Buffer build issues

kdembler:
I updated @joystream/js/content to not use Node built-in Buffer and then any manual injecting is no longer necessary, everything seems to be working both in dev and built versions.


drillprop (approved):
LGTM. Tested this locally and on the built version. Managed to fetch and claim a reward. Thanks for fixing this 🙏🏻

Issue #2237, opened 2022-03-08
CO: Custom Thumb user story


dmtrjsg:
### Context

User story for Channel Owner wrt Playlist creation and mgt

### Scope
- [ ] Custom Thumbnail for Playlist

### Designs

[Link to Figma](https://www.figma.com/file/XfKZ5ZSavt88rVfiKf0tBv/Playlist%3A-Channel-owner?node-id=970%3A16745)

kdembler:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Atlas-600edbb8799cc700103bc2d7/issues/Joystream/atlas/2237?planning-poker) with ZenHub @drillprop @eldiegod @rafalpawlow

PR #3586, opened 2023-01-02
📲 Custom playlist US

WRadoslaw:
#2237


packages/atlas/src/components/DraggableComponent/DraggableComponent.styles.ts
```diff
@@ -0,0 +1,13 @@
+import styled from '@emotion/styled'
+
+import { sizes } from '@/styles'
+
+export const Wrapper = styled.div`
+  display: flex;
+  gap: ${sizes(3)};
+  cursor: grab;
+
+  svg {
```
drillprop: I would avoid styling like this.  Let's create `StyledSvgActionDrag` and apply `align-self: center` to it.

packages/atlas/src/components/DraggableComponent/DraggableComponent.styles.ts
```diff
@@ -0,0 +1,13 @@
...
+    align-self: center;
+  }
+`
```
drillprop, line 36: The component should have some `hover` and `draggable` state: 
https://www.figma.com/file/XfKZ5ZSavt88rVfiKf0tBv/My-playlists-page-%26-Playlist-workspace?node-id=905%3A9427&t=WyGuCWg3Z3ZFL3mB-4

Not sure if the latter is doable, but we should definitely change background on hover.

packages/atlas/src/components/DraggableComponent/DraggableComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { XYCoord, useDrag, useDrop } from 'react-dnd'
+
+import { SvgActionArrowBottom, SvgActionArrowTop, SvgActionDrag, SvgActionMore, SvgActionTrash } from '@/assets/icons'
+import { Button } from '@/components/_buttons/Button'
+import { ContextMenu } from '@/components/_overlays/ContextMenu'
+
+import { Wrapper } from './DraggableComponent.styles'
+
+type DraggableComponentProps = {
+  id: unknown
```
drillprop: Why unknown? Could we use type `string` here?

packages/atlas/src/components/_video/VideoListItem/VideoListItem.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+  gap: ${({ variant }) => (variant === 'small' ? 'unset' : sizes(2))};
+  display: grid;
+`
+
+export const ThumbnailImage = styled.img`
+  height: 100%;
+  object-fit: contain;
+`
+
+export const ThumbnailBackground = styled.div`
```
drillprop: I don't think we should create ThumbnailBackground - we already have a `VideoThumbnail` component and we should use it here.

packages/atlas/src/components/_video/VideoListItem/VideoListItem.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+  background: ${cVar('colorBackground')};
+  position: relative;
+`
+
+export const PillContainer = styled.div`
```
drillprop: The VideoThumbnail component has a prop called `slots` and we can pass a pill to it.

packages/atlas/src/components/_video/VideoListItem/VideoListItem.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const Wrapper = styled.div<{ variant: 'small' | 'large' }>`
+  width: 100%;
+  display: flex;
+  gap: ${sizes(4)};
+  height: ${({ variant }) => (variant === 'small' ? '80px' : '111px')};
+  position: relative;
+  overflow: hidden;
+  cursor: pointer;
+  padding: ${sizes(3)} 0;
```
drillprop: There is no hover effect on these list items. https://www.figma.com/file/XfKZ5ZSavt88rVfiKf0tBv/My-playlists-page-%26-Playlist-workspace?node-id=905%3A8978&t=WyGuCWg3Z3ZFL3mB-4

packages/atlas/src/components/_video/VideoListItem/VideoListItem.tsx
```diff
@@ -0,0 +1,101 @@
...
+  onClick?: () => void
+  isActive?: boolean
+  variant?: 'small' | 'large'
+  className?: string
+}
+
+export const VideoListItem = ({ id, onClick, isActive, className, variant = 'small' }: VideoListItemProps) => {
+  const { video, loading } = useBasicVideo(id ?? '', {
+    skip: !id,
+    onError: (error) => SentryLogger.error('Failed to fetch video', 'VideoTile', error, { video: { id } }),
```
drillprop:
```suggestion
    onError: (error) => SentryLogger.error('Failed to fetch video', 'VideoListItem', error, { video: { id } }),
```

packages/atlas/src/views/playground/PlaygroundLayout.tsx
```diff
@@ -42,6 +43,7 @@ const playgroundRoutes = [
   { path: 'reactions-comments', element: <PlaygroundReactionsComments />, name: 'Reactions & comments' },
   { path: 'iframe', element: <PlaygroundIframe />, name: 'Iframe' },
   { path: 'captcha', element: <PlaygroundCaptcha />, name: 'Captcha' },
+  { path: 'pl-custom-thumbnail', element: <PlaygroundCustomPLThumbnail />, name: 'PL: Custom thumbnail' },
```
drillprop: Possibly I'm nitpicking here, but I would name this differently, for the element PlaygroundPlaylistWorkspace and for the name just Playlist workspace

packages/atlas/src/views/studio/PlaylistWorkspace/PlaylistWorkspace.tsx
```diff
@@ -0,0 +1,299 @@
...
+  })
+  const thumbnail = watch('thumbnail')
+  const { video } = useBasicVideo(playlistVideos[0] ?? '', {
+    skip: !playlistVideos[0],
+    onError: (error) =>
+      SentryLogger.error('Failed to fetch video', 'VideoTile', error, { video: { id: playlistVideos[0] } }),
+  })
+  const { url: thumbnailPhotoUrl } = useAsset(video?.thumbnailPhoto)
+
+  // const hasUnsavedAssets = dirtyFields.thumbnail?.cropId || false
```
drillprop, line 90: Will this be needed in the future? If not let's remove this

packages/atlas/src/views/studio/PlaylistWorkspace/PlaylistWorkspace.tsx
```diff
@@ -0,0 +1,299 @@
...
+  const thumbnailAsset = useRawAsset(thumbnail?.cropId || null)
+  const originalThumbnailAsset = useRawAsset(thumbnail?.originalId || null)
+  // const firstVideoAsset = useRawAsset(video?.thumbnailPhoto?.id || null)
```
drillprop, line 93: same as above

packages/atlas/src/views/studio/PlaylistWorkspace/PlaylistWorkspace.tsx
```diff
@@ -0,0 +1,299 @@
...
+    ) {
+      setShouldFallbackThumbnail(false)
+      updateFallbackThumbnail()
+    }
+  }, [
+    video,
+    thumbnail,
+    getValues,
+    setValue,
+    thumbnailPhotoUrl,
```
drillprop: Are we need these dependencies here?

packages/atlas/src/views/studio/PlaylistWorkspace/PlaylistWorkspace.tsx
```diff
@@ -0,0 +1,299 @@
...
+    shouldFallbackThumbnail,
+    updateFallbackThumbnail,
+    playlistVideos,
```
drillprop: same as above

packages/atlas/src/views/studio/PlaylistWorkspace/PlaylistWorkspace.tsx
```diff
@@ -0,0 +1,299 @@
...
+        setValue(
+          'thumbnail',
+          {
+            ...currentThumbnailValue,
+            ...{ cropId: video.thumbnailPhoto?.id ?? null, originalId: null, url: thumbnailPhotoUrl },
+          },
+          { shouldDirty: true }
+        )
+      }
+      setValue('thumbnail', { ...currentThumbnailValue, ...{ cropId: null, originalId: null } }, { shouldDirty: true })
```
drillprop: Do we call `setValue` here twice? If there is a `video` we will call the second `setValue` anyway, right? Should we add `else` here?  
I'm not sure if this will cause bugs.

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+          placeholder="Search videos to add"
+        />
+        {loading ? (
+          <div style={{ width: '100%' }}>
+            {Array.from({ length: 3 }, (_, idx) => (
+              <VideoListItemLoader key={idx} variant="small" />
+            ))}
+          </div>
+        ) : (
+          <div>
```
drillprop: Not sure, because the design doesn't specify this well, but I think the list should have its own scrollbar. The search box should be always visible and placed on top when you scroll. Right now this is not a pleasant UX IMHO: 
image


I would consult this with a designer. This design could suggest that it should work like that(look at the last item):  https://www.figma.com/file/XfKZ5ZSavt88rVfiKf0tBv/My-playlists-page-%26-Playlist-workspace?node-id=905%3A9397&t=WyGuCWg3Z3ZFL3mB-4

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+      onError: (error) => SentryLogger.error('Failed to fetch videos', 'MyVideosView', error),
+    }
+  )
+
+  useEffect(() => {
+    if (show) {
+      setSelectedVideos(initialSelected)
+    }
+    // need to update selected videos when modal opens
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
drillprop: I'm a little worried that it could cause bugs. What if `initalSelected` change?

drillprop: Do we actually need this effect?

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useAuthorizedUser } from '@/providers/user/user.hooks'
+import { SentryLogger } from '@/utils/logs'
+
+import { Wrapper } from './VideoSelectorDialog.styles'
+
+type VideoSelectorDialogProps = {
+  onHide: () => void
+  show: boolean
+  onSelect: (arg: string[]) => void
+  initialSelected?: string[]
```
drillprop: Can we rename this? `initiallySelectedVideoIds`?

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+const INITIAL_FIRST = 50
+
+export const VideoSelectorDialog = ({ onHide, show, onSelect, initialSelected = [] }: VideoSelectorDialogProps) => {
+  const [selectedVideos, setSelectedVideos] = useState<string[]>(initialSelected)
```
drillprop: same here - `selectedVideoIds`?

packages/atlas/src/views/studio/PlaylistWorkspace/PlaylistWorkspace.tsx
```diff
@@ -0,0 +1,299 @@
...
+    caption: 'Visible only with link',
+  },
+]
+// todo: should be transfered to the context later on
+export type PlaylistWorkspaceProps = {
+  show: boolean
+  onHide: () => void
+}
+
+export const PlaylistWorkspace = ({ show, onHide }: PlaylistWorkspaceProps) => {
```
drillprop: I guess we forgot to mention that, but we're used to using `FC` for all our components i.e `export const Component: FC<ComponentProps>`. Can we use it for the sake of consistency?

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+type VideoSelectorDialogProps = {
+  onHide: () => void
+  show: boolean
+  onSelect: (arg: string[]) => void
+  initialSelected?: string[]
+}
+const INITIAL_FIRST = 50
+
+export const VideoSelectorDialog = ({ onHide, show, onSelect, initialSelected = [] }: VideoSelectorDialogProps) => {
```
drillprop: same as above

packages/atlas/src/components/DraggableComponent/DraggableComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+  children: ReactNode
+}
+
+type DragItem = {
+  index: number
+  id: string
+  type: string
+}
+
+export const DraggableComponent = ({
```
drillprop: same as above above

packages/atlas/src/components/_inputs/ImageUploadAndCrop/ImageUploadAndCrop.tsx
```diff
@@ -0,0 +1,143 @@
...
+type ImageUploadAndCropProps = {
+  onImageChange: (arg: ImageInputFile | null) => void
+  editMode?: boolean
+  disabled?: boolean
+  className?: string
+  file?: ImageInputFile
+  maxSize?: number
+}
+
+export const ImageUploadAndCrop = ({
```
drillprop: same as above

packages/atlas/src/components/_video/VideoListItem/VideoListItem.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+type VideoListItemProps = {
+  id?: string
+  onClick?: () => void
+  isActive?: boolean
+  variant?: 'small' | 'large'
+  className?: string
+}
+
+export const VideoListItem = ({ id, onClick, isActive, className, variant = 'small' }: VideoListItemProps) => {
```
drillprop: same as above

packages/atlas/src/components/_video/VideoListItem/VideoListItemLoader.tsx
```diff
@@ -0,0 +1,16 @@
+import { SkeletonLoader } from '@/components/_loaders/SkeletonLoader'
+import { sizes } from '@/styles'
+
+import { SkeletonTextWrapper, Wrapper } from './VideoListItem.styles'
+
+export const VideoListItemLoader = ({ variant }: { variant: 'small' | 'large' }) => {
```
drillprop: same as above

drillprop (changes requested):
Let's adjust RWD to match the designs as well https://www.figma.com/file/XfKZ5ZSavt88rVfiKf0tBv/My-playlists-page-%26-Playlist-workspace?node-id=1114%3A37729&t=WyGuCWg3Z3ZFL3mB-4

packages/atlas/src/views/studio/PlaylistWorkspace/PlaylistWorkspace.tsx
WRadoslaw: Basing on video upload it will be needed, that's why i left it as comment

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
WRadoslaw: It depends on what behavior we want to have (eventually I can introduce some callback to track videos when the modal is closed). If modal should remember what videos are selected, then yes. If we want the modal to always show videos as unselected at the beginning then it can be removed. WDYT

WRadoslaw: > I'm a little worried that it could cause bugs. What if `initalSelected` change?

We only want to update the state when modal reopens, since it's mounted all the time I don't see another cleaner option

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+      primaryButton={{
+        text: 'Add videos',
+        disabled: !selectedVideoIds.length,
+        onClick: () => {
+          onSelect(selectedVideoIds.map((videoId) => [videoId, createId()]))
+          setSelectedVideoIds([])
+          onHide()
+        },
+      }}
+      secondaryButton={{ text: 'Cancel', onClick: onHide }}
```
drillprop: once someone clicks cancel I think we should:
- unselect all the videos
- clear the search

Right now you can click cancel and open the dialog again and you'll still see the same videos checked, which I don't think is a good user experience

packages/atlas/src/views/studio/PlaylistWorkspace/VideoSelectorDialog/VideoSelectorDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+          <ListWrapper>
+            {Array.from({ length: 8 }, (_, idx) => (
+              <VideoListItemLoader key={idx} variant="small" />
+            ))}
+          </ListWrapper>
+        ) : (
+          <ListWrapper>
+            {edges?.length
+              ? edges.map(({ node: { id } }) => (
+                  <VideoListItem
```
drillprop, line 100: I think I said this already, there is no hover effect on the list item. Designers use here ListItem component 
Zrzut ekranu 2023-01-16 o 19 41 58
And if you inspect ListItem, you'll notice that this component has a hover effect defined: 
image

Also, can we use our component here? https://atlas-storybook-git-dev-joystream.vercel.app/?path=/story/other-listitem--default

drillprop (changes requested):
I still have some concerns about dialog. Besides that, really stunning work! Thanks for improving drag and drop! 🎉

packages/atlas/src/components/_video/VideoListItem/VideoListItem.styles.ts
```diff
@@ -0,0 +1,78 @@
...
+  flex-direction: column;
+  position: relative;
+  overflow: hidden;
+  cursor: pointer;
+  padding: ${sizes(3)} ${sizes(4)};
+
+  ${media.xs} {
+    flex-direction: row;
+
+    > *:first-child {
```
drillprop: Could we avoid using first-child here and replace it with first-of-type? Not a big deal, but emotion is complaining about this: 
image

drillprop (approved):
LGTM!

Issue #2252, opened 2022-03-09
CW: Playlist Details View Components


dmtrjsg:
### Context

Playlists content viewer side

### Scope

- [ ] Playlist details card

Thumbnail image
Meta
Description
Channel Owner + Follow
Copy playlist link

- [ ] List of videos on the right
+ Infinite scroll

- [ ] Handle empty state (no videos) [Figma > ](https://www.figma.com/file/f6pcaMR7bryFmszfrzTqJM/Playlist-details-page?node-id=336%3A2174)

Out of scope ⛔ 
- [ ] Numerated list (e.g. drop numbers)

### Designs

🎨  [Link to Figma > ]([Link to Figma>](https://www.figma.com/file/f6pcaMR7bryFmszfrzTqJM/Playlist-details-page?node-id=927%3A51669)

Preview: 

<img width="817" alt="Screenshot 2022-03-09 at 16 43 39" src="https://user-images.githubusercontent.com/93924525/157488815-eafc4319-b3e8-4d5d-9eaa-b533bc83aa06.png">

Preview Empty List State:

<img width="862" alt="Screenshot 2022-03-09 at 17 15 20" src="https://user-images.githubusercontent.com/93924525/157495082-e9379701-5ed4-4751-8242-43017aad76be.png">

kdembler:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Atlas-600edbb8799cc700103bc2d7/issues/Joystream/atlas/2252?planning-poker) with ZenHub @drillprop @eldiegod @rafalpawlow

PR #3590, opened 2023-01-03
📛 Playlist details

WRadoslaw:
#2252 

⚠️ review after #3586


packages/atlas/package.json
```diff
@@ -69,6 +69,8 @@
     "react": "^18.2.0",
     "react-beautiful-dnd": "^13.1.1",
     "react-datepicker": "^4.8.0",
+    "react-dnd": "^16.0.1",
+    "react-dnd-html5-backend": "^16.0.1",
```
drillprop: We don't need this packages anymore

packages/atlas/src/components/_video/VideoListItem/VideoListItem.styles.ts
```diff
@@ -4,7 +4,29 @@ import styled from '@emotion/styled'
 import { ListItem } from '@/components/ListItem'
 import { cVar, media, sizes } from '@/styles'
 
-export const DetailsWrapper = styled.div<{ variant: 'small' | 'large' }>`
+export type VideoListItemVariants = 'small' | 'medium' | 'large'
+
+export const getVideoVaraintDimensions = (variant: VideoListItemVariants): { width: number; height: number } => {
```
drillprop:
```suggestion
export const getVideoVariantDimensions = (variant: VideoListItemVariants): { width: number; height: number } => {
```
typo

packages/atlas/src/views/viewer/PlaylistView/PlaylistDetails.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <ChannelLink id="1" followButton />
+        </ColumnGapBlock>
+      </InfoContainer>
+      <RowGapBlock>
+        {Array(10)
+          .fill(1)
+          .map((_, index) => (
+            <CustomVideoListItemWrapper key={index}>
+              <Counter>{index + 1}</Counter>
+              <StyledVideoListItem id={String(index)} variant={xsMatch && !smMatch ? 'medium' : 'large'} />
```
drillprop: Doesn't match the designs:
image
image

Check the position of the title and description.
Also, I think we shouldn't use here the ListItem, we should use the VideoTile component here. I think we already have a horizontal variant implemented if I'm not mistaken.
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3630%3A118313&t=DnHdi9dR8x1yog5s-4
We shouldn't have a hover effect in this particular case as well.

packages/atlas/src/views/viewer/PlaylistView/PlaylistDetails.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <ColumnGapBlock justify="space-between" padding={6}>
+          <ChannelLink id="1" followButton />
+        </ColumnGapBlock>
+      </InfoContainer>
+      <RowGapBlock>
+        {Array(10)
+          .fill(1)
+          .map((_, index) => (
+            <CustomVideoListItemWrapper key={index}>
+              <Counter>{index + 1}</Counter>
```
drillprop, line 57: The gap between the counter and the tile is too big. Should be 16px 
image

packages/atlas/src/views/viewer/PlaylistView/PlaylistDetails.styles.ts
```diff
@@ -0,0 +1,90 @@
...
+`
+
+export const Counter = styled.div`
+  border-radius: 50%;
+  width: ${sizes(6)};
+  height: ${sizes(6)};
+  display: grid;
+  place-items: center;
+  color: ${cVar('colorTextMuted')};
+  background-color: rgb(187 217 246 / 0.13);
```
drillprop: use should use `cVar` here.

drillprop (changes requested):

packages/atlas/src/views/viewer/PlaylistView/PlaylistDetails.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <ChannelLink id="1" followButton />
+        </ColumnGapBlock>
+      </InfoContainer>
+      <RowGapBlock>
+        {Array(10)
+          .fill(1)
+          .map((_, index) => (
+            <ColumnGapBlock key={index} gap={0}>
+              <Counter>{index + 1}</Counter>
+              <StyledVideoListItem id={String(index)} variant="large" isInteractive={false} />
```
drillprop: As I said on discord earlier, we should use here the VideoTile component instead of VideoListItem

drillprop (changes requested):

drillprop (approved):

PR #3602, opened 2023-01-09
Ypp syncing integration

### LINKED ISSUE #3589: Integration for leaving the program and turning off Ypp syncing
- [ ] turn off syncing from the settings tab
- [ ] leave the program

drillprop:
Fix #3589

### DIFF
--- a/packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
+++ b/packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
@@ -14,9 +14,10 @@ import { useChannelsStorageBucketsCount } from '@/providers/assets/assets.hooks'
 import { useConfirmationModal } from '@/providers/confirmationModal'
 import { useBloatFeesAndPerMbFees, useJoystream } from '@/providers/joystream/joystream.hooks'
 import { useSnackbar } from '@/providers/snackbars'
+import { TX_SIGN_CANCELLED_SNACKBAR_TIMEOUT } from '@/providers/transactions/transactions.config'
 import { useTransaction } from '@/providers/transactions/transactions.hooks'
 import { useUser } from '@/providers/user/user.hooks'
-import { SentryLogger } from '@/utils/logs'
+import { ConsoleLogger, SentryLogger } from '@/utils/logs'
 import { useGetYppSyncedChannels } from '@/views/global/YppLandingView/YppLandingView.hooks'
 import { StyledActionBar } from '@/views/viewer/EditMembershipView/EditMembershipView.styles'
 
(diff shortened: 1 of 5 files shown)


packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
```diff
@@ -126,12 +236,7 @@ export const YppDashboardSettingsTab = () => {
             label="Category of imported videos"
             description="Choose a category to be assigned to the imported videos by default. You can change it for each video later once it’s imported."
           >
-            <Select
-              disabled={currentChannel?.shouldBeIngested}
-              items={categoriesSelectItems}
-              onChange={setCategory}
-              value={category}
-            />
+            <Select items={categoriesSelectItems} onChange={setCategoryId} value={categoryId} />
```
WRadoslaw: Shouldn't we disable the Select when the category is already chosen?

https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=3542%3A89393&t=YX2PO34nrAi7tXy2-0

drillprop: We changed that, and you should be able to update the category.

packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
```diff
@@ -29,31 +46,54 @@ const categoriesSelectItems: SelectItem[] =
...
-        `${atlasConfig.features.ypp.youtubeSyncApiUrl}/channels/${currentChannel.joystreamChannelId}`,
-        {
-          category,
-          isSync,
-        }
+        `${atlasConfig.features.ypp.youtubeSyncApiUrl}/channels/${currentChannel.joystreamChannelId}/ingest`,
+        { message, signature }
       )
 
       if (data.status === 200) {
```
WRadoslaw, line 110: Looks like refetch is not working as intended. After changing the category footer is still visible.

image

packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
```diff
@@ -156,7 +261,7 @@ export const YppDashboardSettingsTab = () => {
         secondaryButton={{
           text: 'Cancel',
           onClick: () => {
-            setCategory(currentChannel?.videoCategoryId)
+            setCategoryId(currentChannel?.videoCategoryId)
```
WRadoslaw:
```suggestion
            setCategoryId(currentChannel?.videoCategoryId)
            setIsSync(currentChannel?.shouldBeIngested)
```

packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
WRadoslaw: Same applies for `isSync`

WRadoslaw (changes requested):

WRadoslaw (approved):
Works as expected!

Btw, WDYT about introducing some loading state to the settings tab?
RN, on entry the form is empty and after request is done it gets filled, which doesn't look good. If so let's create an issue. @toiletgranny @drillprop 

image

image

toiletgranny:
Thanks for flagging, did not account for the settings taking that much time to load. Assuming we can not fetch the settings before the user goes specifically into the "Settings" tab, my quick recommendation are as follows:

1. If the loading is expected to take less than a second on average, I'd recommend a simple solution: reducing the opacity of the settings form to something like `.5` or even just `.75` and disabling any cursor events within that area.
2. If it's more likely to take more than a second, then I'd recommend using the Loader component (the largest variant or the slightly smaller one) and displaying that instead of the form's settings.

PR #3605, opened 2023-01-10
Fix balance tooltip issues

### LINKED ISSUE #3604: Incorrect display of locked funds
CleanShot 2023-01-10 at 10 35 10@2x

2 issues here:
1. The number doesn't fit in and it weirdly wraps around
2. Funds under vesting lock are also shown as "Invitation lock", maybe worth considering just changing that to "Locked funds"?

drillprop:
Fix #3604

### DIFF
--- a/packages/atlas/src/components/_overlays/MemberDropdown/BalanceTooltip.styles.ts
+++ b/packages/atlas/src/components/_overlays/MemberDropdown/BalanceTooltip.styles.ts
@@ -5,7 +5,7 @@ import { cVar, sizes } from '@/styles'
 
 export const TooltipWrapper = styled.div`
   /* sizes(6) - tooltip horizonthal padding */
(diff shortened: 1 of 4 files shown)


packages/atlas/src/components/_overlays/MemberDropdown/BalanceTooltip.styles.ts
```diff
@@ -5,7 +5,7 @@ import { cVar, sizes } from '@/styles'
 
 export const TooltipWrapper = styled.div`
   /* sizes(6) - tooltip horizonthal padding */
-  width: calc(224px - ${sizes(6)});
+  width: calc(232px - ${sizes(6)});
```
drillprop, line 8: Consulted with the designer. We decided to increase a bit width of BalanceTooltip

WRadoslaw (approved):

Issue #3319, opened 2022-10-06
Mention Content Focus for Atlas GW onboarding flow


dmtrjsg:
## Context
Multiple gateways set up covered here will not rule out content-vertical focus, as a way to build up a congruent value proposition by external operators.

* https://github.com/Joystream/marketing/issues/289

Gleev, operated by JSG as the prime example of this case is planned to focus on Web3/ Crypto vertical only.

For the benefit of end users of Gleev the idea is to inform them on all entry points to the app, about the specific focus of the app, be it content, business model, cause or any other point of differentiation.

This issue attempts to capture all relevant places 

YPP adaptation is covered in a separate issue:

* https://github.com/Joystream/marketing/issues/292

:point_up: Importantly the designs (and implementation) must be done in such a way that will make it easy for external operators to configure this content to their purposes without rebuilding the pages in a custom ways. 

## Scope
1.  Introduce a global variable allowing gateway operators to describe the app's content focus so that it can be later referenced in the UI. Ideally, the content focus description is contained within 25 characters (at most), and a few examples of usage are provided for a gateway operator, e.g.:
   -  "Your `{{contentFocus}}` creator journey starts here"
        Example: _Your `web3 & crypto` creator journey starts here_
   - "Welcome to `{{appName}}`, a video platform focused on `{{contentFocus}}`  content."
        Example: _Welcome to `Gleev`, a video platform focused on `web3 & crypto` content._
        **>> this will be hardcoded on Gleev side.**    
        
2. Introduce an **App KV** component based on the designs, docs, and changelog, which [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=10156%3A185036&t=swmhnBC9qu8VfCqm-4)
   > **Note**
   > @drillprop Please let me know on Discord once you get to this, as we need to discuss the matter of keeping illustration assets in the Figma file. Thanks!
3. Introduce a **Welcome dialog** component based on the designs, docs, and changelog, which [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=10153%3A183208&t=swmhnBC9qu8VfCqm-4),
4. Update the **Sign in dialog** component based on the designs, docs, and changelog, which [can be found here](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=6322%3A126310&t=swmhnBC9qu8VfCqm-4),
[...]

PR #3606, opened 2023-01-10
🌫️ Feature/content focus

WRadoslaw:
#3319


packages/atlas/src/components/WelcomeDialog/WelcomeDialog.tsx
```diff
@@ -0,0 +1,44 @@
...
+      contentClassName="welcome-content"
+      size="small"
+    >
+      <Wrapper>
+        <AppKV />
+        <StyledText variant="h500" as="p" color="colorTextStrong">
+          Welcome to {atlasConfig.general.appName}
+        </StyledText>
+        <StyledText variant="t200" as="p" color="colorCoreNeutral200">
+          {atlasConfig.general.appName} is a {atlasConfig.general.appContentFocus} content-focused video platform that's
```
drillprop: according to the design, `appContentFocus` should have different color 
image

packages/atlas/src/components/AppKV/AppKV.styles.ts
```diff
@@ -0,0 +1,38 @@
...
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+`
+
+export const Fade = styled.div`
+  position: absolute;
+  inset: 0;
+  top: 40%;
+  background: linear-gradient(180deg, rgb(24 28 32 / 0) 0%, #181c20 100%);
```
drillprop: we should use her `cVar(colorBackground)` for `#181c20` and we could replace `rgb(24 28 32 / 0)` with `transparent`(I think)

packages/atlas/src/components/AppKV/AppKV.styles.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  ${media.sm} {
+    top: 52%;
```
drillprop: why 52% and not 128px? I think the fade height is always the same independently of window size.

packages/atlas/src/components/AppKV/AppKV.styles.ts
```diff
@@ -0,0 +1,38 @@
...
+  gap: ${sizes(6)};
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+`
+
+export const Fade = styled.div`
+  position: absolute;
+  inset: 0;
+  top: 40%;
```
drillprop: shouldn't we just put this on the bottom - i.e. set `bottom:0`

packages/atlas/src/components/AppKV/AppKV.styles.ts
```diff
@@ -0,0 +1,38 @@
...
+`
+
+export const LogoWrapper = styled.div`
+  position: absolute;
+  inset: 0;
+  display: flex;
+  gap: ${sizes(6)};
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
```
drillprop, line 26: This should have a z-index set. Because the fade is covering the letters.

drillprop (changes requested):

drillprop (changes requested):
Could you also add an entry to the changelog? 🙏🏻

packages/atlas/src/components/AppKV/AppKV.styles.ts
WRadoslaw: Leftover, when fade was the same size as the image (pretty big since image is stretched as hell, so fade started too low)

packages/atlas/src/components/AppKV/AppKV.styles.ts
WRadoslaw: Nice catch 👀

drillprop (approved):

Issue #3584, opened 2022-12-28
Use react-query to get information from ypp backend


drillprop:
As the title says

PR #3614, opened 2023-01-12
🧑‍💻 Introduce react-query in all places

WRadoslaw:
#3584


drillprop (commented):
Can we introduce react-query mutation to upload a membership avatar as well?  The function is inside `packages/atlas/src/utils/image.ts`

Besides that, looks good to me

drillprop (approved):

PR #3640, opened 2023-01-18
Included platform royalty in NFT selling details

### LINKED ISSUE #3621: We should inform user about `nftPlatformFeePercentage` when user is listing nft
So this happened to me during testing gleev:
1. I listed nft for 100 joy for a fixed price.
2. Someone bought this nft for 100 joy.
3. I went to my channel and checked my channel account and surprisingly noticed that I received 98 joy, not 100 joy. 

This happened because every time someone buys NFT we're taking the so-called `nftPlatformFeePercentage`. In this particular case, it was set to 2%.  It's displayed only on Purchase View https://www.figma.com/file/wjeAOZq77pSUFZsYXJqKIB/NFT-purchase-view?node-id=1236%3A26860&t=K9VBQmZPqx2cwp5V-4 
image

The problem is this information should be displayed to a seller as well because he's actually losing some money when he lists his videos. I checked the designs and we don't inform the user that he won't receive all the money. We only inform about the creator's royalties, not platform royalties.  
https://www.figma.com/file/xm6vBdSnQDroc57rGfGtPM/Video-workspace?node-id=2393%3A85860&t=q9EUF0iw0C5Dg1jL-4

attemka:
closes #3621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Welcome dialog when content focus is present
 - Fallback state for video hero and video category hero
+- Platform royalty in NFT selling details
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


drillprop (approved):
LGTM!

drillprop (changes requested):
Sorry, I forgot about one thing. Could you add this change to the `CHANGELOG.md` under `unreleased`?

drillprop (approved):

Issue #3641, opened 2023-01-18
Sign up modal is showing when you enter the site 


drillprop:
To reproduce.
1. Sign in on firefox
2. Refresh the site(sometimes you need to do this few times)
3. You'll get the "Select wallet" popup

Expected: 
You shouldn't see the "sign-in modal" when entering the site, If you already signed in.

PR #3643, opened 2023-01-19
🧇 Retry wallet connection

WRadoslaw:
For firefox, Polkadot extension doesn't load as fast as talisman, which results in the wallet not connecting automatically.
#3641

### DIFF
--- a/packages/atlas/src/providers/user/user.provider.tsx
+++ b/packages/atlas/src/providers/user/user.provider.tsx
@@ -54,7 +54,8 @@ export const UserProvider: FC<PropsWithChildren> = ({ children }) => {
   const signIn = useCallback(
     async (
       walletName?: string,
-      mobileCallback?: ({ onConfirm }: { onConfirm: () => void }) => void
+      mobileCallback?: ({ onConfirm }: { onConfirm: () => void }) => void,
+      invokedAutomatically?: boolean
     ): Promise<boolean> => {
       let accounts = []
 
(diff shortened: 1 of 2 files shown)


packages/atlas/src/providers/user/user.provider.tsx
```diff
@@ -67,7 +68,7 @@ export const UserProvider: FC<PropsWithChildren> = ({ children }) => {
       }
 
       try {
-        const initializedAccounts = await initSignerWallet(walletName)
+        const initializedAccounts = await retryPromise(() => initSignerWallet(walletName), 500, 2000)
```
drillprop: image
I'm getting an error, when I want to authorize the app, the error won't disappear until I refresh the site. This didn't happen on `dev`

To reproduce: 
1. Run `yarn atlas:dev` on a different port for example `yarn atlas:dev --port 4333`. This will be helpful to trigger the authorization popup
2. Click connect wallet 
3. Select Polkadot extension and click Select Wallet
4. The authorization popup should show
5. Wait 2 seconds, and you'll notice the error above
6. Once you authorize, the error stays and won't disappear until you refresh the site.

drillprop (changes requested):

packages/atlas/src/providers/user/user.provider.tsx
drillprop: Actually, I just found out you can trigger the authorization popup easier.
1. Disconnect the wallet
2. Click the Polkadot extension icon in your browser
3. Click the cog icon to open the settings, then click manage website access
4. Find localhost:3000 or 127.0.0.1:3000 and click the trash button
5. Connect the wallet again

drillprop (approved):
Looks good to me!

attemka (approved):

Issue #3622, opened 2023-01-13
Display Negative Transferable balance


dmtrjsg:
## Context

When channel gets created while membership wallet balance is at 0, the bloat bond is still taken from the trasnferable balance. (While the tx fees are covered by Invitation lock).

This results in the transferable balance going below 0. 

## Scope
- [ ] Display negative balance with tooltip. 

Tooltip content:
"Your transferable balance is negative. This happened because the channel creation bloat bond, deposit required for the storage allowance, was paid from your transferable balance while it was 0. You can still upload videos, mint NFTs like and comment other videos, and the transaction fees for these actions will be covered from the invitation lock."

## Additional info

Related issue in Pioneer: 
- https://github.com/Joystream/pioneer/issues/4076

KubaMikolajczyk:
The tooltip for this special scenario is designed and can be found under the link below. I took an opportunity to rewrite the text proposed above to make it shorter to fit in such a small space. 

📄Link to figma changelog: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=9258%3A167766&t=SCdWqjqTo1pLHFnI-4

WRadoslaw:
@dmtrjsg 
After taking the flow: 
- Create new account (on Atlas)
- Create channel
- Upload video

I've noticed that balances coming from `api.derive.balances.all` are still positive.
(These are entries from balances object and mapped to -> [key, value in string if BN, bool which is true when val is negative]) 
![image](https://user-images.githubusercontent.com/92513933/212884737-036a5803-4de4-4eeb-8bb8-1fb9e4bbd0a1.png)

On Pioneer's success modal for token transfer we are able to conclude that balance was negative only with a knowledge of amount transferred and current transferable balance
```
  const toNow = toBalance.transferable
  const toBefore = toNow.sub(amount)
```
[link](https://github.com/Joystream/pioneer/blob/f27ab516c018396987748f043d1a3771b2efd445/packages/ui/src/accounts/modals/TransferModal/TransferSuccessModal.tsx#L40)

The question is whether there is another way to check if balance on given address is negative.

PR #3646, opened 2023-01-19
❎ Negative balance

WRadoslaw:
#3622

### DIFF
--- a/packages/atlas/src/components/JoyTokenIcon/JoyTokenIcon.tsx
+++ b/packages/atlas/src/components/JoyTokenIcon/JoyTokenIcon.tsx
@@ -29,6 +29,7 @@ export type JoyTokenIconProps = {
   size?: JoyTokenIconSize
   className?: string
   withoutInformationTooltip?: boolean
+  isNegative?: boolean
 }
 
 const VARIANT_SIZE_COMPONENT_MAPPING: Record<JoyTokenIconVariant, Record<JoyTokenIconSize, ElementType>> = {
@@ -63,6 +64,7 @@ export const JoyTokenIcon: FC<JoyTokenIconProps> = ({
   size = 16,
   className,
   withoutInformationTooltip,
+  isNegative,
 }) => {
   const ref = useRef<HTMLDivElement>(null)
   return (
@@ -80,6 +82,7 @@ export const JoyTokenIcon: FC<JoyTokenIconProps> = ({
         hasShadow={!['regular', 'gray'].includes(variant)}
         className={className}
         variant={variant}
+        isNegative={isNegative}
       />
     </>
   )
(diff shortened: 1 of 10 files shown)


drillprop (changes requested):
1. image

The total is not correct in this case? 
Actually, if I understand this correctly: 
- The invitation lock should be constant. So it should be always 100 000 in my case in the screenshot above(this is the value that is given by the faucet)?
- Total balance should be locked invitational + transferable balance. 

But please confirm this with Dima, I'm a little confused about what should we show here. 

2. You forgot to add information about the negative balance: 
image

dmtrjsg:
1️⃣ 
Keep the label as  "Locked Funds". That field should only display the locked funds that can be spent on tx fees. 

Update tooltip to "Your transferable balance is negative because the channel creation bond was paid from it while it was 0. You can still upload videos and perform other actions, but the transaction fees will be covered by the invitation lock as part of your locked funds, and it cannot be transferred."

2️⃣ ⚠️ Check why the transferable balance gets reduced with every tx signed, like "Like", Comment and reaction to the video. The only part that needs to be paid from transferable balance when it is <= 0 is the bloat bonds, which are taken for channel creation and (maybe) a comment.

WRadoslaw:
About

> 1️⃣ Keep the label as "Locked Funds". That field should only display the locked funds that can be spent on tx fees.
> 
> Update tooltip to "Your transferable balance is negative because the channel creation bond was paid from it while it was 0. You can still upload videos and perform other actions, but the transaction fees will be covered by the invitation lock as part of your locked funds, and it cannot be transferred."
> 
> 2️⃣ ⚠️ Check why the transferable balance gets reduced with every tx signed, like "Like", Comment and reaction to the video. The only part that needs to be paid from transferable balance when it is <= 0 is the bloat bonds, which are taken for channel creation and (maybe) a comment.

About the 2nd point. 

Right now it looks like the transferable balance should be going down with each TX fee paid. Since Invitation lock works like a debt, it means that every token spend from lock will need to be paid back. 
Eg. 
My inv lock - 100 // Balance 0
I've spent for fees - 20 // Balance - 20
My friend sent me - 30 // Balance - 10

dmtrjsg:
Solution: 
1. We _show_ transferable balance at 0 all the time. 
2. When transferable balance goes to negative zone, we display tooltip
3. Tooltip has one balance only "Spendable balance". Spendable balance = Invitation lock - debt on transferble balance. 
4. We remove "invitation lock" and total balance from the tooltip. 
5. Tooltip text says: "You don't have any tokens to spend on NFTs or transfer, but there is a balance you can spend on paying of transaction fees for creating channels, posting videos, adding comments and reactions. This balance needs to be repaid before accumulating transferable balance."

drillprop (changes requested):
I see a tooltip in the account where is a transferable balance, and I think I shouldn't. 
image


I tested this on mainnet environment with account
`j4X5B4Dsgeb43yzpC8PbxSodaS7G3uYxNZbRs1fW9UNEYKfp3`
 This is my balance on pioneer. I don't have an invitation lock on this account, only a vesting lock
image

drillprop (approved):
LGTM but Let's wait for final approval from @dmtrjsg

dmtrjsg:
pls don't get mad at me 😬  but after a final review I feel that to effectively solve the problem of informing the users on why they receive less JOY from NFT sale or transfer we need to: 

1. Also show tooltip when transferable balance = 0. And invitation lock is not spent.  (currently tooltip shows only when lock balance is spent on at least one transaction).
2. Display transferable balance and joy token icon in red, when transferable balance is non-positive.  
3. Slightly update copy to "Your transferable balance is zero, meaning you cannot buy NFTs or transfer tokens. There is a spendable balance from faucet, which you can still spend on paying transaction fees for creating channels, posting videos, adding comments and reactions. This balance needs to be repaid before accumulating transferable balance."

cc @WRadoslaw @drillprop

dmtrjsg:
Perfect, thanks a mil @WRadoslaw and @attemka 🙏

Issue #3647, opened 2023-01-20
Lots of requests are sent on my videos page, when videos are syncing? Syncing video tiles not working correctly


drillprop:
This could be hard to reproduce because it didn't happen frequently for me, but I noticed this already few times. I'm not sure what causes this. Try to go to `my videos`, then switch tabs, go to my uploads view, and then go back to `my videos` to reproduce this.
  
When the `state` of the syncing videos is`UploadFailed` or `UploadStarted`(not super sure about the second one) tiles act weirdly, they are switching back from `upload failed` to `auto-sync`. Besides that, we're sending a massive amount of `GetFullVideos` requests(hundreds per minute), when this is happening. 

If you need a google account to test, I can provide you email and password, just dm me.

PR #3655, opened 2023-01-20
🍶 Filter out YT private videos and turn on polling only for uploading videos

WRadoslaw:
#3647


packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
```diff
@@ -47,11 +47,19 @@ export const DELAYED_FADE_CLASSNAME = 'delayed-fade'
 
 export const VideoTilePublisher: FC<VideoTilePublisherProps> = memo(
   ({ id, onEditClick, onDeleteVideoClick, onReuploadVideoClick, onMintNftClick, titlesInSync }) => {
+    const [videoTitleMap, setVideoTitleMap] = useState('')
     const { video, loading } = useFullVideo(id ?? '', {
       skip: !id,
       onError: (error) => SentryLogger.error('Failed to fetch video', 'VideoTilePublisher', error, { video: { id } }),
-      pollInterval: titlesInSync?.length ? 1000 : undefined,
+      pollInterval: titlesInSync?.length && titlesInSync.includes(videoTitleMap) ? 1000 : undefined,
```
drillprop: Could we increase the interval a bit here, to three seconds at least? I don't think we should refetch that frequently. 

Also, let's store this in some global variable to not repeat this in MyVideosView.

packages/atlas/src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -68,7 +68,10 @@ export const MyVideosView = () => {
     () => axios.get<YppVideoDto[]>(`${atlasConfig.features.ypp.youtubeSyncApiUrl}/channels/${channelId}/videos`),
     {
       enabled: !!channelId,
-      refetchInterval: (data) => (data?.data.some((resource) => resource.state === 'UploadStarted') ? 1000 : false),
+      refetchInterval: (data) =>
+        data?.data.some((resource) => resource.state === 'UploadStarted' && resource.privacyStatus !== 'private')
+          ? 1000
```
drillprop: Let's use a global variable from the above

packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
```diff
@@ -47,11 +47,19 @@ export const DELAYED_FADE_CLASSNAME = 'delayed-fade'
 
 export const VideoTilePublisher: FC<VideoTilePublisherProps> = memo(
   ({ id, onEditClick, onDeleteVideoClick, onReuploadVideoClick, onMintNftClick, titlesInSync }) => {
+    const [videoTitleMap, setVideoTitleMap] = useState('')
```
drillprop, line 51: What means `videoTitleMap`? In the end, It's not a map, it's a string 😅 . I'm little confused reading this code, and what this effect below is doing. Could you explain?

drillprop (changes requested):

packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
WRadoslaw: `...Map` because this is mapped value of video title, we can change that naming np. 

The problem here is that from ypp backend we don't receive joystream video id, so we need to refetch videos by the title of video. This effect allows to access video title in `pollInterval`.

Issue #3654, opened 2023-01-20
Removing and adding collaborator when signing/opting out to ypp program


drillprop:
We've noticed two problems during the YPP flow, that need to be resolved somehow. 

## Signing
Every time user is signing into the youtube partner program, the user needs to: 
1. Send a request to the YPP backend
2. Once the request is successful we're asking the user to sign in a transaction to add a collaborator to his channel. This is needed to allow ypp backend to add videos to the user's channel.  
We're doing these steps in this order. 

The problem is, that it's possible for the user to send a request(do the first step) and then ignore the signing transaction. This will result in not syncing videos, but the user will be still registered in the YPP backend. The user won't know what's exactly happening, and his youtube videos won't be added to his atlas channel.

We could I think resolve this issue in a couple of ways: 
1. Fix this only in the backend. Once the user sends the request, the backend could after a few minutes check if the collaborator was set and if that didn't happen it could remove the user from the YPP program. 
The downside of this is the user could not register what happened and why he's not registered. But I'm assuming most of the users will sign the transaction and everything should be fine on their side
2. Change the designs and add more steps. We could add more steps to the sign-in flow. We could inform the user that he needs to add a collaborator, add a step for signing the transaction, wait for this transaction, and then show another step when the user could finish and send a proper request to the YPP backend. We will need to also inform the user, that if something goes wrong when sending the request he will need to remove the collaborator and we will provide a button to do this. 
3. We could add a banner in the dashboard and just inform the user that he didn't have a collaborator set in the YPP program. And inside this banner, we could add buttons, where he could take an action - add a collaborator or opt out from the program(he will need to only sign the message to do this) 

I'm open to other suggestions.

## ~Opting out~
~The same applies partly to opting out.~
~During opting out, the user needs to~
~1. Sign the message to send the authorized request to ypp backend~
~2. Sign the transaction to remove the collaborator~

~If he does only the first step. The collaborator will stay - not sure, but not a big problem I guess? 
If he does only the second step, he'll remove collaborator and he won't be able to sync his videos, but the backend will still treat him as a registered user.~

~I have no idea how we could tackle this, unfortunately, so I'm open to proposals~

attemka:
I would prefer p.3 here, but it should be clearly noticeable for the user that he needs to take some action in order to make the system work correctly. So maybe it should be an auto-opening modal (which might be a little bit too annoying) or a message which will overlap the videos section with an action button next to it. Track this state should be easy as well I suppose.

[...]

PR #3662, opened 2023-01-24
Fix issue with doubled popup

drillprop:
Fix https://github.com/Joystream/atlas/issues/3589#issuecomment-1397411654

Also partly fix opting out https://github.com/Joystream/atlas/issues/3654 The problem no longer occurs. If the user doesn't sign a message and remove the collaborator, he won't be able to opt out, because we won't send the request to the backend until he does both actions.

### DIFF
--- a/packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
+++ b/packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
@@ -156,43 +156,45 @@ export const YppDashboardSettingsTab = () => {
         timestamp: Date.now(),
       }
 
-      const signaturePromise = joystream.signMessage({
+      closeModal()
+      const signature = await joystream.signMessage({
         data: JSON.stringify(message),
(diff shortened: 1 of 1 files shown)


packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardSettingsTab.tsx
```diff
@@ -156,43 +156,45 @@ export const YppDashboardSettingsTab = () => {
...
         })
+
+        if (completed) {
+          const data = await axios.put(
+            `${atlasConfig.features.ypp.youtubeSyncApiUrl}/channels/${currentChannel.joystreamChannelId}/optout`,
+            { message, signature }
+          )
+          if (data.status === 200) {
+            displaySnackbar({
+              title: 'You left the progam',
```
WRadoslaw:
```suggestion
              title: 'You left the program',
```

WRadoslaw (changes requested):
Just a typo fix

WRadoslaw (approved):

Issue #3540, opened 2022-12-15
Implement Suspended


dmtrjsg:
## Context

Epic: 
- https://github.com/joystream/atlas/issues/2973

Backend issue: 
- https://github.com/Joystream/youtube-synch/issues/34
PR
-  https://github.com/Joystream/youtube-synch/pull/44

## Scope

- [ ] Display correct variation of dashbaord for suspended channels

[Design link> ](https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1797%3A99499&t=ri1k5FPPdCUA8CIv-1)

[Link to Figma > ](https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1797%3A99499&t=ri1k5FPPdCUA8CIv-1)

- [ ] On YPP dashboard page load show notification about Channel suspended. Only check this status when YPP dashboard loads.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1203868726963296/1203870652534023) by [Unito](https://www.unito.io)

dmtrjsg:
Tested with Channel 47
Seed: `damage seven rhythm cherry drive reopen kitten palace mirror rocket pool gaze`


Expected result:

<img width="807" alt="Screenshot 2023-02-01 at 15 52 16" src="https://user-images.githubusercontent.com/93924525/216093977-f6f5b568-15a9-4aa7-8f07-1f9116c9d7dc.png">

Actual Result: 

Dashboard displayed
<img width="807" alt="Screenshot 2023-02-01 at 15 52 16" src="https://user-images.githubusercontent.com/93924525/216094980-bebf40a8-f5eb-43b9-9874-ac65ccd34e2a.png">

Upon trying to sign up again: unknown error is displayed in the toaster
<img width="1003" alt="Screenshot 2023-02-01 at 15 57 06" src="https://user-images.githubusercontent.com/93924525/216095213-84646fad-db19-4c6e-b413-39e96d88d80c.png">


__

I suspect that may be due to the Is_suspended table column being deprecated now and status is governed from the same column as Opted out:

<img width="233" alt="Screenshot 2023-02-01 at 15 58 03" src="https://user-images.githubusercontent.com/93924525/216095595-1bc84133-ca2b-4b90-a6d2-55c15d9a73c5.png">

[...]

PR #3713, opened 2023-02-01
Fix suspended view in YppDashboard

drillprop:
Fix https://github.com/Joystream/atlas/issues/3540#issuecomment-1412300705
Also:
- used react-query to get synced channels
- fix some errors about invalid HTML in YppDashboard

### DIFF
--- a/packages/atlas/src/views/global/YppLandingView/YppLandingView.hooks.ts
+++ b/packages/atlas/src/views/global/YppLandingView/YppLandingView.hooks.ts
@@ -35,13 +35,7 @@ export type YppSyncedChannel = {
 export const useGetYppSyncedChannels = () => {
   const { activeMembership, membershipsLoading, isAuthLoading, channelId } = useUser()
   const location = useLocation()
-  const {
-    data: syncedChannels,
-    isLoading,
-    refetch,
-  } = useQuery('ypp-synced-channels', () => getSyncedChannels(), {
-    refetchOnWindowFocus: false,
-  })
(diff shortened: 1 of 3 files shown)


packages/atlas/src/views/global/YppLandingView/YppLandingView.hooks.ts
```diff
@@ -34,27 +34,24 @@ export type YppSyncedChannel = {
 
 export const useGetYppSyncedChannels = () => {
   const { activeMembership, membershipsLoading, isAuthLoading, channelId } = useUser()
-  const [isLoading, setIsLoading] = useState(true)
   const location = useLocation()
-  const [syncedChannels, setSyncedChannels] = useState<YppSyncedChannel[]>([])
+  const { data: syncedChannels, isLoading, refetch } = useQuery('ypp-synced-channels', () => getSyncedChannels())
```
attemka, line 38: note that by default react-query will refetch all the queries every time user re-enters the tab. This could cause higher server load if the number of active queries is high. You can read about it [here](https://react-query-v3.tanstack.com/guides/window-focus-refetching)

WRadoslaw:
@drillprop @attemka To be fair I think we could set `refetchOnWindowFocus` to false in the provider default config and turn it on only for specific queries that would need it. 

We can add it in the constructor for `queryClient` in `packages/atlas/src/CommonProviders.tsx`

WRadoslaw (approved):

PR #3716, opened 2023-02-02
removed payouts dashboard for YPP

### LINKED ISSUE #3706: Hide my payments dashboard for YPP release
ssia

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1203868726963296/1203870592708229) by [Unito](https://www.unito.io)

attemka:
removed payouts dashboard for YPP
closes #3706

### DIFF
--- a/packages/atlas/src/components/_navigation/SidenavStudio/SidenavStudio.tsx
+++ b/packages/atlas/src/components/_navigation/SidenavStudio/SidenavStudio.tsx
@@ -39,12 +39,16 @@ const studioNavbarItems: NavItemType[] = [
     expandedName: 'My uploads',
     to: absoluteRoutes.studio.uploads(),
   },
-  {
-    icon: <SvgSidebarPayments />,
-    name: 'Payments',
-    expandedName: 'My payments',
-    to: absoluteRoutes.studio.payments(),
-  },
+  ...(atlasConfig.features.ypp.googleConsoleClientId
+    ? []
+    : [
+        {
+          icon: <SvgSidebarPayments />,
+          name: 'Payments',
+          expandedName: 'My payments',
+          to: absoluteRoutes.studio.payments(),
+        },
+      ]),
   {
     icon: <SvgSidebarToken />,
     name: 'Token',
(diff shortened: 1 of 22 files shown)


packages/atlas/src/components/_navigation/SidenavStudio/SidenavStudio.tsx
```diff
@@ -39,12 +39,16 @@ const studioNavbarItems: NavItemType[] = [
...
+  ...(atlasConfig.features.ypp.googleConsoleClientId
+    ? []
+    : [
+        {
+          icon: <SvgSidebarPayments />,
+          name: 'Payments',
+          expandedName: 'My payments',
+          to: absoluteRoutes.studio.payments(),
+        },
+      ]),
```
drillprop: I don't think we should just hide the payments from the navigation. On gleev we still have payments in navigation, although it's not working - we have a placeholder screen there.
image

Removing the payments nav item together with the placeholder screen could suggest to the users that we abandoned the payments feature when we release YPP.

drillprop (changes requested):
Unfortunately, this is not enough. If I switch the environment to atlas-dev, then I connect to the wallet and go to the studio, the app will break.
image

Remember that we want to use the dev environment for testing.

Besides that, we should also change the targetDevEnv inside `packages/atlas/src/providers/environment/store.ts`

dmtrjsg:
+ there's one event that is taken from Ephesus and it needs to be removed from Graph QL

packages/atlas/src/components/_navigation/SidenavStudio/SidenavStudio.tsx
drillprop: You brought back the placeholder view but didn't bring back the nav item in the sidebar.

drillprop (changes requested):
Also, let's update the targetDevEnv inside `packages/atlas/src/providers/environment/store.ts` to 'development'. 

This will set the `development` environment as default(currently it's `next`)

drillprop (approved):
LGTM

Issue #3700, opened 2023-01-30
Payments: Remove Member wallet widget


dmtrjsg:
## Context

Link to design issue:
- https://github.com/Joystream/atlas/issues/3635

[LInk to Figma> ](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=1547%3A19291&t=6unuV24KglrUpI25-1)

## Scope
- [ ] Layout change for widgets with two in a row + Removed membership balance. ℹ️ No need to show avatar on the channel balance widget.
- [ ] On the first time visit we should display the [welcome banner>](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=2648%3A29515&t=6unuV24KglrUpI25-1) 
- [ ] Placeholder screen for cases when no transactions yet, [link to Figma> ](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=2648%3A29634&t=6unuV24KglrUpI25-1)
☝️ ⚠️ Copy update required for the placeholder component: 
Here you will see proceedings to your channel balance from sold NFTs and royalties, claimed council rewards, direct payments from members to your channel and withdrawals from channel balance. 
<img width="324" alt="Screenshot 2023-01-30 at 17 50 06" src="https://user-images.githubusercontent.com/93924525/215554765-7d21deab-fa58-4bec-a083-2c96860e820c.png">

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1203868726963296/1203920964981336) by [Unito](https://www.unito.io)

dmtrjsg:
Tested, widget removed ✅ 

<img width="1300" alt="Screenshot 2023-02-07 at 20 12 32" src="https://user-images.githubusercontent.com/93924525/217300057-0ae7fa94-86c4-45d2-9988-c5615057df3b.png">

PR #3718, opened 2023-02-02
🏴󠁧󠁢󠁷󠁬󠁳󠁿 Payments widgets changes

WRadoslaw:
#3700

### DIFF
--- a/packages/atlas/src/components/Table/Table.styles.ts
+++ b/packages/atlas/src/components/Table/Table.styles.ts
@@ -58,10 +58,8 @@ export const EmptyTableContainer = styled.div`
 `
 
 export const EmptyTableHeader = styled(Text)`
-  h5 {
-    margin-top: ${sizes(10)};
-    text-align: center;
-  }
+  margin-top: ${sizes(10)};
+  text-align: center;
 `
 
 export const EmptyTableDescription = styled(Text)`
--- a/packages/atlas/src/components/Table/Table.styles.ts
+++ b/packages/atlas/src/components/Table/Table.styles.ts
@@ -47,3 +47,22 @@ export const Td = styled(Text)`
 export const StyledPagination = styled(Pagination)`
   margin: ${sizes(6)} 0;
 `
+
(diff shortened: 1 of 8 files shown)


attemka:
seems that the `PayoutsWelcomeDialogContent` should be changed according to the [figma designs](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=321%3A203&t=IH98CI57oMlkypDC-0)
image

packages/atlas/src/components/Table/Table.styles.ts
```diff
@@ -47,3 +47,22 @@ export const Td = styled(Text)`
...
+
+  p {
+    text-align: center;
+    margin-top: ${sizes(2)};
+  }
+
+  h5 {
+    margin-top: ${sizes(10)};
+    text-align: center;
+  }
```
drillprop: I'm probably nitpicking here, but I would avoid targeting tags in CSS whenever we can. Let's just create: 
```tsx
const EmptyTableTitle = styled(Text)`
// styles
`
const EmptyTableDescription = styled(Text)`
// styles
`
```

attemka: @drillprop agree, would be easier to navigate in component by reading the styleds names

attemka (approved):

packages/atlas/src/components/Table/Table.styles.ts
```diff
@@ -47,3 +47,24 @@ export const Td = styled(Text)`
...
+  justify-items: center;
+  margin: ${sizes(10)} auto;
+  padding: 0 ${sizes(4)};
+`
+
+export const EmptyTableHeader = styled(Text)`
+  h5 {
+    margin-top: ${sizes(10)};
+    text-align: center;
+  }
```
drillprop:
```suggestion
  margin-top: ${sizes(10)};
  text-align: center;
```

drillprop (changes requested):

drillprop (approved):

Issue #3701, opened 2023-01-30
Payments history: Updated Table structure


dmtrjsg:
## Context

Link to design issue:
- https://github.com/Joystream/atlas/issues/3635

[LInk to Figma> ](https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?node-id=1547%3A19291&t=6unuV24KglrUpI25-1)

## Scope
1. Change of structure: date/ type/ sender/ amount and description
☝️ Type requires new icons, different for every type. 
For payments from direct members, us channel avatar. For council rewards use the "avatar" used in the designs. For NFTs, display avatar of the member or channel who purchased the NFT. 
2. Create the new pagination component for the table, but keep the pagination on the client side.

PR #3736, opened 2023-02-06
👔 New table footer

WRadoslaw:
#3701

### DIFF
--- a/packages/atlas/src/components/Table/Table.tsx
+++ b/packages/atlas/src/components/Table/Table.tsx
@@ -1,18 +1,21 @@
+import { useState } from 'react'
 import { Column, usePagination, useTable } from 'react-table'
 
+import { TablePagination } from '@/components/TablePagination'
 import { Text } from '@/components/Text'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
 
-import { StyledPagination, TableBase, Td, Th, Thead, Wrapper } from './Table.styles'
+import { TableBase, Td, Th, Thead, Wrapper } from './Table.styles'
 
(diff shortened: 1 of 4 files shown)


packages/atlas/src/components/Table/Table.tsx
```diff
@@ -20,15 +20,16 @@ export type TableProps<T = object> = {
...
   emptyState?: {
     title: string
     description: string
     icon: ReactElement
   }
+  initialPageSize?: number
 }
 
-export const Table = <T extends object>({ columns, data, title, pageSize = 20, emptyState }: TableProps<T>) => {
+export const Table = <T extends object>({ columns, data, title, initialPageSize = 20, emptyState }: TableProps<T>) => {
```
attemka: I think it would be better to have this default page size in a different const

packages/atlas/src/components/Table/Table.tsx
```diff
@@ -96,8 +97,14 @@ export const Table = <T extends object>({ columns, data, title, pageSize = 20, e
         </EmptyTableContainer>
       ) : null}
 
-      {data.length > pageSize && (
-        <StyledPagination onChangePage={gotoPage} page={pageIndex} itemsPerPage={pageSize} totalCount={data.length} />
+      {data.length < pageSize && (
```
attemka: so if I'll have 30 payments with 20 perPage limit I wont be able to paginate through them?

WRadoslaw: Leftover after enforcing pagination to show for development

attemka (approved):

Issue #3731, opened 2023-02-05
Update the Popover component


toiletgranny:
The goal of this task is to update the Popover component based on the designs which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=510%3A89428&t=ajscHzTgfZgedU13-4).

Changelog:
- Set all Button component instances' width to fluid in the Popover footer.

Please note that since this design component hasn't been updated in a long time, its documentation is not aligned with our current standards. That's why the changelog is included above.

![image](https://user-images.githubusercontent.com/35307309/216811706-d95fc281-5667-4ae4-9abe-78d0dea28d3e.png)

Issue #3730, opened 2023-02-03
Update the NFT tile details component


toiletgranny:
The goal of this task is to update the NFT tile details component based on the designs, docs, and changelog which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3630%3A138998&t=Uei2UCINRrNXX4FR-4).

![image](https://user-images.githubusercontent.com/35307309/216668166-4f7a5eb9-59f0-4cc5-be5d-f40a7db82e70.png)

Issue #3673, opened 2023-01-26
Introduce a Filter button component


toiletgranny:
The goal of this task is to drop the Badge property from the implementation of the Button component, whose sole purpose was to serve the filtering use case, and instead create a standalone Filter button component based on the designs and docs which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=11315%3A214123&t=kPcjYHdGlYA4UbwZ-4).

Since this task requires updating a legacy component, I suggest it be tackled by @drillprop, as he knows the history of the Filter button component.

![image](https://user-images.githubusercontent.com/35307309/214811405-e30040b9-44ff-46ba-9a66-0d2233e174c3.png)

Issue #3672, opened 2023-01-26
Introduce a Toggle button group component


toiletgranny:
The goal of this task is to introduce a Toggle button group component based on the designs and docs which [can be found in Figma](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=11292%3A213721&t=kPcjYHdGlYA4UbwZ-4).

![image](https://user-images.githubusercontent.com/35307309/214777934-e8321511-c372-4466-bb2d-01b186235239.png)

PR #3747, opened 2023-02-08
✖️ Grouped design update

WRadoslaw:
#3665 
#3730 
#3731 
#3663 
#3666 
#3672 
#3673


packages/atlas/src/components/Banner/Banner.tsx
```diff
@@ -24,11 +48,11 @@ export const Banner: FC<BannerProps> = ({ title, description, className, icon, d
   }
 
   return (
-    <BannerWrapper className={className}>
+    <BannerWrapper size={size} className={className}>
       {title && (
         <BannerHeader>
           {icon && <IconWrapper>{icon}</IconWrapper>}
-          <BannerText as="h3" variant="h400">
+          <BannerText as="h3" variant="h400" color="colorTextStrong">
```
drillprop: the variant should be h300

packages/atlas/src/components/Banner/Banner.tsx
```diff
@@ -24,11 +48,11 @@ export const Banner: FC<BannerProps> = ({ title, description, className, icon, d
...
             {title}
           </BannerText>
           {dismissibleId && (
```
drillprop: Probably we should also fix the spacing between the text and the close button. The design has a space between the close button and a block of text. 
image
In our implementation close button is above text:
image

packages/atlas/src/components/_overlays/Dialog/Dialog.styles.ts
```diff
@@ -111,22 +111,31 @@ export const Footer = styled.div<FooterProps>`
...
-    margin-left: ${sizes(2)};
     display: grid;
-    justify-content: flex-end;
+    ${(props) =>
+      !props.stretchButtons &&
+      `
+       margin-left: ${sizes(2)};
+       justify-content: flex-end; 
+    `}
+
```
drillprop: I would move this to the `css` block - we have this our style guide: https://github.com/Joystream/atlas/blob/master/docs/styleguide.md#conditional-styles

packages/atlas/src/components/_overlays/DialogPopover/DialogPopover.tsx
```diff
@@ -48,6 +48,7 @@ const _DialogPopover: ForwardRefRenderFunction<PopoverImperativeHandle, DialogPo
         primaryButton={primaryButton}
         secondaryButton={secondaryButton}
         size="compact"
+        stretchButtons
       >
         {children}
       </StyledDialog>
```
drillprop, line 54: Could we add a story with just one button? It was hard for me to test that, I had to manually edit the popover in the code.

packages/atlas/src/components/_buttons/Button/Button.styles.ts
```diff
@@ -52,40 +52,46 @@ const buttonVariantStyles = ({ variant, iconOnly }: ButtonBaseStyleProps): Seria
       return css`
         background-color: ${cVar('colorCoreBlue500')};
         color: ${cVar('colorTextStrong')};
+        box-shadow: inset 0 0 0 1px ${cVar('colorCoreNeutral800Lighten')};
 
         &:hover,
         &:focus {
-          background-color: ${cVar('colorCoreBlue700')};
+          background-color: ${cVar('colorCoreBlue400')};
+          box-shadow: inset 0 0 0 1px ${cVar('colorCoreNeutral600Lighten')};
```
drillprop: As part of the button update, we should also update all the colors to colors that are using descriptive names. So for example `colorCoreBlue400` should be  `colorBackgroundPrimary`. We didn't have this updated because when we were applying the first design tokens to the button we didn't have tokens with descriptive names. 

Why this is important? 
If the designer changed the background's primary color to something else for example red color, we would need to find every instance of colorBlue400 and replace it with colorRed400.

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <Anchor>
+        {size === 'medium' && isOverflowing && (
+          <StyledButton
+            onClick={handleArrowScroll('left')}
+            size="small"
+            variant="tertiary"
+            icon={<SvgActionChevronL />}
+          />
+        )}
+      </Anchor>
```
drillprop: In this place should be some kind of gradient effect. You didn't implement it. We have something similar in the tabs component, not sure if it's easily applicable here as well. 
Design:
image
Yours: 
image

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <OptionWrapper
+        onMouseDown={onMouseDown}
+        ref={optionWrapperRef}
+        shadowSide={shadowsVisible.left ? 'left' : 'right'}
+      >
+        {options.map((option) => (
+          <Button
```
drillprop: Buttons in the group should have size `small`

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+            key={option}
+            fullWidth={size === 'small'}
+            variant={option !== value ? 'tertiary' : 'secondary'}
+            onClick={() => onChange(option)}
+          >
+            {capitalizeFirstLetter(option)}
```
drillprop: Is this needed actually? I know that we usually capitalize letters in our buttons, but on the other hand, you didn't add `capitalizeFirstLetter` in the button component, and there is no mention of this in the design.

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+            icon={<SvgActionChevronL />}
+          />
+        )}
+      </Anchor>
+      <OptionWrapper
+        onMouseDown={onMouseDown}
+        ref={optionWrapperRef}
+        shadowSide={shadowsVisible.left ? 'left' : 'right'}
+      >
+        {options.map((option) => (
```
drillprop: When I removed two buttons in storybook I still see the arrows. Not sure if it's caused by the storybook. 
If yes, could you add the variant with two buttons to easily test this?

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const ToggleButtonGroup = <T extends string>({
+  label,
+  size,
+  options,
+  value,
+  onChange,
+}: ToggleButtonGroupProps<T>) => {
+  const optionWrapperRef = useRef<HTMLDivElement>(null)
+  const [isOverflowing, setIsOverflowing] = useState<boolean>(false)
+  const { onMouseDown } = useDraggableScroll(optionWrapperRef, { direction: 'horizontal' })
```
drillprop, line 31: I'm getting errors in the console:
image
I think it's related to this library, not sure if it's easily fixable.

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.styles.ts
```diff
@@ -0,0 +1,39 @@
...
+import { cVar, sizes } from '@/styles'
+
+import { ToggleButtonGroupProps } from './ToggleButtonGroup'
+
+export const Container = styled.div<{ size: ToggleButtonGroupProps<''>['size'] }>`
+  display: grid;
+  grid-template-columns: auto auto 1fr auto;
+  padding: ${sizes(1)};
+  gap: ${sizes(1)};
+  border: 1px solid ${cVar('colorCoreNeutral800Lighten')};
```
drillprop: you should use here `Border/Default/Muted/Alpha`

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }, [isOverflowing])
+
+  const handleArrowScroll = (direction: 'left' | 'right') => () => {
+    const optionGroup = optionWrapperRef.current
+    if (!optionGroup || !isOverflowing) {
+      return
+    }
+
+    const addition = (direction === 'left' ? -1 : 1) * (optionGroup.clientWidth / 2)
+    optionGroup.scrollLeft = optionGroup.scrollLeft + addition
```
drillprop: Not super required, but is it possible to apply smooth scrolling here?

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+            fullWidth={size === 'small'}
+            variant={option !== value ? 'tertiary' : 'secondary'}
+            onClick={() => onChange(option)}
+          >
+            {capitalizeFirstLetter(option)}
+          </Button>
+        ))}
+      </OptionWrapper>
+      <Anchor>
+        {size === 'medium' && isOverflowing && (
```
drillprop: I don't understand how sizes work with arrows here. For instance, we should turn arrows off if the size is small? What if the buttons overflow? That doesn't make sense to me. Could you consult this with a designer? 

image

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { capitalizeFirstLetter } from '@/utils/misc'
+
+import { Anchor, Container, Label, OptionWrapper } from './ToggleButtonGroup.styles'
+
+export type ToggleButtonGroupProps<T extends string> = {
+  options: T[]
+  value?: T
+  label?: string
+  size?: 'small' | 'medium' | 'large'
+  onChange: (type: T) => void
```
drillprop: I would add `className` here as well.

packages/atlas/src/components/FilterButton/FilterButton.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+      },
+    ],
+    label: 'Label',
+    icon: <SvgActionShoppingCart />,
+  },
+} as Meta<FilterButtonProps>
+
+const Template: StoryFn<FilterButtonProps> = (args) => {
+  const [value, setValue] = useState<number[]>([])
+  return <FilterButton {...args} selected={value} onApply={setValue} />
```
drillprop, line 29: I don't see the counter when I'm selecting items. 
Not sure, but this could be caused by the storybook. Could you make this work?

packages/atlas/src/components/FilterButton/FilterButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { CheckboxProps } from '@/components/_inputs/Checkbox'
+import { CheckboxGroup } from '@/components/_inputs/CheckboxGroup'
+import { DialogPopover } from '@/components/_overlays/DialogPopover'
+
+export type FilterButtonProps = {
+  options: CheckboxProps[]
+  selected?: number[]
+  onApply: (ids: number[]) => void
+  label?: string
+  icon?: ReactNode
```
drillprop, line 13: I would add className here as well.

packages/atlas/src/components/FilterButton/FilterButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export const FilterButton: FC<FilterButtonProps> = ({ label, icon, options, onApply, selected = [] }) => {
```
drillprop: shouldn't we apply this new component to current filters? For instance in NftsView. If not could you create issue to add them to our views?

packages/atlas/src/components/FilterButton/FilterButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+    onApply([])
+    triggerRef.current?.click()
+  }
+
+  return (
+    <DialogPopover
+      trigger={
+        <Button
+          ref={triggerRef}
+          icon={selected?.length ? <Counter>{selected.length}</Counter> : icon}
```
drillprop, line 47: The Icon should always have the same color: `Text/Default/Muted`. I think we could create StyledButton here instead and safely apply: 
```css
  > svg > path {
    fill: colorMuted
  }
```

drillprop (changes requested):

packages/atlas/src/components/FilterButton/FilterButton.stories.tsx
WRadoslaw: Filters get applied after `Apply` click

drillprop: Oh yeah, my bad 👍🏻  - didn't check that

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
WRadoslaw: Question is how you removed these buttons? Pre or post render? Overflow check fire only once on render.

drillprop: I removed them by clicking `(-)` button in the `options` controls
image 
I'm not sure how the storybook behaves in this case. It thought that it should rerender, but it seems not(?)

packages/atlas/src/components/Banner/Banner.tsx
```diff
@@ -24,41 +49,50 @@ export const Banner: FC<BannerProps> = ({ title, description, className, icon, d
...
-            />
+          {actionButton && (
+            <ActionButton {...actionButton} variant="primary" _textOnly>
+              {actionButton.text}
+            </ActionButton>
           )}
-        </BannerDescription>
-      )}
+        </div>
+        {dismissibleId && (
```
drillprop, line 77: image
There is doubled CloseButton on no title variant

packages/atlas/src/components/FilterButton/FilterButton.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+import { cVar } from '@/styles'
+
+export const Counter = styled.div`
+  height: 16px;
+  width: 16px;
+  background-color: ${cVar('colorBackgroundPrimary')};
+  border-radius: 50%;
+  text-align: center;
+  font: ${cVar('typographyDesktopT100')};
+  line-height: 16px !important;
```
drillprop: Why do we need to change the line height here and why is this set to `!important`? I removed the property and I don't see the difference. The computed value is also 16px. 

image

packages/atlas/src/components/_inputs/ToggleButtonGroup/ToggleButtonGroup.tsx
drillprop: I don't think this still works correctly: 
this is how buttons are displayed in sizes `small` and `medium. 
image

packages/atlas/src/components/Tabs/Tabs.styles.ts
```diff
@@ -20,7 +20,7 @@ type TabsGroupProps = {
   }
 }
 
-const getMaskImage = ({ shadowsVisible }: TabsGroupProps) => {
+export const getMaskImage = ({ shadowsVisible }: TabsGroupProps) => {
```
drillprop: Let's move it to the helpers or utils if this is used in two different components.

packages/atlas/src/components/_buttons/Button/Button.tsx
```diff
@@ -64,10 +66,11 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>((props, ref) =>
       disabled={disabled}
       aria-disabled={disabled}
       variant={variant}
-      iconOnly={!!iconOnly}
+      iconOnly={iconOnly}
```
drillprop: Missed that before, but it looks like iconOnly button variant is oversized: 
large: 56px vs 48
medium: 46px vs 40 
small: 34px vs 32

drillprop (changes requested):

attemka (commented):
@WRadoslaw  [navigation arrows](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=11292%3A213721&t=3ftGnFWy1HT5HlyH-0) for ToggleButtonGroup are missing?
image

WRadoslaw:
> @WRadoslaw [navigation arrows](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=11292%3A213721&t=3ftGnFWy1HT5HlyH-0) for ToggleButtonGroup are missing? image

Nope, just need to reload storybook after switch to medium. Check to see if button overflow is fired once on render.

drillprop (changes requested):
I still see one inconsistency in the toggle button group.
The padding on the left is bigger than on the right. This makes the component asymmetric.

image

image
image

Possibly this gap was meant for the label. Is the label required and will it be always present in ToggleButtonGroup?

drillprop (approved):

attemka:
@WRadoslaw final [change](https://discord.com/channels/811216481340751934/1035122951126798386/1077563519836373002) in this ticket

PR #3766, opened 2023-02-10
Fix issue with not being able to opt out

drillprop:
The Ypp backend API changed recently and we should send timestamps as iso strings, not a number to update settings and opt-out.


drillprop:
Not ready yet, we're still sending the wrong data when we're opting out. Checking this with Zeeshan

drillprop:
should work now

attemka (approved):

Issue #3754, opened 2023-02-09
Update T&Cs styling


drillprop:
## Context

Finalising T&Cs

## Scope

- [ ] Update headers structure to 

h1: Terms & Conditions
h2: Key Information; YPP
h3: Section headers

- [ ] Two diff font sizes used in markup and Tiers component 14/16px.  Needs to be 16 everywhere.

___________

✅ Already merged in a separate pR:
- [ ] Make hyperlinks clickable: 
![image](https://user-images.githubusercontent.com/51168865/217753536-6c6c7e11-5de5-49b6-a7d2-e660cdd09433.png)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1203868726963296/1203935643509260) by [Unito](https://www.unito.io)

dmtrjsg:
@WRadoslaw hey sorry to jump in, but there was also the point about the headers sctructure we discussed on the standup that was meant to go to the same issue 😬 

I am also awaiting Zeeshan's confirmation re mandatory scopes, and if confirmed we may need to update the YT sync subsection. This is now pending, based on the outcome of: 
- https://github.com/Joystream/youtube-synch/issues/144

Issue #3723, opened 2023-02-03
Remove Payouts widget from dashboard


dmtrjsg:
ssia

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1203868726963296/1203891363354819) by [Unito](https://www.unito.io)

dmtrjsg:
1️⃣  ~When payout widget was hidden, the remaining two remained with the original positioning on the page. I was expecting the same treatment for the widgets as we have in the Payments - stretching both to occupy the full row.. Is it worth fixing?~

<img width="1322" alt="Screenshot 2023-02-09 at 16 59 33" src="https://user-images.githubusercontent.com/93924525/217819906-1d8e8f78-46ec-4b24-a487-36cc519a62b6.png">


2️⃣ 
Also super small one: can we pls change the copy on the payout banner to 

"Payouts are made on a weekly basis, every Friday, for the previous calendar week. Your first payment will involve the reward for the sign up of 5K JOY after your channel is verified."

Otherwise it is split into 2 lines unecessarily..

PR #3769, opened 2023-02-10
TnC and banners fix

attemka:
https://github.com/Joystream/atlas/issues/3754
https://github.com/Joystream/atlas/issues/3540
https://github.com/Joystream/atlas/issues/3723


dmtrjsg (approved):
lezzgo

dmtrjsg:
@attemka 

Looking good, few small things: 

T&Cs
- Rename T&Cs to "Terms and Conditions"

-Remove "Key information" header from Under Youtube Partnership Programme, since we are going with only 2 levels of headers, this either needs to be changed to 3 levels or simply removed (which is prob better). 

Everything else looks perfect 👍

PR #3777, opened 2023-02-13
Fix displaying wrong font color in VideoTileDetails

### LINKED ISSUE #3765: Views number has the wrong color
image
https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=3630%3A116416&t=Z2hdToPjAYQIYKL3-4

It should have `Text/Default/Default`

drillprop:
Fix #3765

### DIFF
--- a/packages/atlas/src/components/NumberFormat/NumberFormat.tsx
+++ b/packages/atlas/src/components/NumberFormat/NumberFormat.tsx
@@ -61,12 +61,7 @@ export const NumberFormat = forwardRef<HTMLHeadingElement, NumberFormatProps>(
       withTooltip &&
       ((format === 'short' && (internalValue > 999 || hasDecimals)) || (format === 'dollar' && hasDecimals))
     const content = (
-      <StyledText
-        {...textProps}
-        color={isNegative || bnValue.isNeg() ? 'colorTextError' : undefined}
-        variant={variant}
-        ref={mergeRefs([ref, textRef])}
-      >
+      <StyledText {...textProps} variant={variant} ref={mergeRefs([ref, textRef])}>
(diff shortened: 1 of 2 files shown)


WRadoslaw (commented):
I don't see how this PR is related to the issue

drillprop:
> I don't see how this PR is related to the issue

Doesn't this PR fix this? I think it does.

So the bug was introduced in this PR: 
https://github.com/Joystream/atlas/pull/3646/files#diff-6675b318866edea32f9b06aaee584c0967c82e8bd38613fbe09e702ec2e073e0 

These two lines were making a difference: 
```
        {...textProps}
        color={isNegative || bnValue.isNeg() ? 'colorTextError' : undefined}
```
If you look closely you'll notice that we were overwriting the `color` prop. The textProps already contained color, and then we were adding the property once again. No matter what color you passed to `NumberFormat` the value was always either `colorTextError` or `undefined`. 
Then, If the color was undefined, we were picking `colorTextStrong` as default, because that's how it's set in the Text component -https://github.com/Joystream/atlas/blob/43fe320016b150a36f70b8d484041c2a6e9d59cc/packages/atlas/src/components/Text/Text.styles.ts#L50

As a result, we had lighter color on the views counter. Possibly we also had the wrong colors applied on other NumberFormats instances 

Hope that makes sense

drillprop:
@attemka why did you apply `status:need changes` label?

drillprop:
@WRadoslaw as we discussed on the call, I brought back the `colorTextError` change - if the value is negative or prop `isNegative` is passed, we're displaying red color.

attemka (approved):

PR #3866, opened 2023-02-17
🦩 Clear button overflow fix

### LINKED ISSUE #3788: Search's close button overlap on member avatar on smaller screens
The bug:
image

To reproduce: 
1. make sure that your screen is smaller than 1024 px
2. click the button with magnifying glass
3. type something 
4.  hit escape

WRadoslaw:
#3788

### DIFF
--- a/packages/atlas/src/components/Searchbar/Searchbar.styles.ts
+++ b/packages/atlas/src/components/Searchbar/Searchbar.styles.ts
@@ -29,7 +29,7 @@ export const Input = styled.input`
 
 export const CancelButton = styled(Button)`
   position: absolute;
-  right: ${sizes(4)};
+  right: 0;
   border: none;
   margin-right: ${sizes(2)};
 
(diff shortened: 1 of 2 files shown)


drillprop (changes requested):
Actually, this is not how it is supposed to work. Maybe I should be more clear on the issue.
Check the designs:
https://www.figma.com/file/8JhE2GxIjjLP3WU37xPnDH/Search-results-page?node-id=5%3A35603&t=PgJryRQfBH6iI1Zi-4

When the search bar is opened we shouldn't show the member icon and the notification button on the right. 
This should work exactly the same as it works for unregistered users. 

https://user-images.githubusercontent.com/51168865/220046478-d432e77a-d026-4150-aa4c-cd1dc101a99b.mov

drillprop (changes requested):
This is almost correct, but still, I see one issue 😅 
Now you're hiding buttons every time independently of screen size. We should hide the buttons only when the size is smaller than `1024 px`. 
On the bigger screens, we should keep them since there is plenty of space for the search input and the buttons and it doesn't make sense to hide the latter.

drillprop (approved):
japanology-perfect-gif

now it's perfect!

attemka (approved):

Issue #3519, opened 2022-12-12
Wrap channel and video creation in `AppAction` messages


kdembler:
Part of https://github.com/Joystream/joystream/issues/4307

1. Let the app operator provide optional `appId` via `atlas.config.yml`
2. If `appId` is provided, whenever user wants to create a new channel or video, we wrap that extrinsic in `AppAction`, get signature from Orion and send the `AppAction` as extrinsic metadata.

dmtrjsg:
<img width="819" alt="Screenshot 2023-01-09 at 18 08 06" src="https://user-images.githubusercontent.com/93924525/211389144-6d72428b-44b2-408f-97e1-d5f90ffabd7a.png">

PR #3875, opened 2023-02-23
🉑 App actions

WRadoslaw:
#3519


packages/atlas/src/views/studio/VideoWorkspace/VideoWorkspace.hooks.ts
```diff
@@ -34,6 +37,10 @@ export const useHandleVideoWorkspaceSubmit = () => {
   const { joystream, proxyCallback } = useJoystream()
   const startFileUpload = useStartFileUpload()
   const { channelId, memberId } = useAuthorizedUser()
+  const { data: channelVideosCount } = useGetVideoCountQuery({
+    skip: true,
```
attemka: @WRadoslaw will it change later or should it be smth like `!channelId`?

WRadoslaw: Yeah, it should be `!channelId || !atlasConfig.general.appId`

attemka (changes requested):
approved after statement mentioned above is implemented

attemka (approved):

Issue #3874, opened 2023-02-23
Applying filters on 2nd or 3rd page gives you empty results


dmtrjsg:
![Screenshot 2023-02-23 at 17.43.16.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/6cfa1e71-9f36-4c35-b494-4e24a0f85a77)

PR #3883, opened 2023-02-28
🌳 Reset pagination on nft filter apply

WRadoslaw:
#3874

### DIFF
--- a/packages/atlas/src/views/viewer/NftsView/NftsView.tsx
+++ b/packages/atlas/src/views/viewer/NftsView/NftsView.tsx
@@ -134,7 +134,11 @@ export const NftsView: FC = () => {
           </div>
           {smMatch && sortingNode}
         </HeaderContainer>
-        <FiltersBar {...filtersBarLogic} activeFilters={['nftStatus', 'date-minted', 'other']} />
+        <FiltersBar
+          {...filtersBarLogic}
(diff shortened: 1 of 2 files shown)


packages/atlas/src/views/viewer/NftsView/NftsView.tsx
```diff
@@ -134,7 +134,11 @@ export const NftsView: FC = () => {
           </div>
           {smMatch && sortingNode}
         </HeaderContainer>
-        <FiltersBar {...filtersBarLogic} activeFilters={['nftStatus', 'date-minted', 'other']} />
+        <FiltersBar
```
attemka, line 137: should we add the `onAnyFilterSet` for the other `FiltersBar` component usages in a separate PR?

attemka (approved):

packages/atlas/src/views/viewer/NftsView/NftsView.tsx
WRadoslaw: I think so, most of the components that use FilterBar are using it in a way that makes it impossible to handle it with just onAnyFilterChange tho. 

Flow is reversed, so Filterbar is implemented in the parent component while pagination lives strictly inside children (most often it's tab content), so the only way I see to enforce pagination reset on children when we have an event in the parent is either `useEffect` when `totalCount / filters` change or using `useImperativeHandle`.

PR #3891, opened 2023-03-03
changed query for fetching distribution buckets

### LINKED ISSUE #3886: Video page takes 15 seconds to load
## Context
New videos take 15-20s  to load, which is in the dead zone from usability pov:
- https://gleev.xyz/video/917

The performance was not great last week (circa 4-6 seconds to load) but deteriorated dramattically. 

We've looked into it on the standup and concluded that fetching distributor node details takes 15 sec. Bad requests were found on QN logs. Orion instance was moved recently so that may have contributed to it. 

Issue requires further investigation. 

## Scope
- [ ] Investigate and define the root cause of the issue
- [ ] Suggest treatment on how to fix it

attemka:
fixes #3886


drillprop (approved):
Looks good to me!

WRadoslaw (approved):

PR #3903, opened 2023-03-07
ytSynch thumbnails fix

### LINKED ISSUE #3762: Synced videos still appearing as failed videos
As the title says. 

Video tiles with variant `auto-sync` are displayed for a few seconds and then, once the video is uploaded, the tiles are marked as `upload failed`. 

After the page refresh tiles are correct - i.e. they have status uploaded and you see the correct thumbnail and you can watch the video.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1203868726963296/1203945359351603) by [Unito](https://www.unito.io)

attemka:
#3762 
when user has a YT video being synced via ytSynch, there is a short period of time (~5s), when video and thumbnail are already uploaded, but not accepted yet. Assuming ytSynch assets will pass the requirements, I've removed the asset upload error for ytSynch videos


packages/atlas/src/components/_video/VideoTilePublisher/VideoTilePublisher.tsx
```diff
@@ -121,10 +119,13 @@ export const VideoTilePublisher: FC<VideoTilePublisherProps> = memo(
     const hasThumbnailUploadFailed =
       (video?.thumbnailPhoto &&
         !video.thumbnailPhoto.isAccepted &&
-        uploadThumbnailStatus?.lastStatus !== 'completed') ||
+        uploadThumbnailStatus?.lastStatus !== 'completed' &&
+        !video.ytVideoId) ||
```
drillprop, line 123: image

So this is how it should work? For some time user won't be able to see his thumbnail at all? This is far from perfect to me. 

Also, what will happen if the user decides to change the thumbnail on the atlas side, he'll update the new image and the upload fails? He won't be able to know what's happening, because `hasThumbnailUploadFailed` won't be false if the ytVideoId is set.

attemka: resolved on the call

drillprop (approved):

PR #3906, opened 2023-03-08
livesession tweaks

attemka:


packages/atlas/src/AnalyticsManager.tsx
```diff
@@ -29,7 +29,6 @@ export const AnalyticsManager: FC = () => {
 
     ls.init(atlasConfig.analytics.livesession.id, {
       keystrokes: true,
-      rootHostname: atlasConfig.analytics.livesession.rootHostname,
```
drillprop, line 32: If we don't need it we should remove this from the atlas config and from the config schema.

drillprop (changes requested):

drillprop (approved):

Issue #3908, opened 2023-03-08
Refactor useNftsConnection query


WRadoslaw:
Issue https://github.com/Joystream/atlas/issues/3893 is resolved by #3909 only partially. 

The current implementation in the app utilizes a cursor approach to fetch next results, which is good for infinite loading and similar, but not perfect when it comes to normal pagination, as we allow user to jump straight to the last page.

Assuming we first fetch 10 of total 1000 NFTs, if user clicks the last page button, then we have no other option than just to exec the same query almost 100 times or fetch 990 items inside one query (we can only operate on cursors, but we don't know them before they are already fetched).

Since we don't ever display NFTs in an infinite list I suggest switching from the `ownedNftsConnection` query to just `ownedNfts` based on `offset` and `limit`.

@attemka @drillprop Lemme know in case I'm missing something and if this could be fixed easier.

Issue #3893, opened 2023-03-06
NFTs are not loading when you go to the last page


drillprop:
Steps to reproduce. 
1. go to https://gleev.xyz/nfts
2. click the last page on the pagination component
3. Nfts are not loading and the browser sends a lot of GetNftsConnection requests

https://user-images.githubusercontent.com/51168865/223103425-3cab500f-aeb1-4d67-97a9-c2d4c3f09436.mov

It's worth checking how this is working in other places where we use the Pagination component

PR #3909, opened 2023-03-08
🤼 Nft last page issue

WRadoslaw:
#3893 
#3908


packages/atlas/src/views/viewer/NftsView/NftsView.tsx
```diff
@@ -78,20 +78,25 @@ export const NftsView: FC = () => {
...
+      }
+      if (totalCount <= nfts.length) {
+        return
+      }
+
+      const dataNeededForPage = pageToCheck * tilesPerPage + tilesPerPage
+
+      if (dataNeededForPage > nfts.length) {
+        fetchMore({
+          variables: { ...variables, after: pageInfo?.endCursor, first: dataNeededForPage - nfts.length },
```
drillprop: We don't want to fetch all the nfts when you go to the last page.
Here's the request, when you click the last page. It took 16 seconds to load:
image
image

I think we should use here `ownedNfts` because it allows offset-base pagination.

I'm wondering why we used the cursor-based pagination here initially, though. I think there was a reason why ownedNftsConnection is used here, but can't recall why 🤔

drillprop (changes requested):

packages/atlas/src/views/viewer/NftsView/NftsView.tsx
drillprop: Oh, my bad. I didn't notice that you were aware of the pagination issue and you already created the issue: https://github.com/Joystream/atlas/issues/3908 

Anyway, should we apply the fix that you include in this PR anyway? I mean, It kinda fixes the issue, but it's just a workaround(and this workaround introduces another issue). Should we just move to offset-based pagination already? 
@attemka I need another pair of eyes here, wdyt?

packages/atlas/src/hooks/useNfts.ts
```diff
@@ -0,0 +1,26 @@
...
+  })
+
+  const { data: nftData, ...rest } = useGetNftsQuery(baseOptions)
+
+  return {
+    nfts: nftData?.ownedNfts,
+    totalCount: paginationData?.ownedNftsConnection.totalCount,
+    ...rest,
+  }
+}
```
drillprop: We should move this to `packages/atlas/src/api/hooks/nfts.ts`

packages/atlas/src/utils/data.ts
```diff
@@ -17,3 +17,17 @@ export const promisify =
   <T>(fn: (...args: unknown[]) => T) =>
   (...args: Parameters<typeof fn>) =>
     new Promise((resolve) => resolve(fn(...args))) as Promise<T>
+
+export const repeat = <T = undefined>(count: number, data?: T) => Array.from({ length: count }, () => data)
```
drillprop: I'm not too fond of the name of the function. I had to double-check what it does. Also, we're creating a video platform here, so repeat could be mistaken with some other function related to video playback. 
Maybe let's just name it `arrayFrom` or `arrayFromLength`? This is the first thing that comes to my mind.

drillprop (changes requested):
Found two minor things, besides that great work, a really exquisite solution for the problem! 👍🏻

drillprop (approved):
LGTM!

packages/atlas/src/components/_comments/CommentEditHistory/CommentEditHistory.tsx
```diff
@@ -19,7 +20,7 @@ export const CommentEditHistory: FC<CommentEditHistoryProps> = ({ originalCommen
   const { commentEdits, loading } = useCommentEdits(originalComment?.id)
   const { url: memberAvatarUrl, isLoadingAsset } = useMemberAvatar(originalComment?.author)
 
-  const placeholderItems = Array.from({ length: 3 }, () => ({ id: undefined }))
+  const placeholderItems = createPlaceholderData(3)
```
attemka, line 23: IMO it would be great to use the const for this number like its implemented in `ExpandableChannelsList.tsx`. Also applied to the rest of the hardcoded numbers across the PR. wdyt?

attemka (commented):
approved besides minor suggestion

PR #3955, opened 2023-03-20
Fix issue with redundant requests when YPP url is not set

### LINKED ISSUE #3951: What will happen with Atlas when YPP params will not get set
What will happen with Atlas when the following are not provided:

1. Youtube Backend URL
2. Youtube Collaborator member
3. Google Console API ID

_ 

Check if YPP module shows up
Check the studio
Check the sign up and general navigation

ER: YPP content will be disabled if Collaborator ID is not set

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1203868726963296/1204216520502039) by [Unito](https://www.unito.io)

drillprop:
Fix #3951 

I also handled the error, when the collaborator member id is not set.


drillprop:
~Draft. We shouldn't show `ConnectWithYtModal` as well~

attemka (approved):

Issue #3775, opened 2023-02-13
Marketplace (Hero) Carousel


dmtrjsg:
## Context
- https://github.com/Joystream/atlas/issues/3686

## Scope

AC:
- If there are less than 2 video NFTs in the Featured section, Marketplace carousel card components should be left-aligned.
- Card States: Default + Hover; Hover Transition link [> ](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)
- Interaction: transitions [link >](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1) 
- Navigation: looped  [link>](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)
- Countdown progress (white line for 30 seconds) [link>](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)
- Loading State (video thumb prev) [link > ](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)
- [x] #3782

🚫 Out of scope:
- Component states: NFT, Verified, Whitelisted [link > ](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)
(this is for NFT Carousel with small thumbs)

## Designs
[Link to FIgma > ](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)

Preview

<img width="641" alt="Screenshot 2023-02-13 at 11 48 45" src="https://user-images.githubusercontent.com/93924525/218400785-30f5d23b-f597-4938-a7ea-a1d969b6e40d.png">

dmtrjsg:
Duplicating arrays 👍

dmtrjsg:
[...]

PR #3959, opened 2023-03-21
🏳️  Nft carousel

WRadoslaw:
#3775 

⚠️ There are still few issues:

- Idk why, but on first page load items on edges (when at item 0 -> item on the left) are displaying as skeletons even tho they are loaded. To showcase:
1. Go into playground
2. Choose carousel and hit F5
3. Navigate left (item should be a skeleton and on entry will load normally)
4. After changing to any other playground and coming back to carousel (without F5) same flow as before will work perfectly

- Sometimes there is a gap between fade and video in the background even tho both elements are the same width


packages/atlas/src/api/hooks/nfts.ts
```diff
@@ -174,3 +175,12 @@ export const useNftHistory = (
     ...rest,
   }
 }
+
+export const useFeaturedNfts = () => {
```
WRadoslaw, line 199: This is only a placeholder till orion work is merged

packages/atlas/src/components/Gallery/Gallery.tsx
```diff
@@ -12,14 +13,13 @@ export type GalleryProps = PropsWithChildren<{
...
   className?: string
   seeAllUrl?: string
+  children: ReactNode[] | ReactNode
 }> &
-  CarouselProps
+  Omit<CarouselProps, 'children'>
 
-export const Gallery: FC<GalleryProps> = ({ title, className, seeAllUrl, ...carouselProps }) => {
+export const Gallery: FC<GalleryProps> = ({ title, className, seeAllUrl, children, ...carouselProps }) => {
   // TODO: this is the only place in the app that requires refs to buttons. Once we refactor this component, we can remove forwardRef from buttons
```
WRadoslaw: Remove

packages/atlas/src/components/Gallery/Gallery.tsx
```diff
@@ -42,29 +42,29 @@ export const Gallery: FC<GalleryProps> = ({ title, className, seeAllUrl, ...caro
...
       </GridHeadingContainer>
-      <Carousel
-        {...carouselProps}
-        prevArrowRef={prevArrowRef}
-        nextArrowRef={nextArrowRef}
-        ref={carouselRef}
-        itemWidth={350}
-      />
+      <Carousel {...carouselProps}>
+        {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
```
WRadoslaw: Fix

packages/atlas/src/components/NftCarousel/components/MarketplaceCarouselCard/NftCarouselDetails.tsx
```diff
@@ -0,0 +1,158 @@
...
+                )}
+                {nftDetails.topBid && (
+                  <DetailsContent
+                    tileSize="big"
+                    caption="TOP BID"
+                    content={nftDetails.topBid}
+                    icon={<JoyTokenIcon size={16} variant="regular" />}
+                  />
+                )}
+                <div>{nftDetails.endsAt?.toLocaleString()}</div>
```
WRadoslaw: Implement countdown

packages/atlas/src/components/Carousel/Carousel.tsx
```diff
@@ -1,110 +1,34 @@
...
-import { GliderProps, useGlider } from '@/components/Glider'
-
-import { Container, Dots, GliderContainer, Track } from './Carousel.styles'
+import { StyledSwiper } from '@/components/Carousel/Carousel.styles'
 
+export type SwiperInstance = SwiperType
 export type CarouselProps = {
   className?: string
-  arrowPosition?: number
+  // arrowPosition?: number
```
attemka: lets delete it completely if we'll not going to use it?

attemka (changes requested):
works well, I've just noticed 2 things:
1. keyboard arrow controls doesn't work anymore. I don't think that many users will use that, but lets enable it if its a quick fix.
2. we need to add the callback in the end of the video playback to switch to the next video. Currently video just ends

attemka (approved):

PR #3961, opened 2023-03-22
Release v1.4.0

attemka:


WRadoslaw (approved):

packages/atlas/atlas.config.yml
```diff
@@ -2,6 +2,7 @@ general:
   appName: 'Atlas' # Application name - used in the copy throughout the app, in index.html, open graph meta tags, etc
   appDescription: 'The streaming platform empowering viewers, creators, and builders. Built on and operated by the Joystream blockchain and DAO.' # Application description - used in index.html meta tags
   appTagline: 'The streaming platform empowering viewers, creators, and builders. Built on and operated by the Joystream blockchain and DAO.'
+  appId: '$VITE_APP_ID' # App ID for Apps as first-class citizens
```
drillprop, line 5: the app breaks, because we didn't provide appId.
image

Should we at least add `VITE_APP_ID` with a blank string in the env as a fallback?

drillprop (changes requested):

packages/atlas/atlas.config.yml
drillprop: or maybe that should be null?

PR #3993, opened 2023-03-29
Ephesus QN tests

attemka:


packages/atlas/src/components/_overlays/AdminModal/AdminModal.tsx
```diff
@@ -170,12 +171,7 @@ const EnvTab: FC = () => {
   return (
     <VerticalSpacedContainer>
       <FormField label="Environment">
-        <Select
-          items={environmentsItems}
-          onChange={handleEnvironmentChange}
-          value={targetDevEnv}
-          disabled={BUILD_ENV === 'production'}
```
drillprop: I think we shouldn't remove the disabled flag here. I don't think we should allow people to change the environment.

packages/atlas/src/config/env.ts
```diff
@@ -29,8 +29,6 @@ export const availableEnvs = () => {
 export const readEnv = (name: string, required = true, direct = false): string => {
   const fullName = direct
     ? getEnvName(name)
-    : BUILD_ENV === 'production'
-    ? getEnvName(`PRODUCTION_${name}`)
```
drillprop: same here

drillprop (changes requested):

packages/atlas/src/components/_overlays/AdminModal/AdminModal.tsx
attemka: So we've decided that it'll be nice to let the community do some testing, this is why i've unlocked the env changing feature. Any visible cons of that?

drillprop: This will still be possible, for example, we could share to community URL https://atlas-git-dev-joystream.vercel.app/  or https://atlas-git-master-joystream.vercel.app/ but with the ENV variable set to `development`

I'm worried, that someone will change the environment on gleev, and he'll forget how to switch this back, or he'll be using another environment thinking that this is a production one.

drillprop: and actually, it was possible to switch env on https://atlas-git-dev-joystream.vercel.app/ in the past. I don't understand why we changed that.

dmtrjsg:
@drillprop there is certainly a benefit from flexibility of choosing the env. This is only triggered via shortcut that general users would not know how to. So I don't see a problem in enabling this selector for our team internal purposes.

packages/atlas/src/components/_overlays/AdminModal/AdminModal.tsx
attemka: the difference between these instances is in the categories in the config. They'll be broken in Atlas if I'll update env vars.
Regarding switching env on atlas-dev, it happened because the 'production' env was used to create a prod bundle, and it's also disabling the functionality above

drillprop (approved):
I don't agree with these changes, but if you guys feel we need that, I won't argue 👍🏻  LGTM

Issue #3974, opened 2023-03-24
Showing reqs modal before creating channel


dmtrjsg:
For new members signing up to YPP show requirements modal after "Create channel and sign up" button is clicked, and before redirect to studio.

PR #3994, opened 2023-03-29
👨‍🚀 Show requirements before channel creation

WRadoslaw:
#3974

### DIFF
--- a/packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
+++ b/packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
@@ -368,7 +368,7 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
             text: isFetchingData
               ? 'Please wait...'
               : activeMembership?.channels.length
-              ? 'Select channel'
+              ? 'Continue'
               : 'Create new channel',
             disabled: isFetchingData,
             onClick: () => verifyChannelRequirements(),
--- a/packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
+++ b/packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
@@ -343,10 +343,20 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
     if (data?.unsyncedChannels?.length) {
       setSelectedChannelId(data?.unsyncedChannels[0].id)
     }
-    if (data?.unsyncedChannels?.length && data?.unsyncedChannels.length > 0) {
+    if (data?.unsyncedChannels?.length && data?.unsyncedChannels.length > 1) {
       onChangeStep('select-channel')
+      return
     }
-  }, [activeMembership?.channels, navigate, onChangeStep, refetchYppSyncedChannels, setSelectedChannelId])
+
+    handleAuthorizeClick(data?.unsyncedChannels?.[0].id)
+  }, [
+    activeMembership?.channels,
+    handleAuthorizeClick,
+    navigate,
+    onChangeStep,
+    refetchYppSyncedChannels,
+    setSelectedChannelId,
+  ])
 
   const authorizationStep = useMemo(() => {
     switch (currentStep) {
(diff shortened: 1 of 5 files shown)


attemka (approved):

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -319,15 +324,56 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
...
+          component: <YppAuthorizationRequirementsStep requirements={requirements} />,
+        }
       case 'select-channel':
         return {
           title: 'Select channel',
           description: `Select the ${APP_NAME} channel you want your YouTube channel to be connected with.`,
           primaryButton: {
-            text: 'Select channel',
-            onClick: () => onChangeStep('requirements'),
+            // text: 'Select channel',
```
drillprop:
```suggestion
```

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -319,15 +324,56 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
...
+            text: 'Authorize with YouTube',
+            // onClick: () => onChangeStep('requirements'),
```
drillprop:
```suggestion
```

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -319,15 +324,56 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
...
+    })
+    if (data?.currentChannel) {
+      navigate(absoluteRoutes.studio.ypp())
+      return
+    }
+
+    if (data?.unsyncedChannels?.length) {
+      setSelectedChannelId(data?.unsyncedChannels[0].id)
+    }
+    if (data?.unsyncedChannels?.length && data?.unsyncedChannels.length > 0) {
```
drillprop: As far as I remember we should show `select-channel` step only if the member has more than one channel https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=1637-121102&t=bi75D366Rqaa2LFi-4

packages/atlas/src/views/global/YppLandingView/YppLandingView.tsx
```diff
@@ -46,12 +45,9 @@ export const YppLandingView: FC = () => {
...
   const shouldContinueYppFlow = useYppStore((store) => store.shouldContinueYppFlow)
 
-  const channels = activeMembership?.channels
-
-  const { unsyncedChannels, isLoading, currentChannel, refetchYppSyncedChannels } = useGetYppSyncedChannels()
-
+  const { unsyncedChannels, isLoading, currentChannel } = useGetYppSyncedChannels()
   const isYppSigned = !!currentChannel
-  const hasAnotherUnsyncedChannel = isYppSigned && !!unsyncedChannels?.length
+  const hasAnotherUnsyncedChannel = isYppSigned && !!unsyncedChannels?.length //&& currentChannel.joystreamChannelId === channelId
```
drillprop:
```suggestion
  const hasAnotherUnsyncedChannel = isYppSigned && !!unsyncedChannels?.length
```
for removal

drillprop (changes requested):

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
WRadoslaw: Nice catch! Do you think we should change the button text? Now we always display `Select channel`, even tho if there is only 1 channel we will go straight to google auth. Maybe smth more generic, like `Proceed`?

image

drillprop: `Authorize with YouTube` ? This is what we have in the design at least
image

WRadoslaw: It's not really true, since now it's either Auth or select channel and we don't have information which one till user clicks the button

drillprop: We agreed together that we use `Continue` in this case

drillprop (approved):

Issue #3752, opened 2023-02-08
[USERSNAP] I tried to comment on someones comment while waiting for my prior Like was still processing, i.e. the spinner was stil going.  It turns out, this is not possible, so I got a big dialog saying "you have to wait". I guess that is OK in principl...


dmtrjsg:
**Sender**: None
**Taken at**: https://gleev.xyz/video/472
**Fields**:
- Expected Result: I tried to comment on someones comment while waiting for my prior Like was still processing, i.e. the spinner was stil going.

It turns out, this is not possible, so I got a big dialog saying &#34;you have to wait&#34;. I guess that is OK in principle, at least for now, however the bigger problem is that all my text is now gone!!! I can&#39;t wait and try to submit again when like has been finalied.

FYI: I also think there is a seprate bug of likes often never getting finalized when you are on the screen, it just keeps spinning, only then to be finalized when you refresh.
- Houston, we have a problem.: Comment lost!

![SCREENSHOT](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/408e879e-3521-4433-aff0-49eaa9ff0880-annotated_fe2ce3e8-fc90-4266-a2f1-bc992d062421.png?etag=f21c685f49dc6fd270465b28e3ebedcd)
*[Download original image](https://resources.usersnap.com/company/f9f719df-a929-4461-bb6e-46dbfbceffb7/datapoint_screenshot/408e879e-3521-4433-aff0-49eaa9ff0880-annotated_fe2ce3e8-fc90-4266-a2f1-bc992d062421.png?etag=f21c685f49dc6fd270465b28e3ebedcd)*

**Browser**: Firefox 109 (macOS Catalina)
**Screen size**: 5120x1440
**Browser size**: 1955x1130
**[Open #30 in Usersnap](https://app.usersnap.com/l/feedback/1ece8c25-76cc-46b3-be0e-4cc99fb38a7a?utm_medium=integration&utm_campaign=integration&utm_content=open)**

Powered by **[Usersnap](https://usersnap.com/?utm_source=product&amp;utm_medium=poweredbylink&amp;utm_campaign=github_entry)**.

dmtrjsg:
In this issue let's fix losing the text on the reply

In the new issue: 
- [ ] Queue transactions and let submit the reply/ comment without blocking this tx

dmtrjsg:
[...]

PR #3995, opened 2023-03-29
😺 Make video reactions non-blocking

WRadoslaw:
#3752


attemka (approved):

drillprop (changes requested):
We should remove the ongoing transaction modal as well. It should be inside `transaction.hooks.ts` I don't think we need it anymore.

WRadoslaw:
> We should remove the ongoing transaction modal as well. It should be inside `transaction.hooks.ts` I don't think we need it anymore.

I think it's good to have the user sign action before trying to perform the next one. I guess some people might be confused why nothing is processed with having 3 wallet signing modals open on another monitor or workspace. I know we show the snack bar to proceed in extension, but there is always the possibility that one will somehow miss it.

drillprop (approved):

Issue #3989, opened 2023-03-29
Confusing pagination


traumschule:
https://gleev.xyz/channel/7712?tab=Videos

It looks like the pagination is broken: two pages are selected at the same time.
~If page 8/12 has no videos the user should not be able to select it.~
~If it's a loading in process thing it shouldn't say "No videos in this channel".~

![pagination](https://user-images.githubusercontent.com/31551045/228525070-5d6d5b2f-88d9-4ea9-a13a-e3571a0c07da.png)

drillprop:
Thanks for reporting:
>If page 8/12 has no videos the user should not be able to select it.
>If it's a loading in process thing it shouldn't say "No videos in this channel".

fixed in https://github.com/Joystream/atlas/pull/3985

> It looks like the pagination is broken: two pages are selected at the same time.

Will fix this. This is a minor CSS issue. I would consider using :focus-visible here - https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

PR #4001, opened 2023-03-30
📗 Correct pagination button styles

WRadoslaw:
#3989

### DIFF
--- a/packages/atlas/src/components/Pagination/Pagination.styles.ts
+++ b/packages/atlas/src/components/Pagination/Pagination.styles.ts
@@ -60,10 +60,13 @@ export const PaginationButton = styled(Button)<PaginationButtonProps>`
   user-select: none;
 
   :hover,
-  :focus-visible,
-  :active {
+  :focus-visible {
     background-color: ${cVar('colorBackgroundStrongAlpha')};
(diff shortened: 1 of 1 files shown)


drillprop:
I don't think this PR fixes the issue.
image

`1` :hover style
`2` is currently the selected page
Both of them have the same color because when the button is clicked it has `:focus` state applied. The solution would be using `:focus-visible` instead of regular `:focus` 

According to the design, these colors should be different https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=622-89056&t=mEqEkFs9vGDfFceA-4

image

drillprop (changes requested):
I don't think this PR fixes the issue.
image

`1` :hover style
`2` is currently the selected page
Both of them have the same color because when the button is clicked it has `:focus` state applied. The solution would be using `:focus-visible` instead of regular `:focus` 

According to the design, these colors should be different https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=622-89056&t=mEqEkFs9vGDfFceA-4

image

drillprop (approved):

Issue #4019, opened 2023-04-03
Category pages is displaying wrong videos


drillprop:
As the title says. Check the URL below, you'll notice that we're displaying all the videos, not the ones that belong to the category. 
https://gleev.xyz/category/2
Possibly we're overwriting `where` arguments in the CategoryView

I think it's worth doing it together with https://github.com/Joystream/atlas/issues/4010

Issue #4010, opened 2023-03-31
Remove blockedDataObjectIds, blockedVideosIds and blockedChannelIds from atlas config


drillprop:
Since we released Orion v2, these fields are no longer needed, because we can exclude content using new orion mutations. 
These fields are hard to maintain and they caused a few bugs in the past. Let's just remove them

PR #4023, opened 2023-04-03
🎪 Fix showing all videos in category

WRadoslaw:
#4010 
#4019

### DIFF
--- a/packages/atlas/src/config/contentFilter.ts
+++ b/packages/atlas/src/config/contentFilter.ts
@@ -1,31 +1,4 @@
-import { ChannelWhereInput, InputMaybe, VideoWhereInput } from '@/api/queries/__generated__/baseTypes.generated'
-import { atlasConfig } from '@/config/config'
-
-import { allUniqueVideoCategories } from './categories'
-
-export const createChannelWhereObjectWithFilters = (
-  channelWhereInput?: ChannelWhereInput | null
-): ChannelWhereInput => {
-  return {
-    ...channelWhereInput,
-  }
-}
-
-export const createVideoWhereObjectWithFilters = (
-  videoWhereInput?: VideoWhereInput | InputMaybe<VideoWhereInput>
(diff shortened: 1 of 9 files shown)


packages/atlas/src/config/contentFilter.ts
```diff
@@ -3,19 +3,11 @@ import { atlasConfig } from '@/config/config'
 
 import { allUniqueVideoCategories } from './categories'
 
-type ContentFilter = string[]
-
-export const filteredVideoIds: ContentFilter = atlasConfig.content.blockedVideoIds
-export const filteredChannelIds: ContentFilter = atlasConfig.content.blockedChannelIds
-
 export const createChannelWhereObjectWithFilters = (
```
drillprop: I'd remove this function as well, it's not doing anything actually.

packages/atlas/src/config/contentFilter.ts
```diff
@@ -25,14 +17,11 @@ export const createVideoWhereObjectWithFilters = (
...
 
-    // filter by video ids
-    ...(filteredVideoIds.length ? { id_not_in: filteredVideoIds } : {}),
-
     // filter by channel ids
     channel: createChannelWhereObjectWithFilters(videoWhereInput?.channel),
 
     // filter by category
-    ...(!atlasConfig.content.showAllContent
+    ...(atlasConfig.content.showAllContent
```
drillprop: I would also remove `showAllContent` from the config. This option is also configurable in Orion. So I don't think we should keep it

drillprop: Also, let's remove `createVideoWhereObjectWithFilters` as well.

drillprop (changes requested):

drillprop (approved):
Thanks for cleaning this up, looks good to me!

attemka (approved):

PR #4024, opened 2023-04-03
Fix nft carousel, fix the issue with not setting the app id

drillprop:
Diff could look scary, but I only renamed one function(useMemberAvatar to getMemberAvatar) and most of the changes come from this refactor. 
The important changes are inside 
- NftCarouselDetails.tsx
- app.ts
- channels.graphql

Please also double-check if it's possible to create a video and channel that has an app id assigned.


WRadoslaw (approved):
Nice 💯

attemka (approved):

Issue #3793, opened 2023-02-15
Section header component


attemka:
[Figma design](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=11419%3A264823&t=F0JOQ8XTKU4zfsHy-0)

Implementation details: 

- [ ] two media breakpoints supported
- [ ] mobile\desktop view supported
- [ ] search is debounced
- [ ] filters component uses shared type
- [ ] custom content shouldn't brake the view

PR #4038, opened 2023-04-06
Section component

### LINKED ISSUE #3808: Section component
[Figma design](https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=11419%3A246478&t=PbJDzXs8a4Zgi5Sx-0)

Implementation details

- [ ] tabs content should be lazy loaded 
- [ ] only current tab content should be rendered

drillprop:
Fix #3808, #3793


packages/atlas/src/components/FilterButton/FilterButton.tsx
```diff
@@ -1,50 +1,72 @@
...
-  const handleSelection = (num: number) => {
-    setLocalSelection((prev) => {
-      if (prev.includes(num)) {
-        return prev.filter((prevNum) => prevNum !== num)
-      } else {
-        return [...prev, num]
+  const handleCheckboxSelection = (num: number) => {
+    const selected = options.map((option, idx) => {
+      if (num === idx) {
+        return { ...option, selected: true }
```
WRadoslaw:
```suggestion
        return { ...option, selected: !option.selected}
```

packages/atlas/src/components/MobileFilterButton/MobileFilterButton.tsx
```diff
@@ -0,0 +1,154 @@
...
+    if (filterIdxToEdit === -1) {
+      return
+    }
+    const newFilters = filters.map((filter, filterIdx) => {
+      if (filterIdx === filterIdxToEdit) {
+        return {
+          ...filter,
+          options: filter.options?.map((option, optionIdx) => {
+            if (num === optionIdx) {
+              return { ...option, selected: true }
```
WRadoslaw: Same as in `FilterButton`

packages/atlas/src/components/MobileFilterButton/index.ts
```diff
@@ -0,0 +1 @@
+export * from './MobileFilterButton'
```
WRadoslaw, line 1: Could we add a story for this?

packages/atlas/src/hooks/useHorizonthalFade.ts
```diff
@@ -0,0 +1,100 @@
...
+      setIsOverflow(hasOverflow)
+      setScrollWidth(ref.current.scrollWidth)
+
+      if (callback)
+        callback({ hasOverflow, clientWidth: ref.current.clientWidth, scrollWidth: ref.current.scrollWidth })
+    }
+
+    if (ref.current) {
+      if ('ResizeObserver' in window) {
+        new ResizeObserver(trigger).observe(ref.current)
```
WRadoslaw: Shouldn't we cleanup on unmount?

WRadoslaw (changes requested):
First of all, nice job. That's a lot of details to cover. After the first click through:

1. On stories I cannot deselect checkboxes inside the filter button - not sure if it's just the problem with story implementation or smth in component
image

2. On `With title` story (between 800-950px viewport width) toggle button group is useless and doesn't make much sense, maybe we should hide it when search is open and we are below MD and still in one column
image

drillprop:
> First of all, nice job. That's a lot of details to cover. After the first click through:
> 
> 1. On stories I cannot deselect checkboxes inside the filter button - not sure if it's just the problem with story implementation or smth in component
>    image
> 2. On `With title` story (between 800-950px viewport width) toggle button group is useless and doesn't make much sense, maybe we should hide it when search is open and we are below MD and still in one column
>    image

fixed

WRadoslaw (approved):

WRadoslaw (changes requested):
Just noticed that on checkbox group on the mobile filter behaves as a radio group

drillprop:
> Just noticed that on checkbox group on the mobile filter behaves as a radio group

fixed as well

WRadoslaw (approved):

Issue #2981, opened 2022-07-07
Channel Balance for public view


dmtrjsg:
channel balance for public view

Is super important. It should be very prominent.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202570962318926) by [Unito](https://www.unito.io)

KubaMikolajczyk:
Exploration for channel balance 👇
🎥 Loom video (Channel balance starts at 19:15): https://www.loom.com/share/5595fb504d334ad99ac9478b2cf22810
📄 Figma files: https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=1412%3A83542

bedeho:
Thank you for this iteration @KubaMikolajczyk , a pleasure to review as always <3. Also let me commend you on all the variations you have tried and all the effort and evaluation that went into that, it really shows.

1. I really _love_ the clean menu free view, lets keep that.
2. Many of the header shots were really beautiful and compelling, but I think the last one indeed brought home the message as well. The one thing I would be concerned about would be whether people would discover that this center-line is adjustable, and whether everyone will understand what it means. I think this is not a blocker at this point, lets go with the last one you proposed.
3. Combining the examples with the focus of the section really made most sense, so other verified creators with the header, and recently paid with the rewards.
4. Steps section was really great, not sure you even need the detailed visuals, just focusing on the steps and a short title can bring the message home clearly.
5. On the channel balance subject: I think that it probably does deserve more attention than the final location you advised, its arguably more important than the set of NFT collectors for the channel. I also think that it may need a bit more space than you have chosen, because I suspect it may need a label. Simply showing a balance leaves a little bit too much unsaid for the large number of users who are going to be new to the whole idea of a channel having a balance, and what that even means.

dmtrjsg:
@bedeho just a few points of consideration on point 5, perhaps that may win you over to starting with a version that Kuba has suggested. Perhaps increase the size/ and have a slightly offset positioning from its exact one, in the last variation but overall I think this may actually work well 🤞 

> more attention than the final location you advised.. more important than NFTs

I think this location is beneficial from the pov of replicating it in the Channel Info section on every video posted by this channel. If we agree that it is beneficial to include it there, here are my two suggestions where that would look not out of place and easily discoverable: 

[...]

PR #4054, opened 2023-04-07
🍎 Public channel balance

WRadoslaw:
#2981

### DIFF
--- a/packages/atlas/src/providers/joystream/joystream.hooks.ts
+++ b/packages/atlas/src/providers/joystream/joystream.hooks.ts
@@ -172,14 +172,14 @@ export const useSubscribeAccountBalance = (
   )
 
   useEffect(() => {
-    if (!activeMembership?.controllerAccount || !joystream) {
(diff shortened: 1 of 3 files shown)


dmtrjsg:
With Kuba earlier we agreed to show a joy tooltip when hovering over the JOY icon. For this to take effect component update in Figma is required. I think we should not wait for this and ship as-is and update the icon later as enhancement. 

For the time being the only currency is JOY so it should be understandable.. Becomes more of an issue when CRTs are out so let's release as built 👍

drillprop (changes requested):
It's not working when you're not signed up. Channel balance is always 0 until you login. Tested on `channel/90`

drillprop (dismissed):
Now it works, but on the other hand, the storybook build is not working. Can we investigate why? Do we want to merge the PR, which will break the storybook?

drillprop:
@attemka could you redeploy this and check if this is a one-off case?

attemka (approved):

PR #4080, opened 2023-04-13
Content type modal

### LINKED ISSUE #4051: Display modal informing users of Web3 content upon first video upload in the studio
## Context
We have a growing number of channels that upload non Web3 content to Gleev and then enquire why their channel is taken down

## Scope
- [ ] Add modal to the first upload of video that informs users about Web3 focus and that channels who do not comply this may be taken down from Gleev. 
- [ ] This should be some sort of an image rather than text as everyone needs to read it

@ Dev modal needs to be configurable
- [ ] Add this to config as URL to image, if not added we don't show this popup modal up

drillprop:
Fix #4051


drillprop:
~Draft, I noticed in the one scenario modal is not showing~

WRadoslaw (commented):
Good overall, but shouldn't we base it on the information that the user has no videos uploaded? Now, even if I have 10 videos, but the storage got cleared I'll see this modal.

drillprop:
> Good overall, but shouldn't we base it on the information that the user has no videos uploaded? Now, even if I have 10 videos, but the storage got cleared I'll see this modal.

I'm not sure, to be honest. @dmtrjsg @attemka wdyt?

Should we show this dialog if someone already created a video in gleev? I think there are still some users that already uploaded videos and they are not aware that the gleev is web3 focused app

attemka:
@drillprop I think it worth fetching a query with `totalVideosUploded` like we're doing for apps metadata.

@WRadoslaw good catch, thx.

attemka (approved):

dmtrjsg:
Yes sounds good, it can be in the local storage and shown to everyone at least once 👍 @drillprop

PR #4087, opened 2023-04-17
Featured nft section, marketplace view

### LINKED ISSUE #3776: Featured NFTs Section
## Context
- https://github.com/Joystream/atlas/issues/3686

## Design
Preview of section
Screenshot 2023-02-13 at 12 03 39

[Link to Figma components> ](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)

## Scope

AC:
- The “Featured” section contains video NFTs selected by the gateway operator, ordered by:
Auction timer, ascending (if available),
Views,
Upcoming sales at the end.

- If a video NFT is not on sale, or is not on an upcoming sale, it should not be included in this section.
The section should not be displayed if there are less than 2 video NFTs that can be displayed.

- Component states: NFT, Verified, Whitelisted [link > ](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=467%3A36434&t=TO7oWuZTnWUYZ7pM-1)

drillprop:
Fix #3776 

to test it, go to FeaturedNftSection and adjust the number of items, for example:
```tsx
  // const items = loading ? createPlaceholderData(10) : nfts ?? []
  const items = createPlaceholderData(15)
```


WRadoslaw (approved):
LGTM

drillprop:
Actually, I think I broke the section with the carousel with all these changes. Draft

attemka (approved):

drillprop:
To test this, go to https://atlas-jf3dq6nt5-joystream.vercel.app/marketplace


My goal was to make the carousel overflow its container. Like on the screens below: 
image
To accomplish this I adjusted a few styles and set paddings and negative margins based on the position of the carousel. The main problem is we're using an external library called Swiper and if we change something in terms of container sizing the library is miscalculating things(I suppose)

 The problems with the current implementation: 
 - If we want to make the carousel overflow its container it's impossible to adjust the number of columns - the current approach is always to use grid "auto". To make this possible we need to specify the fixed width of every item and I suppose we need to specify this for every breakpoint. Currently, it's 300 px for all breakpoints.
 - button in the header is not switching to disabled once you reach the end. Don't know why
image
- when you click the next slide button, once you reach the end, the last click on the button moves the carousel just a tiny amount. 
- the last item in the carousel is not well aligned with the edge of the container: 
image
- every prop change for the swiper requires a refreshing site. You cannot change the prop dynamically

drillprop:
In the end, I didn't manage to accomplish the design requirements in terms of the carousel. The carousel won't be overflowing its container. 
@attemka or @WRadoslaw perhaps It's worth rereview this PR once again because I added some changes in terms of marketplace layout

packages/atlas/src/api/queries/nfts.graphql
```diff
@@ -40,8 +40,18 @@ query GetNftsConnection(
   }
 }
 
-query GetFeaturedNfts {
-  ownedNfts(limit: 5, orderBy: [createdAt_DESC]) {
+query GetFeaturedNftsVideos {
+  ownedNfts(
+    limit: 5
```
attemka: why this value is hardcoded?

packages/atlas/src/api/queries/nfts.graphql
```diff
@@ -40,8 +40,18 @@ query GetNftsConnection(
...
+    orderBy: [createdAt_DESC]
+    where: {
+      isFeatured_eq: true
+      transactionalStatus: {
+        isTypeOf_in: ["TransactionalStatusAuction", "TransactionalStatusBuyNow"]
+        auction: { isCompleted_eq: false }
```
attemka: auction might also be cancelled, so we need to ensure that isCancelled == false

attemka: also, why you're not using `useFeaturedNfts` query here?

packages/atlas/src/views/viewer/MarketplaceView/FeaturedNftsSection/FeaturedNftsSection.tsx
```diff
@@ -0,0 +1,109 @@
...
+            spaceBetween: mdMatch ? 24 : 16,
+            breakpoints: responsive,
+          }}
+        />
+      )}
+      {activeChannel && activeChannel.totalVideosCreated > 0 && (
+        <Banner
+          title="How to get featured?"
+          icon={<StyledSvgAlertsInformative24 />}
+          description="The Gleev team handpicks featured video NFTs as a way to recognize and promote high-quality content. To increase your chances of getting your NFT featured on the marketplace, upload videos related to web3 & crypto and make sure your NFT is up for sale."
```
attemka: replace "Gleev" with `appName`, "web3 & crypto" with `appContentFocus`

packages/atlas/src/views/viewer/MarketplaceView/FeaturedNftsSection/FeaturedNftsSection.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <Section
+          headerProps={{
+            start: {
+              type: 'title',
+              title: 'Featured',
+            },
+          }}
+          contentProps={{
+            type: 'carousel',
+            children: items.map((nft, idx) => <NftTileViewer nftId={nft.id} key={`${nft.id}-${idx}`} />),
```
attemka: this keying system seems not obvious for me. Makes sense if it'll be either `nft.id` or `idx`, but why both?

attemka (changes requested):

packages/atlas/src/api/queries/nfts.graphql
drillprop: > also, why you're not using `useFeaturedNfts` query here?

What do you mean here? I'm still using it, I only renamed it. I think the `FeaturedNftsVideos` name is more suitable because it's fetching NFTs together with videos data(the query is used for NFTs video carousel in marketplace)

attemka (approved):

Issue #3780, opened 2023-02-13
All NFTs section


dmtrjsg:
## Context
- https://github.com/Joystream/atlas/issues/3686

## Scope
- Add section itself
- Counter of NFTs matching the filters (inclusive of empty filters)
- Loading state (Upon applying a filter, for the time of NFTs being fetched, a skeleton loader is be used) 
- Empty fallback state [(link>)](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=474%3A170242&t=TO7oWuZTnWUYZ7pM-1)
- Lazy load: Infinite loading and virtualisation

## Design
⚠️ depends on filter button component upd: 
- https://github.com/joystream/atlas/issues/3673

[Link to Figma> ](https://www.figma.com/file/zQYsZXJo6jmbqcj1kRRmDG/Marketplace-page?node-id=472%3A97553&t=TO7oWuZTnWUYZ7pM-1)

Preview: 
<img width="727" alt="Screenshot 2023-02-13 at 12 17 11" src="https://user-images.githubusercontent.com/93924525/218405673-bf874b6c-2f57-43b2-a591-e406bbdc4b19.png">

PR #4090, opened 2023-04-17
🈂️ Nft section

WRadoslaw:
#3780


packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+]
+
+const FILTERS: SectionFilter[] = [
+  {
+    name: 'price',
+    type: 'range',
+    label: 'Price',
+    icon: <SvgActionSettings />,
+    range: { min: undefined, max: undefined },
+  },
```
drillprop, line 56: I see a two issues here: 
1. You can't change price in mobile view:  
image
2. Once you apply the filter the button doesn't change. You should see some indication that filter is applied, otherwise, it would be very bad UX. Currently, once prices are applied you still see the cogwheel icon 
image

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+  { label: 'Paid promotional material', selected: false, applied: false, value: 'promotional' },
+  { label: 'Mature content rating', selected: false, applied: false, value: 'mature' },
+]
+
+const FILTERS: SectionFilter[] = [
+  {
+    name: 'price',
+    type: 'range',
+    label: 'Price',
+    icon: <SvgActionSettings />,
```
drillprop: The label and the icon are wrong: 
image

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+    range: { min: undefined, max: undefined },
+  },
+  {
+    name: 'status',
+    label: 'Status',
+    icon: <SvgActionAuction />,
```
drillprop: The icon is wrong here as well 
image

packages/atlas/src/components/FilterButton/FilterButton.tsx
```diff
@@ -13,85 +15,157 @@ export type FilterButtonOption = {
...
+
+    return (
+      <DialogPopover
+        className={className}
+        flipEnabled
+        appendTo={document.body}
+        trigger={
+          <StyledButton
+            ref={triggerRef}
+            icon={counter ? <Counter>{counter}</Counter> : icon}
```
drillprop, line 92: I noticed that we should use different colors for icons. This should be  `Text/Default/Muted`. Could you fix this as well?  
image

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+            value: order === OwnedNftOrderByInput.CreatedAtDesc ? 'Newest' : 'Oldest',
+            onChange: (order) =>
+              setOrder(order === 'Oldest' ? OwnedNftOrderByInput.CreatedAtAsc : OwnedNftOrderByInput.CreatedAtDesc),
+          },
+        },
+      }}
+      contentProps={{
+        type: 'grid',
+        minChildrenWidth: 250,
+        children:
```
drillprop, line 162: The fallback is not centered on bigger viewports: 
image

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    setHasAppliedFilters(Boolean(minPrice || maxPrice || isPromotional || isMature || mappedStatus.length))
+
+    return {
+      lastSalePrice_gte: minPrice ? String(minPrice) : undefined,
+      lastSalePrice_lte: maxPrice ? String(maxPrice) : undefined,
+      ...(mappedStatus.length
+        ? {
+            transactionalStatus: {
+              isTypeOf_in: mappedStatus,
```
drillprop: Filters are not working for statuses correctly. Open auction returns no results even though there are open-type auctions. I think only buy now and not for sale is working correctly, not sure about timed auctions.

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+    range: { min: undefined, max: undefined },
+  },
+  {
+    name: 'status',
+    label: 'Status',
+    icon: <SvgActionAuction />,
+    type: 'checkbox',
+    options: NFT_STATUSES,
+  },
+  { name: 'other', type: 'checkbox', options: OTHER, label: 'Other', icon: <SvgActionSettings /> },
```
drillprop, line 64: Not super sure, but in the previous implementation we had a label `Exclude` here, I think we should have the same.  
image

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+            options: ['Newest', 'Oldest'],
+            value: order === OwnedNftOrderByInput.CreatedAtDesc ? 'Newest' : 'Oldest',
+            onChange: (order) =>
+              setOrder(order === 'Oldest' ? OwnedNftOrderByInput.CreatedAtAsc : OwnedNftOrderByInput.CreatedAtDesc),
+          },
+        },
+      }}
+      contentProps={{
+        type: 'grid',
+        minChildrenWidth: 250,
```
drillprop, line 161: The number of columns doesn't match the design in pretty much every breakpoint. I did some changes to section content in my PR to be able to use columns instead of the `auto-fill` approach: https://github.com/Joystream/atlas/pull/4095/files#diff-402c9530ab8bdd71ca29b619535773a25754dd7fd4263349a041041f276acd75R6

Not sure how we should proceed. Maybe, if we merge your pull request first, then I will apply those changes to your section. However, if we merge mine PR first, you will implement the grid in your pull request. WDYT?

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
drillprop: Actually found the correct copy in the design: 
image

packages/atlas/src/components/FilterButton/FilterButton.tsx
```diff
@@ -13,85 +15,157 @@ export type FilterButtonOption = {
...
+          />
+        )}
+      </DialogPopover>
+    )
   }
 
-  const handleClear = () => {
-    onChange(options.map((option) => ({ ...option, selected: false, applied: false })))
-    triggerRef.current?.click()
+  if (props.type === 'range') {
```
drillprop, line 125: I think it's nice to have: 
image

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const placeholderItems = loading || isLoading ? createPlaceholderData(limit) : []
+  const nftsWithPlaceholders = [...(nfts || []), ...placeholderItems]
+  return (
+    <Section
+      headerProps={{
+        onApplyFilters: setFilters,
+        start: {
+          type: 'title',
+          title: 'All NFTs',
+          nodeEnd: totalCount ? (
```
drillprop: According to the design, we should show 0 as well: 
image

drillprop (changes requested):

packages/atlas/src/components/Section/SectionFooter/SectionFooter.tsx
```diff
@@ -43,7 +45,11 @@ export const SectionFooter = (props: SectionFooterProps) => {
   const { ref, inView } = useInView()
 
   useEffect(() => {
-    if ((props.type === 'infinite' || (props.type === 'load' && isSwitchedToInfinite)) && inView && !isLoading) {
+    if (
+      ((props.type === 'infinite' && !props.reachedEnd) || (props.type === 'load' && isSwitchedToInfinite)) &&
```
drillprop: shouldn't we move `props.reachedEnd` to the end of this condition? I could be wrong, but I think if the footer has type `load`, we should use `props.reachedEnd` as well

drillprop (changes requested):

packages/atlas/src/components/Section/SectionFooter/SectionFooter.tsx
WRadoslaw: I covered this in Channel section PR, so no reason to duplicate code

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
WRadoslaw: We can merge yours and I'll fix it here

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,201 @@
...
+    const priceFilter = filters.find((filter) => filter.name === 'price')
+    const minPrice = priceFilter?.range?.appliedMin
+    const maxPrice = priceFilter?.range?.appliedMax
+
+    setHasAppliedFilters(
+      Boolean(minPrice || maxPrice || isPromotionalExcluded || isMatureExcluded || mappedStatus.length)
+    )
+
+    const commonFilters = {
+      lastSalePrice_gte: minPrice ? String(minPrice) : undefined,
```
drillprop: Last sale price don't seem to be working correctly. Tomato nft last sale price was 5000 joys and I still see it in the view: 
 image

I checked the network request and it seems you don't convert min price into hapi: 
image

Check the lastPrice value in Orion 
image

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,201 @@
...
+                auctionType: {
+                  isTypeOf_eq: option.value,
+                },
+              },
+            }
+          }
+
+          return { isTypeOf_eq: option.value }
+        }, [] as OwnedNftWhereInput['transactionalStatus'][])
+        .filter((filter) => !!filter) ?? []
```
drillprop:
```suggestion
        }, [] as OwnedNftWhereInput['transactionalStatus'][]) ?? []
```
I think, this `Array.filter` is redundant here

drillprop (changes requested):
Found minor things. Besides these, looks good

drillprop (approved):
LGTM!

packages/atlas/src/components/AllNftSection/AllNftSection.tsx
```diff
@@ -0,0 +1,201 @@
...
+                      hasAppliedFilters && (
+                        <Button variant="secondary" onClick={() => setFilters(FILTERS)}>
+                          Clear all filters
+                        </Button>
+                      )
+                    }
+                  />
+                </FallbackContainer>,
+              ]
+            : nftsWithPlaceholders.map((nft, idx) => <NftTileViewer key={(nft.id ?? '') + idx} nftId={nft.id} />),
```
attemka: what is the idea behind this key naming system? If idx is included in the key you don't really need anything else there? if you want to proceed with the id in the key you might need to have something like `nft.id ?? idx`

Issue #4088, opened 2023-04-17
Google Oauth branding guidelines


dmtrjsg:
## Context

Request from google verification team:

If your app uses Profile and Email scopes for Google user account sign-in, make sure that this sign-in button complies with the [Google Sign-in Branding Guidelines.](https://developers.google.com/identity/branding-guidelines) 

Also ensure that any in-app branding (including in any buttons used to launch the Google Product OAuth integration) comply with the [Google APIs Branding Guidelines](https://about.google/brand-resource-center/guidance/apis/) . If your app uses any of the following products, please comply with product-specific branding guidelines:  

YouTube API

## Scope

1️⃣  Add Youtube logo and Google logo to YPP landing page. 
2️⃣  Add info to YPP landing page "Gleev uses Google OAuth to get access to your public profile and account email address as part of sign up flow, and integrates with YouTube API to obtain details about your YouTube channel data, such as followers and video statistics. 
3️⃣   Add google branded button to the modal which triggers Oauth screen

I think its safer to use Google provided button for that.

Assets are available on the linked page. Includes sign-In buttons in PNG, SVG, EPS, and Sketch formats, in many resolutions. Includes base, hover and press states.

These files require the [Roboto font](https://developers.google.com/identity/branding-guidelines#font).

KubaMikolajczyk:
All requirements are implemented into designs ✅
https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?node-id=3338-84773&t=Mk6227nURJ9RE5ux-4

dmtrjsg:
Button style: https://developers.google.com/identity/branding-guidelines

dmtrjsg:
@WRadoslaw let's asdd "sign up with Google to the Landing page in all places instead of Sign In. 

And push the changes... this is most likely what they meant and based on some more research on similar sites who sign up with Youtube that's what they've done. 

+ We don't need to show Youtube and Google accounts next to all sing up buttons on the landingj page, but all of them need to be "sign up with google".


I've also fired an email to the google team but most likely thats what they meant.. 

cc @attemka @KubaMikolajczyk

PR #4096, opened 2023-04-18
🍼 Google button on YPP

WRadoslaw:
#4088


package.json
```diff
@@ -82,5 +82,8 @@
   "engines": {
     "node": "^16"
   },
-  "packageManager": "yarn@3.2.1"
+  "packageManager": "yarn@3.2.1",
+  "dependencies": {
+    "react-google-button": "^0.7.2"
+  }
```
drillprop: why didn't you install it only for the `atlas` package? Why do we need it here?

packages/atlas/src/assets/google/googleLogo.webp
drillprop, line 1: Like I said in the standup. I think we could use SVG here. You can add it either by yourself in the codebase or ask Kuba to add them to Figma - somewhere here https://www.figma.com/file/1uxugLps6oooI38PGmdrn9/Figma-API--%3E-Storybook?node-id=637-379&t=29guQf48Bguay3R0-4 or in illustrations.

The latter would be better, but I'm fine if you just add this to assets.

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -152,7 +157,7 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
     }
     if (currentStep === 'requirements' && hasMoreThanOneChannel) {
       setYtRequirementsErrors([])
-      onChangeStep('select-channel')
+      // onChangeStep('select-channel')
```
drillprop: to remove or to keep it?

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -326,66 +331,65 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({
...
-    setSelectedChannelId,
-  ])
-
   const authorizationStep = useMemo(() => {
     switch (currentStep) {
-      case 'requirements':
+      case 'requirements': {
+        const getPrimaryButton = () => {
+          if (isLoading) {
+            return <Spinner />
```
drillprop: why do we use the Spinner now? Previously we had the button with the text `'Please wait...'` and this is what we have in other places in the app, so for the sake of consistency, I would use regular button

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationSteps/YppAuthorizationRequirementsStep/YppAuthorizationRequirementsStep.tsx
```diff
@@ -15,15 +16,18 @@ type YppAuthorizationRequirementsStepProps = {
...
+        {requirements.map((item, idx) => (
+          <ListItem key={idx} as="li" variant="t200" color="colorText">
+            <TickWrapper fulfilled={item.fulfilled}>
+              {item.fulfilled ? <SvgActionCheck /> : <SvgActionClose />}
+            </TickWrapper>
+            <Paragraph>{item.text}</Paragraph>
+          </ListItem>
+        ))}
+      </StyledList>
+      <Banner description="Gleev uses Google OAuth to get access to your public profile and account email address as part of sign up flow, and integrates with YouTube API to obtain details about your YouTube channel data, such as followers and video statistics." />
```
drillprop: we should use here appName

drillprop (changes requested):

packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
WRadoslaw: The problem is that we used to display "please wait", cause only the text changed back then. Imho using button as a loading state when in the end button can be switched to Google one is weird. I would go either with a spinner, skeleton or disabled textbutton

packages/atlas/src/components/_buttons/GoogleButton/GoogleButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import { SvgGoogleSmallLogo } from '@/assets/icons/GoogleSmallLogo'
+
+interface GoogleButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
+  label?: string
+  disabled?: boolean
+}
+export const GoogleButton = ({ label, ...rest }: GoogleButtonProps) => {
+  return (
+    <Container as="button" {...rest}>
```
drillprop: This is a button, we don't need `as` here
```suggestion
    <Container {...rest}>
```

packages/atlas/src/components/_buttons/GoogleButton/GoogleButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const IconContainer = styled.div`
+  background-color: #fff;
+  height: 100%;
+  display: grid;
+  place-items: center;
+`
+
+const Container = styled.button`
```
drillprop: I would name this StyledGoogleButton or something similar with a name that contains the word `button`. It's not a container

packages/atlas/src/components/_buttons/GoogleButton/GoogleButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  display: inline-flex;
+  align-items: center;
+  text-align: center;
+  padding: 0;
+  box-shadow: 0 2px 4px 0 rgb(0 0 0 / 0.25);
+  font-size: 16px;
+  border-radius: 1px;
+  transition: background-color 0.218s, border-color 0.218s, box-shadow 0.218s;
+  font-family: Roboto, arial, sans-serif;
```
drillprop, line 37: We don't have available Roboto font in the app, so the rendered font will be Arial: 
image
I'm a little worried that Google team could not accept this

drillprop (dismissed):

attemka (approved):

PR #4110, opened 2023-04-21
💪 Fix ypp for multiple channels

WRadoslaw:


packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
drillprop, line 1: There is the same issue in line 465 😬

drillprop (changes requested):

drillprop (approved):

attemka (approved):

Issue #3917, opened 2023-03-13
Introduce infinite scrolling to member activities


drillprop:
Similar to https://github.com/Joystream/atlas/issues/3913

https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile-page?node-id=919%3A133673&t=rtZws3QHMZoPihYu-4

PR #4125, opened 2023-04-26
♾️ Infinite pagination on activities 

WRadoslaw:
#3917

### DIFF
--- a/packages/atlas/src/api/client/cache.ts
+++ b/packages/atlas/src/api/client/cache.ts
@@ -248,6 +248,9 @@ const channelCacheFields: CachePolicyFields<keyof FullChannelFieldsFragment> = {
 }
 
 const cache = new InMemoryCache({
+  possibleTypes: {
+    NftOwner: ['NftOwnerChannel', 'NftOwnerMember'],
+  },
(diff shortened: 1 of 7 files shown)


packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -187,3 +218,21 @@ export const MemberActivity: FC<MemberActivityProps> = ({
     </section>
   )
 }
+
+// children: [
```
attemka: could be removed as already implemented?

packages/atlas/src/api/client/cache.ts
```diff
@@ -248,6 +248,9 @@ const channelCacheFields: CachePolicyFields<keyof FullChannelFieldsFragment> = {
 }
 
 const cache = new InMemoryCache({
+  possibleTypes: {
+    NftOwner: ['NftOwnerChannel', 'NftOwnerMember'],
+  },
```
drillprop, line 253: ❤️

drillprop (approved):
Besides the comment from Artem, LGTM!

Issue #4064, opened 2023-04-12
UI modifications: Homepage


attemka:
[Figma design](https://www.figma.com/file/FU2k3hoyr9rnuf5wrWAtiJ/Homepage?node-id=1-14408&t=RunXEi9Acc6Djvz9-0)

PR #4139, opened 2023-05-03
👍 New homepage

WRadoslaw:
⚠️ To be decided whether removed components should be removed or kept in case of some future use case

#4064


packages/atlas/src/views/viewer/HomeView.tsx
```diff
@@ -1,92 +1,36 @@
...
-        {!followedLoading && followedChannelsVideosCount ? (
-          <InfiniteVideoGrid
-            title="Followed channels"
-            videoWhereInput={{ channel: { id_in: channelIdIn } }}
-            ready={!followedLoading}
-            onDemand
-            titleLoader
-          />
-        ) : null}
         <InfiniteVideoGrid
```
drillprop: I would use the `Section` component here since we are using it already in other places. Currently, we have two components that do practically the same thing, I would remove InfiniteVideoGrid and all related logic to it entirely(I think it's only used in HomeView anyway)

drillprop (changes requested):

packages/atlas/src/views/viewer/HomeView.tsx
```diff
@@ -1,96 +1,51 @@
...
-  if (followedError) {
-    return <ViewErrorFallback />
-  }
+  const { columns, fetchMore, pageInfo, tiles } = useInfiniteVideoGrid({
+    query: GetBasicVideosConnectionDocument,
+    variables: {
+      limit: 200,
+      orderBy: [VideoOrderByInput.VideoRelevanceDesc, VideoOrderByInput.CreatedAtDesc, VideoOrderByInput.ViewsNumDesc],
+    },
+  })
```
drillprop, line 23: let's apply `publicVideoFilter` from `packages/atlas/src/config/contentFilter.ts`. Currently, I see a lot of videos on the homepage that are not playable or their thumbnail is missing.
image

drillprop (changes requested):
You'll need to resolve conflicts as well. 

Besides my small comment looks good.

attemka (approved):

drillprop (approved):
looks good to me as well

PR #4149, opened 2023-05-05
Better handling media error

### LINKED ISSUE #4129: Unsupported media tracking \ UI
Currently, if media format is not supported by browser, generic asset download test error is thrown.

Add logging for media open error.
Good to have: notify user in the interface.

drillprop:
Fix #4129 

test this firefox with videos 3319 or 1692(production)

### DIFF
--- a/packages/atlas/src/api/client/index.ts
+++ b/packages/atlas/src/api/client/index.ts
@@ -109,6 +109,12 @@ const createApolloClient = () => {
               logDistributorPerformance(resolvedUrl, eventEntry)
               return resolvedUrl
             } catch (err) {
+              if (err instanceof MediaError) {
+                SentryLogger.error('Error during asset download test, media is not supported', 'AssetsManager', err, {
+                  asset: { parent, resolvedUrl, mediaError: err },
+                })
+                return null
+              }
               bannedDistributorUrls[distributorUrl] = (bannedDistributorUrls[distributorUrl] || 0) + 1
               if (err instanceof TimeoutError) {
                 AssetLogger.logDistributorResponseTimeout(eventEntry)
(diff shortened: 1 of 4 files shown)


packages/atlas/src/providers/assets/assets.helpers.ts
```diff
@@ -74,7 +74,12 @@ export const testAssetDownload = (url: string, type: DataObjectType): Promise<nu
       video.addEventListener('loadeddata', resolve)
       video.addEventListener('canplay', resolve)
       video.addEventListener('progress', resolve)
-      video.addEventListener('error', reject)
+      video.addEventListener('error', (err) => {
+        if (err.target) {
+          reject((err.target as HTMLVideoElement).error)
+        }
+        reject(err)
```
attemka: I cant find the usecase to reject it twice, should we switch to if-else?

packages/atlas/src/api/client/index.ts
```diff
@@ -109,6 +109,12 @@ const createApolloClient = () => {
               logDistributorPerformance(resolvedUrl, eventEntry)
               return resolvedUrl
             } catch (err) {
+              if (err instanceof MediaError) {
+                SentryLogger.error('Error during asset download test, media is not supported', 'AssetsManager', err, {
```
traumschule, line 113: There was logging directly by the plugin before that the media type isn't supported. This change adds a rather generic console error. Without checking the log common users still have no clue what is wrong and where to look , the video just won't play without specific error message.

drillprop: I don't think that's correct.  [This change](https://github.com/Joystream/atlas/pull/4149/files#diff-7cd3fa81e7c8a9694c1b2265bb611db21ee1cec0f4e6d53336a97e5d1be482f8R296) will make sure users will see the screen error with a message: 
```
The video could not be played because the data is corrupted or the encoding is not supported. This issue has been logged. If you need support, reach out to our community on Discord.
```
I think that's enough.

attemka (approved):

PR #4154, opened 2023-05-08
Avatar component maintenance

### LINKED ISSUE #3115: Update sizes in Avatar component
Designs: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?node-id=9109%3A171159

- [ ] remove redundant ones
- [ ] remove media queries in some size variants
- [ ] make sure that everything works fine on a desktop and mobile. 

I would also opt not to use string names for size variants. I think we should replace them with just numbers - e.g `bid` variant would be just `24` and `small` variant would be just `40`

drillprop:
Fix #3115 

Designs https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?type=design&node-id=9109-171185&t=G8ylPSMzZrkzQZPD-4

### DIFF
--- a/packages/atlas/src/components/Avatar/Avatar.styles.ts
+++ b/packages/atlas/src/components/Avatar/Avatar.styles.ts
@@ -1,22 +1,13 @@
 import isPropValid from '@emotion/is-prop-valid'
-import { SerializedStyles, css } from '@emotion/react'
+import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 
 import { SvgActionAddImage, SvgActionEdit, SvgIllustrativeFileFailed } from '@/assets/icons'
 import { SvgAvatarSilhouette } from '@/assets/illustrations'
 import { SkeletonLoader } from '@/components/_loaders/SkeletonLoader'
-import { cVar, media, square, zIndex } from '@/styles'
+import { cVar, square, zIndex } from '@/styles'
 
-export type AvatarSize =
-  | 'preview'
-  | 'cover'
-  | 'default'
-  | 'fill'
-  | 'bid'
-  | 'small'
-  | 'channel'
-  | 'channel-card'
-  | 'extra-small'
+export type AvatarSize = 136 | 104 | 88 | 40 | 32 | 24
 
 type ContainerProps = {
   size: AvatarSize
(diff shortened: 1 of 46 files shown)


attemka (commented):
IMO using the enum will be a little bit better for the readability of code, as it'll be easier to find and change the usages of the size.
In this case we'll be using something like `size={AvatarSizes.small}` instead of `size={24}`

@drillprop @WRadoslaw thoughts?

drillprop:
> IMO using the enum will be a little bit better for the readability of code, as it'll be easier to find and change the usages of the size. In this case we'll be using something like `size={AvatarSizes.small}` instead of `size={24}`
> 
> @drillprop @WRadoslaw thoughts?

but how you would name them? I used number values because we use numbers in Figma.
image

And every time I see an Avatar component in Figma and I want to add Avatar in the implementation I don't need to check which number matches the string/enum value. That was the point of this issue and PR - to **not use** string values, but number values.

attemka:
> > IMO using the enum will be a little bit better for the readability of code, as it'll be easier to find and change the usages of the size. In this case we'll be using something like `size={AvatarSizes.small}` instead of `size={24}`
> > @drillprop @WRadoslaw thoughts?
> 
> but how you would name them? I used number values because we use numbers in Figma. image
> 
> And every time I see an Avatar component in Figma and I want to add Avatar in the implementation I don't need to check which number matches the string/enum value. That was the point of this issue and PR - to **not use** string values, but number values.

then 2 options are there for having it as a type, but I'm not sure if either of them are good enough.

Option 1:
```
type AvatarSizesMap = Map<number, number>
const AvatarSizes: AvatarSizesMap = new Map([[24,24], [32,32], [40,40]]);

size={AvatarSizes.get(24)}
```

Option 2: 
```
enum AvatarsSizes {
  's24' = 24,
  's32' = 32,
  's40' = 40 
}

size={AvatarsSizes.s24}

```

In both cases its easy to e.g find all the usages of avatar with size === 24

drillprop:
> > > IMO using the enum will be a little bit better for the readability of code, as it'll be easier to find and change the usages of the size. In this case we'll be using something like `size={AvatarSizes.small}` instead of `size={24}`
> > > @drillprop @WRadoslaw thoughts?
> > 
> > 
> > but how you would name them? I used number values because we use numbers in Figma. image
> > And every time I see an Avatar component in Figma and I want to add Avatar in the implementation I don't need to check which number matches the string/enum value. That was the point of this issue and PR - to **not use** string values, but number values.
> 
> then 2 options are there for having it as a type, but I'm not sure if either of them are good enough.
> 
> Option 1:
> 
> ```
> type AvatarSizesMap = Map<number, number>
> const AvatarSizes: AvatarSizesMap = new Map([[24,24], [32,32], [40,40]]);
> 
> size={AvatarSizes.get(24)}
> ```
> 
> Option 2:
> 
> ```
> enum AvatarsSizes {
>   's24' = 24,
>   's32' = 32,
>   's40' = 40 
> }
> 
> size={AvatarsSizes.s24}
> ```
> 
> In both cases its easy to e.g find all the usages of avatar with size === 24

Option 1 - no, just no 😅

Option 2 - is better, but is it worth it? What's the difference`AvatarsSizes.s24` and simply using`24`? Both of them are type-safe. In both cases, you'll get nice intellisense(in your proposition you would need to import AvatarSize first).  Wasn't my solution just simpler and more straightforward?

attemka:
> > > > IMO using the enum will be a little bit better for the readability of code, as it'll be easier to find and change the usages of the size. In this case we'll be using something like `size={AvatarSizes.small}` instead of `size={24}`
> > > > @drillprop @WRadoslaw thoughts?
> > > 
> > > 
> > > but how you would name them? I used number values because we use numbers in Figma. image
> > > And every time I see an Avatar component in Figma and I want to add Avatar in the implementation I don't need to check which number matches the string/enum value. That was the point of this issue and PR - to **not use** string values, but number values.
> > 
> > 
> > then 2 options are there for having it as a type, but I'm not sure if either of them are good enough.
> > Option 1:
> > ```
> > type AvatarSizesMap = Map<number, number>
> > const AvatarSizes: AvatarSizesMap = new Map([[24,24], [32,32], [40,40]]);
> > 
> > size={AvatarSizes.get(24)}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Option 2:
> > ```
> > enum AvatarsSizes {
> >   's24' = 24,
> >   's32' = 32,
> >   's40' = 40 
> > }
> > 
> > size={AvatarsSizes.s24}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In both cases its easy to e.g find all the usages of avatar with size === 24
> 
> Option 1 - no, just no 😅
> 
> Option 2 - is better, but is it worth it? What's the difference`AvatarsSizes.s24` and simply using`24`? Both of them are type-safe. In both cases, you'll get nice intellisense(in your proposition you would need to import AvatarSize first). Wasn't my solution just simpler and more straightforward?

e.g. I want to find all usages of the avatar with size === 24 across the project, it could be hard to do that if I'll just put `size={24}` in the search bar, especially if there will be other usages of the size with consts. For the enum, you'll just entering `AvatarSizes.s24` and you're sure that its the usage of the avatar size only. It's also easier to modify these sizes this way, if needed.

attemka (approved):

Issue #4136, opened 2023-05-03
Prompt Login on all transactions attempted


dmtrjsg:
## Context
- https://github.com/Joystream/atlas/issues/4132

## Design
- [Link to Figma > ](https://www.figma.com/file/C3hvYPVkZ7TcjAsXfqizSD/User-accounts?node-id=2097%3A58521&t=LGFKNZcf1pUf7jac-1)

See links from the link above for individual cases

## Scope
- [ ] Create a new component that prompts users to connect when trying to interact with the platform. 
- [ ] Content of the component is customisable, depending on the actions
- [ ] Actions: 
- Like video
- Dislike video
- Comment video
- Reply to comment on video
- Follow video
- Report video
- Follow channel
- Report channel
- Bid on NFT
- Buy Now NFT

Out of scope: ⛔ 
- Sell token
- Buy token

PR #4164, opened 2023-05-10
🔑 Protected actions

WRadoslaw:
#4136


packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -336,16 +337,25 @@ export const VideoView: FC = () => {
...
-              Share
-            </Button>
+            <ProtectedActionWrapper
+              title="You want to report this video?"
+              description="Log in to report harmful content"
+            >
+              <Button variant="tertiary" icon={<SvgActionShare />} onClick={handleShare}>
+                Share
+              </Button>
+            </ProtectedActionWrapper>
```
drillprop: It doesn't make sense to add a wrapper on the share button.

drillprop: Also, the copy is not correct

packages/atlas/src/components/_nft/NftWidget/NftWidget.tsx
drillprop, line 1: I still see the old dialog when I click buy now 
image

drillprop (changes requested):
We should also remove the old `Connect walle to continue` modal. I think it's safe to do it. 
Besides these small issues, looks good to me.

drillprop (approved):

PR #4168, opened 2023-05-10
anonymous user interaction

### LINKED ISSUE #4138: Anonymous user interaction
## Context
- https://github.com/Joystream/atlas/issues/4132

## Scope

 - [ ] Obtaining a token as an anonymous user ([API](https://github.com/Lezek123/orion/blob/user-accounts/src/auth-server/docs/Apis/DefaultApi.md#anonymousauth))
 - [ ] Storing and checking for token in localStorage
 - [ ] Using the token for the allowed queries ([figma](https://www.figma.com/file/C3hvYPVkZ7TcjAsXfqizSD/User-accounts?node-id=2097-57732&t=cxZAzAd347ZrhuRP-0))

drillprop:
Fix #4138 

~I think there might be a bug when sending userId to auth API. The session cookie is not set in this case, but I'm still confirming this with Leszek~ Fixed

### DIFF
--- a/packages/atlas/src/.env
+++ b/packages/atlas/src/.env
@@ -20,28 +20,21 @@ VITE_YOUTUBE_SYNC_API_URL=https://52.204.147.11.nip.io
 VITE_YOUTUBE_COLLABORATOR_MEMBER_ID=18
 
 # Production env URLs
-VITE_PRODUCTION_ORION_AUTH_URL=https://atlas-dev.joystream.org/orion-auth/api/v1/anonymous-auth 
+VITE_PRODUCTION_ORION_AUTH_URL=
 VITE_PRODUCTION_ORION_URL=https://orion.joystream.org/graphql
 VITE_PRODUCTION_QUERY_NODE_SUBSCRIPTION_URL=wss://orion.joystream.org/graphql
 VITE_PRODUCTION_NODE_URL=wss://rpc.joystream.org:9944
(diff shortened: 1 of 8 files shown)


packages/atlas/src/MainLayout.tsx
```diff
@@ -52,6 +57,14 @@ export const MainLayout: FC = () => {
     },
     onExitClick: () => closeDialog(),
   })
+  const firstRender = useRef(true)
+  // make sure that userId is set in localstorage and its up to date
+  useEffect(() => {
```
WRadoslaw: Wdyt about creating new HOC, kinda like provider, I think that maybe we don't even need it to be react context.  We would keep there whole Auth logic, which will get bigger with each PR

packages/atlas/src/.env
```diff
@@ -20,18 +20,28 @@ VITE_YOUTUBE_SYNC_API_URL=https://52.204.147.11.nip.io
 VITE_YOUTUBE_COLLABORATOR_MEMBER_ID=18
 
 # Production env URLs
+VITE_PRODUCTION_ORION_AUTH_URL=https://atlas-dev.joystream.org/orion-auth/api/v1/anonymous-auth
```
WRadoslaw: I don't think `anonymous-auth` should be part of this URL.

WRadoslaw (approved):

Issue #4133, opened 2023-05-02
 Log in for existing members


dmtrjsg:
## Context
- https://github.com/Joystream/atlas/issues/4132

## Design
- [Link to Figma > ](https://www.figma.com/file/C3hvYPVkZ7TcjAsXfqizSD/User-accounts?node-id=2076%3A106315&t=oBUYdL7AMN4eijwo-1)

## Scope
- [ ] Change CTA in Top Nav to Login instead of connect wallet
- [ ] Login modal
- [ ] Show/hide password
- [ ] Logging in in progress (:warning: here designs will slightly change to include progress spinner)
- [ ] Once user logs in no conf toasters
- [ ] Errors on email not found
- [ ] Form Validation
- [ ] captcha

PR #4182, opened 2023-05-13
🔐 Login modal

WRadoslaw:
#4133


packages/atlas/src/hooks/useLogIn.ts
```diff
@@ -0,0 +1,112 @@
...
+    const decryptedSeed = aes256CbcDecrypt(encryptedSeed, cipherKey, Buffer.from(cipherIv, 'hex'))
+    const keypair = keyring.addFromMnemonic(
+      entropyToMnemonic(Buffer.from(decryptedSeed.slice(2, decryptedSeed.length), 'hex'))
+    )
+    const payload = {
+      joystreamAccountId: keypair.address,
+      gatewayName: 'Gleev',
+      timestamp: time,
+      action: 'login',
+    }
```
drillprop, line 86: I can't sign in currently, you need to apply recent changes on the Orion side

 image


https://github.com/Lezek123/orion/blob/user-accounts/docs/developer-guide/tutorials/authentication-api.md#log-in-to-user-account-using-e-mail-and-password

packages/atlas/src/components/_auth/LogInModal/LogInModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+          ? {
+              text: 'Sign up',
+            }
+          : undefined
+      }
+      additionalActionsNode={!isLoading && <Button variant="tertiary">Close</Button>}
+    >
+      {!isLoading ? (
+        <Container>
+          <StyledAppLogo variant="short-monochrome" />
```
drillprop, line 86: Once my PR with the sign-up flow is merged we could use the `AuthenticationModalStepTemplate` that I introduced(it's similar to the component that we hade in SignInModal), could you add a small to-do comment to add this later?

packages/atlas/src/components/_auth/LogInModal/LogInModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    setIsLoading(false)
+  }
+
+  return (
+    <DialogModal
+      show
+      primaryButton={
+        !isLoading
+          ? {
```
drillprop: It would be much more readable if we use the condition close to on each field. ie:

```
 primaryButton={{
        text: !isLoading ? 'Log in' : 'Waiting...',
        disabled: isLoading,
        onClick: () =>
          handleSubmit((data) => {
            onLogin(data.email, data.password)
          })(),
      }}

```

packages/atlas/src/components/_auth/LogInModal/LogInModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+    resolver: (data, ctx, options) => {
+      const resolver = zodResolver(
+        z.object({
+          email: z.string().email(),
+          password: z.string().min(1, 'Please provide password'),
+        })
+      )
+
+      return resolver(data, ctx, options)
+    },
```
drillprop:
```suggestion
    resolver: zodResolver(
      z.object({
        email: z.string().email(),
        password: z.string().min(1, 'Please provide password'),
      })
    ),
```

this should be the same

packages/atlas/src/components/_auth/LogInModal/LogInModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+  })
+
+  const onLogin = async (email: string, password: string) => {
```
drillprop: `handleLogin` seems a better name for it. We usually use `on...` naming for the function that is passed by props

packages/atlas/src/hooks/useLogIn.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+type LogInHandler = {
+  data: {
+    accountId: string
+  } | null
+  error?: LogInErrors
+}
+
+export const useLogIn = () => {
```
drillprop, line 77: I don't think that we need to create a hook, the function doesn't use any state or other "react toys", this could be just `logIn`.

drillprop (changes requested):

packages/atlas/src/hooks/useLogIn.ts
WRadoslaw: I think we might need to set some global state after logging in, so I will keep it till it clears out

drillprop (approved):

PR #4187, opened 2023-05-15
New query for tiles

attemka:


drillprop (commented):
It's much better now, but is it possible to adjust this even more? How significantly would load time change if we remove those fields: 
- for video: 
```
reactionsCount
commentsCount
```
- for channel(you would need to introduce another fragment):
```
 followsNum
  rewardAccount
  channelStateBloatBond
```
?
I don't think they are required for the video tile

drillprop (changes requested):
Now we fetch every tile separately. So, this is worse than before, previously we got the information about videos from the Apollo cache and we didn't need to send more requests.  

I would either revert the recent commit or introduce some changes that will make this work correctly.

attemka:
closing because another solution was implemented

Issue #4197, opened 2023-05-17
Enhancement of the scroll mechanics on homepage


dmtrjsg:
Position the div in absolute and make an offset of 1.5 tiles from the bottom

PR #4205, opened 2023-05-17
⏰ Better loading ux

WRadoslaw:
#4197


drillprop (commented):
To be honest I don't like it at all. In my opinion, this change is making this worse than before. I'm getting the impression that now I'm staring at loading tiles more often than I used to.

Anyway, I didn't request the change, and I didn't have problems with how it worked before, so I'll leave the final decision to @dmtrjsg

dmtrjsg:
>I'm getting the impression that now I'm staring at loading tiles more often than I used to.

Same for me, so let's go with what's already on dev, but @WRadoslaw thanks for trying!

WRadoslaw:
@dmtrjsg  @drillprop It's because the fix on `useInfiniteGrid` is not in place, and the threshold for featchMore is on skeletons, instead of tiles

drillprop (approved):
oh, so your branch didn't have [this fix](https://github.com/Joystream/atlas/pull/4184)? 

Now it's better, the change is really subtle but I like it. Let's wait for final approval from @dmtrjsg

dmtrjsg:
Love it ❤️

PR #4213, opened 2023-05-19
Fix overlay issue on the video and some refactoring

drillprop:
- also renamed the component
- fix emotion error in the console


WRadoslaw (approved):
lgtm

PR #4217, opened 2023-05-22
Fix weird carousel animation

### LINKED ISSUE #4216: Weird animation transition when user switch from home to marketplace
Reproducible only on bigger viewports for some reason, it doesn't happen to me on localhost. 

https://github.com/Joystream/atlas/assets/51168865/04b96bbf-8f5b-4882-b9d5-12bb78def5f3

I'm guessing that we're applying transition all on one of the elements of the NFT tile.

drillprop:
Fix #4216 

The bug was reproducible only on Chrome, everything was fine for me on Safari and Firefox. The actual fix is applying `width: 100%` on `LimitedWidthContainer`, but it's completely a mystery to me, why this is working.

I added some small adjustments in other places on the fly. Made sure, that we don't animate padding on tiles for example.

### DIFF
--- a/packages/atlas/src/components/LimitedWidthContainer/LimitedWidthContainer.tsx
+++ b/packages/atlas/src/components/LimitedWidthContainer/LimitedWidthContainer.tsx
@@ -2,12 +2,13 @@ import styled from '@emotion/styled'
 
 import { sizes } from '@/styles'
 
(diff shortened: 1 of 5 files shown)


WRadoslaw (approved):
Looks good

Issue #4201, opened 2023-05-17
Remove followers count from recently verified channels carousel on YPP landing page


dmtrjsg:
![Screenshot 2023-05-17 at 16.15.16.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/a8ec2673-58d5-459f-b4df-c4b8dede3480)

drillprop:
Closing, because we decided to not remove followers after all

PR #4218, opened 2023-05-22
♂️ Hide followers on YPP

WRadoslaw:
#4201


drillprop (approved):
Looks good to me 👍🏻 

On the other hand, should we still remove the follows count? It looks like the problem doesn't exist anymore, all verified channels have at least 300 followers(bot attack?) @dmtrjsg what do you think?

drillprop:
After a quick chat with @dmtrjsg we decided to not remove follows count from a channel card

Issue #4157, opened 2023-05-09
Share video dialog problem


WRadoslaw:
When share video dialog is open and the user clicks on the backdrop it will trigger the video play/pause even tho dialog is open and does not close.

![image.png](https://images.zenhubusercontent.com/61b33b1e8ddbc86c62bdcd28/0f2c83f8-d29a-4985-a80a-fba825b3a2de)

PR #4221, opened 2023-05-23
🍧 Fix ShareVideo modal problem

WRadoslaw:
#4157

### DIFF
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
@@ -938,16 +938,16 @@ const VideoPlayerComponent: ForwardRefRenderFunction<HTMLVideoElement, VideoPlay
             )}
           </>
         )}
-        <VideoShare
-          onCloseShareDialog={handleCloseSharingDialog}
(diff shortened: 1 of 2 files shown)


packages/atlas/src/components/_video/VideoPlayer/VideoPlayer.tsx
```diff
@@ -938,16 +938,16 @@ const VideoPlayerComponent: ForwardRefRenderFunction<HTMLVideoElement, VideoPlay
...
       </div>
+      <VideoShare
+        onCloseShareDialog={handleCloseSharingDialog}
+        isEmbedded={isEmbedded}
+        currentTime={videoTime}
+        isFullScreen={isFullScreen}
+        videoId={videoId}
+        isShareDialogOpen={isShareDialogOpen || isSharingOverlayOpen}
+        videoTitle={video?.title}
+      />
```
drillprop: actually, VideoShare needs to be inside this div. There is a button that allows you to share the video in full-screen mode when the video is embedded. If you move Videoshare outside this div, the Videoshare overlay won't open. 

image

drillprop (changes requested):

drillprop (approved):

PR #4270, opened 2023-05-31
Change sentry replays only to log YPP page actions

attemka:


packages/atlas/src/MainLayout.tsx
```diff
@@ -53,6 +54,20 @@ export const MainLayout: FC = () => {
     onExitClick: () => closeDialog(),
   })
 
+  useEffect(() => {
+    const stopReplay = async () => await SentryLogger.replay?.stop()
+
+    if (location.pathname === absoluteRoutes.viewer.ypp()) {
```
drillprop, line 64: why can't we just move this to ypp view?

packages/atlas/src/MainLayout.tsx
```diff
@@ -53,6 +54,20 @@ export const MainLayout: FC = () => {
     onExitClick: () => closeDialog(),
   })
 
+  useEffect(() => {
```
drillprop, line 58: We should return when the sentry is not set in atlas.config.yml

drillprop (changes requested):

packages/atlas/src/MainLayout.tsx
attemka: this was the initial Idea, but we have async stopReplay, which would be tricky to execute on component unmount

packages/atlas/src/MainLayout.tsx
attemka: good catch, added this.

drillprop (approved):

Issue #4238, opened 2023-05-24
Other Member's Profile


dmtrjsg:
## Context
- https://github.com/Joystream/atlas/issues/4132

## Design

[Link to Figma>](https://www.figma.com/file/yhZpTHdf1sxJx13uRZ71GV/Membership-profile?type=design&node-id=2961%3A88257&t=eUlvDxfWiX54aLi5-1)

[Loom>](https://www.loom.com/share/dc3ab342a7c74aa781311470dcb24741 )  12:58 

## Scope
- [ ] Update filter and sorting to new component
- [ ] Update component for Statistics (nfts count)
- [ ] Updated Empty state
- [ ] Update how address copying works

Potentially out of scope:
- [ ] Extend sorting with "By price"
- [ ] Categories filter

PR #4271, opened 2023-06-01
📝 Changes on member profile

WRadoslaw:
#4238


packages/atlas/src/components/_buttons/CopyAddressButton/CopyAddressButton.tsx
```diff
@@ -20,25 +19,25 @@ export const CopyAddressButton: FC<CopyAddressButtonProps> = ({ address, classNa
...
-    <StyledText
-      as="button"
-      variant={size === 'big' ? 't300' : 't100'}
-      color="colorText"
-      className={className}
-      onClick={handleCopyAddress}
-    >
-      {shortenString(address, 6, 4)}
-      <Tooltip text="Copy account address" placement="top">
+    <StyledTooltip hideOnClick={false} text={copyButtonClicked ? 'Copied!' : 'Copy account address'} placement="top">
```
drillprop: According to designs, this should be `Copy wallet address' now

drillprop: Also, I would change the position of the tooltip if it's possible. Placement should be top left or left 
image

image

packages/atlas/src/components/_buttons/CopyAddressButton/CopyAddressButton.tsx
```diff
@@ -20,25 +19,25 @@ export const CopyAddressButton: FC<CopyAddressButtonProps> = ({ address, classNa
     if (!address) {
       return
     }
-    copyToClipboard(address, 'Account address copied to clipboard')
+    copyToClipboard(address)
     setCopyButtonClicked(true)
     setTimeout(() => {
       setCopyButtonClicked(false)
     }, 3000)
```
drillprop: A small thing, but according to the design icon should change after 2 seconds. Also, it's a little buggy. Sometimes the text changes immediately, maybe we should move the timeout to useEffect?  Tooltips like acting weird in our app, if it's hard to fix, let's leave it as it is.

drillprop (changes requested):
We need small adjustments regarding the tooltip, otherwise looks good. 

Also, you need to resolve conflicts.

packages/atlas/src/views/viewer/MemberView/MemberActivity.tsx
```diff
@@ -112,7 +112,7 @@ export const MemberActivity: FC<MemberActivityProps> = ({
   return (
     <section>
       {!loading && items.length === 0 ? (
-        <EmptyFallback title="No activity" subtitle="Go out there and explore!" variant="small" />
+        <EmptyFallback title="No activity" subtitle="Go out there and explore!" variant="large" />
```
drillprop:
```suggestion
        <EmptyFallback title="No activity" subtitle="This member hasn’t done anything yet." variant="large" />
```

drillprop (changes requested):

packages/atlas/src/components/_buttons/CopyAddressButton/CopyAddressButton.tsx
WRadoslaw: I'll just block onClick for the duration of the timeout

drillprop (approved):
LGTM!

attemka (approved):

Issue #4263, opened 2023-05-29
Glue all user accounts components


WRadoslaw:
## Todo

- [ ] Add triggers for the login modal in the correct place
- [ ] Integrate new handlers from auth provider to login
- [ ] Verify whether basic flows work as expected (signup, login with email, login with wallet)
- [ ] Integrate protected action with auth provider
- [ ] Verify session persistence on re-entry
- [ ] Nice to have - introduce AuthenticationModalStepTemplate for every auth modal

PR #4282, opened 2023-06-05
🥊 Glue auth components

WRadoslaw:
#4263


packages/atlas/src/providers/auth/auth.store.ts
```diff
@@ -1,25 +1,24 @@
 import { createStore } from '@/utils/store'
 
+type AuthModals = 'logIn' | 'externalLogIn' | 'signUp'
+
 export type AuthStoreState = {
   anonymousUserId: string | null
-  signInModalOpen: boolean
-  signUpModalOpen: boolean
+  authModalOpen?: AuthModals
```
drillprop: Possibly I'm exaggerating here, but I would rename this to something different. The name `authModalOpen` suggests that this is a boolean, not a literal type. It could be confusing to use.

packages/atlas/src/components/_auth/AuthModals/AuthModals.tsx
```diff
@@ -0,0 +1,18 @@
...
+  switch (authModalOpen) {
+    case 'logIn':
+      return <LogInModal />
+    case 'externalLogIn':
+      return <ExternalSignInModal />
+    case 'signUp':
+      return <SignUpModal />
+    default:
+      return null
+  }
```
drillprop, line 17: Because of this switch statement once you close the modal there is no ending transition animation. The modal closes immediately.

packages/atlas/src/components/_auth/ExternalSignInModal/ExternalSignInModal.tsx
```diff
@@ -103,25 +79,42 @@ export const ExternalSignInModal: FC = () => {
             hasNavigatedBack={false}
           />
         )
+      case ModalSteps.NoMembership: {
+        return (
+          <ExternalSignInModalStepTemplate
```
drillprop: Correct me if I'm wrong, but `ExternalSignInModalStepTemplate` is the same component as `AuthenticationModalStepTemplate`, can we use one and get rid of the other one?

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -309,9 +309,9 @@ export const VideoView: FC = () => {
                   placement="top-start"
                   offsetY={8}
                   delay={[1000, null]}
-                  text={`${formatDate(video.createdAt)} at ${format(video.createdAt, 'HH:mm')}`}
+                  text={`${formatDate(new Date(video.createdAt))} at ${format(new Date(video.createdAt), 'HH:mm')}`}
```
drillprop, line 312: why do we need this change? I'm pretty sure that date is formatted correctly in video view.

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
```diff
@@ -309,9 +309,9 @@ export const VideoView: FC = () => {
...
                 >
-                  {formatDate(video.createdAt)}
+                  {formatDate(new Date(video.createdAt))}
```
drillprop, line 314: same as above

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.hooks.ts
```diff
@@ -239,6 +239,7 @@ export const useCreateEditChannelSubmit = () => {
         onCompleted?.()
       }
     },
+    // eslint-disable-next-line no-sparse-arrays
```
drillprop: We don't break the rule here, do we? Should we remove this ignore?

packages/atlas/src/providers/auth/auth.provider.tsx
```diff
@@ -196,25 +214,43 @@ export const AuthProvider: FC<PropsWithChildren> = ({ children }) => {
...
 
+  const handleLogout = useCallback(async () => {
+    try {
+      await axios.post(`${ORION_AUTH_URL}/logout`, {}, { withCredentials: true })
+      handleAnonymousAuth(anonymousUserId).then((userId) => {
+        setAnonymousUserId(userId ?? null)
+      })
+      refetch()
+      setCurrentUser(null)
+      setEncodedSeed(null)
```
drillprop: should we await the refetch? Did you do this deliberately? I mean, did you want to set the current user and seed to null regardless?

packages/atlas/src/providers/auth/auth.provider.tsx
```diff
@@ -131,27 +134,35 @@ export const AuthProvider: FC<PropsWithChildren> = ({ children }) => {
...
         payload.joystreamAccountId = keypair.address
         signatureOverPayload = u8aToHex(keypair.sign(JSON.stringify(payload)))
       }
 
       if (params.type === 'external') {
         payload.joystreamAccountId = params.address
-        signatureOverPayload = await params.sign(JSON.stringify(payload))
+        try {
+          signatureOverPayload = await params.sign(JSON.stringify(payload))
+        } catch (_) {
```
drillprop:
```suggestion
        } catch {
```
As far as I remember you can omit parentheses.

packages/atlas/src/providers/auth/auth.provider.tsx
```diff
@@ -131,27 +134,35 @@ export const AuthProvider: FC<PropsWithChildren> = ({ children }) => {
...
+          setIsAuthenticating(false)
+          return {
+            data: null,
+            error: LogInErrors.SignatureCancelled,
+          }
```
drillprop: Are you sure that this will always throw `SignatureCancelled` error?

packages/atlas/src/providers/auth/auth.provider.tsx
```diff
@@ -1,25 +1,30 @@
...
 import { AES, enc, lib, mode } from 'crypto-js'
 import { FC, PropsWithChildren, createContext, useCallback, useContext, useMemo, useState } from 'react'
 
-import { useGetCurrentAccountQuery } from '@/api/queries/__generated__/accounts.generated'
+import { GetCurrentAccountQuery, useGetCurrentAccountLazyQuery } from '@/api/queries/__generated__/accounts.generated'
 import { atlasConfig } from '@/config'
 import { ORION_AUTH_URL } from '@/config/env'
-// import { ViewErrorFallback } from '@/components/ViewErrorFallback'
 import { useMountEffect } from '@/hooks/useMountEffect'
+// import { ViewErrorFallback } from '@/components/ViewErrorFallback'
```
drillprop: to remove?

packages/atlas/src/components/_overlays/MemberDropdown/MemberDropdownNav.tsx
```diff
@@ -280,17 +272,10 @@ export const MemberDropdownNav: FC<MemberDropdownNavProps> = ({
                   {
                     asButton: true,
                     label: 'My profile',
-                    onClick: onCloseDropdown,
+                    onClick: onCloseDropdown ?? (() => undefined),
```
drillprop:
```suggestion
                    onClick: onCloseDropdown,
```
redundant?

drillprop (changes requested):
Amazing work with gluing this together. I managed to create a channel and post a video without using any extension 🤯 . Some things that I found when I was testing this:

1. I already told you about this on discord, but I'll post this here as well. Deployment preview is not working because we're getting error from polkadot
image

2. I think I broke this, but there are dividers in a few modal, and they shouldn't be here
image
image

3. I didn't go any errors when I used email that is already used in this step: 
image
4. I don't think we should show the error-type snackbar. 
image
Please check what we have currently: 
image

5. Once I created a video without extension and transaction modal disappeared I didn't have a new video in my videos. It seems that we didn't refetch user videos.

6. I'm experiencing multiple redirects when I refresh the studio(signed user with extension)

https://github.com/Joystream/atlas/assets/51168865/dadd4e79-2669-43d0-a9a8-98bbc2261cce


7. Once you switch the channel to a member there is a weird gap in the member dropdown

image

**EDIT** - Actually I can't reproduce this anymore, so possibly this can be ignored.

8. Once I signed up I wasn't automatically logged in. 


I'm fine if we do these fixes in follow-up PRs, please create issues on GH if you feel that we should tackle these separately.

packages/atlas/src/views/viewer/VideoView/VideoView.tsx
WRadoslaw: Not sure why, but sometimes on VideoView, on initial load date is returned as a string instead of the date

drillprop: We're transforming it to the desired format in the cache.tsx(`createDateHandler`). So it should work, but I'm fine with your changes, if you found some edge cases.

packages/atlas/src/views/studio/CreateEditChannelView/CreateEditChannelView.hooks.ts
WRadoslaw: Not sure :D Looks like IDE added it by itself

drillprop (changes requested):
There is another issue. If someone doesn't have a wallet installed he'll see this: 
image

drillprop (changes requested):
Few bugs:
- If a user provides an email that already exists, and then corrects the email and signs up again, he won’t be logged in automatically after signing up a second time


- found another redirecting bug. Steps to reproduce:
1. log into your account using your wallet. The account needs to have a channel
2. Go to the studio
3. Logout
4. Click login CTA 
5. Use a different account that doesn’t have a channel
6. The app is constantly redirecting from studio to sign-in, and I’m ending up in an infinite loop of redirecting 

I think we don’t clear the channelId from local storage after logging out. This could be a reason

- You forgot to add a handler to this button 
image

drillprop (approved):
Looks good to me, nice 🎉 !

Issue #4289, opened 2023-06-05
Swap Membership and Email in the Sign up flow


dmtrjsg:
This is required to do the requests async and avoid saving all inputs in the YPP and firing a cascade of requests in the very end.

PR #4317, opened 2023-06-12
🦢 Swap sign up step

WRadoslaw:
#4289


packages/atlas/src/components/_auth/AuthModals/AuthModals.tsx
```diff
@@ -1,18 +1,22 @@
...
-  switch (authModalOpenName) {
-    case 'logIn':
-      return <LogInModal />
-    case 'externalLogIn':
-      return <ExternalSignInModal />
-    case 'signUp':
-      return <SignUpModal />
-    default:
-      return null
+  const debouncedAuthModalOpenName = useDebounceValue(authModalOpenName, 2_000)
```
drillprop: I would remove that, it doesn't work well, IMO it's worse than before

packages/atlas/src/components/_auth/ExternalSignInModal/ExternalSignInSteps/ExternalSignInModalEmailStep.tsx
```diff
@@ -34,30 +38,61 @@ export const ExternalSignInModalEmailStep: FC<SignInModalEmailStepProps> = ({
   const { handleSubmit, setError, formState, register } = useFormContext<{ email: string }>()
   const handleConfirm = useCallback(async () => {
     const account = await joystream?.selectedAccountId
-    if (!joystream?.signMessage || !account || !memberId) return
+    if (!joystream?.signMessage || !account || !memberId || !formState.isValid) return
```
drillprop: As far as I know, we don't need to add another check here. If `formState` is invalid it shouldn't run handleSubmit, this is handled by react-hook-form

packages/atlas/src/components/_auth/SignUpModal/SignUpSteps/SignUpEmailStep.tsx
```diff
@@ -78,6 +81,12 @@ export const SignUpEmailStep: FC<SignUpEmailStepProps> = ({
     })
   }, [handleGoToNextStep, isEmailAlreadyTakenError, setPrimaryButtonProps])
 
+  useMountEffect(() => {
+    if (isEmailAlreadyTakenError) {
+      setError('email', { message: 'Email already taken' })
```
drillprop:
```suggestion
      setError('email', { message: 'Email already taken.' })
```

drillprop: For consistency, we are using dots on every validation message.

packages/atlas/src/hooks/useCreateMember.ts
```diff
@@ -37,16 +52,22 @@ type FaucetParams = {
...
+  onSuccess?: (amountOfTokens?: number) => void
+  onError?: (step?: RegisterError) => void
+}
+
 export const useCreateMember = () => {
   const { handleLogin } = useAuth()
-  const { refetchUserMemberships } = useUser()
-  const [emailAlreadyRegisteredMemberId, setEmailAlreadyRegisteredMemberId] = useState('')
+  // const { refetchUserMemberships } = useUser()
+  // const [emailAlreadyRegisteredMemberId, setEmailAlreadyRegisteredMemberId] = useState('')
```
drillprop: to remove?

packages/atlas/src/hooks/useCreateMember.ts
```diff
@@ -199,20 +161,67 @@ export const useCreateMember = () => {
...
+      onStart?.()
+
+      await cryptoWaitReady()
+      const keypair = keyring.addFromMnemonic(data.mnemonic)
+
+      const address = keypair.address
+
+      try {
+        // const { data: memberShipData } = await refetchUserMemberships()
+        // const lastCreatedMembership = memberShipData.memberships[memberShipData.memberships.length - 1]
```
drillprop: to remove as well? I know you asked me about this and we had a condition here, but can't recall why we actually needed that, so let's just remove this

drillprop (changes requested):

packages/atlas/src/components/_auth/AuthModals/AuthModals.tsx
WRadoslaw: New approach in commit `CR fixes`

packages/atlas/src/components/_auth/SignUpModal/SignUpSteps/SignUpEmailStep.tsx
```diff
@@ -53,9 +54,11 @@ export const SignUpEmailStep: FC<SignUpEmailStepProps> = ({
     register,
     handleSubmit,
     control,
+    setError,
     formState: { errors },
   } = useForm<EmailStepForm>({
     criteriaMode: 'all',
+    mode: 'onBlur',
```
drillprop, line 61: why did we change the mode here?

WRadoslaw: After setting the field error when email is taken, it wouldn't revalidate the field, so even if the user has changed the email, the error would still be visible

drillprop (approved):

Issue #4190, opened 2023-05-15
Forgot password flow 


dmtrjsg:
From Orion's perspective, there are no passwords, there are only Gateway accounts and the associated Joystream accounts. If you want to log in to a Gateway account, you always need to provide a signature from one of the associated Joystream accounts.

On the Atlas side, you can either sign the message by retrieving a seed from user's email and password or ask the user to sign it themselves, using an external wallet.

You can also:
1. Ask user for the e-mail, seed and a new password, so that you can re-encrypt the seed and allow the user to continue using e-mail and password to log in, without the need to install an extension.
2. Allow the user to change password once they are already signed in (which would be also available for users that remember their password, but just want to change it), so that they can install an external wallet, import seed, sign in, change password and continue using e-mail + password to sign in, but this would require migrating everything to a new account. 
However it may be challenging on the Atlas side to figure out whether a user is using an external wallet because they already switched to this more secure approach or because they just imported their Atlas-generated seed (which has limited security) to an external wallet.

dmtrjsg:
Going for this> Ask user for the e-mail, seed and a new password, so that you can re-encrypt the seed and allow the user to continue using e-mail and password to log in, without the need to install an extension.

meaning:
On login screen add forgot password> enter email + seed > new password twice > show new seed (THIS IS THE NEW ONE!).

dmtrjsg:
@Lezek123  So we decided to go forward with the following flow:

On login screen add forgot password option > (user clicks), on new modal enter email + seed > next modal  enters new password twice >  (here the seed gets regenerated)> Atlas shows **new** seed to user to save. 

Question: is that already implemented on Orion?

Lezek123:
@dmtrjsg I think the new seed generation is not strictly necessary, there are actually 2 possible approaches:

## Re-encrypting the old seed

1. User provides `email`, `seed` and `newPassword`
2. Atlas encrypts the seed using a `cipherKey` derived from `email` and `newPassword` and posts the encrypted seed to Orion (`POST /artifacts`)
3. User can now login using the `email` and `newPassword`, as Atlas will be able to retrieve the encrypted `seed` from `Orion` and decrypt it using the provided data. The same blockchain account is still used, as the seed doesn't change.

**Pros of this approach:**
- No migration required

**Cons of this approach:**
- Someone who knows the old password can still log in to the user's account, so there are now effectively 2 valid passwords for 1 account (this is another issue which comes from account/artifacts data serpartion, cc: @bedeho)

[...]

PR #4325, opened 2023-06-13
🧠 Forgot password flow

WRadoslaw:
#4190


packages/atlas/src/components/_auth/ForgotPasswordModal/ForgotPasswordModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+            loader
+            hasNavigatedBack
+          />
+        )}
+      </FormProvider>
+    </DialogModal>
+  )
+}
+
+// dwad!aA@13
```
drillprop: 👽 ✋🏻

packages/atlas/src/components/_auth/ForgotPasswordModal/steps/NewPasswordStep.tsx
```diff
@@ -0,0 +1,55 @@
...
+          icon: isFieldVisible[name] ? <SvgActionHide /> : <SvgActionShow />,
+          onClick: () => setIsFieldVisible((fields) => ({ ...fields, [name]: !fields[name] })),
+        },
+      }
+    },
+    [isFieldVisible, register]
+  )
+
+  return (
+    <AuthenticationModalStepTemplate hasNavigatedBack={false} title="Change password" subtitle="NewPasswordStep">
```
drillprop:
```suggestion
    <AuthenticationModalStepTemplate hasNavigatedBack={false} title="Change password" subtitle="Set your new password.">
```

packages/atlas/src/components/_inputs/FormField/FormField.tsx
```diff
@@ -43,6 +44,7 @@ export const FormField = memo(
         tooltip,
         error,
         disableErrorAnimation,
+        footerNode,
```
drillprop: To be honest I don't like the fact that we're adding another prop to the component just to add the `Forgot password?` link. I don't see any other situation where we will need this in the future. Why don't we just put the link below the FormField component?

packages/atlas/src/providers/auth/auth.helpers.ts
```diff
@@ -99,9 +101,54 @@ export const decodeSessionEncodedSeedToMnemonic = async (encodedSeed: string) =>
...
   return hasAtleastOneWallet ? 'externalLogIn' : 'logIn'
 }
+
+type EncryptSeedParams = {
+  seed: string
+  email: string
+  password: string
+}
+
+export const encryptSeed = async ({ seed, email, password }: EncryptSeedParams) => {
```
drillprop: The name is misleading. We will return from this function encryption artficats, not only seed. 
I actually created a very similar function here https://github.com/Joystream/atlas/pull/4326/files#diff-6dc99f6da6a82344e4dd005fd116a64ae23fd9022aaddefdb48bb42577f7942aR17
And I called it prepareEncryptionArtifacts(I took the name from the orion)

packages/atlas/src/components/_auth/ForgotPasswordModal/ForgotPasswordModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+      } catch (error) {
+        displaySnackbar({
+          title: 'Something went wrong',
+          description: `We encountered unexpected error. Please try again.`,
+          iconType: 'error',
+        })
+        SentryLogger.error('Failed to change password', 'ForgotPasswordModal', error)
+      } finally {
+        setCurrentStep(ForgotPasswordStep.NewPasswordStep)
+      }
```
drillprop, line 173: Correct me if I'm wrong, but we don't want to change the step when there is an error.

packages/atlas/src/components/_auth/ForgotPasswordModal/ForgotPasswordModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+      const loginSignature = await keypair.sign(JSON.stringify(loginPayload))
+      const loginResponse = await loginRequest(u8aToHex(loginSignature), loginPayload)
+      const accData = await lazyCurrentAccountQuery()
+
+      if (accData.data?.accountData.email !== data['EmailAndSeedStep'].email) {
+        form.setError(`${ForgotPasswordStep.EmailAndSeedStep}.email`, {
+          message: 'Provided email do not match mnemonic.',
+        })
+        logoutRequest().catch(() => undefined)
```
drillprop: Why `catch(() => undefined)`? Isn't it the same as not adding a catch at all?

packages/atlas/src/components/_auth/ForgotPasswordModal/ForgotPasswordModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+      SentryLogger.error('Failed to login when recovering password', 'ForgotPasswordModal', error)
+    } finally {
+      setIsLoading(false)
+    }
+  }
+
+  const handleNewPasswordStep = useCallback(
+    async (data: ForgotPasswordModalForm) => {
+      if (!accountId) {
+        return
```
drillprop, line 122: I would throw an error here, and wrap this whole code with try-catch, not only the axios part, there are a lot of things that can go wrong here.

packages/atlas/src/components/_auth/LogInModal/LogInModal.styles.ts
```diff
@@ -1,9 +1,14 @@
...
 import { sizes } from '@/styles'
 
 export const Container = styled.div`
   display: flex;
   flex-direction: column;
   gap: ${sizes(6)};
 `
+
+export const ForgotPasswordText = styled(TextButton)`
+  margin: 0 0 0 auto;
```
drillprop: `margin-left: auto`  would be much easier to read

drillprop (changes requested):

packages/atlas/src/components/_auth/LogInModal/LogInModal.styles.ts
WRadoslaw: I needed to reset other margins to 0, that's the reason

packages/atlas/src/components/_auth/ForgotPasswordModal/ForgotPasswordModal.tsx
WRadoslaw: If we don't user will be stuck on laoding modal -> line 119

drillprop: ok, that make sense

packages/atlas/src/components/_auth/LogInModal/LogInModal.styles.ts
drillprop: so you're saying that TextButton has some default margins applied?

packages/atlas/src/components/_auth/ForgotPasswordModal/ForgotPasswordModal.tsx
WRadoslaw: Nope, this way it won't trigger catch which is there to handle login error

WRadoslaw: I'll add sentry there as well for consistency

drillprop (approved):

PR #4327, opened 2023-06-13
Conditional import of analytics plugins

attemka:


drillprop (changes requested):
Did you test that? I was trying to build with some dummy value set in `VITE_OPTIMIZE_ID` and I'm getting `undefined`

packages/atlas/plugins/index.ts
```diff
@@ -118,3 +118,19 @@ export const EmbeddedFallbackPlugin: PluginOption = {
...
   },
 }
+
+export const OptimizePlugin: PluginOption = {
+  name: 'optimize-init-plugin',
+  transformIndexHtml: {
+    enforce: 'pre',
+    transform: (html) => {
+      const optimizeEnv = 'VITE_OPTIMIZE_ID'
+      const optimizeId = process.env[optimizeEnv]
```
drillprop:
```suggestion
      const optimizeId = process.env[optimizeEnv] || loadEnv('production', path.join(process.cwd(), 'src'))[optimizeEnv]
```
should help

drillprop (changes requested):

drillprop (approved):

Issue #4322, opened 2023-06-13
Huge request is sent on the video page


drillprop:
To reproduce: 
1. Open the network tab in the dev tools
2. Go to any video that belongs to the channel with many videos. For example https://gleev.xyz/video/9585
3. Check the topmost query(sort by size)
4. You'll notice that we're pretty much fetching all the videos from that channel(and we don't need to)

This query is called possibly in the `More from {channel}` section, so worth investigating this component in particular

PR #4328, opened 2023-06-13
🧮 Videos over-fetching

WRadoslaw:
#4322

A new way of suggestion: 
1. Fetch a newer video than the current one 
2. If there is no newer video, fetch the oldest

### DIFF
--- a/packages/atlas/src/components/_video/VideoPlayer/VideoOverlay.tsx
+++ b/packages/atlas/src/components/_video/VideoPlayer/VideoOverlay.tsx
@@ -37,7 +37,18 @@ export const VideoOverlay: FC<VideoOverlayProps> = ({
   currentVideoCreatedAt,
 }) => {
   const [randomNextVideo, setRandomNextVideo] = useState<BasicVideoFieldsFragment | null>(null)
-  const { videos, refetch } = useBasicVideos(
+  const { videos: newerVideos, loading: loadingNewestVideos } = useBasicVideos({
+    limit: 1,
+    orderBy: VideoOrderByInput.ChannelCreatedAtAsc,
+    where: {
+      ...publicVideoFilter,
+      channel: {
+        id_eq: channelId,
+      },
+      createdAt_gt: currentVideoCreatedAt,
+    },
+  })
+  const { videos: olderVideos } = useBasicVideos(
     {
       limit: 1,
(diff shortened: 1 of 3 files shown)


packages/atlas/src/components/_video/VideoPlayer/VideoOverlay.tsx
```diff
@@ -31,28 +34,41 @@ export const VideoOverlay: FC<VideoOverlayProps> = ({
...
+  const { videos, refetch } = useBasicVideos(
     {
+      limit: 1,
+      orderBy: VideoOrderByInput.ChannelCreatedAtAsc,
       where: {
+        ...publicVideoFilter,
         channel: {
           id_eq: channelId,
         },
+        createdAt_gt: currentVideoTimestamp,
```
drillprop: The name `currentVideoTimestamp` is really misleading. At first glance, I thought you mean the current time of the actual video(in the video player). Can we rename this to something more intuitive - maybe even `currentVideoCreatedAt`?

drillprop (changes requested):

packages/atlas/src/components/_video/VideoPlayer/VideoOverlay.tsx
attemka: agree, should be easier to match to `video.createdAt` prop this way

packages/atlas/src/components/_video/VideoPlayer/VideoOverlay.tsx
```diff
@@ -31,28 +34,41 @@ export const VideoOverlay: FC<VideoOverlayProps> = ({
...
-    if (!videos?.length || videos.length <= 1) {
+    if (!videos?.length || videos.length === 0) {
+      refetch({
+        where: {
+          channel: {
+            id_eq: channelId,
+          },
+          createdAt_lt: currentVideoCreatedAt,
+        },
+      })
```
drillprop: Sorry, I missed that before. I tested this on video /video/552 and I think this is not working correctly. Channel has two videos and it's not switching to the other one.

drillprop (changes requested):

attemka (approved):

drillprop (approved):
LGTM as well

Issue #4287, opened 2023-06-05
Create Channel


dmtrjsg:
## Context
- #4132 

### Design
[Link to Figma> ](https://www.figma.com/file/zi5siPUXme0Fivq2i40XBd/My-channel?type=design&node-id=1763%3A90805&t=7m71IXTFprodj3ZA-1)

[Loom> ](https://www.loom.com/share/cfe74250346c451c9f68e5e6195b3a03)

## Scope
- [ ] New button in the member Drop down
- [ ] New button in the Side Nav
- [ ] Updated create channel modal
- [ ] Success screen = public view of channel with two additional buttons..

PR #4334, opened 2023-06-19
🌵 Channel creation modal

WRadoslaw:
#4287


packages/atlas/src/components/_channel/CreateChannelModal/CreateChannelModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+          darkBackground
+          hasNavigatedBack={false}
+          title="Create channel"
+          backgroundImage={watch('avatar')?.croppedUrl || undefined}
+          subtitle={
+            <Description as="p" variant="t200" color="colorText">
+              Publish your videos and sell them as NFTs with your own channel.
+            </Description>
+          }
+          // hasNavigatedBack={hasNavigatedBack}
```
drillprop: to remove

packages/atlas/src/views/studio/MyChannelView/tabs/GeneralTab/GeneralTab.tsx
drillprop, line 1: can we include `channel` in the name of the file? I spent a minute to figure out where is the view where we actually edit the channel

packages/atlas/src/views/studio/MyChannelView/tabs/GeneralTab/GeneralTab.tsx
```diff
@@ -217,7 +216,7 @@ export const GeneralTab = ({ actionBarPortal }: { actionBarPortal: RefObject<HTM
             feeLoading={feeLoading}
             primaryButton={{
               text: 'Publish changes',
-              onClick: handleSubmit,
+              onClick: () => handleSubmit(),
```
drillprop: ActionBar should be always visible here

packages/atlas/src/hooks/useChannelForm.ts
```diff
@@ -189,6 +184,71 @@ export const useChannelForm = (props: FormType) => {
...
+          channelId = result.channelId
+        },
+        setValue,
+        () => {
+          if (atlasConfig.features.ypp.googleConsoleClientId && atlasConfig.features.ypp.youtubeSyncApiUrl) {
+            setTimeout(() => setShowConnectToYtDialog(true), 2000)
+          }
+
+          // not sure why, but even tho Orion should be in sync initial refetches return empty state
+          // that's why I've introduced this logic to make sure state is in sync before taking actions
```
drillprop: not sure, but I think we should refetch memberships on line 218? This is where onTxSync is called. Could you double-check if this will work?

drillprop (changes requested):
The channel creation works well, I didn't spot any major problems. I noticed some issues on the edit channel page though: 
- There is no placeholder in description: 
image
- We don't have a view channel feature? Should we tackle this in separate PR? I don't see any issues for this
image
- Edit cover image should have different background: 
image
- We shouldn't hide ActionBar in the form. The action bar should be always visible, and we should show the tooltip when there are no changes. I added a tooltip already in my PR, so maybe don't bother to implement this for now. Maybe for now just remove `{!hideActionBar && ...` condition
- Let's remove the notification tab for now. 
- There is not a lot of space at the bottom, probably worth discussing with Kuba 
image

drillprop (changes requested):
Actually, there is one major issue, you can actually create a channel without providing a channel title(didn't know that's possible 🤯 ). We should add validation here.

packages/atlas/src/hooks/useChannelForm.ts
WRadoslaw: I did on the first version, but it doesn't make sense what should, since `onTxSync` is called even before `onCompleted` and the problem is the Orion is still returning old number of channels for membership.  I would expect a worse outcome when dispatching request even earlier.

drillprop: ok, let's leave it as it is then

packages/atlas/src/hooks/useChannelForm.ts
```diff
@@ -43,12 +42,25 @@ type FormType = EditChannelProps | NewChannelProps
 const isEditType = (props: FormType): props is EditChannelProps => props.type === 'edit'
 const DEFAULT_LANGUAGE = atlasConfig.derived.popularLanguagesSelectValues[0].value
 
+const channelSchema = z.object({
+  title: z
+    .string()
+    .min(1, 'Title is required')
```
drillprop:
```suggestion
    .min(1, 'Title is required.')
```

packages/atlas/src/hooks/useChannelForm.ts
```diff
@@ -43,12 +42,25 @@ type FormType = EditChannelProps | NewChannelProps
...
+    .min(3, 'Title should be at least 3 characters long')
```
drillprop:
```suggestion
    .min(3, 'Title should be at least 3 characters long.')
```

packages/atlas/src/hooks/useChannelForm.ts
```diff
@@ -43,12 +42,25 @@ type FormType = EditChannelProps | NewChannelProps
...
+    .max(40, 'Title can be only 40 characters long'),
```
drillprop:
```suggestion
    .max(40, 'Title can be only 40 characters long.'),
```

packages/atlas/src/hooks/useChannelForm.ts
```diff
@@ -43,12 +42,25 @@ type FormType = EditChannelProps | NewChannelProps
...
+  isPublic: z.boolean(),
+  description: z.string().optional(),
+  language: z.any(),
+  avatar: z.any(),
+  cover: z.any(),
```
drillprop, line 55: Why `z.any`? We know how the data structure for these should look like

packages/atlas/src/views/studio/MyChannelView/tabs/GeneralTab/StudioChannelGeneralTab.tsx
```diff
@@ -67,6 +68,27 @@ export const GeneralTab = ({ actionBarPortal }: { actionBarPortal: RefObject<HTM
...
+    description:
+      'You have unsaved changes which are going to be lost if you change the tab. Are you sure you want to continue?',
+    type: 'warning',
+    primaryButton: {
+      text: 'Discard changes',
+      onClick: () => {
+        closeDialog()
+        reset()
+        displaySnackbar({
+          title: 'Change were discarded',
```
drillprop:
```suggestion
          title: 'All changes were discarded,
```

packages/atlas/src/views/studio/MyChannelView/tabs/GeneralTab/GeneralTab.styles.ts
drillprop, line 1: You didn't rename this file, and the directory

packages/atlas/src/views/studio/MyChannelView/MyChannelView.tsx
```diff
@@ -14,6 +17,7 @@ const TABS = ['General', 'Notifications'] as const
 export const MyChannelView = () => {
```
drillprop, line 19: Can we remove the notifications tab two lines above? It's not used for now. We can add this later.

packages/atlas/src/views/studio/MyChannelView/MyChannelView.tsx
```diff
@@ -35,10 +39,21 @@ export const MyChannelView = () => {
...
     <>
       <NoGlobalPaddingWrapper>
-        <PageTabs tabs={TABS.map((tab) => ({ name: tab }))} onSelectTab={handleChangeTab} selected={currentTab} />
+        <PageTabs
+          tabs={TABS.map((tab) => ({ name: tab }))}
+          trailingContent={
+            channelId && (
+              <Button variant="secondary" to={absoluteRoutes.viewer.channel(channelId)} icon={<SvgActionShow />}>
+                View channel
+              </Button>
```
drillprop, line 54: the button should be outside of page tabs on smaller screens 
image

packages/atlas/src/views/studio/MyChannelView/tabs/GeneralTab/StudioChannelGeneralTab.tsx
```diff
@@ -210,27 +235,41 @@ export const GeneralTab = ({ actionBarPortal }: { actionBarPortal: RefObject<HTM
           </InputsWrapper>
         </EntitySettingTemplate>
       </form>
```
drillprop: I mentioned this in my previous review. Let's add padding-bottom on the form(I think this should be 52px according to the designs), the spacing feels awkward to me currently

packages/atlas/src/views/studio/MyChannelView/MyChannelView.tsx
```diff
@@ -35,10 +39,21 @@ export const MyChannelView = () => {
   return (
     <>
       <NoGlobalPaddingWrapper>
-        <PageTabs tabs={TABS.map((tab) => ({ name: tab }))} onSelectTab={handleChangeTab} selected={currentTab} />
+        <PageTabs
```
drillprop, line 47: we should provide a <title> for this page.
```
  const headTags = useHeadTags('My channel')
```

and in the component 
```
    <>
    {headTags}
      <NoGlobalPaddingWrapper>
```

drillprop (changes requested):

packages/atlas/src/hooks/useChannelForm.ts
WRadoslaw: But we don't need to validate them with Zod. And if I don't include it in the schema Zod will strip these params from submit object.

drillprop: ok, sounds reasonable 👍🏻

drillprop (changes requested):
Last thing.
We're missing spacing here: 
image

drillprop (changes requested):
~There is another problem - I can't upload channel assets on your PR. Not sure if it's related to your changes.~

Edit: it looks like storage providers are not working in the dev environment

drillprop (changes requested):
Ok, I see one more problem with this flow. Previously when someone was uploading the avatar or cover, we were setting avatar on the member dropdown and in the other views immediately. 
Now user needs to refresh the site to see his changes. This is a bad experience because a user is never sure if he managed to change assets or not.

drillprop (approved):

Issue #4115, opened 2023-04-22
Combined member & channel drop down confusing


bedeho:
# Background

For the longest time I have really struggled with this drop down in the app, but I was not able to articulate why.

![Screenshot 2023-04-22 at 19 48 27](https://user-images.githubusercontent.com/437292/233799204-867147b0-9554-4fd3-96e5-d0ae67f3fe97.png)

I would go there, and expect to see some things pertaining to my membership or my channel, but it would show other things, and I would see tons of identical avatars, like this

<img width="280" alt="Screenshot 2023-04-22 at 19 51 16" src="https://user-images.githubusercontent.com/437292/233799309-6235ebc8-c443-4ca7-9f06-7958e4583a39.png">

this is from the studio view, on how I even have _two_ avatars in the top menu. The avatars inside the drop down are actually two distinct buttons, while at the top they are one. Likewise, when you need to switch, the button you are looking for is dimmed out inside the dropdown, so you are extremely likely to miss it, as I have many times. This is a bit of a mess.

# Proposal

- Stop showing avatars inside the drop down, ever.
- Show one or two avatars inside the top menu, depending on whether user is in member or studio mode, and make each two distinc buttons, which show two distinct drop downs with distinct content.

KubaMikolajczyk:
The proposed solution from me is ready to be reviewed ✅ 

In this Loom, I'll be showing you around the production and discussing how we can simplify the navigation for our users. I'll be covering topics such as the dropdowns for membership and channel, creating new channels, and logging out. I'll also be comparing our navigation to YouTube's and proposing a solution for having two different dropdowns. Chapters include: Viewer Experience, Studio Experience, YouTube Comparison, Future Navigation, and Creating New Channels 

(Oh and this description was generated by Loom's AI preety nice right? :D )

🎥 Loom video: https://www.loom.com/share/106d728ae23743d9a034b704db235846
📄 Figma explorations: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?type=design&node-id=4645%3A137109&t=86KTOo8xrUn4wowt-1

dmtrjsg:
Loved it ❤️  nice and simple and not too many options in the menu as well, which is good. Agreed with the argumentation to keep only a single drop down in distinct views. 

Small remarks:
1. Log in / log out -> I suggest to go with Sign in / Sign out
[...]

PR #4337, opened 2023-06-19
🐉 Dropdown rework

WRadoslaw:
#4115


drillprop (changes requested):
I recommend watching Kuba's Loom video where he goes through all the changes. I noticed you ignored some of them in the implementation. 

- The position of the member dropdown and the animation should be different. Kuba explains this in the video. This is also included in the design documentation. 
About the position: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?type=design&node-id=3201-114134&t=wEdC0h1ctWvhrhr2-4
About animation: https://www.figma.com/file/Pf31tuYpozYmpq163U2ho8/Web-components?type=design&node-id=13531-345343&t=wEdC0h1ctWvhrhr2-4
(I think we should remove react-spring animation and switch to a more lightweight CSSTransition here, it should be relatively easy)
- The border radius should increase to a large one. 
- The address holder looks oversized, I think it should be smaller. 
- The spacing is wrong in the top part. Compare to designs: 

image
image

packages/atlas/src/components/_buttons/CopyAddressButton/CopyAddressButton.tsx
```diff
@@ -4,15 +4,16 @@ import { Tooltip } from '@/components/Tooltip'
...
 import { shortenString } from '@/utils/misc'
 
-import { StyledSvgActionCheck, StyledSvgActionCopy, StyledText } from './CopyAddressButton.styles'
+import { Container, StyledSvgActionCheck, StyledSvgActionCopy, StyledText } from './CopyAddressButton.styles'
 
 export type CopyAddressButtonProps = {
   address: string
   className?: string
   size?: 'small' | 'big'
+  trucate?: boolean
```
drillprop:
```suggestion
  truncate?: boolean
```

packages/atlas/src/components/_buttons/CopyAddressButton/CopyAddressButton.tsx
```diff
@@ -28,17 +29,13 @@ export const CopyAddressButton: FC<CopyAddressButtonProps> = ({ address, classNa
...
-      as="button"
-      variant={size === 'big' ? 't300' : 't100'}
-      color="colorText"
-      className={className}
-      onClick={handleCopyAddress}
-    >
-      {shortenString(address, 6, 4)}
+    <Container onClick={handleCopyAddress}>
+      <StyledText as="button" variant={size === 'big' ? 't300' : 't100'} color="colorText" className={className}>
+        {trucate ? address : shortenString(address, 6, 4)}
```
drillprop:
```suggestion
        {truncate ? address : shortenString(address, 6, 4)}
```

drillprop (changes requested):
Thanks for taking the time to fix the animation, works great now. 

We should remove the tooltip, I don't think we need it, and it also shows wrong information when you're in the studio(the tooltip shows it's a member). 
image

The avatar here is also clickable, but clicking the component doesn't do anything, so maybe it should be static here?

drillprop (approved):

Issue #4343, opened 2023-06-21
Update YPP Landing page


dmtrjsg:
## Context
Re-evaluation of rewards based on JOYSTREAM token price on MEXC

## Scope

- [ ] Update YPP landing page rewards
- [ ] Update Dashboard rewards
For Tier1:
Sign up: 20 USD
Referral: 10 USD
Video Sync: 3 USD
- [ ] Display rewards in USD* ⚠️ since the feed is not connected to, let's not display live JOYStream values, but only show USD
- [ ] Add clarification on the currency of payment: * rewards are paid in JOYSTREAM tokens based on the exchange rate on the date of payout
- [ ] Make Tier 3 popularity selected by default (with 100$ per sign up)

- [ ] Multipliers stay the same ✅

KubaMikolajczyk:
We currently have (in designs at least) the exact msg about rewards being paid in JOY tokens right below those cards look at the screenshot - do you think that this info is not sufficient? 
![image](https://github.com/Joystream/atlas/assets/31144205/c747fc8c-d5d3-47ac-a4f7-eec1aed717b4)

dmtrjsg:
Thanks that would do just fine 👍  @KubaMikolajczyk

PR #4345, opened 2023-06-22
Ypp rewards

attemka:
#4343

### DIFF
--- a/packages/atlas/atlas.config.yml
+++ b/packages/atlas/atlas.config.yml
@@ -59,7 +59,7 @@ features:
       - title: Sign Up to YouTube Partner Program
         showInDashboard: false # Optional. If false the reward will be shown only on YouTube partner program landing page. Default true
         shortDescription: Connect your YouTube channels via a simple step-by-step flow and get your first reward.
-        baseAmount: 5000 # Base amount that will be multiplied by tier multiplier
+        baseAmount: 20 # Base amount that will be multiplied by tier multiplier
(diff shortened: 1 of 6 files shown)


packages/atlas/atlas.config.yml
```diff
@@ -59,7 +59,7 @@ features:
       - title: Sign Up to YouTube Partner Program
         showInDashboard: false # Optional. If false the reward will be shown only on YouTube partner program landing page. Default true
         shortDescription: Connect your YouTube channels via a simple step-by-step flow and get your first reward.
-        baseAmount: 5000 # Base amount that will be multiplied by tier multiplier
+        baseAmount: 20 # Base amount that will be multiplied by tier multiplier
```
drillprop: ok, I see two problems with this: 
 - what if we will want to reintroduce `baseAmount` as joy in the future? 
 - we should explicitly say here that this value should be provided in dollars or even rename the variable to `baseAmountInUSD`. Maybe it's worth keeping `baseAmount` and setting these to null and introducing another variable - `baseAmountInUSD`?

drillprop (changes requested):

packages/atlas/atlas.config.yml
attemka: @drillprop I did the same, but in the benefit card, as it now accepts dollar amount or joy amount or both. Check the changes in BenefitCard.tsx

drillprop (approved):

PR #4347, opened 2023-06-23
YPP rewards fix

attemka:


drillprop (dismissed):
1. The position of the text seems off to me on larger views: 
image
2. Can we apply scrolling animation on that element? It seems this is the only element that's not moving here
3. Tooltip works only on click. Should this works like that? 
4. I would change the position of the tooltip, seems off to me as well. 
image

dmtrjsg (approved):

PR #4349, opened 2023-06-23
Segment events 

attemka:


packages/atlas/src/hooks/useSegmentAnalytics.ts
```diff
@@ -12,8 +12,133 @@ const useAnalytics = () => {
...
+  )
+
+  const followChannel = useCallback(
+    (channelId: string) => {
+      analytics.track('channel followed', {
+        channelId,
+      })
+    },
+    [analytics]
+  )
```
drillprop: All these function names suggest that these functions don't have anything with analytics. It could be easily mistaken for some actual app functionality like followChannel. 
I would add a suffix at the end, for example, `followChannelAnalytics`.

packages/atlas/src/MainLayout.tsx
```diff
@@ -6,6 +6,7 @@ import { StudioLoading } from '@/components/_loaders/StudioLoading'
 import { CookiePopover } from '@/components/_overlays/CookiePopover'
 import { atlasConfig } from '@/config'
 import { BASE_PATHS, absoluteRoutes } from '@/config/routes'
+import useAnalytics from '@/hooks/useSegmentAnalytics'
```
drillprop: It was missed before, but we should avoid default exports. Also, let's name the hook `useSegmentAnalytics`

drillprop (changes requested):
Asking, because I'm not actually sure. Will these functions be actually called, when there is no segment id provided?

packages/atlas/src/hooks/useSegmentAnalytics.ts
attemka: good idea, how about `trackFollowChannel`?

drillprop: sure, sounds good to me.

attemka:
> Asking, because I'm not actually sure. Will these functions be actually called, when there is no segment id provided?

regarding this: tested it locally, they weren't. Will check additionally on dev

drillprop (approved):

PR #4357, opened 2023-06-24
Update meta-server user agent rule

traumschule:
based on https://github.com/Joystream/atlas/blob/master/packages/atlas-proxy-worker/src/handler.ts#L2C4-L2C222

Note: the current nginx rule doesn't catch discordbot for example. The replacement based on the linked handler does.


attemka:
@traumschule thx for the PR. please target the `dev` env at first in such PRs so it can be released properly.

dmtrjsg:
@traumschule please kindly add some description in terms of context and desired outcome to the PRs in the future otherwise it makes it harder to prioritise and execute in the duly manner. 

Thank you in advance.

attemka (changes requested):
see above

attemka (approved):

Issue #4258, opened 2023-05-29
Fiat Denomination of Prices: API for USD rates compatible with Coinmarketcap


dmtrjsg:
## Scope
- [ ] Compare the price feed APIs from CMC ([link to documentation](https://coinmarketcap.com/api/documentation/v1/))

## Context
With `Carthage` release, all components that had JOY prices also displayed such prices in FIAT denomination. We chose a single currency for this - USD.

Background epic
- https://github.com/Joystream/atlas/issues/2607

Joy tokens are likely to get listed in June, which will enable displaying FIAT denomination of prices in Atlas (gleev). for 

- asset prices (NFTs)
- tx fees
- channel and member account balances
- tokens transfer and withdrawal dialogues
- channel payment history
- member and channel payment notifications

The exchanges where JOY token is listed is likely to change over time, and in order to save us from implementing their APIs, the suggestion is to rely on CoinMarketCap instead. This issue is about integrating to CMC for the USD/currency_x exchange rate, so that once listed JOY/UDS feeds can be enabled quickly.

WRadoslaw:
1. Top selling channels
![image](https://github.com/Joystream/atlas/assets/92513933/d3aa592f-7715-40c7-8982-f6dc193d3352)

2. NFT tile
![image](https://github.com/Joystream/atlas/assets/92513933/965d8dcd-015c-46d1-92c4-310e77481fb9)
![image](https://github.com/Joystream/atlas/assets/92513933/f90aecd4-9b5c-4610-a874-30499cc4da2b)

3. Payments widgets
![image](https://github.com/Joystream/atlas/assets/92513933/190f8013-e791-497f-8dda-271956461228)

4. Member dropdown
![image](https://github.com/Joystream/atlas/assets/92513933/0be4f61f-3d3b-48eb-b5ed-a72e920f017f)

dmtrjsg:
@WRadoslaw could we place the denomination right under joy values like you did in the top selling channels in other places as well?  as Iooks great there and nothing else is needed imho 👍

KubaMikolajczyk:
It looks fine in those places listed here ✅ I also added some fiat denomination designs here in design issue : https://github.com/Joystream/atlas/issues/4368#issuecomment-1612801348

dmtrjsg:
@KubaMikolajczyk 

If you don't mind we will go ahead and release the current version, but as a follow up we'd need your steering on these two cases; 

1. NFT main hero carousel


[...]

PR #4140, opened 2023-05-04
Make nft widget collapsible

drillprop:
Fix #4124

PR #4370, opened 2023-06-28
💸 Fiat prices

WRadoslaw:
#4258

### DIFF
--- a/packages/atlas/atlas.config.yml
+++ b/packages/atlas/atlas.config.yml
@@ -24,7 +24,7 @@ storage:
 
 joystream:
   tokenTicker: 'JOY' # Ticker for the token used in the app
-  tokenPriceFeedUrl: null # URL for the token price feed - used to display token price in the app
+  tokenPriceFeedUrl: 'https://status.joystream.org/price' # URL for the token price feed - used to display token price in the app
   alternativeNodes:
     - name: 'Jsgenesis (Europe/UK)'
       url: 'wss://testnet-rpc-3-uk.joystream.org'
--- a/packages/atlas/src/components/Fee/Fee.tsx
+++ b/packages/atlas/src/components/Fee/Fee.tsx
@@ -49,6 +49,7 @@ export const Fee: FC<FeeProps> = ({
             variant={variant}
             color={loading ? 'colorTextMuted' : color}
             withTooltip
+            withDenomination="after"
             format="short"
             margin={{ right: 1 }}
           />
--- a/packages/atlas/src/components/NumberFormat/NumberFormat.tsx
+++ b/packages/atlas/src/components/NumberFormat/NumberFormat.tsx
@@ -23,9 +23,11 @@ export type NumberFormatProps = {
   displayedValue?: string | number
   isNegative?: boolean
   icon?: ReactNode
-  withDenomination?: boolean | 'horizontal' | 'vertical'
+  withDenomination?: boolean | 'horizontal' | 'vertical' | 'before' | 'after'
   denominationAlign?: 'left' | 'right'
 } & Omit<TextProps, 'children' | 'variant'>
+
+const TEXT_DENOMINATION_ALIGMENTS: NumberFormatProps['withDenomination'][] = ['before', 'after']
 
 export const NumberFormat = forwardRef<HTMLHeadingElement, NumberFormatProps>(
   (
@@ -55,8 +57,8 @@ export const NumberFormat = forwardRef<HTMLHeadingElement, NumberFormatProps>(
     let tooltipText
     switch (isNegative || bnValue.isNeg() ? 'full' : format) {
       case 'short':
-        formattedValue = internalValue ? (internalValue > 0.01 ? formatNumberShort(internalValue) : `< 0.01`) : 0
-        formattedDenominatedValue = fiatValue ? (fiatValue > 0.01 ? formatNumberShort(fiatValue) : `< 0.01`) : 0
+        formattedValue = internalValue ? (internalValue > 0.01 ? formatNumberShort(internalValue) : `<0.01`) : 0
+        formattedDenominatedValue = fiatValue ? (fiatValue > 0.01 ? formatNumberShort(fiatValue) : `<$0.01`) : 0
         tooltipText = formatNumber(internalValue)
         break
       case 'full':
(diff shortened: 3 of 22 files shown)


attemka:
@WRadoslaw 
1. Inconsistent alignment looks kinda strange at marketplace page. Top selling channels table is the only place that I found which has the denomination price not under the joystream icon.
Screenshot 2023-06-30 at 12 52 56

2. price tooltip on NFT price - should it even be there if we have 4 digits already, if it should then should have that many digits? Also, there's a strange behaviour of it:
gif-1

drillprop:
> Also, there's a strange behaviour of it:
>    gif-1

I'm almost sure, that I already fixed this in https://github.com/Joystream/atlas/pull/4140. This probably was happening because the BuyNow component was recreating every 1 second. Worth double checking though.

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
```diff
@@ -19,8 +22,13 @@ export type NumberFormatProps = {
   variant?: TextVariant
   displayedValue?: string | number
   isNegative?: boolean
+  icon?: ReactNode
+  withDenomination?: boolean | 'horizontal' | 'vertical' | 'before' | 'after'
+  denominationAlign?: 'left' | 'right'
 } & Omit<TextProps, 'children' | 'variant'>
 
+const TEXT_DENOMINATION_ALIGMENTS: NumberFormatProps['withDenomination'][] = ['before', 'after']
```
drillprop:
```suggestion
const TEXT_DENOMINATION_ALIGNMENTS: NumberFormatProps['withDenomination'][] = ['before', 'after']
```
typo

packages/atlas/src/components/TablePaymentsHistory/TablePaymentsHistory.tsx
```diff
@@ -174,16 +173,15 @@ const Type = ({ type }: { type: PaymentType }) => {
...
+    <StyledNumberFormat
+      icon={<StyledJoyTokenIcon variant="gray" error={isNegative} />}
+      variant="t200-strong"
+      as="p"
+      value={tokenAmount}
+      format="short"
+      color={isNegative ? 'colorTextError' : 'colorTextStrong'}
+      withDenomination
+      denominationAlign="right"
+    />
```
drillprop, line 184: The price in dollars is misaligned in the table 
image
I'm getting the impression that it depends on the value, so probably something wrong with styling of table cells

packages/atlas/src/components/Fee/Fee.tsx
```diff
@@ -49,6 +49,7 @@ export const Fee: FC<FeeProps> = ({
             variant={variant}
             color={loading ? 'colorTextMuted' : color}
             withTooltip
+            withDenomination="after"
             format="short"
             margin={{ right: 1 }}
           />
```
drillprop, line 55: In the action bar:
- The color of dollars is different. It's always `Muted`
- There should be a tooltip for dollars as well
image
- The spacing between eee and the Information component is incorrect on smaller devices 
image
image

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
```diff
@@ -68,14 +86,69 @@ export const NumberFormat = forwardRef<HTMLHeadingElement, NumberFormatProps>(
...
+              <Denomination
+                align={denominationAlign}
+                className="denomination"
+                as="span"
+                color={bnValue.isNeg() || isNegative ? 'colorTextError' : 'colorText'}
+                variant="t100"
+                ref={mergeRefs([ref, textRef])}
+              >
+                {formattedDenominatedValue !== '<$0.01' ? '$' : ''}
+                {formattedDenominatedValue}
```
drillprop, line 146:
``` 
{formattedDenominatedValue !== '<$0.01' ? `$${formattedDenominatedValue}` : ''}
``` 
?

packages/atlas/atlas.config.yml
```diff
@@ -24,7 +24,7 @@ storage:
 
 joystream:
   tokenTicker: 'JOY' # Ticker for the token used in the app
-  tokenPriceFeedUrl: null # URL for the token price feed - used to display token price in the app
+  tokenPriceFeedUrl: 'https://status.joystream.org/price' # URL for the token price feed - used to display token price in the app
```
drillprop, line 27: if you set `tokenPriceFeedUrl` to null we shouldn't show the price in dollars at all. Now we're showing 0 in a lot of places
image
image
image

drillprop (changes requested):
1. I found that we don't have denomination for last price in NFT widget. Probably worth adding 
image

Video `/video/507`

2. I think prices in the rows above the `you will pay` should have different color. From user pov he won't be able to differentiate what's more important.   
image

And I think that how it should do this according the designs 
image

Fun fact:
The designs for the Summary Row were created by Michael Scott himself 
image

packages/atlas/src/components/Fee/Fee.tsx
WRadoslaw: Spacing problem is only valid for Galaxy Fold phones, which we don't really support (for this device width most of the UI is broken, starting with an empty homepage)

drillprop: > Spacing problem is only valid for Galaxy Fold phones, which we don't really support (for this device width most of the UI is broken, starting with an empty homepage)

oh, you're right, my bad. We don't support viewports below 320px.

packages/atlas/src/components/NumberFormat/NumberFormat.tsx
WRadoslaw: Formatted value can be just a number `100` etc. or `<0.01` with JOY there is no problem since the ticker is behind. For USD ticker is in front, so for normal value we just add `$`, but for lower than 0.01 we insert `$` in the switch case, so we cannot add it here as well. It's either this conditional or adding `$` in every switch case

drillprop (changes requested):
There is a small layout shift in the purchase drawer. Probably worth adjusting the skeleton loader there: 

https://github.com/Joystream/atlas/assets/51168865/9ad81e34-4fce-4d71-954b-9c00f6510a06

Besides this tiny issue, looks good to me 👍🏻

drillprop (approved):

PR #4371, opened 2023-06-28
Additional segment events

attemka:


packages/atlas/src/hooks/useSegmentAnalytics.ts
```diff
@@ -47,14 +54,57 @@ export const useSegmentAnalytics = () => {
...
+        totalLength,
+        fullScreen,
+        quality,
+      })
+    },
+    [analytics]
+  )
+
+  const trackVideoPlaybackPaused = useCallback(
+    (videoId: string, channelId: string, title: string, totalLength: number, fullScreen: boolean, quality: string) => {
```
drillprop: I would avoid functions that have more than 3 or 4 parameters. It's bug-prone, you could easily pass the arguments in the wrong order(especially if most of them are in the same type). I would create a function that receives only one object with properties - videoId, channelId, title... etc

packages/atlas/src/hooks/useSegmentAnalytics.ts
```diff
@@ -47,14 +54,57 @@ export const useSegmentAnalytics = () => {
...
+        totalLength,
+        fullScreen,
+        quality,
+      })
+    },
+    [analytics]
+  )
+
+  const trackVideoPlaybackResumed = useCallback(
+    (videoId: string, channelId: string, title: string, totalLength: number, fullScreen: boolean, quality: string) => {
```
drillprop: same here

packages/atlas/src/hooks/useSegmentAnalytics.ts
```diff
@@ -47,14 +54,57 @@ export const useSegmentAnalytics = () => {
...
+        totalLength,
+        fullScreen,
+        quality,
+      })
+    },
+    [analytics]
+  )
+
+  const trackVideoPlaybackCompleted = useCallback(
+    (videoId: string, channelId: string, title: string, totalLength: number, fullScreen: boolean, quality: string) => {
```
drillprop: same here

packages/atlas/src/MainLayout.tsx
```diff
@@ -59,6 +59,17 @@ export const MainLayout: FC = () => {
...
+    // had to include this timeout to make sure the page title is updated
+    setTimeout(
+      () =>
+        trackPageView(
+          document.title,
+          'viewer',
+          (location.pathname === absoluteRoutes.viewer.ypp() && searchParams.get('referrer')) || undefined
+        ),
+      1000
+    )
```
drillprop: maybe it would be easier to put this in `useHeadTags`? Don't know. If we want to keep this as it I would add cleanup function in the useEffect and clear timeout.

packages/atlas/src/views/studio/StudioLayout.tsx
```diff
@@ -77,6 +79,11 @@ const StudioLayout = () => {
     }
   }, [closeUnsupportedBrowserDialog, openUnsupportedBrowserDialog])
 
+  useEffect(() => {
+    // had to include this timeout to make sure the page title is updated
+    setTimeout(() => trackPageView(document.title, 'studio', undefined), 1000)
+  }, [location.pathname, trackPageView])
```
drillprop, line 87: same here

drillprop (changes requested):

packages/atlas/src/MainLayout.tsx
attemka: not very obvious place to put this track function and too many extra renders there, so I've just added a destructor

drillprop (approved):

PR #4377, opened 2023-06-30
improve AssetImage loading

drillprop:
Context: I noticed that for basically every image in the app we had two "loading states": 
- first, when data was loading, 
- second, when the image was resolving.

Besides that, the behavior of these states was inconsistent - for example In Thumbnail, for the first loading, we had CSSTransition applied, but for the second loading state we had not. This was leading to an unpleasant experience.

Summary of changes:
- added isLoading prop
- applied CSSTransition inside of AssetImage
- added imgPlaceholder optional prop, which will show when the image is not resolved


WRadoslaw (approved):
LGTM, nice work

attemka (approved):

Issue #4376, opened 2023-06-30
Benchmark user connection for asset resolution


WRadoslaw:
## Problem

After changing asset logic for non-blocking for quicker providers, we decided to keep the loading timeout at 2 seconds. This will cause problems when one's internet connection is slow enough to exceed this limit, the result of a such situation would be a page without any assets and just alternative text for a given image. 

## Possible Solution

A possible solution for this issue would be introduction of connection benchmarking when a user enters the page. The benchmarking would happen asynchronously, so it wouldn't block the displaying of the assets. 

This benchmark should take random video and try to fetch its assets with a really long timeout. The performance of this action should be measured and general asset timeout should be adjusted to it. 

If the user's time to resolve an asset would be quicker than config timeout, then we should not take any actions or even consider decreasing general timeout for the sake of UX. Otherwise, if performance is slower than config timeout, we should take action of increasing the general timeout, so the user is not served with a page with some blue text instead of an actual asset.

PR #4385, opened 2023-07-03
📈 User connection benchmarking

WRadoslaw:
#4376 

⚠️ It needs to be decided what interval should be used by benchmark. We can even extract it to env/config if we want it not hardcoded.

### DIFF
--- a/packages/atlas/src/providers/assets/assets.provider.tsx
+++ b/packages/atlas/src/providers/assets/assets.provider.tsx
@@ -169,17 +169,11 @@ export const OperatorsContextProvider: FC<PropsWithChildren> = ({ children }) =>
         const startTime = performance.now()
         await promiseRace
         const msDuration = performance.now() - startTime
-        console.log(
-          'ben time',
-          msDuration,
-          msDuration > atlasConfig.storage.assetResponseTimeout,
-          msDuration < atlasConfig.storage.assetResponseTimeout / 2
-        )
         if (
           msDuration > atlasConfig.storage.assetResponseTimeout ||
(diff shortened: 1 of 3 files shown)


packages/atlas/src/providers/assets/assets.provider.tsx
```diff
@@ -145,6 +151,39 @@ export const OperatorsContextProvider: FC<PropsWithChildren> = ({ children }) =>
...
+      if (thumbnail) {
+        const promiseRace = getFastestImageUrl(thumbnail.resolvedUrls)
+        const startTime = performance.now()
+        await promiseRace
+        const msDuration = performance.now() - startTime
+        if (
+          msDuration > atlasConfig.storage.assetResponseTimeout ||
+          msDuration < atlasConfig.storage.assetResponseTimeout / 2
+        ) {
+          setUserBenchmarkTime(msDuration * 1.5)
```
attemka: this won't be overwritten once set, as it always comparing it to the config const. I think it should be a variable with the val = `userBenchmarkTime || atlasConfig.storage.assetResponseTimeout`

attemka: also, to exclude time spikes (in both ways) I think it might be good to take and average value between current `userBenchmarkTime || atlasConfig.storage.assetResponseTimeout` and new `msDuration`.  So It'll look smth like this: `setUserBenchmarkTime((userBenchmarkTime || atlasConfig.storage.assetResponseTimeout + msDuration) * 0.75) // / 2 * 1.5 === * 0.75`

attemka (changes requested):

attemka (approved):

PR #4430, opened 2023-07-13
Fixed missing channelId on YPP signup

attemka:


packages/atlas/src/views/global/YppLandingView/YppAuthorizationModal/YppAuthorizationModal.tsx
```diff
@@ -255,12 +256,15 @@ export const YppAuthorizationModal: FC<YppAuthorizationModalProps> = ({ unSynced
...
           setActiveChannel(channelId)
+          createdChannelId.current = channelId
         },
         onCompleted: async () => {
           await refetchUserMemberships()
 
           const channelCreationResponse = await yppSignChannelMutation({
-            ...(selectedChannelId ? { joystreamChannelId: parseInt(selectedChannelId) } : {}),
+            ...(selectedChannelId || createdChannelId.current
+              ? { joystreamChannelId: parseInt(selectedChannelId || createdChannelId.current || '') }
```
WRadoslaw: `parseInt` from `''` will return `NaN`

attemka: @WRadoslaw thats true, but then it won't pass the ternary condition above? `selectedChannelId || createdChannelId.current` either of them should be truthy value

attemka: anyways, I've put it into a separate value to make the code look cleaner

WRadoslaw (approved):

PR #4456, opened 2023-07-19
Segment events modification

attemka:


packages/atlas/src/views/viewer/ViewerLayout.tsx
```diff
@@ -57,7 +57,7 @@ const locationToPageName = {
   '/channel': 'Channel',
   '/video': 'Video',
   '/membership/edit': 'Edit membership',
-  '/membership': 'Member',
+  '/member/': 'Member',
```
WRadoslaw, line 60: Do we need an extra `/` at the end?

WRadoslaw (approved):

packages/atlas/src/views/viewer/ViewerLayout.tsx
attemka: as we're using `includes` I just want to avoid the situation when it matches `member*` URL, e.g. `membership/*` and for this URL we will always have the Id of the member (`member/[id]`), so slash will exclude all the other options

PR #4469, opened 2023-07-20
Set default segment value to null

attemka:


packages/atlas/src/hooks/useSegmentAnalytics.ts
```diff
@@ -35,13 +35,15 @@ export const useSegmentAnalytics = () => {
 
   const identifyUser = useCallback(
     (email = 'no data') => {
+      if (!analytics) return
```
WRadoslaw: It will never be null if we used the correct hook

packages/atlas/src/providers/segmentAnalytics/segment.provider.tsx
```diff
@@ -12,7 +12,7 @@ interface AnalyticsProviderProps {
 }
 
 const defaultAnalyticsContext = {
-  analytics: new AnalyticsBrowser(),
+  analytics: null,
```
WRadoslaw: Whole `ctx` should be `null` .

`... = createContext(null)`

and we reassign it in provider

packages/atlas/src/hooks/useSegmentAnalytics.ts
attemka: > It will never be null if we used the correct hook

but if I don't wont to have segment at all and I don't have a key this will stay as null:
`() => ({ analytics: writeKey ? AnalyticsBrowser.load({ writeKey }) : null }),`

WRadoslaw (approved):

Issue #3401, opened 2022-11-02
Component: split ratio slider


dmtrjsg:
### Context
- https://github.com/Joystream/atlas/issues/3385

#### Design:
[LInk to Figma> ](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/CRT-Creator?node-id=5%3A4315)

Preview:
<img width="418" alt="Screenshot 2022-11-02 at 18 56 22" src="https://user-images.githubusercontent.com/93924525/199523596-dc6a1167-5261-4b44-bffd-14e572c8e485.png">

polikosi:
Works good.
Revenue share with holders - every 10% to 100%
Annual creator reward - every 2% to 30%

![image](https://github.com/Joystream/atlas/assets/80429191/78d1cc2f-28f0-463f-ba8c-0ea7620fd363)

polikosi:
> Works good. Revenue share with holders - every 10% to 100% Annual creator reward - every 2% to 30%
> 
> ![image](https://user-images.githubusercontent.com/80429191/284356728-78d1cc2f-28f0-463f-ba8c-0ea7620fd363.png)

everything is the same as it was then

PR #4470, opened 2023-07-20
🛝 CRT Ratio Slider

thesan:
- #3401

Story: https://atlas-storybook-git-fork-thesan-feature-ratio-slider-joystream.vercel.app/?path=/story/inputs-slider-ratioslider--ratio-slider

### DIFF
--- a/packages/atlas/src/components/_inputs/Slider/RatioSlider.tsx
+++ b/packages/atlas/src/components/_inputs/Slider/RatioSlider.tsx
@@ -43,12 +43,12 @@ export const RatioSlider = forwardRef<HTMLInputElement, Props>(
         <Track xmlns="http://www.w3.org/2000/svg">
           <circle className="knob" cx={valuePercent} cy={sizes(3)} r={sizes(2)} />
 
-          <mask id="myMask">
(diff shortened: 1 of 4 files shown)


packages/atlas/src/components/_inputs/Slider/RatioSlider.tsx
```diff
@@ -0,0 +1,78 @@
+import { ChangeEventHandler, forwardRef, useMemo, useState } from 'react'
+
+import { sizes } from '@/styles'
+
+import { Range, Track, Wrapper } from './RatioSlider.styles'
+
+type Props = {
```
WRadoslaw: For the sake of code unity let's go with the `{component name}Props` naming convention for props, plus we can export it and import it in the story to not duplicate the code, plus if smth changes on the component we would receive some errors on story to fix it.

WRadoslaw (commented):
Wow, nice approach and work!

thesan:
> Wow, nice approach and work!

Thanks. I had just started implementing the draggable element when realized I could just use the range input directly, and to my surprised it works on mobile browsers !

attemka (approved):

Issue #4488, opened 2023-07-24
Don't send orion queries until auth token is obtained


attemka:
Currently, hooks will be using the orion queries even before anonymous-auth\user auth succeded, which will result in the 400 errors from orion. 
We need to check if the token is not yet obtained and:
1) pause all orion queries
2) switch all components which require data fetching to loading mode

PR #4497, opened 2023-07-25
❎ Avoid sending unauthorized errors to sentry

WRadoslaw:
#4488

### DIFF
--- a/packages/atlas/src/api/client/index.ts
+++ b/packages/atlas/src/api/client/index.ts
@@ -1,10 +1,12 @@
 import { ApolloClient, ApolloLink, FetchResult, HttpLink, Observable, split } from '@apollo/client'
+import { onError } from '@apollo/client/link/error'
 import { GraphQLWsLink } from '@apollo/client/link/subscriptions'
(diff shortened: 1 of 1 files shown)


packages/atlas/src/api/client/index.ts
```diff
@@ -23,6 +25,33 @@ const delayLink = new ApolloLink((operation, forward) => {
...
+    'Apollo Error',
+    'Error Link',
+    {
+      graphQLErrors,
+      networkError,
+    },
+    {
+      data: {
+        operation,
+        response,
```
attemka, line 48: posting this as an object makes it impossible to see orions message (error.response?.data.message) in sentry. Lets either stringify it or include message as a separate field

attemka: posting this as an object makes it impossible to see orions message (error.response?.data.message) in sentry. Lets either stringify it or include message as a separate field

attemka (changes requested):

attemka (approved):

PR #4511, opened 2023-07-27
Set referrer channel as default value

attemka:


packages/atlas/src/components/_ypp/YppReferralBanner/YppReferralBanner.tsx
```diff
@@ -57,7 +57,9 @@ export const YppReferralBanner: FC<YppReferralBannerProps> = ({ className }) =>
 
   // persist referrer id in store
   useEffect(() => {
-    setReferrerId(channel ? channel.id : null)
+    if (channel) {
+      setReferrerId(channel ? channel.id : null)
```
WRadoslaw: Do we need conditional if we checked whether the channel is defined?

WRadoslaw (approved):

Issue #4513, opened 2023-07-27
DEV: YPP dashboard not showing for the channel that is signed up


dmtrjsg:
## Context

Issues with YPP flow - getting to YPP dashboard. ℹ️  THE DB IS CLEAN on dev so you can use your YT channels to sign up.

## Problem

1️⃣ 

ChannelID: https://dev.gleev.xyz/channel/224?tab=Videos
email: 1gleevypp04@gmail.com
pass: 1gleevypp04@gmail.comA

ER: when this user is signed in and channel 224 is selected YPP Page shows "go to dashboard" on the viewer side; and /dashboard page in the studio
AR: Sign in with Google button shows on the landing page

⚠️ Update: when user goes to studio via flow > My channel > Manage videos > YPP everything shows correctly.. So expected result is for this state of the app to be updated without going to studio right away. 

I think this may be connected with the legacy flow where user was sent to studio dashboard after sign up was successful, and this is not happenning by default. 

__
2️⃣ 
Just tried to sign up again, and did it fast

user: 1gleevypp05@gmail.com
pass: 1gleevypp05@gmail.comA

1. Indeed the user was not redirected to studio after the sign up confirmation modal  (I clicked continue and the modal just closed)
<img width="1395" alt="Screenshot 2023-07-27 at 18 10 03" src="https://github.com/Joystream/atlas/assets/93924525/2461eb29-bc1b-408d-961c-80b3bd68dcb4">

2. When trying to sign in, the error says membership not found.

PR #4516, opened 2023-07-28
🔥 Fix sign up flawed flow

WRadoslaw:
#4513

### DIFF
--- a/packages/atlas/src/components/_auth/SignUpModal/SignUpModal.tsx
+++ b/packages/atlas/src/components/_auth/SignUpModal/SignUpModal.tsx
@@ -105,104 +105,77 @@ export const SignUpModal = () => {
     [currentStep]
   )
 
+  const handleOrionAccountCreation = useCallback(() => {
+    if (!memberRef.current) {
+      throw new Error('MemberID ref is empty, do a check before calling handleOrionAccountCreation function')
+    }
+
+    return createNewOrionAccount({
+      data: { ...signUpFormData.current, memberId: memberRef.current },
+      onError: (error) => {
+        if (error === RegisterError.EmailAlreadyExists) {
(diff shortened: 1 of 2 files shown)


packages/atlas/src/components/_auth/SignUpModal/SignUpModal.tsx
```diff
@@ -106,102 +105,83 @@ export const SignUpModal = () => {
...
+      onError: (error) => {
+        if (error === RegisterError.EmailAlreadyExists) {
+          setEmailAlreadyTakenError(true)
+          goToStep(SignUpSteps.SignUpEmail)
+          return
+        }
+        if (error === RegisterError.MembershipNotFound) {
+          setTimeout(() => {
+            handleOrionAccountCreation()
+          }, 10_000)
```
attemka, line 124: I think timeouts should be consistent across the steps. We have 10s here...

packages/atlas/src/components/_auth/SignUpModal/SignUpModal.tsx
```diff
@@ -248,28 +228,28 @@ export const SignUpModal = () => {
...
-          handlePasswordStepSubmit(signUpFormData.password)
+          handlePasswordStepSubmit(signUpFormData.current.password)
         }
         syncState.current = 'synced'
-      }, 20_000)
-
-      if (newMemberId) {
-        setMemberId(newMemberId)
-      }
+      }, 15_000)
```
attemka, line 248: ... and 15s here. My approach would have beef to introduce the const (so we can easily adjust it later), but if not, then let's just make them the same

attemka (changes requested):

attemka (approved):

PR #4518, opened 2023-07-28
Custom schema fetch with authorisation

attemka:
#4539


thesan (approved):
LGTM! I really need this PR to be merged to complete #4538 because I have been generating the schema from `atlas-dev.joystream.org` which I'm guessing has an older schema.

Issue #4457, opened 2023-07-19
Replace carousel with recently verified to Recently Paid Channels


dmtrjsg:
## Context

Recently verified channels do not accrue enough followers to be attractive on the landing page

## Scope
- [ ] Replace carousel with recently verified to Paid Channels on YPP landing page
(https://github.com/Joystream/joystream-org/issues/640#issuecomment-1632554063)
- [ ] Display Total Rewards to channels as value in the payout cards
- [ ] Update the cards to include the USD denomination of the total paid out. [Use this design as guidance >]

:warning: :NB: we had the recently paid out channels in the original designs so can use [this card](https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?type=design&node-id=1817%3A76425&mode=design&t=11LkrgIvKq1LuXrN-1) with the adjustment for the FIAT prices.
![Screenshot 2023-07-19 at 17.21.03.png](https://images.zenhubusercontent.com/6197874e1913711bf1f6ad31/119a8d34-9853-4dc8-afb5-66ebdb4af7c9)

PR #4538, opened 2023-07-31
🤑 Display recently Paid Channels on the YPP carousel

thesan:
- #4457


packages/atlas/src/api/hooks/channel.ts
```diff
@@ -81,6 +86,38 @@ export const useBasicChannels = (
   }
 }
 
+type ChannelPayment = { channel: BasicChannelFieldsFragment; amount: BN }
+export const useRecentlyPaidChannels = (): { channels: ChannelPayment[] | undefined; loading: boolean } => {
+  const aMonthAgo = useMemo(() => startOfDay(subMonths(Date.now(), 1)), [])
```
WRadoslaw: As discussed

packages/atlas/src/providers/assets/assets.provider.tsx
```diff
@@ -46,7 +46,7 @@ type OperatorsContextValue = {
   fetchStorageOperators: () => Promise<BagOperatorsMapping>
   userBenchmarkTime: number | null
 }
-const OperatorsContext = createContext<OperatorsContextValue | undefined>(undefined)
+export const OperatorsContext = createContext<OperatorsContextValue | undefined>(undefined)
```
WRadoslaw, line 49: Why do we need to export it now?

thesan: It makes it easier to mock it in the story [here](https://github.com/Joystream/atlas/blob/d5c82c3118790078b91f7403e3366ec50a7c26e4/packages/atlas/src/components/_channel/ChannelCard/PaidChannelCard.stories.tsx#L45). Using the provider requires setting `ApolloProvider` and `QueryClientProvider`, while just passing an empty object value to `OperatorsContext.Provider` is good enough to run the story.

thesan:
Fetching all payment events (493 atm) result in a 133KB transfer. Is this acceptable for Orion for a page that probably get significant traffic ?

image

I set a 2000 limit so it should 4x long term.

thesan:
Trimming the queried fields results in a ~6x transferred data reduction:
image

attemka (approved):
lgtm. Great job with this request weight reduction, seems much better now

PR #4546, opened 2023-08-01
😶‍🌫️ Introduce auth epoch

WRadoslaw:


thesan (approved):
Looks like it should work but it feels like there should be a better way long term. Like all requests return a timestamp in the response headers this could be used to flag the few users with wrong system times.

Also out of curiosity why do the timestamps get subtracted 30sec ?

attemka:
> Like all requests return a timestamp in the response headers this could be used to flag the few users with wrong system times.

The only drawback of this approach I can think of atm is the availability of this external endpoint for the user. Is there anything else?

WRadoslaw:
> Looks like it should work but it feels like there should be a better way long term. Like all requests return a timestamp in the response headers this could be used to flag the few users with wrong system times.
> 
> Also out of curiosity why do the timestamps get subtracted 30sec ?

The problem is that this timestamp needs to be signed with a private key before it's sent, or I'm missing something in what you mean. 

There was a problem that even with the correct time on the device Orion API would sometimes return the error that timestamp is in the future. It was just better to move it -30s instead of trying to debug something that occurs only sometimes plus timestamp is just a check and is not saved anywhere on BE.

WRadoslaw:
> > Like all requests return a timestamp in the response headers this could be used to flag the few users with wrong system times.
> 
> The only drawback of this approach I can think of atm is the availability of this external endpoint for the user. Is there anything else?

Fortunately, even if this endpoint is not available it will default quickly to the `Date.now`, which is all we have rn

thesan:
> The problem is that this timestamp needs to be signed with a private key before it's sent, or I'm missing something in what you mean.

@WRadoslaw no what I meant was to use the responses of other requests. It was just a general idea but a more concrete application would be e.g: to have some function run on every gql response (even 400) to check the response header for time difference and store the weirdly high ones somewhere (context, local storage...). So when it's time to sign timestamps there's no need to rely on a 3rd party.

> The only drawback of this approach I can think of atm is the availability of this external endpoint for the user. Is there anything else?

@attemka yeah I agree this definitely should be merged, it just feels weird to me to long term rely on a 3rd party api call for all users for something as basic as getting the time.

Issue #4512, opened 2023-07-27
Handle the case of informing the users that sync starts after the channel is verified


dmtrjsg:
## Context
- https://github.com/Joystream/youtube-synch/issues/224

## Scope
Upon success of YT sign up we need to 
- [ ]  Add info about the sync to the splash screen
- [ ]  Update the copy of the info bar in the YPP dashboard

"Your channel needs to get verified before content syncing starts.  It normally takes 12-48 hours for channels to get verified."

KubaMikolajczyk:
The changes can be found here ✅ : https://www.figma.com/file/oQqFqdAiPu16eeE2aA5AD5/YouTube-Partner-Program?type=design&node-id=4497-96722&mode=design&t=DrBv5OcngjlM64b3-4

PR #4555, opened 2023-08-02
YPP sync copy change

attemka:
#4512


WRadoslaw (approved):
Looks good. One thing wonders me, rn on dev `enrollmentReward` is incorrect. We should display the reward in USD, but currently, we have 5k token units. Should we adjust it here, or it's taken care of in Gleev?

attemka:
@WRadoslaw good catch. I discussed with @dmtrjsg and updated the banner. Thx for noticing

Issue #3387, opened 2022-11-01
Create CRT Flow Custom Params


dmtrjsg:
### Context
- https://github.com/Joystream/atlas/issues/3384

### Scope
- [ ] Custom assurance options summon additional options

[Link to Figma> ](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/CRT-Creator?node-id=178%3A50883)

<img width="738" alt="Screenshot 2022-11-01 at 18 09 36" src="https://user-images.githubusercontent.com/93924525/199253411-cc7a9691-f6e9-4a99-b651-03e3908941df.png">

polikosi:
tested on: https://atlas-git-crt-joystream.vercel.app/studio/crt

❌Second stage (token assurances -> Custom -> First payout)
It should be 100% maximum, no more
![image](https://github.com/Joystream/atlas/assets/80429191/067a196f-6f62-4470-b92d-d7792406837d)
❌cannot be a negative number
![image](https://github.com/Joystream/atlas/assets/80429191/dc6d4e37-2b92-4dcc-b47a-dbbfbc1e6cb1)

polikosi:
✅Custom assurance options summon additional options
⚠️But cannot be more or less than 100%
![image](https://github.com/Joystream/atlas/assets/80429191/b59cc35b-d599-416d-896d-dad47f83f7e2)

polikosi:
✅Custom assurance options summon additional options

Out of scope
⚠️But cannot be more or less than 100%

polikosi:
Remove the addition and subtraction signs. 
Make only numbers and commas/dots available
![image](https://github.com/Joystream/atlas/assets/80429191/4bbaf664-0f5e-4529-b841-c37d7d045d1d)

WRadoslaw:
[...]

Issue #3385, opened 2022-11-01
Create CRT Flow


dmtrjsg:
### Context

- https://github.com/Joystream/atlas/issues/3384

### Scope
- Welcome screen  [link to figma>](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/CRT-Creator?node-id=80%3A37156)
- Create token page scaffolding
- First stage from the flow (name, access, revenue share, creator award)
- Second stage (token assurances, only presets)
- Token Summary Screen
- All tooltips on all stages

Out of scope ⛔ 
- Previews on the left hand side
- Custom assurance inputs
- Submit token extrinsic

### Design
[Link to Figma user story > ](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/CRT-Creator?node-id=79%3A37125)

polikosi:
tested on: https://atlas-git-crt-joystream.vercel.app/studio/crt

I think there should be a description. The same problem with Secure, Risky and Custom token assurances
![image](https://github.com/Joystream/atlas/assets/80429191/eb199844-67e1-46c3-b78a-7829b4f67725)

There should be no chart here. 
![image](https://github.com/Joystream/atlas/assets/80429191/777b9f9a-6238-4e25-88fe-b715aa9c08ea)

polikosi:
Token Summary Screen
Selected "Secure" on Second stage. I expected see here information about cliff, vesting period and first payout. The same problem when I selected "Safe". Displays the information about the cliff, the vesting period, and the first payout only when I select Custom.
![image](https://github.com/Joystream/atlas/assets/80429191/1b9c42dc-a7d9-4dc1-9188-d874190255a6)

According to the design "Transaction" should be separated by a line
![image](https://github.com/Joystream/atlas/assets/80429191/550b051d-3027-45ee-87b7-c923b84f8a71)

polikosi:
![image](https://github.com/Joystream/atlas/assets/80429191/56efd1de-5521-463e-bb2a-29f68875b850)

PR #4564, opened 2023-08-03
🇩🇪 Create token flow

WRadoslaw:
#3385 
#3387

### DIFF
--- a/packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
+++ b/packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
@@ -96,11 +96,7 @@ export const SetupTokenStep = ({ setPrimaryButtonProps, onSubmit, form }: SetupT
         description="Define the share of your channel revenue that goes to yourself vs shared with your token holders.
 Recommended values — Channel: 80%, Holders:20%. "
       >
-        <Controller
-          name="revenueShare"
-          control={control}
-          render={({ field }) => <RatioSlider id="revenueShare" {...field} />}
-        />
+        <Controller name="revenueShare" control={control} render={({ field }) => <RatioSlider {...field} />} />
       </FormField>
       <FormField
         label="Annual creator reward"
@@ -112,11 +108,7 @@ Recommended values — Channel: 80%, Holders:20%. "
         }}
         description="Define your own reward for managing the tokens. 10% means that if you have 10000k of tokens exist and this amount does not change, an additional 1k tokens will get minted and added to your wallet gradually over the course of 1 year."
       >
-        <Controller
-          name="creatorReward"
-          control={control}
-          render={({ field }) => <RatioSlider id="creatorReward" {...field} />}
-        />
+        <Controller name="creatorReward" control={control} render={({ field }) => <RatioSlider {...field} />} />
       </FormField>
     </CrtFormWrapper>
   )
--- a/packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
+++ b/packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
@@ -49,7 +49,7 @@ export const SetupTokenStep = ({ setPrimaryButtonProps, onSubmit, form }: SetupT
   })
 
   return (
-    <CrtFormWrapper title="Set up your token" subtitle="Enter basic token information and settings." titleLink="">
+    <CrtFormWrapper title="Set up your token" subtitle="Enter basic token information and settings." learnMoreLink="">
       <FormField
         label="Name"
         description="Choose 3 letter name for your token to be displayed on your token page, tokens marketplace and in your buyers’ portfolio."
--- a/packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
+++ b/packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
@@ -1,6 +1,7 @@
 import { Controller, useForm } from 'react-hook-form'
 
 import { SvgActionLock, SvgActionUnlocked } from '@/assets/icons'
+import { Text } from '@/components/Text'
 import { SetupStepForm } from '@/components/_crt/CreateTokenDrawer/CreateTokenDrawer.types'
 import { CrtFormWrapper } from '@/components/_crt/CrtFormWrapper'
 import { FormField } from '@/components/_inputs/FormField'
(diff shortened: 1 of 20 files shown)


packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.tsx
```diff
@@ -0,0 +1,194 @@
...
+              tooltip={{
+                text: 'Do you want your tokens to be gradually available for you to sell, sending the signal to your audience that this project is aimed on long term success? Then choose longer vesting.',
+              }}
+              error={errors.vesting?.message}
+            >
+              <Controller
+                name="vesting"
+                control={control}
+                render={({ field }) => <Select items={vestingOptions} {...field} />}
+              />
```
thesan, line 103: Shouldn't the validation insure that vesting can't be lower than cliff ?

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.tsx
```diff
@@ -0,0 +1,194 @@
...
+            </FormField>
+            {customVesting && customVesting !== '0' ? (
+              <FormField
+                label="First payout"
+                description="A portion of your own tokens that will be released to you right after cliff period."
```
thesan, line 108: I thought this was what followers would receive "now"

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.tsx
```diff
@@ -0,0 +1,194 @@
...
+                tooltip={{
+                  text: 'Do you want to send the signal to your token buyers that only a portion of all created tokens is possible to get sold and the rest will get unlocked over time, signalling about long term goals of your project? Then we advise you to choose amount less than 50% here.',
+                }}
+                error={errors.firstPayout?.message}
+              >
+                <Controller
+                  name="firstPayout"
+                  control={control}
+                  render={({ field }) => (
+                    <TokenInput
```
thesan, line 118: Does this component insure that the value can't be lower than 0 or bigger than 100 ?

Also I can't find the design of these fields but IMO this should the first field. The way I understand it if it's set to 100%: none of the other fields matters. Then I would see the vesting and then the cliff.

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.utils.ts
```diff
@@ -0,0 +1,113 @@
...
+  },
+  {
+    value: '3',
+    name: '3 month',
+  },
+  {
+    value: '6',
+    name: '6 month',
+  },
+]
```
thesan, line 63: The vesting values seems too low the Secure type can't be reproduced with a slightly different cliff or part of the tokens available directly.

Shouldn't these be arbitrary number of months instead of a select ?

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
```diff
@@ -0,0 +1,123 @@
...
+      <FormField
+        label="Revenue share with holders"
+        description="Define the share of your channel revenue that goes to yourself vs shared with your token holders.
+Recommended values — Channel: 80%, Holders:20%. "
+      >
+        <Controller
+          name="revenueShare"
+          control={control}
+          render={({ field }) => <RatioSlider id="revenueShare" {...field} />}
+        />
```
thesan: I don't remember if I tested it but since `RatioSlider` can work as an uncontrolled component doesn't it work without being wrapped in `<Controller ...` ?
```suggestion
        <RatioSlider {...register('revenueShare')} id="revenueShare" />
```

packages/atlas/src/components/_inputs/Slider/RatioSlider.tsx
```diff
@@ -14,7 +15,7 @@ type RatioSliderProps = {
 }
 
 export const RatioSlider = forwardRef<HTMLInputElement, RatioSliderProps>(
-  ({ min = 0, max = 100, step = 10, defaultValue = 50, value: controlledValue, onChange }, ref) => {
+  ({ min = 0, max = 100, step = 10, defaultValue = 50, value: controlledValue, onChange, id = '' }, ref) => {
```
thesan: Setting an id is a good fix but IMO most dev would expect it to be set on the `<input id={id}` too.

Also for when the `id` isn't set I think it should not get a default value. Instead:
```js
const internalId = useMemo(() => id ?? createId(), [id])
```
could be used on:
```jsx
<mask id={`cutout-mask-${internalId}`}>
```

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.tsx
thesan: Actually in case there are no designs for this yet why not using the `RatioSlider` here too ?

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/SetupTokenStep.tsx
thesan: @WRadoslaw actually that probably won't work since the input itself is controlled

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.tsx
WRadoslaw: Discussed on daily

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.tsx
WRadoslaw: Discussed on daily

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.tsx
WRadoslaw: Discussed on daily

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenIssuanceStep/TokenIssuanceStep.utils.ts
WRadoslaw: Discussed on daily

thesan (approved):

Issue #4575, opened 2023-08-07
Follow up enabler of source map deployment


dmtrjsg:
Optimise the build for sourcemaps to unlock debugging of Sentry errors

PR #4580, opened 2023-08-07
💹 Fix chunk sizes

WRadoslaw:
#4575

### DIFF
--- a/packages/atlas/src/MainLayout.tsx
+++ b/packages/atlas/src/MainLayout.tsx
@@ -108,12 +108,14 @@ export const MainLayout: FC = () => {
   return (
     <>
       <CookiePopover />
+      {/*<Suspense fallback={<div>adawd</div>}>*/}
       <Routes>
         <Route path={BASE_PATHS.viewer + '/*'} element={<ViewerLayout />} />
         <Route path={BASE_PATHS.legal + '/*'} element={<LegalLayout />} />
         <Route path={BASE_PATHS.studio + '/*'} element={<LoadableStudioLayout />} />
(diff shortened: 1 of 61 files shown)


packages/atlas/src/MainLayout.tsx
```diff
@@ -108,12 +108,14 @@ export const MainLayout: FC = () => {
   return (
     <>
       <CookiePopover />
+      {/*<Suspense fallback={<div>adawd</div>}>*/}
```
attemka, line 111: suspense will be included when it'll be supported by apollo client

packages/atlas/src/components/StudioEntrypoint/StudioEntrypoint.tsx
```diff
@@ -35,3 +35,5 @@ export const StudioEntrypoint: FC<StudioEntrypointProps> = ({ enterLocation }) =
 
   return <StudioLoading />
 }
+
+export default StudioEntrypoint
```
attemka, line 39: we should't change it in this PR, but generally we should avoid default export in files

attemka (approved):

packages/atlas/src/components/StudioEntrypoint/StudioEntrypoint.tsx
WRadoslaw: `React.lazy` requires default export

packages/atlas/src/MainLayout.tsx
WRadoslaw: This should be deleted, since I introduced it in `Viewer` and `Studio` layout

PR #4683, opened 2023-08-17
Segment followup fixes

attemka:


packages/atlas/src/.env
```diff
@@ -1,10 +1,10 @@
 # This file is committed. Do not store secrets here
 
 # should be set to "production" for production builds
-VITE_ENV=development
+VITE_ENV=production
```
WRadoslaw: Why do we switch to production on dev

WRadoslaw (approved):

Issue #3826, opened 2023-02-16
CRT Viewer: Token Page - Header Widget


dmtrjsg:
## Context
- #3384 

## Scope

Creator token description
-  Header widget
- Tooltips on Hover

## Design

[Link to Figma>](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=1963%3A105418&t=XWyT31UxvE7yKN66-1)

Preview: 
<img width="583" alt="Screenshot 2023-02-16 at 17 27 55" src="https://user-images.githubusercontent.com/93924525/219377730-874ac5a0-f8ef-4cbc-b36c-52c5520ac982.png">

ivanturlakov:
Tested on https://atlas-o35urjwe5-joystream.vercel.app/channel/39?tab=Token

⚠️ Header widget
-Short description duplicates information from the About section
-"invitу only" badge seems to be hardcoded, as it shows up regardless of the settings when the token is created

<img width="1327" alt="Screenshot 2023-12-06 at 16 02 01" src="https://github.com/Joystream/atlas/assets/46903215/2bc07966-0c9b-4b4b-88dc-91ec8210e3b5">

✅ Tooltips
<img width="343" alt="Screenshot 2023-12-06 at 16 08 52" src="https://github.com/Joystream/atlas/assets/46903215/46e6aa66-8880-4117-98cd-190e4d6fc88e">
<img width="266" alt="Screenshot 2023-12-06 at 16 09 00" src="https://github.com/Joystream/atlas/assets/46903215/efad7041-d212-41c1-bdd5-bca1b0c9dd20">
<img width="353" alt="Screenshot 2023-12-06 at 16 09 08" src="https://github.com/Joystream/atlas/assets/46903215/5687b42a-c5d0-416d-a317-e8327ef8b1ad">
<img width="259" alt="Screenshot 2023-12-06 at 16 09 15" src="https://github.com/Joystream/atlas/assets/46903215/2aa0f99b-8bf0-49a6-a47b-ab6c4341e951">

WRadoslaw:
> -Short description duplicates information from the About section

@KubaMikolajczyk there is no short description field in the schema and form for the CRT, should we just delete it?

PR #4693, opened 2023-08-18
👨 CRT header widget

WRadoslaw:
#3826


attemka (approved):

thesan (commented):
In term of code it looks good but it looks different from the design [on figma](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel?type=design&node-id=1963-105439). Has there been changes since then ?

thesan:
@WRadoslaw 
The last differences are:
- The joy icons on the percentage (revenue share and annual reward) which don't appear in the design.
- "Token name" showing on hover on the token (I think it should be just a title attribute).

Here's the figma design:
image
And the story:
image

Also there is a conflict.

thesan (commented):
@WRadoslaw 
The last differences are:
- The joy icons on the percentage (revenue share and annual reward) which don't appear in the design.
- "Token name" showing on hover on the token (I think it should be just a title attribute).

Here's the figma design:
image
And the story:
image

Also there is a conflict.

packages/atlas/src/components/_crt/CrtBasicInfoWidget/CrtBasicInfoWidget.tsx
```diff
@@ -1,64 +1,45 @@
...
 
-export const CrtBasicInfoWidget = ({ creatorReward, revenueShare, totalRevenue, name }: CrtBasicInfoWidgetProps) => {
+export const CrtBasicInfoWidget = ({ name, details, titleColor }: CrtBasicInfoWidgetProps) => {
+  const titleRef = useRef<HTMLSpanElement | null>(null)
   const smMatch = useMediaMatch('sm')
-
   return (
     <WidgetTile
-      title={`$${name || 'ABC'}`}
+      title={<span ref={titleRef}>${name ?? 'ABC'}</span>}
```
thesan: Somehow the tooltip doesn't to show for me. Why not just:
```suggestion
      title={<span title="Token name">${name ?? 'ABC'}</span>}
```

WRadoslaw: It's just because I used ref and the component does not rerender after it's set. The `title` attribute imo is too slow, especially since all of our tooltips appear almost instantly.

Replace useRef for useState

Issue #3833, opened 2023-02-16
CRT Viewer: Buy Token - Chart Variations


dmtrjsg:
## Context
- #3384 

## Scope
- [ ] Chart for cliff, vesting, first payout
- [ ] [Interaction](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=2003%3A127310&t=XWyT31UxvE7yKN66-1)
- [ ] [Tooltips](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=2004%3A129512&t=XWyT31UxvE7yKN66-1)

## Design

[Link to Figma> ](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=2003%3A123614&t=XWyT31UxvE7yKN66-1)

Preview: 
<img width="392" alt="Screenshot 2023-02-16 at 17 53 10" src="https://user-images.githubusercontent.com/93924525/219383348-da13fc9e-ac82-4e56-ab8e-6dd670d63b43.png">

Issue #3830, opened 2023-02-16
CRT Viewer: Token Page - Buy Token On Sale


dmtrjsg:
## Context
- #3384 

## Scope

- [ ] Token on sale widget
- [ ] Step 1 - how much to buy (excl Max?)
- [ ] Step 2 - Cliff and Payouts chart and params 
- [ ] [All validations: link > ](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=2000%3A122810&t=XWyT31UxvE7yKN66-1)

☝️ T&Cs are scrollable

Out of scope: 
Submit TX 
Conf modal
Buy max 
Chart for Vesting and Cliff

## Design

[Link to Figma> ](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=1903%3A117720&t=XWyT31UxvE7yKN66-1)

Preview: 

<img width="201" alt="Screenshot 2023-02-16 at 17 45 18" src="https://user-images.githubusercontent.com/93924525/219381535-10e98886-3acd-4b1b-8615-fa18a451d77a.png">
<img width="297" alt="Screenshot 2023-02-16 at 17 45 21" src="https://user-images.githubusercontent.com/93924525/219381548-f0f4ee1e-d652-4def-926a-5e8a6acd3e04.png">
<img width="286" alt="Screenshot 2023-02-16 at 17 45 27" src="https://user-images.githubusercontent.com/93924525/219381565-71e9c354-a5b5-426c-81af-650eb4222866.png">

ivanturlakov:
Tested on https://atlas-o35urjwe5-joystream.vercel.app/

@dmtrjsg Sale tokens option is disabled on testing env, there is another env where we can test?

<img width="1169" alt="Screenshot 2023-12-07 at 15 17 02" src="https://github.com/Joystream/atlas/assets/46903215/d6f7609c-1962-4e5e-8bcf-697464700ce5">

PR #4728, opened 2023-08-23
💸 Buy token on sale modal

WRadoslaw:
#3830 
#3833

### DIFF
--- a/packages/atlas/src/components/_crt/BuySaleTokenModal/BuySaleTokenModal.tsx
+++ b/packages/atlas/src/components/_crt/BuySaleTokenModal/BuySaleTokenModal.tsx
@@ -46,6 +46,7 @@ export const BuySaleTokenModal = ({ tokenId, onClose }: BuySaleTokenModalProps) 
       onExitClick={onClose}
       primaryButton={primaryButtonProps}
       secondaryButton={secondaryButton}
+      noContentPadding={activeStep === BUY_SALE_TOKEN_STEPS.terms}
     >
       {activeStep === BUY_SALE_TOKEN_STEPS.form && (
         <BuySaleTokenForm
--- a/packages/atlas/src/components/_crt/BuySaleTokenModal/BuySaleTokenModal.tsx
+++ b/packages/atlas/src/components/_crt/BuySaleTokenModal/BuySaleTokenModal.tsx
@@ -1,0 +1,89 @@
+import { useMemo, useRef, useState } from 'react'
+import { flushSync } from 'react-dom'
+import { CSSTransition, SwitchTransition } from 'react-transition-group'
+
+import { BuySaleTokenForm } from '@/components/_crt/BuySaleTokenModal/steps/BuySaleTokenForm'
+import { BuySaleTokenTerms } from '@/components/_crt/BuySaleTokenModal/steps/BuySaleTokenTerms'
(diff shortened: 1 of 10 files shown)


packages/atlas/src/components/_crt/BuySaleTokenModal/BuySaleTokenModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, [activeStep])
+
+  const commonProps = {
+    setPrimaryButtonProps,
+  }
+
+  return (
+    <DialogModal
+      title={activeStep !== BUY_SALE_TOKEN_STEPS.success ? 'Buy $JBC' : undefined}
```
attemka: should fetch the token name by id?

packages/atlas/src/components/_crt/BuySaleTokenModal/BuySaleTokenModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+      onExitClick={activeStep !== BUY_SALE_TOKEN_STEPS.success ? onClose : undefined}
+      dividers={activeStep === BUY_SALE_TOKEN_STEPS.terms}
+      show
+      primaryButton={primaryButtonProps}
+      secondaryButton={secondaryButton}
+      noContentPadding={activeStep === BUY_SALE_TOKEN_STEPS.terms}
+      confetti={activeStep === BUY_SALE_TOKEN_STEPS.success && smMatch}
+    >
+      {activeStep === BUY_SALE_TOKEN_STEPS.form && (
```
attemka, line 62: ofc we can do like that, but I think it would be better to keep the unified code style and use `switch(activeStep)` here. wdyt?

packages/atlas/src/components/_crt/BuySaleTokenModal/steps/BuySaleTokenForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {
+        title: 'Tokens on sale',
+        content: (
+          <NumberFormat
+            value={tokensOnSale}
+            as="p"
+            variant="t200"
+            withDenomination="before"
+            withToken
+            customTicker="$JBC"
```
attemka: since we don't have any data maybe it'll be good to add mocked data instead of hardcoded values? By doing this, transition to real data could be smoother?

packages/atlas/src/components/_crt/BuySaleTokenModal/steps/BuySaleTokenSuccess.tsx
```diff
@@ -0,0 +1,59 @@
...
+            />
+          </LottieContainer>
+        )}
+      </IllustrationWrapper>
+      <ContentWrapper>
+        <Text variant="h500" as="h2" margin={{ bottom: 2 }}>
+          Congratulations on your purchase
+        </Text>
+        <Text variant="t200" as="p" color="colorText">
+          You are now an {tokenName} holder - you can buy, sell, transfer your token. You can also stake when creator
```
attemka, line 53: e.g here you have a var for token name

attemka (changes requested):
idk regarding hardcoded data, let me know your thoughts

Issue #3836, opened 2023-02-16
CRT Viewer: Token Page - Sell tokens on Market


dmtrjsg:
## Context
- #3384 

## Scope

- [ ] UI and Submit Sale tx for AMM 

## Design

[Link to figma> ](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=1903%3A117692&t=XWyT31UxvE7yKN66-1)

Preview: 
<img width="286" alt="Screenshot 2023-02-16 at 18 03 54" src="https://user-images.githubusercontent.com/93924525/219386325-d4faaef5-33f2-4225-a520-f4728502058a.png">

ivanturlakov:
Tested on https://atlas-o35urjwe5-joystream.vercel.app/channel/23?tab=Token

Design
<img width="602" alt="Screenshot 2023-12-09 at 17 25 15" src="https://github.com/Joystream/atlas/assets/46903215/736a8845-71d2-484a-bb4a-31797aee232e">

- Placeholder design a bit missed(input field)
- Tooltips is empty
<img width="498" alt="Screenshot 2023-12-09 at 17 21 08" src="https://github.com/Joystream/atlas/assets/46903215/7548485b-251a-434b-b23c-1773cadec15f">
<img width="498" alt="Screenshot 2023-12-09 at 17 21 17" src="https://github.com/Joystream/atlas/assets/46903215/e8d27c93-a582-44ac-90e9-61ee536b2d03">

Tx Failed
https://polkadot.js.org/apps/?rpc=wss://3.123.129.217.nip.io/ws-rpc#/explorer/query/0x5a8f6f453e7f65853c1f71691062f3c05862bc8182ccb1dffaa589dccf5b9255
<img width="1440" alt="Screenshot 2023-12-09 at 17 21 40" src="https://github.com/Joystream/atlas/assets/46903215/8ef65de2-acca-49f5-9d3c-f5e3007c8b7a">

PR #4753, opened 2023-08-25
🧬 Sell token modal

WRadoslaw:
#3836

### DIFF
--- a/packages/atlas/src/components/_crt/SellTokenModal/SellTokenModal.tsx
+++ b/packages/atlas/src/components/_crt/SellTokenModal/SellTokenModal.tsx
@@ -1,0 +1,121 @@
+import { useMemo, useState } from 'react'
+
+import { FlexBox } from '@/components/FlexBox/FlexBox'
+import { Information } from '@/components/Information'
+import { JoyTokenIcon } from '@/components/JoyTokenIcon'
+import { NumberFormat } from '@/components/NumberFormat'
+import { Text } from '@/components/Text'
+import { TextButton } from '@/components/_buttons/Button'
+import { FormField } from '@/components/_inputs/FormField'
+import { TokenInput } from '@/components/_inputs/TokenInput'
+import { DetailsContent } from '@/components/_nft/NftTile'
+import { DialogModal } from '@/components/_overlays/DialogModal'
+import { useMediaMatch } from '@/hooks/useMediaMatch'
+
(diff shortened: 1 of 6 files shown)


packages/atlas/src/components/_crt/SellTokenModal/SellTokenModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+      {
+        title: 'You will get',
+        content: (
+          <NumberFormat
+            value={tokens || 0}
+            as="p"
+            variant="t200"
+            withDenomination="before"
+            withToken
+            customTicker="$JBC"
```
attemka: you're selling $JBC here, so you should receive `${atlasConfig.joystream.tokenTicker}`?

attemka: same thing with align as in token buy modal

attemka (changes requested):

packages/atlas/src/components/_crt/SellTokenModal/SellTokenModal.tsx
thesan:
```suggestion
            customTicker={`$${title}`}
```

packages/atlas/src/components/_crt/SellTokenModal/SellTokenModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+            content={pricePerUnit}
+            icon={<JoyTokenIcon size={smMatch ? 24 : 16} variant="silver" />}
+            withDenomination
+          />
+          <DetailsContent
+            avoidIconStyling
+            tileSize={smMatch ? 'big' : 'bigSmall'}
+            caption={`YOUR $${title} BALANCE`}
+            content={userBalance}
+            icon={<JoyTokenIcon size={smMatch ? 24 : 16} variant="silver" />}
```
thesan, line 86:
```suggestion
```

But also if I understand correctly `userBalance` is in $JBC but the $USD conversion is based on the current $JOY/$USD rate so the $USD value will be wrong.

WRadoslaw: I made it a separate issue #4761

attemka (approved):

Issue #4749, opened 2023-08-25
Switch the flow to creating membership on chain and Orion in a single go


dmtrjsg:
To prevent users being left out with no Orion membership, chain membership and no registration on YPP

PR #4758, opened 2023-08-28
💞 YPP flow reversal

WRadoslaw:
#4749

### DIFF
--- a/packages/atlas/src/.env
+++ b/packages/atlas/src/.env
@@ -38,7 +38,7 @@ VITE_PRODUCTION_FAUCET_URL=https://faucet.joystream.org/member-faucet/register
 VITE_PRODUCTION_YPP_FAUCET_URL=https://18.184.136.237.nip.io/membership
 
 # Development env URLs - this is the default configuration if VITE_ENV != production
-VITE_DEVELOPMENT_ORION_AUTH_URL=https://atlas-dev.joystream.org/orion-auth/api/v1
+VITE_DEVELOPMENT_ORION_AUTH_URL=https://atlas-dev.joystream.org/api/v1
 VITE_DEVELOPMENT_ORION_URL=https://atlas-dev.joystream.org/orion-api/graphql
 VITE_DEVELOPMENT_QUERY_NODE_SUBSCRIPTION_URL=wss://atlas-dev.joystream.org/orion-v2/graphql
 VITE_DEVELOPMENT_NODE_URL=wss://atlas-dev.joystream.org/ws-rpc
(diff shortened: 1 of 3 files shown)


thesan (approved):
LGTM but I was wondering if something like this :point_down: would be possible instead of relying on `handleOrionAccountCreation` errors to know if the membership was create. Because from my understanding the Faucet does create the membership and Orion needs to query this membership based on something whether it's `ytResponseData.userId` or `signUpFormData`:

```js
const [getMembership, { data, startPolling }] = useGetMembershipLazyQuery({ pollInterval: 10_000 })

useEffect(() => {
  if (!ytResponseData?.userId) return

  getMembership({ variables: where: { /* use the ytt userId or signUpFormData somehow */ } })
  startPolling()
}, [ytResponseData?.userId]

useEffect(() => {
  if (!data.memberships[0].id) return

  handleOrionAccountCreation()
}, [data.memberships[0].id])
```

Would this be viable or am I missing something ?

WRadoslaw:
> 

It would definitely work, but I'm a little bit afraid of keeping this logic in `useEffect` (tbh I'm trying not to use it at all recently), I think it's less reliable and in larger components harder to read/debug effects are not really linked to each other, so you have to know the logic to get the flow. With "event" based flow some1 can just check `what fn is called onSubmit for the last step -> goes to fn definition -> see that after the faucet call is done or timeout it will try to call another fn -> can go to the definition of new fn`. In this case, recursion + setTimeout introduces some complexity, but I'd say it's still more reliant than react hooks (aside from using `useRef` which is for tracking the number of requests, but this is forced by how React components behave).

@attemka I leave the decision to you

thesan:
Yeah I agree the 1st `useEffect` should probably be avoided and instead a fn calling `getMembership` and `startPolling` should be called after the "see that after the faucet call is done or timeout" (it was just easier to demonstrate the idea with a self contained `useEffect`. But AFAIK the second `useEffect` would be needed due to how the Apollo works.

attemka:
> ```js
> useGetMembershipLazyQuery
> ```

@WRadoslaw  so how about this: 1st hook moved to function as @thesan said, 2nd hook moved to  
```
const [getMembership, { data, startPolling }] = useGetMembershipLazyQuery({ pollInterval: 10_000, onCompleted: (data) => {
  if (!data.memberships[0].id) return
  handleOrionAccountCreation()
} })

```

so both hooks will be eliminated?

WRadoslaw:
> const [getMembership, { data, startPolling }] = useGetMembershipLazyQuery({ pollInterval: 10_000, onCompleted: (data) => {
>   if (!data.memberships[0].id) return
>   handleOrionAccountCreation()
> } })

At this point should I just remove the block sync action if we are polling anyway, or do you want to poll after sync/timeout fires?

attemka:
> At this point should I just remove the block sync action if we are polling anyway, or do you want to poll after sync/timeout fires?

sure, lets just remove it

PR #4820, opened 2023-09-06
Fixed rewards in YPP dashboard

attemka:


thesan (approved):

packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardMainTab.tsx
```diff
@@ -98,20 +98,26 @@ export const YppDashboardMainTab: FC<YppDashboardMainTabProps> = ({ currentTier
...
-          const customMultiplier = reward.customMultiplier
+          const customMultiplier = reward.customMultiplier?.[currentTier]
           const rewardAmount = reward.joyAmount
             ? typeof reward.joyAmount === 'number'
               ? {
                   type: 'number' as const,
-                  amount: reward.joyAmount * (customMultiplier ? customMultiplier[currentTier] : multiplier),
+                  amount:
+                    reward.joyAmount === 0
+                      ? customMultiplier || multiplier
```
WRadoslaw: Why do we want to display the multiplier as an amount if there isn't any reward?

packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardMainTab.tsx
```diff
@@ -98,20 +98,26 @@ export const YppDashboardMainTab: FC<YppDashboardMainTabProps> = ({ currentTier
...
               : { type: 'range' as const, min: reward.joyAmount.min, max: reward.joyAmount.max }
             : null
           const rewardAmountUsd = reward.usdAmount
             ? typeof reward.usdAmount === 'number'
               ? {
                   type: 'number' as const,
-                  amount: reward.usdAmount * (customMultiplier ? customMultiplier[currentTier] : multiplier),
+                  amount:
+                    reward.usdAmount === 0
+                      ? customMultiplier || 0
```
WRadoslaw: Same as above

packages/atlas/src/views/studio/YppDashboard/tabs/YppDashboardMainTab.tsx
attemka: the thing is that this is used to display the rewards with the `baseAmount === 0`. The thing is that usually we'll do the `baseAmount * multiplier`, but in this case it'll result in all rewards will became 0. I don't want to create a separate var for that, but now I think that I can extend the type of baseReward to `number[]` and use the values without a multiplier if its that type. It seems that it'll be more obvious than to use the multiplier values.

attemka: @WRadoslaw please check the updated solution

WRadoslaw (approved):

Issue #3844, opened 2023-02-17
Creator: Revenue Share Tab


dmtrjsg:
## Context
- #3384 

## Scope
- [ ] Revenue Share Tab
- No ongoing sale placeholder widget
- [ ] Revenue Share Widgets
-- Revenue Share ratio

⚠️ only do revenue share ratio widget in this version.

⛔ Out of scope:
-- Current state
-- Channel balance

Changes will be covered here:
- https://github.com/joystream/atlas/issues/4856

## Design
[Link to Figma> ](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/My-token?node-id=787%3A104748&t=9ogR3tZzrka32tXb-1)

Preview: 
<img width="683" alt="Screenshot 2023-02-17 at 14 36 19" src="https://user-images.githubusercontent.com/93924525/219621368-16fdc2c2-de27-4d10-8a33-5f0f1e2a60c4.png">

polikosi:
here should be "No ongoing revenue share"
![image](https://github.com/Joystream/atlas/assets/80429191/075e6b53-436c-40aa-88ca-40e7ead15513)

PR #4867, opened 2023-09-14
🍶 CRT revenue tab

WRadoslaw:
#3844

### DIFF
--- a/packages/atlas/src/components/RatioPreview/RatioPreview.tsx
+++ b/packages/atlas/src/components/RatioPreview/RatioPreview.tsx
@@ -1,0 +1,46 @@
+import styled from '@emotion/styled'
+
+import { FlexBox } from '@/components/FlexBox'
+import { Text } from '@/components/Text'
+import { cVar } from '@/styles'
+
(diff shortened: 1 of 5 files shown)


packages/atlas/src/components/RatioPreview/RatioPreview.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+type RatioPreviewProps = {
+  ratios: [[number, string], [number, string]]
+}
+
+export const RatioPreview = ({ ratios }: RatioPreviewProps) => {
+  const firstRatio = ratios[0]
+  const secondRatio = ratios[1]
+  return (
+    <FlexBox flow="column" gap={3.5}>
```
attemka, line 15: shouldn't you use the `sizes` here?

attemka (approved):

packages/atlas/src/components/RatioPreview/RatioPreview.tsx
WRadoslaw: `Sizes` is used to parse gap value into px

Issue #3798, opened 2023-02-15
Token Start Market flow Chart


dmtrjsg:
## Context

- https://github.com/Joystream/atlas/issues/3384

## Scope
- Chart
⚠️ Chart needs to be a straight line, where A is the slope:
<img width="572" alt="Screenshot 2023-02-15 at 18 36 49" src="https://user-images.githubusercontent.com/93924525/219058619-aeed4c21-e8ef-42de-81bc-bfcfd4f3ce5f.png">

BE required

## Designs

[link> ](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/My-token?node-id=1219%3A159180&t=KiTZ2klK5CxMQJBl-1)

Preview: 
<img width="660" alt="Screenshot 2023-02-15 at 18 34 26" src="https://user-images.githubusercontent.com/93924525/219058599-25cb3ee5-225d-4518-be2e-90e106f82dd6.png">

polikosi:
![image](https://github.com/Joystream/atlas/assets/80429191/546c3842-d0a3-4ffd-b5ea-97da505e3d8b)

dmtrjsg:
@WRadoslaw I believe in this issue the following points would need to be finalised:

- chart title missing "how token price rises with growing total supply"
-  Y axis scale is a bit off (ideally needs to be 1k / 4k rounded up to full thousands.)
-  tooltip on chat hover is missing

Issue #3797, opened 2023-02-15
Token Start Market Flow UI


dmtrjsg:
## Context

- https://github.com/Joystream/atlas/issues/3384

## Scope
- Starting Market Screen One 
-- Price input
-- T&Cs
-- Token widget preview (incl total supply)
- Summary Screen

🚫  Out of scope: 
- Token widget fiat denomination
- Chart
- Submit tx

## Designs

[link> ](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/My-token?node-id=817%3A108160&t=KiTZ2klK5CxMQJBl-1)

Preview: 
<img width="835" alt="Screenshot 2023-02-15 at 18 27 23" src="https://user-images.githubusercontent.com/93924525/219055290-c1c7f810-cfb4-4763-b35f-4f44c47ad246.png">

<img width="867" alt="Screenshot 2023-02-15 at 18 32 08" src="https://user-images.githubusercontent.com/93924525/219056516-81a9a75a-1950-4de1-b24c-6ec14fa15098.png">

polikosi:
- Starting Market Screen One

1. On the first page shoudn't be a chart. I expected to see the "price per unit".
2. When I enter a negative number in the price field, the page crashes.
3. T&Cs  text does not match the design.

- Summary Screen
1. Add text above the graph that describes it


![image](https://github.com/Joystream/atlas/assets/80429191/afa39c88-b791-4965-9b74-fbc8b53f99d2)
![image](https://github.com/Joystream/atlas/assets/80429191/cb948ff2-d585-4ecd-8055-78b8ea1b469b)

https://github.com/Joystream/atlas/assets/80429191/373c7f28-32cb-48b1-bde4-9da1d619e2b6

PR #4884, opened 2023-09-18
Crt market start

attemka:
#3797 
#3798


packages/atlas/src/components/_crt/MarketDrawer/MarketDrawer.tsx
```diff
@@ -0,0 +1,251 @@
...
+    resetField,
+    formState: { isDirty },
+  } = useForm<CrtMarketForm>({
+    defaultValues: {
+      price: DEFAULT_MIN_PRICE,
+      tnc: atlasConfig.legal.crtTnc,
+      isChecked: true,
+    },
+  })
+  const tokenInUsd = (getValues('price') || 0) * (tokenPrice || 0)
```
WRadoslaw: I won't update on input change, RHF works on refs and getValues is lazy. `watch` should fix it

packages/atlas/src/components/_crt/MarketDrawer/MarketDrawer.tsx
```diff
@@ -0,0 +1,251 @@
...
+                    onChange(checked)
+                  }}
+                />
+              )}
+              name="isChecked"
+            />
+            <Controller
+              control={control}
+              render={({ field: { value: tnc, onChange } }) => (
+                <TextArea rows={7} value={tnc} disabled={getValues('isChecked')} onChange={onChange} />
```
WRadoslaw: Same thing as for USD calculation

WRadoslaw: *with disabled logic

packages/atlas/src/components/_crt/MarketDrawer/MarketDrawer.tsx
```diff
@@ -0,0 +1,251 @@
...
+            </Text>
+            <Controller
+              control={control}
+              render={({ field: { value: isChecked, onChange } }) => (
+                <Checkbox
+                  value={isChecked}
+                  label="Keep the default terms & conditions"
+                  onChange={(checked) => {
+                    if (checked) {
+                      resetField('tnc')
```
WRadoslaw: I think it won't bring back the default tnc, but can't check on story

WRadoslaw (changes requested):
1. Chart doesn't match figma color and equasion, missing chart title
image
image
2. Spacing on the summary is slightly off (too wide)
image
3. Same here 
image

packages/atlas/src/components/_crt/MarketDrawer/MarketDrawer.tsx
attemka: Why not? added this call programmatically, field value was reset

WRadoslaw (changes requested):
1. Charts break when there is no token amount
image
2. I can pass on to the summary without providing the amount or providing it lower than required in the copy, the same goes for T&C I can progress leaving to blank

packages/atlas/src/components/_crt/MarketDrawer/steps/MarketStep.tsx
```diff
@@ -0,0 +1,193 @@
...
+}) => {
+  const { tokenPrice } = useJoystream()
+  const {
+    control,
+    handleSubmit,
+    resetField,
+    watch,
+    formState: { isDirty, errors },
+  } = useForm<CrtMarketForm>({
+    defaultValues: {
```
WRadoslaw: This is kinda a problem since useForm is mounted directly on the step and step is getting unmounted if you proceed, data is lost when going back from summary to form. 

You would need to pass submitted data to the form, so you can define default values correctly and use predefined price and tnc as a fallback.

WRadoslaw (changes requested):
Other than that, all good

packages/atlas/src/components/_crt/MarketDrawer/steps/MarketStep.tsx
attemka: @WRadoslaw good catch, thank you!

attemka: @WRadoslaw good catch, thank you!

attemka: @WRadoslaw good catch, thank you!

WRadoslaw (approved):

Issue #3864, opened 2023-02-17
Holder Portfolio: Claim revenue from Revenue Share


dmtrjsg:
## Context
- #3384 

## Scope
- [ ] Revenue share - claim revenue flow
- [ ] Conf message

## Design
[Link to Figma>](https://www.figma.com/file/VvWuJfHYpOPFQrcDVAeHkT/My-portfolio?node-id=1227%3A50780&t=XolD3kk1lOSp2O9T-1)

Preview:

<img width="631" alt="Screenshot 2023-02-17 at 15 48 06" src="https://user-images.githubusercontent.com/93924525/219646242-f940be64-2274-4d43-9890-a5c25c07b743.png">

ivanturlakov:
Tested on https://atlas-dj3xj8za5-joystream.vercel.app/portfolio

⚠️ Revenue share - claim revenue flow

- YOUR_SHARE = 0(expected 0,86), YOUR_TOKENS = 0(expected 55), 

**Before Claim**
<img width="1328" alt="Screenshot 2023-12-21 at 16 45 12" src="https://github.com/Joystream/atlas/assets/46903215/b02b184e-bc59-42ab-89a7-1e669458b19f">

**After Claim**
<img width="1319" alt="Screenshot 2023-12-21 at 16 50 04" src="https://github.com/Joystream/atlas/assets/46903215/e8bcb977-9042-49c4-9386-51d235a5e16e">


✅ Conf message

PR #4906, opened 2023-09-20
🧇 Unlock revenue share

WRadoslaw:
#3864


attemka (approved):

WRadoslaw:
If approved it can be merged, a failing build is caused by submit token tx and it will be fix in #4912

Issue #4826, opened 2023-09-06
YPP 2.0 - Dashboard


dmtrjsg:
## Context
- https://github.com/Joystream/atlas/issues/4772
- https://github.com/Joystream/atlas/issues/4808

## Scope

:one: 
### Add YPP verification status

- Not signed up
- Verification pending
- Verified + Tier 
- Suspended + Reason

:two: 
### YPP Sync Status Upd
- Add number of their videos in the sync backlog 
- Place of the first video in the sync queue 
-  Eta to full sync (of their videos).

:three:
### New Tiers 
[Names can be updated, depends on what we arrive to on the landing page]
70% Iron - amateur quality, follwed by friends or rigged subscribers.
17% Bronze - Decent quality, something that you would follow if interested in the narrow topic and can compromise on quality for the content. 
10% Silver - Popular, good quality, well done, enjoyable to watch even if not interested in the narrow topic.
3% Gold - Absolute gold standard of quality. Professionally done. Recognised influencer status. Large follower audience, likely > 100k. 

:four:
### Suspended Reason
Display rejection reason for suspended channels
- Subpar quality
- Duplicated content
- Unsupported topic
- Program terms exploit

:five:
### Dashboard for all channels
- Display YPP Dashboard for channels not signed up to the program 
- Allow to sign up from dashboard
- Allow to go to landing page (viewer side)
- Anonymous users (not signed in) OR signed in users without channels would need to have their own view of the page

6️⃣ When channels are verified and Tier assigned, display how much they will get for the sign up reward. 

7️⃣ Payments are done the first Friday. Sign up bonus is paid on the first Friday after the date of sign up.

dmtrjsg:
Added point 6, following some requests in the support channel

KubaMikolajczyk:
Hey @dmtrjsg I explored the possible solutions for the issues you listed above and its ready to be reviewed ✅

🎥 Video update: https://www.loom.com/share/73a81cf3ca434813813ecbfc3bddcfd9?sid=d2092250-9abf-4454-8af8-8e376b396b50

📄 YPP Dashboard exploration file: https://www.figma.com/file/mSMwmccDKNDlZ5yVho4WFg/YPP-dashboard-explorations?type=design&node-id=5201-20498&mode=design&t=Uli6qnsriCDxSSgI-4

dmtrjsg:
@KubaMikolajczyk 

Great stuff, loving how its shaping up!

## Comments

### Overall
[...]

PR #4929, opened 2023-09-28
🌍 YPP Dashboard 2.0

WRadoslaw:
#4826


packages/atlas/src/components/_ypp/YppStatusPill/YppStatusDot.styles.ts
```diff
@@ -1,18 +1,18 @@
...
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  gap: ${sizes(2)};
-  background-color: ${cVar('colorCoreNeutral800Lighten')};
-  padding: ${sizes(2)} ${sizes(4)};
-  border-radius: 99px;
-  width: fit-content;
-`
+// export const Container = styled.div`
```
attemka, line 6: remove if not needed

packages/atlas/src/views/studio/MyVideosView/MyVideosView.tsx
```diff
@@ -346,7 +345,7 @@ export const MyVideosView = () => {
       <MintNftFirstTimeModal
         shouldHideNextTime={shouldHideMintModal}
         onShouldHideNextTime={setShouldHideMintModal}
-        show={showMintModal && currentChannel?.yppStatus === 'Verified' && !mintConfirmationDismissed}
+        show={showMintModal && !!currentChannel?.yppStatus.startsWith('Verified') && !mintConfirmationDismissed}
```
attemka, line 348: its either boolean value from `startsWith` or undefined from missing currentChannel, which is also falsy, so why would you need to use the double exclamation mark here?

attemka (changes requested):

packages/atlas/src/views/studio/MyVideosView/MyVideosView.tsx
WRadoslaw: For this modal `show` is a required `boolean`

attemka (approved):

Issue #3835, opened 2023-02-16
CRT Viewer: Buy Tokens on Market - Submit tx


dmtrjsg:
## Context
- #3384 

## Scope

- [ ] Submit purchase tx for AMM

## Design

[Link to Figma>](https://www.figma.com/file/POw4j95jszfIP5bGfas2IC/Channel-page?node-id=1903%3A117678&t=XWyT31UxvE7yKN66-1)

polikosi:
tested on: https://atlas-o35urjwe5-joystream.vercel.app

1. I can't see any confetti when I have successfully bought tokens. 
2. By the design, I expected to see a different photo on the "Congratulations on your purchase" dialog
3. I can't see the snackbar after closing the dialog
![image](https://github.com/Joystream/atlas/assets/80429191/770ee76d-c70e-447c-965b-c2640d2ee339)

Perhaps it was not included:
In the list of owners on the token page, I see only candidate1 (he started the market), but I also bought tokens on candidate2, 3, 7.
![image](https://github.com/Joystream/atlas/assets/80429191/a19e1ec2-578d-4646-8a9b-e6947fbc7711)

polikosi:
1. I can't see the snackbar after closing the dialog
2. it would be nice if the sorting by the amount of tokens was the same as on the "holders" page
![image](https://github.com/Joystream/atlas/assets/80429191/f089a87e-3dad-4e28-9374-efff1455d5dc)

WRadoslaw:
> 1. I can't see the snackbar after closing the dialog

Which dialog?

ivanturlakov:
> Which dialog?

According to design after user clicks on Continue button on "Congrats on your purchase" modal > the Modal should be closed and a snackbar appears on bottom left

<img width="223" alt="Screenshot 2024-01-30 at 19 32 14" src="https://github.com/Joystream/atlas/assets/46903215/8f1fa5c2-738c-4a12-b7a2-af8249751719">

Holders Table > Members links => Member not found

PR #4946, opened 2023-10-03
🥋 Buy token market tx

WRadoslaw:
#3835

### DIFF
--- a/packages/atlas/src/joystream-lib/extrinsics.ts
+++ b/packages/atlas/src/joystream-lib/extrinsics.ts
@@ -71,6 +71,8 @@ type PublicExtrinsic<TxFunction, ReturnValue> = TxFunction extends (...a: infer 
   ? (...a: [...U, ExtrinsicStatusCallbackFn | undefined]) => Promise<ReturnValue>
   : never
 
+const PERMILLS_PER_PERCENTAGE = 10
+
 export class JoystreamLibExtrinsics {
   readonly api: PolkadotApi
   readonly getAccount: AccountIdAccessor
@@ -1126,7 +1128,7 @@ export class JoystreamLibExtrinsics {
           linearVestingDuration: createType('u32', new BN(initialCreatorAllocation.vestingDuration)),
           cliffAmountPercentage: createType(
             'Permill',
-            new BN(initialCreatorAllocation.cliffAmountPercentage)
+            new BN(initialCreatorAllocation.cliffAmountPercentage * PERMILLS_PER_PERCENTAGE)
           ) as number,
         }),
       })
@@ -1134,8 +1136,8 @@ export class JoystreamLibExtrinsics {
 
     const params = createType('PalletProjectTokenTokenIssuanceParameters', {
       initialAllocation,
-      patronageRate: createType('Perquintill', patronageRate) as number,
-      revenueSplitRate: createType('Permill', revenueSplitRate) as number,
+      patronageRate: createType('Perquintill', patronageRate * PERMILLS_PER_PERCENTAGE) as number,
+      revenueSplitRate: createType('Permill', revenueSplitRate * PERMILLS_PER_PERCENTAGE) as number,
       transferPolicy: createType('PalletProjectTokenTransferPolicyParams', 'Permissionless'),
       metadata: prepareCreatorTokenMetadata({ symbol }),
     })
@@ -1171,7 +1173,10 @@ export class JoystreamLibExtrinsics {
       parseInt(tokenId),
       parseInt(memberId),
       amountCast,
-      createType('Option<ITuple<[Permill, u128]>>', [createType('Permill', new BN(0.5)), amountCast]) // percent, number of joy user wants to pay --- default on 0.5%
+      createType('Option<ITuple<[Permill, u128]>>', [
+        createType('Permill', new BN(0.5 * PERMILLS_PER_PERCENTAGE)),
+        amountCast,
+      ]) // percent, number of joy user wants to pay --- default on 0.5%
     )
   }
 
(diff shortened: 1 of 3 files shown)


packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -1163,4 +1163,47 @@ export class JoystreamLibExtrinsics {
...
   }
+
+  purchaseTokenOnMarketTx = async (tokenId: string, memberId: string, amount: string) => {
+    const amountCast = createType('u128', new BN(amount))
+
+    return this.api.tx.projectToken.buyOnAmm(
+      parseInt(tokenId),
+      parseInt(memberId),
+      amountCast,
+      createType('Option<ITuple<[Permill, u128]>>', [createType('Permill', new BN(0.5)), amountCast]) // percent, number of joy user wants to pay --- default on 0.5%
```
thesan: I think 0.5 percent is 5000 permill (or is it 5 ? I'm never sure if `permill` stands for per 1M or 1K).
Also `new BN(0.5)` is `0` as bn.js doesn't support decimal

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -1163,4 +1163,47 @@ export class JoystreamLibExtrinsics {
...
+    return { block }
+  }
+
+  sellTokenOnMarketTx = async (tokenId: string, memberId: string, amount: string) => {
+    const amountCast = createType('u128', new BN(amount))
+    return this.api.tx.projectToken.sellOnAmm(
+      parseInt(tokenId),
+      parseInt(memberId),
+      amountCast,
+      createType('Option<ITuple<[Permill, u128]>>', [createType('Permill', new BN(0.5)), amountCast]) // percent, number of joy user wants to pay --- default on 0.5%
```
thesan: same here

thesan (changes requested):

packages/atlas/src/joystream-lib/extrinsics.ts
WRadoslaw: Hmmm, in issue token transaction we use permill as a normal percentage value and it was parsed correctly by the Orion, plus In Orion tests Ignazio approached it the same. 

But definitely, BN from 0.5 will be a problem since the default value should be 0.5%, I hope `createType('Permill', 0.5)` will work

packages/atlas/src/joystream-lib/extrinsics.ts
WRadoslaw: answered above

packages/atlas/src/joystream-lib/extrinsics.ts
WRadoslaw: Forget the previous response, I decided to map percentages to permills, it seems the most reasonable

thesan (approved):

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -1126,16 +1128,16 @@ export class JoystreamLibExtrinsics {
...
           ) as number,
         }),
       })
     )
 
     const params = createType('PalletProjectTokenTokenIssuanceParameters', {
       initialAllocation,
-      patronageRate: createType('Perquintill', patronageRate) as number,
-      revenueSplitRate: createType('Permill', revenueSplitRate) as number,
+      patronageRate: createType('Perquintill', patronageRate * PERMILLS_PER_PERCENTAGE) as number,
```
thesan: Oh I  just realized this one is `Perquintill`. $`10^{18}`$ [aparently](https://en.wikipedia.org/wiki/Names_of_large_numbers)!

thesan (changes requested):

packages/atlas/src/joystream-lib/extrinsics.ts
thesan: Also are the `patronageRate` and ` revenueSplitRate` percentages ?

WRadoslaw: From rust crate:
image

packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -72,6 +72,7 @@ type PublicExtrinsic<TxFunction, ReturnValue> = TxFunction extends (...a: infer
   : never
 
 const PERMILLS_PER_PERCENTAGE = 10
+const PERQUINTILLS_PER_PERCENTAGE = 100
```
thesan: `Perquintill` are ["parts per Quintillion"](https://docs.rs/sp-runtime/latest/sp_runtime/struct.Perquintill.html) so I think would be:
```ts
const PERQUINTILLS_PER_PERCENTAGE = 10**16
```
($`10^{18 - 2}`$) but  the `PERQUINTILLS_PER_PERCENTAGE` would have to be a `BN` because $`10^{16}`$ is more than `Number.MAX_SAFE_INTEGER` so if `PERQUINTILLS_PER_PERCENTAGE` is a `number`, `BN` will throw an exception for all ratio that aren't extremely low.

thesan (changes requested):

packages/atlas/src/joystream-lib/extrinsics.ts
thesan: Sorry I just saw that (GH had not refresehd). But I'm really confused now...

thesan: Sorry I just saw that (GH had not refresehd). But I'm really confused now...

thesan (approved):

Issue #4874, opened 2023-09-15
Reward withdrawal/claim: disallow Polkadot formatted addresses


mochet:
Background: https://discord.com/channels/811216481340751934/1147348817692340284/1152219779160096849

This is now I think the 2nd or 3rd time that a user has sent funds to a Polkadot address which is on an unsupported exchange. It seems like it would be prudent to verify addresses have Joystream or at least Substrate generic address formatting to help avoid this.

dmtrjsg:
@mochet thanks for raising this issue. We will include the validation of address format to the next release which allows to withdraw to external address from 
- Payouts dashboard withdrawal
- Member drop down transfer modal

Check for 
- length
- prefix
...

Display validation error on the field input 
"Address does not follow Joystream or Substrate Generic format"

dmtrjsg:
@KubaMikolajczyk 

Artem has added a validation for address format but we don't want to simply disallow and block transfers to polkadot generic accounts and show a message  "Address does not follow Joystream format. Make sure that destination address supports Joystream token before transfer." May I ask you to update the transfer/ withdraw modal designs to accommodate for it.

KubaMikolajczyk:
@dmtrjsg  The text field for entering the address together with users interactions with it are captured in user stories ✅ and ready to be reviewed:

📄 Withdrawal: https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?type=design&node-id=2659-35330&mode=design&t=O4AIFrrWeeRr1nuc-4

📄 Withdraw to external wallet: https://www.figma.com/file/ZFRXsyGvnBXnkDUUGc5vyR/My-payments?type=design&node-id=3321-7796&mode=design&t=O4AIFrrWeeRr1nuc-4

Issue #4795, opened 2023-09-02
Rewards: `withdraw` should support withdawing to external accounts in a single flow


mochet:
# Problem
Currently, the UI for creator rewards uses the word `withdraw` which I think is a bit confusing--some users seem to think that this has sent the tokens to some account that they didn't specify (https://discord.com/channels/811216481340751934/1147348817692340284/1147348817692340284)
![image](https://github.com/Joystream/atlas/assets/50830161/4fde7be0-24ad-4d3a-aa16-8fdeac4e2508)

# Solution
Change `withdraw` to `claim`. This is because `withdraw` is commonly used to signify that a transfer between accounts has happened (typical language used on exchanges) and while this is technically true, from a UX standpoint the user is actually `claiming` rewards that go to a preset account.
Once this is done, then the user would `withdraw`.

bedeho:
I agree this is complex, but I think the real solution here is to allow users to withdraw funds directly to an external account if they want to, this is what most people want to do, like move it to ane xchange or something. So ideally, when they click this, they get a choice between "external accoun" and "my membership account". This will under the hood require a batch transaction because its not possible, on-chain, to withdraw to an arbitrary account.

dmtrjsg:
Just linking up quick solution to the ux issue I have drafted before which is in the design bl: 
- https://github.com/Joystream/atlas/issues/4803

If batching txs is possible, then we'd rather go with that ofc (I recall there was some sort of a limitation that prevented us form doing so before).

dmtrjsg:
Scope:
- [ ] Keep the CTA as `Withdraw` and allow to withdraw to external acc in the same modal
- [ ] Use the same modal that is used for Transfer
- [ ] Change the copy in the destination account from "Polkadot wallet account" to "Joystream account address"
- [ ] do point 3 for the transfer modal when transferring from the membership account

PR #4948, opened 2023-10-03
Withdrawal to any polkadot address

attemka:
#4795 
#4874


packages/atlas/src/providers/transactions/transactions.hooks.ts
```diff
@@ -247,7 +249,7 @@ export const useTransaction = (): HandleTransactionFn => {
         })
 
         /* === transaction was successful, do necessary cleanup === */
-        updateStatus(ExtrinsicStatus.Completed)
+        !onTxSuccess && updateStatus(ExtrinsicStatus.Completed)
```
WRadoslaw, line 252: Why do we want to switch the state only when the transaction has NOT succeeded?

attemka: @WRadoslaw  maybe I missed the naming here but heres how it'll work: onTxSuccess should run the chained transaction, so I've hidden the success state of the current one, as it could be confusing for the user. 2nd transaction (or the last one, if many) won't have onTxSuccess, so it'll display the final success state. Maybe I should change the propertyName, but I've ran out of ideas tbh.

WRadoslaw: I think at this point we should just introduce an API that will handle performing batch transactions

WRadoslaw: Idk, this feels pretty weird for me and unintuitive

dmtrjsg:
@attemka 

- Update placeholder copy to "Joystream wallet address"
- Update tooltip and validation copy to
"All substrate generic accounts are supported, but make sure you are transferring to the wallet address that supports JOYSTREAM token in case of withdrawing to exchange deposit address."

packages/atlas/src/providers/transactions/transactions.hooks.ts
attemka: Agreed to implement custom modal approach in future iterations.

WRadoslaw (approved):

Issue #3859, opened 2023-02-17
Holder Portfolio: Buy from Sale/ Sell and Buy from AMM


dmtrjsg:
## Context
- #3384 

## Scope
- [ ] Controls to start 

AMM: 
Sell 
Buy

Links to the same flow as buying/ selling from the token details page, but redirects to portfolio with conf message after

## Design
[Link1>](https://www.figma.com/file/VvWuJfHYpOPFQrcDVAeHkT/My-portfolio?node-id=1202%3A71665&t=XolD3kk1lOSp2O9T-1)
[Link2>](https://www.figma.com/file/VvWuJfHYpOPFQrcDVAeHkT/My-portfolio?node-id=1232%3A55795&t=XolD3kk1lOSp2O9T-1) 
[Link3>](https://www.figma.com/file/VvWuJfHYpOPFQrcDVAeHkT/My-portfolio?node-id=1232%3A55802&t=XolD3kk1lOSp2O9T-1)

Out of scope ⛔ 
Sale: 
Buy

polikosi:
AMM 
Buy:
I can't buy on portfolio page but I can buy from the token page.
![image](https://github.com/Joystream/atlas/assets/80429191/413bf9de-7286-45ff-9756-d36d21d08453)

Sell: 
Doesn't have a Sell button
![image](https://github.com/Joystream/atlas/assets/80429191/0b7f1501-4aab-4579-9a8c-0fa973521178)

dmtrjsg:
@polikosi can you pls add the list of links with bug descriptions as comments to the master QA tracker issue?

polikosi:
Sell:
No "Sell" button
![image](https://github.com/Joystream/atlas/assets/80429191/05c3f51d-3312-45ac-8d4a-e270b5f4c87d)

PR #4954, opened 2023-10-04
🦭 Sell tokens amm transaction

WRadoslaw:
#3859


packages/atlas/src/joystream-lib/extrinsics.ts
```diff
@@ -1163,4 +1163,47 @@ export class JoystreamLibExtrinsics {
...
   }
+
+  purchaseTokenOnMarketTx = async (tokenId: string, memberId: string, amount: string) => {
+    const amountCast = createType('u128', new BN(amount))
+
+    return this.api.tx.projectToken.buyOnAmm(
+      parseInt(tokenId),
+      parseInt(memberId),
+      amountCast,
+      createType('Option<ITuple<[Permill, u128]>>', [createType('Permill', new BN(0.5)), amountCast]) // percent, number of joy user wants to pay --- default on 0.5%
```
attemka: same thing that @thesan mentioned in #4946

attemka (changes requested):

packages/atlas/src/joystream-lib/extrinsics.ts
WRadoslaw: answered in mentioned PR

WRadoslaw: Well, it will be addressed for both transactions in #4946, so this can be considered resolved

attemka (approved):

PR #4967, opened 2023-10-05
Removed video from YPP page

attemka:


attemka:
duplicates https://github.com/Joystream/atlas/pull/4968

PR #4982, opened 2023-10-08
Fixed YPP dashboard status parsing

attemka:


WRadoslaw:
Is there any issue with it? 

I only saw that there were a few logs about `convertUpperCamelToSentence`, but nothing about this split.

WRadoslaw (approved):

Issue #4937, opened 2023-10-02
Maintenance screen for CRTs


dmtrjsg:
## Context
We are adding an option to disable CRTs from the runtime

## Scope
- [ ] Create a version of the maintenance screen for _only_ CRTs (CRTs module is disabled)

KubaMikolajczyk:
@dmtrjsg I have a simple concept for the maintenance screen with key visuals presenting the idea of maintenance being done on the CRTs module ✅

📄 Maintenance changelog: https://www.figma.com/file/xBXWBZMqM0JAydQZ7ZaZ6D/Maintenance-page?type=design&node-id=916-16424&mode=design&t=uG2AAdGErx7uk2r5-4

dmtrjsg:
If its not too much work - we need to add a way for Atlas to actually check if the pallet is disabled. Manual on-off state is also okay if saves us a lot of time. NB: Builders WG QA team will be testing as part of Nara release qa.

dmtrjsg:
Channel was suspended due to misleading content. Example: https://www.youtube.com/watch?v=ga0cLA6AbYk

dmtrjsg:
Confirmed to invoke it manually from config upd.

dmtrjsg:
@WRadoslaw pls add a manual on how to to enable this for the builders

WRadoslaw:
In `atlas.config.yml` change entry `general.crtMaintenanceMode` to `true` to show maintenance and `false` to disable it

PR #4997, opened 2023-10-10
🧰 CRT Maintenance 

WRadoslaw:
#4937


packages/atlas/src/views/studio/CrtWelcomeView/CrtWelcomeView.tsx
```diff
@@ -1,26 +1,28 @@
-import { CrtView } from '@/views/studio/CrtView'
+import { atlasConfig } from '@/config'
+import { CrtMaintenanceView } from '@/views/studio/CrtMaintenanceView'
 
 export const CrtWelcomeView = () => {
-  return (
-    <CrtView />
```
thesan: Why is this removed ?

WRadoslaw: Mb, PR should target CRT branch. That way we no longer need a coming soon view which is what `crtView` was. Users will either see a landing page to create token or dashboard

Issue #3397, opened 2022-11-01
CRT BE integration - Dashboard


dmtrjsg:
### Context

- https://github.com/Joystream/atlas/issues/3390

### Scope
- [ ] Integration with RPC node to fetch total revenue (revenue splits + direct sales)

Powering total revenue widget on the dashboard

<img width="461" alt="Screenshot 2023-09-11 at 18 11 50" src="https://github.com/Joystream/atlas/assets/93924525/dc1b2b09-6502-4011-b633-3a3da5ddedd2">

PR #5024, opened 2023-10-13
🧑‍🚀 Initial CRT dashboard - orion integration

WRadoslaw:
#3397


packages/atlas/src/components/_crt/RevenueShareParticipationWidget.tsx
```diff
@@ -58,22 +73,25 @@ export const RevenueShareParticipationWidget = () => {
                 ENDED ON
               </Text>
               <Text variant="h400" as="h4">
-                {holdersUnlocked}/{holders}
+                {revenueShare.stakers.length}/69420
```
attemka, line 76: should we get rid of the hardcoded value here(and below in the file)?

packages/atlas/src/components/_crt/RevenueShareStateWidget/RevenueShareStateWidget.tsx
```diff
@@ -1,11 +1,14 @@
...
 import { WidgetTile } from '@/components/WidgetTile'
+import { useBlockTimeEstimation } from '@/hooks/useBlockTimeEstimation'
 import { formatDateTime, formatDurationShort } from '@/utils/time'
 
-export const RevenueShareStateWidget = ({ end }: { end?: Date }) => {
-  const status: 'active' | 'past' | 'inactive' = !end ? 'inactive' : end.getTime() < Date.now() ? 'past' : 'active'
-
+export const RevenueShareStateWidget = ({ endsAtBlock }: { endsAtBlock?: number }) => {
+  const { convertBlockToMsTimestamp } = useBlockTimeEstimation()
+  const status: 'active' | 'past' | 'inactive' = !endsAtBlock ? 'inactive' : endsAtBlock < 0 ? 'past' : 'active'
```
attemka: according to the name of the var `endsAtBlock` it would be the name of the block and the logic for showing the past status should be `endsAtBlock - currentBlock < 0`?

packages/atlas/src/components/_crt/RevenueShareParticipationWidget.tsx
WRadoslaw: I'd leave it as it is and fix everything when we will have access to the transferable balances etc

attemka (approved):

PR #5116, opened 2023-10-23
Fixed incorrect withdrawal commissions calculation

attemka:


packages/atlas/src/components/_overlays/SendTransferDialogs/SendFundsDialog.tsx
```diff
@@ -151,9 +151,9 @@ export const SendFundsDialog: FC<SendFundsDialogProps> = ({
             } tokens have been sent over to ${shortenString(data.account, ADDRESS_CHARACTERS_LIMIT)} wallet address`,
           },
           txFactory: async (updateStatus) => {
-            const amount =
-              !isWithdrawalMode &&
-              amountBN.sub(amountBN.add(transferFee).gte(accountBalance) ? transferFee : new BN(0)).sub(accountDebt)
+            const amount = amountBN
+              .sub(amountBN.add(transferFee).gte(accountBalance) ? transferFee : new BN(0))
+              .sub(accountDebt)
```
thesan: If I understand the problem correctly the amount to send depends on the free balance (or transferable). I we get to this point it means that the total balance has enough funds for `withdrawFromChannelBalance`.

So lets say someone want's to withdraw 500 JOY from there channel, and send it to another account. Let's say the transaction fee is 20 for `withdrawFromChannelBalance` and 10 for `sendFunds`:

- *case 1* the account has a dept of 30 JOY (thus free balance of 0):
   - `withdrawFromChannelBalance`: they pay the 20 fee with their invitation lock so they refund 30 + 20 = 50
 on the lock. Which means their intermediate free balance is 500 - 50 = 450.
   - `sendFunds`: they pay 10 fee out of their 450 free balance so they can only afford to send 450 - 10 = 440. Because the free balance is used in priority over the invitation lock, in all possible cases the 2de transaction fee will always come from the free balance.
- *case 2* the account has no debt but only a 25 JOY free balance:
   - `withdrawFromChannelBalance`: they pay the 20 fee out of their 25 free balance, so they have nothing to refund and their free balance is now 500 + 25 - 20 =  505
   - `sendFunds`: they pay  the 10 fee out of their 505 free balance so they can only afford to send 505 - 10 = 495.
- *case 3* the account has 30 JOY or more free balance: they pay both the 20 JOY and 10 JOY fees out of their initial free balance so they send the full amount.

In case 1 knowing the debt is enough but to differentiate between case 2 and 3 the free balance has to be known. So the calculation would be:
```suggestion
            const cost = accountDebt.add(fullFee)
            const amount = amountBN.sub(freeBalance.gte(cost) ? BN_ZERO : cost.sub(freeBalance))
```

I increased the fees in order to make the calculation clearer. But one catch is that if the fees depend on the amount being send, so it might get recalculated infinitely. The fix would be to make the calculation anticipate the fees by accounting for the chain logic.

attemka: @thesan sorry for my long response here.

So it seems to me that the goal of your change is to make sure that the sum of fees for **both** transactions and account debt should be deducted from the transfer amount, and I fully agree with that.

Now, comparing the solution in this PR and in your suggestion, the only difference I can see that in this PR only the `sendFunds` fee was deducted, which is really weird, assuming, that is was made correctly in the hotfix [PR](https://github.com/Joystream/gleev/commit/7f97f542351759b54b99259b3733e3e7070353db), and this is how Gleev works currently.

I'll do a commit to make this PR look like what I've implemented in Gleev, let me know if I didn't get your point and we need some further changes.

WRadoslaw (approved):

Issue #5124, opened 2023-10-24
Change the carousel for recently paid


dmtrjsg:
Accommodate the changes in Orion:
- https://github.com/Joystream/orion/issues/211

PR #5213, opened 2023-11-02
🐐 Show all time most paid channels in the YPP carrousel

thesan:
- #5124

Depends on:
- https://github.com/Joystream/orion/pull/222


packages/atlas/codegen.config.ts
```diff
@@ -1,8 +1,13 @@
 import { CodegenConfig } from '@graphql-codegen/cli'
+import * as assert from 'assert'
 
 import { customSchemaLoader } from './scripts/customSchemaLoader'
 
-const schemaUrl = 'https://orion.gleev.xyz/graphql'
+const NODE_ENV = process.env.NODE_ENV?.toUpperCase() ?? 'PRODUCTION'
```
attemka: I think this should be either `VITE_ENV` or `VITE_DEFAULT_DATA_ENV`

packages/atlas/scripts/customSchemaLoader.ts
```diff
@@ -17,30 +17,34 @@ async function fetchAuthCookie() {
   return response.headers['set-cookie']
 }
 
-export async function customSchemaLoader() {
-  const authCookie = await fetchAuthCookie()
-  const introspectionQuery = getIntrospectionQuery()
+export function customSchemaLoader(schemaUrl: string, authUrl: string) {
+  return async () => {
+    const authCookie = await fetchAuthCookie(authUrl)
```
attemka: isn't the schema available without the auth cookie now?

attemka (changes requested):

packages/atlas/scripts/customSchemaLoader.ts
thesan: Yeah probably actually I just copied the code I added on the notification branch a while ago

packages/atlas/codegen.config.ts
thesan: Yes `VITE_DEFAULT_DATA_ENV` would make a lot of sense

attemka (approved):

Issue #4990, opened 2023-10-09
Remove password suggestion


attemka:
Remove password suggestion

<img width="299" alt="Screenshot 2023-10-24 at 17 43 42" src="https://github.com/Joystream/atlas/assets/93924525/9372e7ed-c0c9-4b1c-98d6-97ca923f523b">
<img width="376" alt="Screenshot 2023-10-24 at 17 42 36" src="https://github.com/Joystream/atlas/assets/93924525/ae4562e5-30f5-4379-a66c-1401bdc80a37">

PR #5222, opened 2023-11-03
🔑 Remove password suggestion

WRadoslaw:
#4990

### DIFF
--- a/packages/atlas/src/hooks/useHidePasswordInInput.tsx
+++ b/packages/atlas/src/hooks/useHidePasswordInInput.tsx
@@ -3,12 +3,13 @@ import { useState } from 'react'
 import { SvgActionHide, SvgActionShow } from '@/assets/icons'
 import { InputProps } from '@/components/_inputs/Input'
 
-export const useHidePasswordInInput = (): [InputProps, () => void] => {
+export const useHidePasswordInInput = (isNew = true): [InputProps, () => void, boolean] => {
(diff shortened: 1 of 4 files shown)


packages/atlas/src/hooks/useHidePasswordInInput.tsx
```diff
@@ -3,18 +3,20 @@ import { useState } from 'react'
...
 import { InputProps } from '@/components/_inputs/Input'
 
-export const useHidePasswordInInput = (): [InputProps, () => void] => {
+export const useHidePasswordInInput = (isNew = true): [InputProps, () => void, boolean] => {
   const [isPasswordVisible, setIsPasswordVisible] = useState(false)
 
   return [
     {
       type: isPasswordVisible ? 'text' : 'password',
+      autoComplete: isPasswordVisible ? 'off' : isNew ? 'new-password' : 'off',
```
attemka, line 12: the thing is that default password generator (at least in chrome) doesn't pass our password rules, so I think we can just use the 'off' value here

attemka (changes requested):

packages/atlas/src/hooks/useHidePasswordInInput.tsx
attemka: resolved on the call

attemka (approved):

PR #5321, opened 2023-11-12
🦖 Swiper performance

WRadoslaw:
#5328 

In the end, I've bumped the package version and adjusted the logic, hopefully, it will improve the performance.

I was planning to switch whole swiper to `cssMode`, but unfortunately it might be more performant, but it;'s not really reliable with our current implementation (I've experienced either some animation lags or jumping more than one slide on arrow click)


WRadoslaw:
After PR gets approved I will revert changes on config and env

attemka (approved):

Issue #3726, opened 2023-02-03
(2nd round Test) Creator Tokens — Owner & Issuer


KubaMikolajczyk:
This includes conducting moderated usability tests on prototypes from "[My token](https://www.figma.com/file/OuywwClsegx2eotkii4ea4/My-token?node-id=5%3A2990&t=ZRHEfgDiUXjFRLqI-1)" file (Owner & Issuer experience). The tests should cover happy path flows for:

Starting a market vs starting a sale
Creating a revenue share as a creator
The expected outcome is a brief summary of key findings, a list of recommendations, and a list of test participants to be paid out.

Depending on the results, the next step is either proceeding finalizing the designs as it is or performing necessary changes on the hi fi designs.

KubaMikolajczyk:
# Overview
This is the final round of moderated tests which primary goal is to find out if users understand everything and are able to perform such actions as:
1. Start a market
2. Start a sale
3. Start a revenue share

# Wallets to pay
(We agreed to pay $40 worth of BCH)
- qzv9yhjnawqrmpgdcufca40nw90e3zhckuswsdu0xc
- qpqg2qwnnmwhlxhnjwtdll7acw2gwfv9cgc6jl8u23
- 12q7HrVEd2i6MkYeaUAZ1EnD36wpfUSMZ9
- qrlcs3ycwlja9vkvztclxrg5tk6rs4vllue7g0clnj

# Recordings links
Session 1: https://drive.google.com/file/d/13TDVQsBb1MUZ8W6bA7-bTbgAfUI4EDyL/view?usp=sharing
Session 2: https://drive.google.com/file/d/17x3U0Evnvxjvcs_vK6zDMSQygPYcaHBx/view?usp=sharing
Session 3:https://drive.google.com/file/d/1sR2QRWusX0PZFK0XeBxsUE2ve741iUP2/view?usp=sharing
Session 4: https://drive.google.com/file/d/1Hn5Lz33wypA4wCd5d33lRGI-I1Vnj20r/view?usp=sharing

# Key findings
### 🟢 No major issue was found. All participants were able to complete all tasks successfully.

- They knew the difference between market and sale and based on a set of questions for eg.: "You want to sell your tokens for a special price but only for a week. Would you choose sale or market?", were able to select the correct event type to perform that action
- They knew how the market works and how the sale works and what charts featured in those flows represent
- They knew what is vesting and cliff and could tell me how this would work
[...]

PR #5412, opened 2023-11-20
🥈 CRT follow up v2

WRadoslaw:
#3726

### DIFF
--- a/packages/atlas/src/components/_crt/CloseMarketModal/CloseMarketModal.tsx
+++ b/packages/atlas/src/components/_crt/CloseMarketModal/CloseMarketModal.tsx
@@ -1,3 +1,6 @@
+import { useApolloClient } from '@apollo/client'
+import { useCallback } from 'react'
+
 import { SvgActionPlay, SvgAlertsWarning24 } from '@/assets/icons'
 import { Banner } from '@/components/Banner'
 import { FlexBox } from '@/components/FlexBox'
@@ -5,6 +8,10 @@ import { NumberFormat } from '@/components/NumberFormat'
 import { Text } from '@/components/Text'
 import { TextButton } from '@/components/_buttons/Button'
 import { DialogModal } from '@/components/_overlays/DialogModal'
+import { useJoystream } from '@/providers/joystream'
+import { useSnackbar } from '@/providers/snackbars'
+import { useTransaction } from '@/providers/transactions/transactions.hooks'
+import { useUser } from '@/providers/user/user.hooks'
 import { cVar } from '@/styles'
 
 const getTokenDetails = (id: string) => ({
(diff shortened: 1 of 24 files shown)


packages/atlas/src/components/_crt/CloseMarketModal/CloseMarketModal.tsx
```diff
@@ -15,26 +22,58 @@ const getTokenDetails = (id: string) => ({
...
+  const { displaySnackbar } = useSnackbar()
+  const { memberId } = useUser()
+  const handleTransaction = useTransaction()
+  const client = useApolloClient()
 
   const hasInsufficientTokens = userCrtToken < tokenRequiredToSell
 
+  const handleCloseAmm = useCallback(async () => {
+    if (!joystream || !channelId || !memberId) {
+      return
```
attemka, line 42: probably we should log in sentry that something went wrong here and user was unable to close the amm

packages/atlas/src/components/_crt/CloseMarketModal/CloseMarketModal.tsx
```diff
@@ -15,26 +22,58 @@ const getTokenDetails = (id: string) => ({
...
+      onTxSync: async () => {
+        client.refetchQueries({ include: 'active' }).then(() => {
+          displaySnackbar({
+            title: 'Market closed successfuly',
+          })
+          onClose()
+        })
+      },
+      onError: () => {
+        displaySnackbar({
```
attemka, line 56: same here, would be good to log an error here

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenSummaryStep.tsx
```diff
@@ -38,7 +38,19 @@ export const TokenSummaryStep = ({ setPrimaryButtonProps, form, onSuccess }: Tok
...
+  const { fullFee } = useFee('issueCreatorTokenTx', [
+    memberId ?? '1',
+    channelId ?? '1',
+    form.name,
+    form.creatorReward,
+    form.revenueShare,
+    {
+      amount: String(form.creatorIssueAmount ?? 0),
+      cliffAmountPercentage: 50,
+      blocksBeforeCliff: 123123,
```
attemka: why should this be hardcoded?

attemka (changes requested):

packages/atlas/src/components/_crt/CloseMarketModal/CloseMarketModal.tsx
attemka: will be implemented in the separate PR

packages/atlas/src/components/_crt/CreateTokenDrawer/steps/TokenSummaryStep.tsx
attemka: discussed on the call

Issue #5338, opened 2023-11-14
Track Referrals page hits with Segment


dmtrjsg:
ssia

Issue #5262, opened 2023-11-07
Upload issue reported by Codefi_keyz


dmtrjsg:
[Discord message>](https://discord.com/channels/811216481340751934/1035131444080680971/1171432683709141042)

<img width="910" alt="Screenshot 2023-11-08 at 13 19 02" src="https://github.com/Joystream/atlas/assets/93924525/1ab7560b-a70c-4f06-ad7c-e3b28d5bcceb">

WRadoslaw:
After investigating the issue, the only problem that I noticed on the Atlas is that we blacklist and try the next operator (out of max 3) even if the connection problem was on the users' side. 

The only thing that I can propose on the Atlas is a check for users' connection after the initial upload request fails, this information can be handled in a few ways:
1. We can introduce an interval that will check if the connection is restored and try to upload the video (from the beginning) without blacklisting the operator.
2. We can abort and just inform the user that because of his connection upload failed, and the one will need to retry an upload after regaining access to the internet.
3. (REQUIRES INFRA CHANGE) Theoretically if the problem is on the user side, we could save the offset on which the upload stopped and when the connection is back we could resume uploading the file.

bedeho:
I think we need to get away from this client side model entirely, it's an artefact of the first model of Atlas where there was no backend. I think we need to bite the bullet and have either Atlas just pick a random operator which is supposed to be suitable, and then retry if it fails, and errors be reported. The fundamental operational side of this is that infrastructure has to be responsie to errors and self-correct, which is not for Orion to explicitly deal with. A v2 version of this is to have Orion be slightly more helpful, and have it suggest operators based on some local state about what is failing or not, but even this seems like a bad path really. What I'm suggesting will possibly lead to more problems afflicting users, but this will force us, the logging and the leads to mature enough to handle this properly.

dmtrjsg:
The solution we discussed further was to introduce some sort of the state for operators kept in Orion, which can persist across user sessions (because right now new session would start uploading to same buckets). 

However this would not solve the issue raised above without changes to the infra which allows to _continue_ uploading the asset when client upload fails. 

After another round of discussions the suggestion was to first introducing the means to distinguish between client and operator failure. In this case we can inform the user and the logging system correctly. 

User perspective:
- When experiencing network issues preventing successful upload - show error message to the user and trying to upload again (from scratch) -

Issues with current logging:
- log failure on each bucket failing (switching and successful upload remains as error)
- log errors teh same way when it comes from client or infra side

[...]

PR #5428, opened 2023-11-21
Updated system logs

attemka:
#5338 
#5262

### DIFF
--- a/packages/atlas/src/providers/uploads/uploads.hooks.ts
+++ b/packages/atlas/src/providers/uploads/uploads.hooks.ts
@@ -173,17 +173,20 @@ export const useStartFileUpload = () => {
         assetsNotificationsCount.current.uploaded[assetKey] =
           (assetsNotificationsCount.current.uploaded[assetKey] || 0) + 1
       } catch (e) {
-        SentryLogger.error('Failed to upload asset', 'uploadsHooks', e, {
-          asset: { dataObjectId: asset.id, uploadOperator },
-        })
         setAssetStatus({ lastStatus: 'error', progress: 0 })
 
         const axiosError = e as AxiosError
-        const networkFailure =
+        const networkFailure = axiosError.isAxiosError && !axiosError.response?.status
+        const providerFailure =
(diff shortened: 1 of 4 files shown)


packages/atlas/src/providers/uploads/uploads.hooks.ts
```diff
@@ -173,17 +173,20 @@ export const useStartFileUpload = () => {
...
-        const networkFailure =
+        const networkFailure = axiosError.isAxiosError && !axiosError.response?.status
+        const providerFailure =
           axiosError.isAxiosError &&
-          (!axiosError.response?.status || (axiosError.response.status >= 400 && axiosError.response.status <= 500))
-        UserEventsLogger.logDistributorError({ dataObjectId: asset.id, distributorId: uploadOperator.id }, e)
-        if (networkFailure) {
+          axiosError.response &&
+          axiosError.response.status >= 400 &&
+          axiosError.response.status <= 500
```
WRadoslaw, line 184: I'm not sure whether this check will be sufficient to determine the side on which the flow failed. By this rule all of the currently reported issues on Sentry will be blamed on the operator, and we still cannot say whether this is true. 

Can we just reuse logic from `ConnectionStatusManager` and when the upload fails, try to benchmark the connection status and apply logs based on this? 

If I remember correctly user needs to drop a connection for at least 9 seconds before upload fails, so I don't think that there will be a case that the user fails the upload, but reconnect before benchmark.

WRadoslaw (changes requested):

packages/atlas/src/providers/uploads/uploads.hooks.ts
attemka: @WRadoslaw receiving actual response code means that the connection to the operator was alive when error occurred. If you'll terminate the connection in the middle of request, axios with throw the `NetworkError` without the response from the server, which will be handled separately.
I'll add the rule to send the user errors only if the error occurred on the user side

WRadoslaw (approved):

PR #5596, opened 2023-12-06
🥉  CRT QA v3

WRadoslaw:


packages/atlas/src/components/_crt/HoldersWidget/HoldersWidget.tsx
```diff
@@ -20,7 +20,7 @@ export const HoldersWidget = ({ holders: _holders, ownerId, totalSupply }: Holde
   const [showModal, setShowModal] = useState(false)
 
   const holders = _holders.map((holder) => ({
-    memberId: ownerId,
+    memberId: holder.member.id,
```
attemka: Unlikely to happen, but still
```suggestion
    memberId: holder?.member?.id || '',
```

packages/atlas/src/views/studio/CrtDashboard/CrtDashboard.tsx
```diff
@@ -25,7 +25,7 @@ export const CrtDashboard = () => {
   const { activeChannel } = useUser()
   const { data } = useGetFullCreatorTokenQuery({
     variables: {
-      id: activeChannel?.creatorToken?.token.id ?? '',
+      id: '18', //activeChannel?.creatorToken?.token.id ?? '',
```
attemka: I think the original code should be restored?

attemka (changes requested):

attemka (approved):

Issue #5571, opened 2023-12-01
"AccountNotSelectedError" error


polikosi:
1. refreshed the home page then I get this error
![image](https://github.com/Joystream/atlas/assets/80429191/9ebbe288-4aae-4a37-939e-273604da7d2d)
2. Go to marketplace https://atlas-o35urjwe5-joystream.vercel.app/marketplace
3. Go to video and place bid 20 joy https://atlas-o35urjwe5-joystream.vercel.app/video/2
4. get this error 
![image](https://github.com/Joystream/atlas/assets/80429191/5513ac03-694b-4577-830c-e05064105916)

PR #5597, opened 2023-12-06
🗡️ Ensure crypto ready before decoding seed

WRadoslaw:
#5571


packages/atlas/src/joystream-lib/lib.ts
```diff
@@ -80,7 +80,7 @@ export class JoystreamLib {
       SentryLogger.error('Missing signer for setActiveAccount', 'JoystreamLib')
       return
     }
-
+    await cryptoWaitReady()
```
attemka, line 83: shouldn't the same line be added to the other usages of this function?

WRadoslaw: What do you mean?

Issue #5490, opened 2023-11-24
CRT Tooltips


dmtrjsg:
### Create Token
<img width="446" alt="Screenshot 2023-12-05 at 15 56 02" src="https://github.com/Joystream/atlas/assets/93924525/b52f071f-954f-40b1-831b-f725b2a21007">

Token assurance

`Different options presented below correspond to different level of token safety projected to your potential buyers. For example choosing risky option does not signal to your buyers that they are safe from the pump and dump scenario, while secure and safe options signal a safer token purchasing option.`

<img width="316" alt="Screenshot 2023-12-05 at 15 55 48" src="https://github.com/Joystream/atlas/assets/93924525/5e00eca4-82da-488a-b7bb-b4e4efaa3586">

Name
`Short name of your token, how it will be displayed in portfolio and marketplace.`
Access
`Token can be accessed and purchased by any member, or strictly limited to members added to your token whitelist.`
Revenue share with holders
`The % ratio of how much channel revenue gets distributed among all token holders vs the % ratio of what get's transferred to channel owners (yours) account when the revenue share is initiated."
Annual creator reward
`Annual creator reward aka patronage, that is earned by channel owner for managing channel.`
Tokens issued in your wallet
`Total number of tokens issued`
Cliff
`Period of time during which no tokens can be issued. After the cliff has passed, initial allocation happens.`
Vesting period
`Period of time after cliff during which all your tokens become vested, meaning that they can be spent or transferred.`
First payot
`Amount of tokens that become transferable, after the clif and before the vesting period starts.`

### Token Dashboard
<img width="291" alt="Screenshot 2023-12-05 at 15 54 59" src="https://github.com/Joystream/atlas/assets/93924525/4cd99a59-3ac2-4238-9538-ae08b869face">
<img width="340" alt="Screenshot 2023-12-05 at 15 54 51" src="https://github.com/Joystream/atlas/assets/93924525/3a2d98aa-6451-431c-b492-0c4950d99f14">
<img width="324" alt="Screenshot 2023-12-05 at 15 54 55" src="https://github.com/Joystream/atlas/assets/93924525/95695a65-1bfc-4cf5-8958-a4f2e190bbdb">

### Token Page
<img width="519" alt="Screenshot 2023-12-05 at 15 57 42" src="https://github.com/Joystream/atlas/assets/93924525/b59eeeb1-bc2e-4173-91b7-fd1aefe3b781">

Total Revenue
[...]

PR #5635, opened 2023-12-11
☁️ New tooltips for CRT

WRadoslaw:
#5490


packages/atlas/src/components/CrtPreviewLayout/CrtPreviewLayout.tsx
```diff
@@ -44,8 +44,7 @@ export const getTokenDetails = (token: FullCreatorTokenFragment) => {
       caption: 'TOTAL REV.',
       content: +token.totalSupply,
       icon: <JoyTokenIcon size={16} variant="silver" />,
-      tooltipText:
-        'This is the total amount of revenue that the channel made from the time it was created. This information can be useful to determine if the creator had high or low income from their content in the past.',
+      tooltipText: 'Total revenue this channel made from DAO earnings, NFT sales and Royalties.',
```
bedeho: royalites is lower case here.

bedeho (changes requested):
Nice job, I will have second look in prod when I can see tooltips in context of UI, but this is low risk issue, I doubt there are any grave mistakes here.

PR #5799, opened 2024-01-26
🌾 POC: Recommendations

WRadoslaw:


WRadoslaw:
## POC testing

This playground is based on public production data. This means that you can use your already created memberships (ONLY EXTERNAL WALLETS), but it will ask you to add a new email, doesn't matter what email you provide, it won't be used anywhere and it won't override your email in Gleev. It is advised not to use studio, YPP and marketplace as they don't impact the system in any way.

The main areas of testing would be the homepage feed (whether it adjusts to the videos that you watched), and the video page (whether the videos proposed on the right side are similar to the currently playing video). 

It would be nice to have dev tools opened during the testing on the network tap and check if as you click through the application the interactions are sent properly. We limit them both per user and video ID and by IP. So if some of the interaction requests will start to fail and you feel that you didn't do a sprint through the UI make a note and share it with us. 

As we don't have many interactions yet and the video count is high, we expect the initial recommendations to be low-quality or irrelevant, but it should get better after a day or two. 

---
Below is an example screenshot that shows how successful video video-clicked request looks

image

---
**In case of breaking issues, that are stopping you from getting recommendations add a comment with a context and a way to reproduce the issue.**

Lastest playground url: https://atlas-ebeoruuu8-joystream.vercel.app/

⚠️ Rember to set the environment to Gleev next in the Admin modal (it can be accessed through the sidebar)
image

kdembler:
Really cool to see this happening!

So what I see right off the bat:
1. Recommendations in video view are always the same
CleanShot 2024-01-26 at 17 37 58@2x

2. After watching one video from a channel, my full home feed are videos only from this channel
CleanShot 2024-01-26 at 17 38 15@2x

WRadoslaw:
> So what I see right off the bat:
> 
> 1. Recommendations in video view are always the same

I've made some changes. It should work better now.

> 2. After watching one video from a channel, my full home feed are videos only from this channel

This is mainly due to the lack of interactions, we need lots more of them. I had similar issue, but it seems to be better today.

PR #5867, opened 2024-02-08
WalletConnect integration

attemka:


package.json
```diff
@@ -82,5 +82,9 @@
   "engines": {
     "node": "^18"
   },
-  "packageManager": "yarn@3.2.1"
+  "packageManager": "yarn@3.2.1",
+  "dependencies": {
+    "@walletconnect/modal": "^2.6.2",
+    "@walletconnect/universal-provider": "^2.11.1"
+  }
```
kdembler: Those should probably be added to atlas package, not monorepo root?

kdembler (commented):
1. We need to figure out the approach to email vs wallet flows. My suggestion is that we keep current Atlas behavior - we show wallet selection only if user has some local wallet (excluding WalletConnect), otherwise show web2 sign in by default. We could add a small link that would open the wallet selection instead, letting users use WalletConnect. I suggest something as simple as below and once we're closer to finalization I will ask for design input:
CleanShot 2024-02-21 at 14 01 04@2x
2. When browsing on mobile:
    a) Let's use the same behaviour as above (so priortize web2 login)
    b) I'm not able to use WalletConnect, when I try to use that option nothing happens.
3. We need some way to destroy WalletConnect session. Right now if I connect a wallet through WalletConnect, I'm not able to "disconnect". Even if I sign out in Atlas and go back to wallet selection, if I select WalletConnect again, my previous session is used, so I don't have an option to connect to a different wallet with WalletConnect without clearing local storage.

packages/atlas/src/providers/wallet/tmpwallet/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './wallet-connect.js'
+export * from './types.js'
+export * from './consts.js'
```
thesan, line 3: Why are these `.js` files ?

packages/atlas/src/providers/wallet/tmpwallet/signer.ts
```diff
@@ -0,0 +1,60 @@
...
+        id: 1,
+        jsonrpc: '2.0',
+        method: 'polkadot_signTransaction',
+        params: { address: payload.address, transactionPayload: payload },
+      },
+    }
+
+    const { signature } = await this.client.request<Signature>(request)
+
+    return { id: ++this.id, signature }
```
thesan, line 41: I'm asking this for the Pioneer implementation but why does the id has to be defined manually ?

Right now in Pioneer the signer just returns the `this.client.request(...` and it seems to work (also I don't define `request.id`, I just followed [the example from the doc](https://docs.walletconnect.com/advanced/multichain/polkadot/dapp-integration-guide#polkadot-example))

thesan (commented):
When I use the Polkadot wallet (with my mainnet controller account) it says that I have "No memberships connected":
image
But it happens on Gleev.xyz too so maybe it's on my side :shrug: 

In any case I agree there should be a way to disconnect a wallet without flushing the localStorage or deleting it through the dev tools.

packages/atlas/src/providers/wallet/tmpwallet/index.ts
attemka: these are used as these files are copied from the actual lib I'm currently working on (paused atm so I'm not sure if I need to finalise it soon). Therefore, I left them untouched

packages/atlas/src/providers/wallet/tmpwallet/signer.ts
attemka: this file has been taken from the WalletConnect implementation an signer is currently left unchanged. I'd say this is used to provide a correct nonce for signed transactions.

packages/atlas/src/providers/wallet/wallet.provider.tsx
```diff
@@ -42,6 +43,41 @@ export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
...
+      SentryLogger.error(
+        'WalletConnect Config error',
+        'wallet.provider.tsx',
+        'No genesis hash found in chain metadata. WalletConnect configuration is incorrect'
+      )
+    }
+    const chainId = `polkadot:${genesisHash?.replace('0x', '').substring(0, 32)}` as `polkadot:${string}`
+
+    const walletConnectParams: WalletConnectConfiguration = {
+      projectId: '33b2609463e399daee8c51726546c8dd',
```
kdembler: Let's pass this through some env variable

packages/atlas/src/providers/wallet/tmpwallet/consts.ts
```diff
@@ -0,0 +1,6 @@
+export const POLKADOT_CHAIN_ID = 'polkadot:91b171bb158e2d3848fa23a9f1c25182'
```
kdembler: I think you have added genesis hash fetch so those consts are not needed anymore I think?

kdembler (commented):
1. If I select WalletConnect option but close the WC modal, I still get "No memberships connected", I think in that case it should go back to wallet selection?

https://github.com/Joystream/atlas/assets/12646744/81926bf4-608e-4f1b-bcaf-4ee3fb126c53

2. If I scan the QR but reject the connection in my wallet, the UI is stuck in "Connecting" state

https://github.com/Joystream/atlas/assets/12646744/544284eb-4b6d-4df1-866d-37f75b4e537d

packages/atlas/src/providers/wallet/tmpwallet/consts.ts
attemka: removed this one, however I've left the default joystream chain Id just in case we'll have some issues with fetching the hash from the RPC (e.g bad connection etc)

packages/atlas/src/components/_auth/ExternalSignInModal/ExternalSignInSteps/ExternalSignInModalWalletStep/ExternalSignInModalWalletStep.tsx
```diff
@@ -87,14 +96,27 @@ export const ExternalSignInModalWalletStep: FC<ExternalSignInModalWalletStepProp
   const selectedWallet = (selectedWalletIdx != null && wallets[selectedWalletIdx]) || null
 
   const handleConfirm = useCallback(async () => {
-    if (!selectedWallet) return
+    if (selectedWalletIdx < wallets.length && !selectedWallet) return
```
kdembler: Why is that new check needed?

packages/atlas/src/components/_auth/ExternalSignInModal/ExternalSignInSteps/ExternalSignInModalWalletStep/ExternalSignInModalWalletStep.tsx
```diff
@@ -146,7 +178,7 @@ export const ExternalSignInModalWalletStep: FC<ExternalSignInModalWalletStepProp
       to: selectedWallet?.installed ? undefined : selectedWallet?.installUrl,
       onClick: selectedWallet?.installed ? handleConfirm : undefined,
     })
-  }, [handleConfirm, isConnecting, selectedWallet, setPrimaryButtonProps])
+  }, [handleConfirm, isConnecting, selectedWallet, selectedWalletIdx, setPrimaryButtonProps, wallets.length])
```
kdembler: Why is that needed if we haven't modified this useEffect?

packages/atlas/src/providers/wallet/wallet.helpers.ts
```diff
@@ -14,7 +16,23 @@ export const getWalletsList = () => {
     return [...acc, new UnknownWallet(walletName)]
   }, [] as UnknownWallet[])
 
-  return [...supportedWallets, ...unknownWallets]
+  const wcWallet = {
+    extensionName: 'WalletConnect',
+    // signer: wcWallet.signer,
+    title: 'WalletConnect',
+    // getAccounts: wcWallet.getAccounts,
```
kdembler: Do we need those commented lines?

packages/atlas/src/providers/wallet/wallet.provider.tsx
```diff
@@ -42,6 +43,41 @@ export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
...
+      )
+    }
+    const chainId = `polkadot:${genesisHash?.replace('0x', '').substring(0, 32)}` as `polkadot:${string}`
+
+    const walletConnectParams: WalletConnectConfiguration = {
+      projectId: atlasConfig.features.members.walletConnectProjectId || '',
+      relayUrl: 'wss://relay.walletconnect.com',
+      metadata: {
+        name: 'Joystream demo app',
+        description: 'Joystream Wallet-Connect',
```
kdembler: This needs to be configurable

packages/atlas/src/providers/wallet/wallet.provider.tsx
```diff
@@ -84,6 +120,36 @@ export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
     },
     [setWallet, setWalletAccounts, setWalletStatus]
   )
+  const signInWithWalletConnect = useCallback(async () => {
+    const walletConnectParams = await getWalletConnectParams()
+    const wcWallet = new WalletConnectWallet(walletConnectParams, atlasConfig.general.appName)
+    await wcWallet.connect()
+    const accounts = await wcWallet.getAccounts()
+    const accountsWithWallet = accounts
+      // .filter(filterUnsupportedAccounts)
```
kdembler: Is this needed?

packages/atlas/src/providers/wallet/wallet.provider.tsx
```diff
@@ -42,6 +43,41 @@ export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
...
+    const chainId = `polkadot:${genesisHash?.replace('0x', '').substring(0, 32)}` as `polkadot:${string}`
+
+    const walletConnectParams: WalletConnectConfiguration = {
+      projectId: atlasConfig.features.members.walletConnectProjectId || '',
+      relayUrl: 'wss://relay.walletconnect.com',
+      metadata: {
+        name: 'Joystream demo app',
+        description: 'Joystream Wallet-Connect',
+        url: '#',
+        icons: ['https://walletconnect.com/static/favicon.png'],
```
kdembler: Shouldn't this be Atlas/Gleev logo? My understanding is that this is passed to the wallet?

packages/atlas/src/providers/wallet/wallet.provider.tsx
```diff
@@ -84,6 +120,36 @@ export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
...
+          wallet: wcWallet,
+          signer: wcWallet.signer,
+        }
+      })
+
+    setWalletAccounts(accountsWithWallet)
+    const storeWallet = {
+      extensionName: 'WalletConnect',
+      logo: {
+        src: 'https://walletconnect.com/static/favicon.png',
```
kdembler, line 152: If we're linking to external logo, why do we need the new `WC-logo.svg`?

packages/atlas/src/components/_auth/ExternalSignInModal/ExternalSignInSteps/ExternalSignInModalWalletStep/ExternalSignInModalWalletStep.tsx
```diff
@@ -87,14 +96,27 @@ export const ExternalSignInModalWalletStep: FC<ExternalSignInModalWalletStepProp
...
   const handleConfirm = useCallback(async () => {
-    if (!selectedWallet) return
+    if (selectedWalletIdx < wallets.length && !selectedWallet) return
 
     setIsConnecting(true)
     setHasError(false)
-    const accounts = await signInToWallet(selectedWallet.extensionName)
+
+    const accounts =
+      selectedWallet?.extensionName === 'WalletConnect'
```
kdembler: We're doing this check with hardcoded name in a couple of places. I think we can use the utility `isWalletConnectWallet` you've created instead?

kdembler (dismissed):

packages/atlas/atlas.config.yml
```diff
@@ -269,6 +269,13 @@ features:
   members:
     avatarServiceUrl: '$VITE_AVATAR_SERVICE_URL' # URL for avatar service - used to upload member's avatar
     hcaptchaSiteKey: '$VITE_HCAPTCHA_SITE_KEY' # Site key for hCaptcha - used to verify users are not bots when creating memberships - depends on hCaptcha being enabled in the faucet
+  walletConnect:
+    walletConnectProjectId: '$VITE_WALLET_CONNECT_PROJECT_ID' # WalletConnect project ID - used to connect to WalletConnect
+    metadata:
+      name: Atlas
+      description: Web3 video streaming platform
+      url: \#
```
kdembler: I think this should be full URL?
IMG_C15FE4F79AC4-1

kdembler: We could just populate it dynamically with the app URL?

packages/atlas/src/providers/wallet/wallet.provider.tsx
```diff
@@ -84,6 +128,36 @@ export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
     },
     [setWallet, setWalletAccounts, setWalletStatus]
   )
+  const signInWithWalletConnect = useCallback(async () => {
+    const walletConnectParams = await getWalletConnectParams()
+    const wcWallet = new WalletConnectWallet(walletConnectParams, atlasConfig.general.appName)
```
WRadoslaw, line 133: Should we even try to initialize WC wallet if there might not be a project ID?

`projectId: walletConnectConfig.walletConnectProjectId || '',`

WRadoslaw (commented):
Should I be able to connect with WC on dev env? 

I installed Zerion wallet which I think only supports ETH. When scanning there is no info on what failed and why. I believe we should cover possibility of a user trying to use wallet without Polkadot support. 

WC modal allows to specify which wallets are displayed by listing supported wallet IDs

packages/atlas/src/providers/wallet/wallet.provider.tsx
attemka: @WRadoslaw yes, we shouldn't, in fact there won't even be an option of WC in the list if you haven't provided the projectId, so this function won't be triggered.

attemka:
@WRadoslaw, I think this wouldn't be possible, as the dev chain is not currently included in the Zerion wallet, so it'll be unknown. There was some discussions on sending a request of adding the dev chain to the subwallet testnet list, yet AFAIK no actions were made there yet.

attemka:
@WRadoslaw update from yesterday's call: for now, adding our testnet to subwallet will be postponed. However, you can still add it manually in the app in order to test the PR on dev.

WRadoslaw (approved):

kdembler (approved):

PR #5910, opened 2024-02-21
🐈‍⬛ Blacklisting logic change

WRadoslaw:


packages/atlas/src/providers/uploads/uploads.hooks.ts
```diff
@@ -176,17 +176,12 @@ export const useStartFileUpload = () => {
...
           axiosError.isAxiosError &&
-          axiosError.response &&
-          axiosError.response.status >= 400 &&
-          axiosError.response.status <= 500
-        if (providerFailure) {
-          SentryLogger.error('Failed to upload asset: provider failure', 'uploadsHooks', e, {
-            asset: { dataObjectId: asset.id, uploadOperator },
-          })
-          UserEventsLogger.logDistributorError({ dataObjectId: asset.id, distributorId: uploadOperator.id }, e)
+          (!axiosError.response?.status || (axiosError.response.status >= 400 && axiosError.response.status <= 500))
```
WRadoslaw: `(!axiosError.response?.status || (axiosError.response.status >= 400 && axiosError.response.status <= 500))`

@kdembler Should we consider codes >500 as a network failure as well and change the provider?

packages/atlas/src/providers/assets/assets.provider.tsx
```diff
@@ -239,9 +249,11 @@ export const useStorageOperators = () => {
         if (includeFailed) {
           return bagOperators
         }
+
         const workingBagOperators = bagOperators.filter(
-          (operator) => !failedStorageOperatorIds.current.includes(operator.id)
+          (operator) => !evaluateStorageOperators().some(([id]) => operator.id === id)
```
kdembler: You call `evaluateStorageOperators` inside `filter` callback, I think it's good enough to do it once?

packages/atlas/src/providers/uploads/uploads.hooks.ts
kdembler: @WRadoslaw As I look at this I don't think we're really able to differentiate between errors caused by a user and those caused by the operator. Do you think we need this `if (networkFailure) {` at all?

packages/atlas/src/providers/assets/assets.provider.tsx
```diff
@@ -51,7 +51,7 @@ OperatorsContext.displayName = 'OperatorsContext'
 export const OperatorsContextProvider: FC<PropsWithChildren> = ({ children }) => {
   const storageOperatorsMappingPromiseRef = useRef<Promise<BagOperatorsMapping>>()
   const [storageOperatorsError, setStorageOperatorsError] = useState<unknown>(null)
-  const failedStorageOperatorIds = useRef<string[]>([])
+  const failedStorageOperatorIds = useRef<[string, number][]>([]) // ID, timestamp of failure
```
kdembler, line 54: This isn't persisted between sessions right?

WRadoslaw: It is not

packages/atlas/src/providers/uploads/uploads.hooks.ts
WRadoslaw: @kdembler I'd say that 500+ responses should definitely trigger blacklisting.

I asked GPT for a snippet for possible handling
```
const makeRequest = async () => {
    try {
        const response = await axios.get('https://example.com/data');
        console.log(response.data);
    } catch (error) {
        if (axios.isAxiosError(error)) {
            const axiosError = error as AxiosError;
            if (axiosError.response) {
                if (axiosError.response.status >= 500) {
                    // Here we can ban storage operator and retry upload
                    console.error('Server error (500+):', axiosError.response.status);
                } else {
                    // Under 500, we cannot distinguish between client and server error, but we need to agree
                    // whether we should try next operator or retry to the same
                    console.error('Unexpected status code:', axiosError.response.status);
                }
            } else if (axiosError.request) { 
                // The request was made but no response was received
                // Generated by GPT - not sure what this would imply - I'd just remove this
                // Theoretically, if there is no response it means that req was rejected or there is a network error
            } else if (axiosError.code === 'ERR_NETWORK') {
                // I'd say here, we should ban operator as well
                console.error('Network error, possibly a CORS issue:', axiosError.message);
            } else {
                // Generated by GPT - not sure what this would imply - I'd just remove this
                console.error('Error setting up the request:', axiosError.message);
            }
        } else {
            // Non-Axios error
            console.error('Non-Axios error:', error);
        }
    }
};
```

kdembler: Yeah, my point is that we may want to consider _any_ error as a basis for blacklisting. Is there something against that?

Issue #6001, opened 2024-03-14
CRT Analytics events


dmtrjsg:
## Context
Product Analytics for CRTs

## Scope

Implement Page views and track events for various parts of the flow.

### View Page Events

Viewed
1️⃣  Token dashboard
2️⃣  Token Market tab
3️⃣  Token Revenue split tab
4️⃣  Channel Token Tab

### Track Events

1️⃣ 
Event = creator clicks on "create token" on Creator landing page
Title =`Token minting flow started`
Attributes = `ChannelId`

2️⃣ 

After token is minted
Title = `Token minted`
Attributes = `Ticker, Initial supply, Safety option (safe, custom, risky)`

3️⃣ 

AMM started 
Title = `Token Market Opened` 
Attribute = `Token ticker, channelId`

AMM closed
Title = `Token Market Closed`
Attribute = `Token ticker, channel`

4️⃣  
Tokens sold on AMM
Titile = `Tokens Market Sale`
Attribute = `Token ticker, channelId, CRT amount, JOY received`

Tokens purchased on AMM
Titile = `Tokens Market Purchase` 
Attribute = `Token ticker, channelId, CRT amount, JOY paid`

dmtrjsg:
QA on dev

@attemka excellent work sir ⭐  a few follow up points that were not included into original scope but popped up as I was testing


# Page Events

1️⃣  🟡  crt landing page undefined
path = /studio/crt
Studio - undefined > `Crt landing`

2️⃣  🟡 Token Page editing undefined
path =studio/crt/edit 
Studio - undefined > `Crt edit`

3️⃣  ➕ New Page Events to add
Page title, name  = Portfolio
url = /portfolio
Attributes = Tokens tab, NFTs tab


## ➕ New Track Events to add
1️⃣ Revenue share started
Properties = channelId, token ticker

2️⃣ Revenue share closed
Properties = channelId, token ticker

_______

❓  Also one thing left which I cannot figure out.. how to send the number of CRts and JOYs as `number` and not a string. Since this would enhance my ability to build reports in Amplitude. Cannot seem to find this in the documentation so will raise a support ticket with them.

dmtrjsg:
Nicely done 👍 @attemka

PR #6040, opened 2024-03-22
Segment analytics for CRT

attemka:
#6001 

few things to note @dmtrjsg:

1. Page visit will show 'CRT Dashboard', selected tab will be in params. Same as we did on YPP Dashboard.
2. Can't track 'Tokens sold on AMM' as this can happen while user is away and we're only tracking user events. Although, 'Tokens purchased on AMM' is tracked and should cover it.

### DIFF
--- a/packages/atlas/src/components/_crt/BuyMarketTokenModal/BuyMarketTokenModal.tsx
+++ b/packages/atlas/src/components/_crt/BuyMarketTokenModal/BuyMarketTokenModal.tsx
@@ -17,6 +17,7 @@ import { DialogProps } from '@/components/_overlays/Dialog'
 import { DialogModal } from '@/components/_overlays/DialogModal'
 import { absoluteRoutes } from '@/config/routes'
 import { useMediaMatch } from '@/hooks/useMediaMatch'
+import { useSegmentAnalytics } from '@/hooks/useSegmentAnalytics'
 import { hapiBnToTokenNumber, tokenNumberToHapiBn } from '@/joystream-lib/utils'
 import { useFee, useJoystream, useSubscribeAccountBalance } from '@/providers/joystream'
 import { useSnackbar } from '@/providers/snackbars'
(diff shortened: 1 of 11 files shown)


packages/atlas/src/components/_crt/BuyMarketTokenModal/BuyMarketTokenModal.tsx
```diff
@@ -164,6 +167,12 @@ export const BuyMarketTokenModal = ({ tokenId, onClose: _onClose, show }: BuySal
         ),
       onTxSync: async () => {
         if (memberTokenAccount?.tokenAccounts.length) {
+          trackAMMTokensPurchased(
+            data?.creatorTokenById?.symbol ?? 'N/A',
```
WRadoslaw, line 172: We shouldn't use symbol as it is not unique

packages/atlas/src/components/_crt/CloseMarketModal/CloseMarketModal.tsx
```diff
@@ -76,6 +78,7 @@ export const CloseMarketModal = ({ onClose, show, channelId, tokenId }: CloseMar
         txFactory: async (updateStatus) =>
           (await joystream.extrinsics).closeAmm(memberId, channelId, proxyCallback(updateStatus)),
         onTxSync: async () => {
+          trackAMMClosed(symbol ?? 'N/A', channelId)
```
WRadoslaw: Same as above

WRadoslaw (changes requested):

WRadoslaw (approved):
lgtm

dmtrjsg:
> Tokens sold on AMM' as this can happen while user is away and we're only tracking user events. Although, 'Tokens purchased on AMM' is tracked and should cover it.

@attemka user can initiate sale via GUI, that's what I was asking for here. I hope that was implemented? Its the same operation as purchased but user just clicks "sell" instead of "purchase".

dmtrjsg (approved):
ok

WRadoslaw:
Looks good

PR #6118, opened 2024-04-05
🧑‍🚀 Action refetching

WRadoslaw:


packages/atlas/src/providers/notifications/notifications.hooks.ts
```diff
@@ -82,15 +87,21 @@ export const useNotifications = (opts?: Pick<QueryHookOptions, 'notifyOnNetworkS
...
     () =>
-      rawNotifications.map(
-        ({ node: { id, createdAt, status, notificationType } }): NotificationRecord => ({
+      rawNotifications.map(({ node: { id, createdAt, status, notificationType } }): NotificationRecord => {
+        const notificationData = parseNotificationType(notificationType as NotificationType)
+        if (notificationData.refetchAction && !refetchedNotifications.has(id)) {
+          const refetchFn = networkUtils[notificationData.refetchAction.name]
+          refetchedNotifications.set(id, 1)
+          refetchFn(...(notificationData.refetchAction.args as Parameters<typeof refetchFn>))
+        }
```
thesan: LGTM!

Personally I would just move this if in a useEffect (with `[notifications, networkUtils]` as the dependencies).

packages/atlas/src/providers/networkUtils/networkUtils.provider.tsx
```diff
@@ -0,0 +1,253 @@
...
+      })
+    },
+    [activeMemberId, client]
+  )
+
+  const refetchAllCommentsSections = useCallback(async () => {
+    await client.refetchQueries({
+      updateCache: (cache) => {
+        cache.evict({
+          fieldName: 'commentsConnection',
```
thesan, line 142: I'm not sure how to improve this... Ideally fetching the comments by video id would be enough for Apollo to figure out the right query to update, but I don't know if it's doable.

thesan (approved):

packages/atlas/src/providers/networkUtils/networkUtils.provider.tsx
WRadoslaw: I've tried to just refetch comments for specific video ID, but I think because we use `orderBy` and other pagination params in the query, a query with simply video ID doesn't match the same cache entry in Apollo and it doesn't update it.

PR #6248, opened 2024-05-01
fix HTML head tags and auto-fullscreen on mobile safari

kdembler:
While investigating why VWO instrumentation doesn't work, I've noticed that almost all tags that are supposed to be in `<head>` are actually in `<body>` because of the `<preference>` tag that was added some time ago. This tag is not valid HTML and is instead intended to be used for `config.xml` for iOS apps. I don't expect it to have any impact for browser apps.

With `<preference>` tag:
CleanShot 2024-05-01 at 10 45 46@2x

Without `<preference>` tag:
CleanShot 2024-05-01 at 10 45 59@2x

It's quite possible that this messed up other things as well (like social previews in some cases), we just didn't notice yet.

### DIFF
--- a/packages/atlas/src/index.html
+++ b/packages/atlas/src/index.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <preference name="AllowInlineMediaPlayback" value="true" />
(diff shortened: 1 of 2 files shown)


kdembler:
Also added `playsInline` attribute to the video on the referrals page so it doesn't go into fullscreen automatically

PR #6372, opened 2024-06-04
🦊 Metamask integration

thesan:
SSIA


thesan:
@kdembler I was not able to test it yet because first [Atlas dev's Orion](https://atlas-next.joystream.org/network/config.json) is down for me and on Atlas next I get: "No memberships connected" even from [dev.gleev.xyz](https://dev.gleev.xyz/) with Alice (which does have a membership :confused:). Could you test with the membership you bought with metamask on mainnet please ?

thesan:
BTW just like with the Pioneer version it doesn't work on FF either for me

Issue #6502, opened 2025-01-14
Only 10 referrals are shown in studio's "referrals" tab


Lezek123:
**Describe the bug**

One of the creators reported that they only see 10 referrals in their Gleev studio (_Studio -> Earn -> Referrals_), even though they invited more:

![image](https://github.com/user-attachments/assets/9942dfac-6710-457f-91c3-0061a2fd4edd)

It can be verified here: https://172.236.19.60.nip.io/channels/66483/referrals, that indeed the number of referrals for this channel is indeed greater than 10.

It seems like the pagination on this page is either lacking or not working properly. Similar issue was also reported for CRT market transactions: https://github.com/Joystream/atlas/issues/6438 

**To Reproduce**

Temporarily hardcode `channelId` in `YppDashboardReferralsTab.tsx` to `66483` (for example)

**Expected behavior**

All referrals should be properly shown and paginated.

PR #6503, opened 2025-01-22
Add YPP referral pagination

DzhideX:
This PR aims to fix the issue described here: https://github.com/Joystream/atlas/issues/6502

In the other hooks that implement this functionality, the pagination was done in the fetch directly through GraphQL. Since this fetches all of the referrals at once, I implemented the pagination functionality within the hook itself after the data is fetched. I also added sorting to show the latest referrals first.

cc @Lezek123


Lezek123 (approved):
LGTM

Issue #6438, opened 2024-08-02
CRTs: Market Transactions page not updating


mochet:
# Problem
This page seems to only show 10 transactions. There have been transactions more recently than this that are not shown.

https://gleev.xyz/studio/crt-dashboard?tab=Market

![image](https://github.com/user-attachments/assets/1bed4845-522e-4cef-9c6f-fcc7e4105f9f)

PR #6505, opened 2025-01-24
Add market transactions pagination

DzhideX:
The PR aims to fix the issue described here: https://github.com/Joystream/atlas/issues/6438

cc @Lezek123

### DIFF
--- a/packages/atlas/src/hooks/useMarketTransactionsPagination.ts
+++ b/packages/atlas/src/hooks/useMarketTransactionsPagination.ts
@@ -26,7 +26,6 @@ export const useMarketTransactionsPagination = ({
   const { data, loading } = useGetFullAmmCurveQuery({
     variables: {
       where: {
-        // TODO: Change this:
         id_eq: ammId,
       },
     },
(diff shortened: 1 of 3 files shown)


packages/atlas/src/hooks/useMarketTransactionsPagination.ts
```diff
@@ -0,0 +1,54 @@
...
+  const [currentPage, setCurrentPage] = useState(0)
+  const [perPage, setPerPage] = useState(initialPageSize)
+
+  const { data, loading } = useGetFullAmmCurveQuery({
+    variables: {
+      where: {
+        id_eq: ammId,
+      },
+    },
+    skip: false,
```
Lezek123:
```suggestion
    skip: !ammId,
```

Lezek123 (changes requested):

Lezek123 (approved):
LGTM, it fixes the original issue, although a better solution would be to fetch only a page of ammTransaction at once, instead of pre-fetching all of them.

